PR #10, opened 2017-08-31
Clean up metrics aggregator. Improve quick start procedure in README.md

becketqin:


becketqin:
@smccauliff I am checking the patch because it has critical fixes. I'll address your comments in a follow up patch.

smccauliff:
Sorry I forgot about this.

becketqin:
No worries. Let me know if you have any comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsReporterSampler.java
```diff
@@ -155,17 +156,15 @@ public void configure(Map<String, ?> configs) {
     consumerProps.setProperty(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, Integer.toString(Integer.MAX_VALUE));
     KafkaCruiseControlUtils.setSslConfigs(consumerProps, configs);
     _metricConsumer = new KafkaConsumer<>(consumerProps);
-    _metricConsumer.subscribe(Pattern.compile(metricReporterTopic), new ConsumerRebalanceListener() {
-      @Override
-      public void onPartitionsRevoked(Collection<TopicPartition> collection) {
-        _metricConsumer.commitSync();
+    _metricConsumer.subscribe(Pattern.compile(metricReporterTopic), new NoOpConsumerRebalanceListener());
```
smccauliff: What's this change about?

becketqin: Previously there could be multiple consumers, so rebalance is possible. But now we only have one consumer thread and we always use timestamp search for the metric messages, so I thought we do not need to have a rebalance listener any more. But I guess it may still be possible for the consumer to drop out of the group in the middle of consumption, maybe we should still commit offsets here.

PR #12, opened 2017-09-01
Remove invalid UTF-8 byte sequence that breaks build on some platforms

bai:
There's an invalid character that prevents build on some platforms: in most of the cases it's in the copyright header right after `(the "License").` but other are also affected.


becketqin:
Thanks for the patch. LGTM.

PR #17, opened 2017-09-02
Fix num valid windows bug. Add back the rebalance listener for CruiseControlMetricReporterSampler.

becketqin:
This is a pretty strait forward patch. Will add LoadMonitor unit tests later.


lindong28:
LGTM

PR #19, opened 2017-09-02
Added a LoadMonitorTest and fixed a few bugs.

becketqin:


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -89,10 +89,23 @@ public synchronized ClusterAndGeneration refreshMetadata(long timeout) {
     return new ClusterAndGeneration(_metadata.fetch(), _metadataGeneration.get());
   }
 
+  /**
+   * Get the metadata maintained by this metadata client.
```
smccauliff: This is kind of useless documentation.  For methods like getX saying that it returns X is not really useful.  Saying something like "X is non-null" or "X may be null" or something like that is useful.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -167,11 +184,14 @@ public LoadMonitorState state() {
     // Get valid snapshot window number and populate the monitored partition map.
     // We do this primarily because the checker and aggregator are not always synchronized.
     int numValidSnapshotWindows = 0;
-    SortedMap<Long, Double> loadSnapshotsWindows = new TreeMap<>();
+    boolean hasInvalidSnapshotWindows = false;
+    SortedMap<Long, Double> loadSnapshotsWindows = new TreeMap<>((w1, w2) -> Long.compare(w2, w1));
```
smccauliff: Why is the ordering of the snapshots being changed?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -83,20 +83,36 @@
...
+    _metadataClient = metadataClient;
+    
     _brokerCapacityConfigResolver = config.getConfiguredInstance(KafkaCruiseControlConfig.BROKER_CAPACITY_CONFIG_RESOLVER_CLASS_CONFIG,
                                                                  BrokerCapacityConfigResolver.class);
     _numSnapshots = config.getInt(KafkaCruiseControlConfig.NUM_LOAD_SNAPSHOTS_CONFIG);
 
     _metricCompletenessChecker = new MetricCompletenessChecker(_numSnapshots);
 
-    _metricSampleAggregator = new MetricSampleAggregator(config, metadata, _metricCompletenessChecker);
+    _metricSampleAggregator = new MetricSampleAggregator(config, metadataClient.metadata(), _metricCompletenessChecker);
```
smccauliff: What happens if metadata client returns null?

smccauliff: Never mind seems like it can't return null.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
smccauliff: You might want to use Comparator.reverseOrder() instead.

becketqin: We changed the window order in MetricSampleAggregator to simplify the logic. So it seems more consistent to keep the same order here. Good suggestion about `Comparator.reverseOrder()`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
becketqin: Yeah, I guess it is more of follow the convention to have a Java doc for each public method...

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -167,11 +185,14 @@ public LoadMonitorState state() {
     // Get valid snapshot window number and populate the monitored partition map.
     // We do this primarily because the checker and aggregator are not always synchronized.
     int numValidSnapshotWindows = 0;
-    SortedMap<Long, Double> loadSnapshotsWindows = new TreeMap<>();
+    boolean hasInvalidSnapshotWindows = false;
+    SortedMap<Long, Double> loadSnapshotsWindows = new TreeMap<>(Comparator.reverseOrder());
     for (long window : _metricSampleAggregator.visibleSnapshotWindows()) {
       double monitoredPartitionsPercentage = windowToMonitoredPercentage.getOrDefault(window, 0.0);
-      if (monitoredPartitionsPercentage >= minMonitoredPartitionsPercentage) {
+      if (monitoredPartitionsPercentage >= minMonitoredPartitionsPercentage && !hasInvalidSnapshotWindows) {
```
smccauliff: So if there is a window that fell below the minimum number of monitored partitions  then this will clamp the number of valid snapshot windows at the number of contiguous windows starting from the most recent window time index.  Why do you want that?

smccauliff:
Other than that one comment this looks good to me.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
becketqin: This is the existing behavior. It is primarily because MetricSampleAggregator does not really skip an invalid window and only take valid windows. I agree the behavior is not ideal and we probably want to consider change this. Maybe we can do that in a separate patch.

becketqin:
@smccauliff Thanks for the review. I created #22 to track the behavior change of handling valid snapshot windows. I'll merge the current PR as it contains important bug fixes.

PR #34, opened 2017-09-14
Improve MetricSampleAggregator and sanity check stability

becketqin:
The patch does the following:
1. check whether old windows should be evicted on every sampling instead of waiting for the next window to be rolled out.
2. Use a adaptive epsilon in the double comparison. This avoids false failure on the big disk utilization.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
@@ -156,18 +156,10 @@ public class MetricSampleAggregator {
                     _windowedAggregatedPartitionMetrics.size());
           _activeSnapshotWindow = snapshotWindow;
           newWindow = true;
-          // We only keep N ready snapshots and one active snapshot, evict the old ones if needed. But do not do this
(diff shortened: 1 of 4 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
```diff
@@ -188,6 +180,23 @@ public boolean addSample(PartitionMetricSample sample, boolean leaderValidation,
     return true;
   }
 
+  private void maybeEvictOldSnapshots() {
+    if (_windowedAggregatedPartitionMetrics.size() > _numSnapshotsToKeep + 1) {
+      synchronized (this) {
```
smccauliff: Its cleaner to just have this method synchronized; one less block of code.

smccauliff: Otherwise LGTM!

becketqin: We are not making the method synchronized to avoid unnecessary locking.

PR #35, opened 2017-09-16
Add an identity map for TopicPartition in the metric aggregator to reduce memory consumption

becketqin:
MetricSampleAggregator has a ConcurrentHashMap for each snapshot window. Currently the key in those ConcurrentHashMaps are different TopicPartition instances for the same partition, which takes a lot of memory. This patch uses an identity map to ensure that there is only one TopicPartition instance for the same partition.


lindong28:
LGTM

PR #38, opened 2017-09-25
Refactor the MetricSampleAggregator to extract the core logic.

becketqin:
The patch create a new project cruise-control-core which will contain the product agnostic logic in the future.


cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/MetricSample.java
```diff
@@ -0,0 +1,103 @@
...
+    if (_sampleTime < 0) {
+      _sampleTime = closingTime;
+    }
+  }
+
+  /**
+   * Validate the metric sample.
+   */
+  public boolean isValid() {
+    boolean completeMetrics = _metrics.size() == Resource.values().length;
```
smccauliff: Resource.values() will always allocate a new array.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/common/Generationed.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+
+package com.linkedin.cruisecontrol.common;
+
+/**
+ * An interface that helps maintain and compare the generation.
+ */
+public interface Generationed<G> {
```
smccauliff: Why bother have a parameterized interface and just have objects that have generation numbers associated with them just return their generation number as a long.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/MetricSample.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class MetricSample<E> {
+  private static final Logger LOG = LoggerFactory.getLogger(MetricSample.class);
+  protected final E _entity;
+  protected final Map<Resource, Double> _metrics;
```
smccauliff: This can just be double[Resource.values().length] which will save lots of space.  null values can be represented as Double.NaN

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/MetricSample.java
```diff
@@ -0,0 +1,103 @@
...
+package com.linkedin.cruisecontrol.monitor.sampling;
+
+import com.linkedin.cruisecontrol.resource.Resource;
+import java.util.HashMap;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class MetricSample<E> {
```
smccauliff: It's not clear what type parameter E is supposed to represent.  @type javadoc can be used to provide documentation.  So is en Entity equivelent to a broker?

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricsForWindow.java
```diff
@@ -0,0 +1,265 @@
...
+  Map<E, AggregatedMetrics> entityMetrics() {
+    return Collections.unmodifiableMap(_entityMetrics);
+  }
+
+  /**
+   * Add a sample to this window
+   * @param sample the sample to add.
+   */
+  void addSample(MetricSample<E> sample, Long generation) {
+    AggregatedMetrics aggMetrics =
```
smccauliff: Sorry if this does not apply.  I've forgotten the threading model that cruise control is using.  _identityEntityMap is just Map<> and so this might be multithread safe.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/MetricSample.java
smccauliff: Also E is used as a key in a map so it needs to implement hashCode() and equals().  This needs to be documented somewhere.

smccauliff: Also I really want the type parameter to be E extends Comparable<E> so I can eventually fulfill the dream of having a deterministic cruise control.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaMetricSampleAggregatorMetadata.java
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  /**
+   * Get the available windows and valid snapshots that meets the monitored percentage requirements.
+   * @param minMonitoredPartitionsPercentage the minimum required monitored percentage.
+   * @param totalNumPartitions the total number of partitions.
+   * @return An {@link AvailableWindows} containing the information of windows that meets the monitored 
+   * partitions percentage requirements.
+   */
+  synchronized AvailableWindows validWindows(double minMonitoredPartitionsPercentage,
```
smccauliff: Shouldn't this logic be extracted from here and put in core?  The problem we have with consistent metrics collection is likely not limited to Kafka.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/common/Generationed.java
becketqin: This is because we have some composite generations. For example, cluster model has a generation which is the combination of metadata generation and the load generation.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/MetricSample.java
becketqin: The metric samples are short lived objects. They will be discarded once we aggregate the data into the metric sample aggregator.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricsForWindow.java
becketqin: The implementation is a ConcurrentHashMap, so it should be safe. I can change the declaration to be ConcurrentMap.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaMetricSampleAggregatorMetadata.java
becketqin: Kafka have a special logic that if all the partitions should be included/excluded all together. The generic logic assumes independence of the entities. So we need special logic here.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/MetricSample.java
smccauliff: That's fine, but it's lots of extra garbage as well.  Number-of-entities x (number-of-metrics + 1) vs number-of-entities.

becketqin:
We have another PR opened for this.

PR #39, opened 2017-09-25
Pass all the configurations to the producer and consumers

becketqin:


becketqin:
The test failure is caused by a flaky test. Ran all the unit tests locally and all the tests have passed.

PR #43, opened 2017-09-28
fix the daemon option in the kafka-cruise-control-start.sh

becketqin:


kafka-cruise-control-start.sh
```diff
@@ -136,7 +136,8 @@ if [ -z "$KAFKA_JVM_PERFORMANCE_OPTS" ]; then
   KAFKA_JVM_PERFORMANCE_OPTS="-server -XX:+UseG1GC -XX:MaxGCPauseMillis=20 -XX:InitiatingHeapOccupancyPercent=35 -XX:+DisableExplicitGC -Djava.awt.headless=true"
 fi
 
-
+DAEMON_NAME="kafka-cruise-control"
+CONSOLE_OUTPUT_FILE=$LOG_DIR/$DAEMON_NAME.out
```
smccauliff: ${LOG_DIR}/${DAEMON_NAME}.out is probably better style, but what is there  should work.  Maybe add quotes to "${LOG_DIR}" since paths can have spaces in them.

PR #45, opened 2017-10-05
Improvement and clean up.

becketqin:
1. Bumped Kafka dependency to 0.10.2.1. So broker version 0.10.0 and above would work.
2. Removed test harness classes. Just depend on li-apache-kafka-clients instead.
3. Added separate configurations for the metrics reporter producer.
4. Added the missing sasl configs.


cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -85,30 +86,36 @@ public void close() {
 
   @Override
   public void configure(Map<String, ?> configs) {
-    String bootstrapServers = (String) configs.get(CRUISE_CONTROL_METRICS_REPORTER_BOOTSTRAP_SERVERS);
-    if (bootstrapServers == null) {
+    Properties producerProps = CruiseControlMetricsReporterConfigs.parseConfigMap(configs);
```
smccauliff: If CruiseControlMetricsRepoterConfigs just extended from the Kafka configs class it would be easy to put all this default logic into that class.  No?

PR #50, opened 2017-10-11
Fix the expected utilization when there is no snapshots.

becketqin:


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Load.java
```diff
@@ -100,14 +100,11 @@ public static int maxNumSnapshots() {
...
-      if (_snapshotsByTime.isEmpty()) {
-        return 0.0;
-      }
-      return Math.max(0, _snapshotsByTime.get(0).utilizationFor(resource));
+    if (_snapshotsByTime.isEmpty()) {
+      return 0.0;
     }
-
-    return Math.max(0, _accumulatedUtilization[resource.id()] / _snapshotsByTime.size());
+    return resource.equals(Resource.DISK) ? Math.max(0, _snapshotsByTime.get(0).utilizationFor(resource)) :
```
radai-rosenblatt: enums can use ==

PR #69, opened 2017-11-03
Avoid holding the lock for too long when running goal violation detection

becketqin:
We are having more and more goals now. We are holding the lock for too long. This patch release the lock between each goal to avoid holding it for all the goals.


smccauliff:
LGTM

PR #72, opened 2017-11-09
Executor should handle deleted topic correctly.

becketqin:


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -311,6 +311,11 @@ private void waitForExecutionTaskToFinish() {
         Cluster cluster = _metadataClient.refreshMetadata().cluster();
         for (ExecutionTask task : _executionTaskManager.tasksInProgress()) {
           TopicPartition tp = task.proposal.topicPartition();
+          if (cluster.partition(tp) == null) {
+            LOG.debug("Proposal {} is marked as finished because the topic has been deleted");
```
efeg: Proposal is not actually logged. Should we log `task.proposal`?

PR #73, opened 2017-11-14
Refactor Rest API:

efeg:
* Remove redundant parameter _with_available_valid_windows_ -- instead use a single parameter _with_available_valid_partitions_ (default is false).
* Raise User Request Exception if request contains invalid endpoint parameters. In the existing version of CC, if user request contains non-existing parameters, CC silently ignores them. As a result, the user might incorrectly assume that her parameter is acknowledged by CC.
* Implement appropriate JSON responses.
* Add `throttle_added_broker` parameter to be used while adding a new, throttled broker.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
@@ -27,9 +27,11 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.HashMap;
+import java.util.Set;
 import java.util.StringJoiner;
 import java.util.stream.Collectors;
 import javax.servlet.ServletException;
@@ -41,6 +43,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+
+import static com.linkedin.kafka.cruisecontrol.servlet.KafkaCruiseControlServlet.EndPoint.*;
 
 
 /**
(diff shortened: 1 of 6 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -143,27 +221,39 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
...
-            getState(request, response);
-            break;
-          default:
-            throw new UserRequestException("Invalid URL for GET");
+        Set<String> validParamNames = VALID_ENDPOINT_PARAM_NAMES.get(endPoint);
+        if (validParamNames != null && !validParamNames.containsAll(request.getParameterMap().keySet())) {
+          // User request specifies parameters that are not a subset of the valid parameters.
+          String errorResp;
+          if (wantJSON(request)) {
+            errorResp = "{\"error\": \"Invalid endpoint parameters in " + endPoint + " get request.\"}";
```
becketqin: We probably want to return the set of invalid parameters to the users. Otherwise it is difficult for the users to know what was going wrong.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -216,19 +306,31 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
-          case REBALANCE:
-            rebalance(request, response);
-            break;
-          case STOP_PROPOSAL_EXECUTION:
-            stopProposalExecution();
-            break;
-          default:
-            throw new UserRequestException("Invalid url for POST");
+        Set<String> validParamNames = VALID_ENDPOINT_PARAM_NAMES.get(endPoint);
+        if (validParamNames != null && !validParamNames.containsAll(request.getParameterMap().keySet())) {
```
becketqin: Ditto above.

Issue #75, opened 2017-11-20
Add an option to blacklist topics from cruise control.


UnsignedLong:
Hi,

we're using https://github.com/andreas-schroeder/kafka-health-check which creates several kafka topics. One broker health topic with one partition assigned to the broker and a "replication-check" topic with one replica per broker.
The problem is - when using the rack-aware goal, cruise control catches an exception because there are more replicas than racks.
Because of this it would be great to be able to blacklist whole topics from cruise control. Blacklisting the topic from partition movement is not sufficient here because the rack-aware goal still throws exceptions.

becketqin:
Thanks for reporting the issue. Yes, the current code does not treat the blacklisted topic separately when doing verification. Will fix it.

efeg:
https://github.com/linkedin/cruise-control/pull/79 should solve this issue.

PR #79, opened 2017-11-23
Enforce exclusion of given topics in goal optimization.

efeg:
This patch fixes the issue https://github.com/linkedin/cruise-control/issues/75.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareCapacityGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareCapacityGoal.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
@@ -34,8 +35,8 @@ import org.slf4j.LoggerFactory;
  * Class for achieving the following hard goals:
  * HARD GOAL#1: Generate leadership and replica movement proposals to push the load on brokers under the capacity limit.
  * HARD GOAL#2: Generate replica movement proposals to provide rack-aware replica distribution.
- * 
- * @deprecated Please use {@link RackAwareGoal}, {@link DiskCapacityGoal}, {@link NetworkInboundCapacityGoal}, 
+ *
+ * @deprecated Please use {@link RackAwareGoal}, {@link DiskCapacityGoal}, {@link NetworkInboundCapacityGoal},
  * {@link NetworkOutboundCapacityGoal}, {@link CpuCapacityGoal} instead.
  */
 @Deprecated
@@ -168,9 +169,10 @@ public class RackAwareCapacityGoal extends AbstractGoal {
    * priority set in the _balancingConstraint.
    *
    * @param clusterModel The state of the cluster.
+   * @param excludedTopics The topics that should be excluded from the optimization proposals.
    */
   @Override
-  protected void initGoalState(ClusterModel clusterModel)
+  protected void initGoalState(ClusterModel clusterModel, Set<String> excludedTopics)
       throws AnalysisInputException, ModelInputException {
     _currentResource = null;
     _balancedResources = new HashSet<>();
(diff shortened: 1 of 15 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareCapacityGoal.java
```diff
@@ -204,7 +206,17 @@ protected void initGoalState(ClusterModel clusterModel)
       }
     }
     // Sanity Check #3 -- i.e. not enough racks to satisfy rack awareness.
-    if (clusterModel.maxReplicationFactor() > clusterModel.numHealthyRacks()) {
+    int maxReplicationFactorOfIncludedTopics = 1;
+    Map<String, Integer> replicationFactorByTopic = clusterModel.replicationFactorByTopic();
+
+    for (Map.Entry<String, Integer> replicationFactorByTopicEntry: replicationFactorByTopic.entrySet()) {
+      if (!excludedTopics.contains(replicationFactorByTopicEntry.getKey())) {
+        maxReplicationFactorOfIncludedTopics =
```
becketqin: Exception could be thrown immediately if the replication factor is larger than the rack number. There is no need to iterate over all the topics anymore.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -48,8 +48,8 @@
   private final double _monitoredPartitionsPercentage;
   private final double[] _clusterCapacity;
   private Load _load;
-  // An integer to keep track of the maximum replication factor that a partition was ever created with.
-  private int _maxReplicationFactor;
+  // The replication factor that each topic in the cluster created with.
+  private Map<String, Integer> _replicationFactorByTopic;
```
becketqin: We could keep the maxReplicationFactor variable as well to avoid potential iteration.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/RackAwareExcludedTopicsTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.slf4j.LoggerFactory;
+
+import static org.junit.Assert.assertTrue;
+
+
+/**
+ * Unit test for testing rack awareness with excluded topics under fixed cluster properties.
+ */
+@RunWith(Parameterized.class)
+public class RackAwareExcludedTopicsTest {
```
becketqin: This test seems a little confusing. 

I think we should have some more straight forward and specific tests. For example:
1. Test a cluster model with 2 racks, a single topic with a single partition, but both replicas are in the same rack. Ensure when the topic is not excluded, the proposals contains the partition. When the topic is excluded, no proposal is generated.
2. Test the cases when an exception should be thrown and when an exception should not be thrown, depending on whether the violated topic is included or not.

Also, the test should be run on all the goals.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
```diff
@@ -135,10 +136,17 @@ protected boolean selfSatisfied(ClusterModel clusterModel, BalancingProposal pro
...
    *
    * @param clusterModel The state of the cluster.
+   * @param excludedTopics The topics that should be excluded from the optimization proposals.
    */
   @Override
-  protected void initGoalState(ClusterModel clusterModel)
+  protected void initGoalState(ClusterModel clusterModel, Set<String> excludedTopics)
       throws AnalysisInputException, ModelInputException {
+
+    // Sanity check to make sure that not all replicas are excluded.
```
becketqin: Is this an exception? It seems we should just do nothing and return, right?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/TopicReplicaDistributionGoal.java
```diff
@@ -123,21 +123,28 @@ protected boolean selfSatisfied(ClusterModel clusterModel, BalancingProposal pro
...
    */
   @Override
-  protected void initGoalState(ClusterModel clusterModel)
+  protected void initGoalState(ClusterModel clusterModel, Set<String> excludedTopics)
       throws AnalysisInputException, ModelInputException {
     _numRebalancedTopics = 0;
     _topicsToRebalance = new ArrayList<>(clusterModel.topics());
+    _topicsToRebalance.removeAll(excludedTopics);
+
+    if (_topicsToRebalance.isEmpty()) {
```
becketqin: ditto above.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/common/RandomCluster.java
```diff
@@ -128,14 +128,13 @@ public static void populate(ClusterModel cluster,
...
     for (int i = 0; i < properties.get(ClusterProperty.NUM_TOPICS).intValue(); i++) {
-      int oldReplicationFactor = metadata.get(i).replicationFactor();
       int randomReplicationFactor = uniformlyRandom(properties.get(ClusterProperty.MIN_REPLICATION).intValue(),
           properties.get(ClusterProperty.MAX_REPLICATION).intValue(), TestConstants.REPLICATION_SEED + i);
       metadata.get(i).setReplicationFactor(randomReplicationFactor);
 
       if (totalTopicReplicas(metadata) > properties.get(ClusterProperty.NUM_REPLICAS).intValue()) {
         // Rollback to previous replicationFactor.
-        metadata.get(i).setReplicationFactor(oldReplicationFactor);
+        metadata.get(i).setReplicationFactor(properties.get(ClusterProperty.MIN_REPLICATION).intValue());
```
becketqin: Just curious why do we need this change?

efeg: To prevent the replication factor value from being set to a value under `MIN_REPLICATION` -- i.e. to the default value of metadata.

PR #80, opened 2017-12-01
Make Cruise Control model more general.

efeg:
This patch is part of a series of patches to make CC more general.
It adds support for the use of general modules.


efeg:
-- This patch is outdated, and the support for making Cruise Control model more general is backlogged for now (closed until we revisit).

PR #82, opened 2017-12-04
Cruise control should not stuck when broker dies during partition movement.

becketqin:


cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -86,6 +116,7 @@ object ExecutorUtils {
 
       // We do not use the ReassignPartitionsCommand here because we want to have incremental partition movement.
       if (newPartitionAssignment.nonEmpty)
+        System.out.println(newPartitionAssignment)
```
efeg: Why do we need to print this? -- i.e. why don't we log it?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -115,6 +115,39 @@ public void testMoveNonExistingPartition() {
...
+      }
+    }
+    int initialLeader0 = (Integer) zkUtils.getLeaderForPartition(topic0, partition).get();
+    int initialLeader1 = (Integer) zkUtils.getLeaderForPartition(topic1, partition).get();
+    _brokers.get(initialLeader0 == 0 ? 1 : 0).shutdown();
+    BalancingProposal proposal0 =
+        new BalancingProposal(tp0, initialLeader0, initialLeader0 == 0 ? 1 : 0, BalancingAction.REPLICA_MOVEMENT);
+    BalancingProposal proposal1 =
+        new BalancingProposal(tp1, initialLeader1, initialLeader1 == 0 ? 1 : 0, BalancingAction.LEADERSHIP_MOVEMENT);
+
```
efeg: Should we also tests for REPLICA_DELETION and REPLICA_ADDITION?

becketqin: Currently those two actions are not generated anywhere. I'll add test when we start to use them.

efeg:
LGTM.

PR #86, opened 2017-12-08
Fix potential misexecution of leadership movements as replica reassignments.

efeg:


cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -88,13 +90,29 @@ object ExecutorUtils {
                 if (!currentReplicaAssignment.contains(destinationBroker)) {
                   if (currentReplicaAssignment.contains(sourceBroker)) {
                     // this is a normal movement.
-                    if (destinationBroker != null)
-                      (currentReplicaAssignment :+ destinationBroker.toInt).filter(_ != sourceBroker) 
-                    else
+                    if (destinationBroker != null) {
```
becketqin: It seems clearer to have the if statements on the balancing action. So we have something like:
```
if (balacning action is replicamovement) {
  getNewReplicasForReplicaMovement
} else if (balancing action is replica addition) {
   getNewReplicasForReplicaAddition
} else if (balancing action is replica deletion) {
   getNewReplicasForReplicaDeletion
} else {
   exception
}```

File: cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala [hunk 2]
```diff
@@ -85,38 +87,126 @@ object ExecutorUtils {
...
-        zkUtils.updatePartitionReassignmentData(newPartitionAssignment)
+      if (newReplicaAssignment.nonEmpty)
+        zkUtils.updatePartitionReassignmentData(newReplicaAssignment)
+    }
+  }
+  def getNewReplicasForReplicaAddition(currentReplicaAssignment: Seq[Int], task: ExecutionTask): Seq[Int] = {
+    val sourceBroker = task.sourceBrokerId()
+    val destinationBroker = task.destinationBrokerId()
+    if (!currentReplicaAssignment.contains(destinationBroker)) {
+      if (currentReplicaAssignment.contains(sourceBroker)) {
```
becketqin: For replica addition, the source should be null. So this branch should just be if source is not null throw exception. If we trust the sanity check we have already made in the constructor of BalancingProposal, we can just skip this check and assume the source is null.

cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -85,38 +87,126 @@ object ExecutorUtils {
...
+        // If the source broker is no longer in the list, just do not add the task to the reassignment.
+        currentReplicaAssignment
+      }
+    }
+  }
+
+  def getNewReplicasForReplicaDeletion(currentReplicaAssignment: Seq[Int], task: ExecutionTask): Seq[Int] = {
+    val sourceBroker = task.sourceBrokerId()
+    val destinationBroker = task.destinationBrokerId()
+    if (!currentReplicaAssignment.contains(destinationBroker)) {
```
becketqin: Similar here, the destination is guaranteed to be null for replica deletion.

cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -85,38 +87,126 @@ object ExecutorUtils {
...
+    }
+  }
+
+  def getNewReplicasForReplicaMovement(currentReplicaAssignment: Seq[Int], task: ExecutionTask): Seq[Int] = {
+    val sourceBroker = task.sourceBrokerId()
+    val destinationBroker = task.destinationBrokerId()
+    if (!currentReplicaAssignment.contains(destinationBroker)) {
+      if (currentReplicaAssignment.contains(sourceBroker)) {
+        // This is a normal movement.
+        if (destinationBroker != null) {
```
becketqin: It is safe to assume both source and destination are not null in this case.

cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -85,38 +87,87 @@ object ExecutorUtils {
...
+                    addTask = false
                   }
-                } else {
-                  if (currentReplicaAssignment.contains(sourceBroker)) {
-                    // The destination broker is already in the list, we just need to filter out the source broker if
-                    // it exists.
-                    currentReplicaAssignment.filter(_ != sourceBroker)
-                  } else {
-                    // If the source broker is no longer in the list, just do not add the task to the reassignment.
+                  // Get new replicas for replica movement.
```
becketqin: Should this be in the else branch? After set addTask=false, we can just return Seq.empty. ditto below.

cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -85,38 +87,87 @@ object ExecutorUtils {
...
+  def getNewReplicasForReplicaMovement(currentReplicaAssignment: Seq[Int], task: ExecutionTask): Seq[Int] = {
+    // Both source and destination brokers are guaranteed to be non-null for replica movement.
+    val sourceBroker = task.sourceBrokerId()
+    val destinationBroker = task.destinationBrokerId()
+    if (!currentReplicaAssignment.contains(destinationBroker)) {
+      if (currentReplicaAssignment.contains(sourceBroker)) {
+        // This is a normal movement.
+        (currentReplicaAssignment :+ destinationBroker.toInt).filter(_ != sourceBroker)
+      } else {
+        // Task must be an add.
```
becketqin: Is this doc and exception right?

cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -85,38 +87,87 @@ object ExecutorUtils {
...
+        throw new RuntimeException(s"Unexpected type for proposal ${task.proposal}. Expected ${BalancingAction.REPLICA_ADDITION}")
+        currentReplicaAssignment :+ destinationBroker.toInt
```
becketqin: Dead code.

cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -85,38 +87,76 @@ object ExecutorUtils {
...
+  def getNewReplicasForReplicaMovement(currentReplicaAssignment: Seq[Int], task: ExecutionTask): Seq[Int] = {
+    // Both source and destination brokers are guaranteed to be non-null for replica movement.
+    val sourceBroker = task.sourceBrokerId()
+    val destinationBroker = task.destinationBrokerId()
+    if (!currentReplicaAssignment.contains(destinationBroker)) {
+      if (currentReplicaAssignment.contains(sourceBroker)) {
+        // This is a normal movement.
+        (currentReplicaAssignment :+ destinationBroker.toInt).filter(_ != sourceBroker)
+      } else {
+        throw new IllegalStateException(s"Unexpected proposal ${task.proposal}. Neither source nor destination broker" +
```
becketqin: Minior: we may want to include the current replica assignment in the exception message as well.

PR #89, opened 2017-12-13
Improve the servlet API by replacing the with_available_valid_partiti…

becketqin:
…ons with data_from parameter. The valid values are with_available_valid_partitions and with_available_valid_windows.


efeg:
LGTM.

PR #93, opened 2017-12-15
Shuffle nodes before getting their capacity from the capacity resolver

efeg:
This enables a capacity resolver to estimate the capacity of the nodes, for which the capacity retrieval has failed (if supported by the capacity resolver). If the call to the `capacityForBroker` fails on the invocation with the first node as the parameter, the capacity resolver would be unable to estimate the remaining nodes; hence, might throw an exception. Node shuffling decreases the likelihood of such failures on consecutive user requests.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -389,7 +390,14 @@ public ClusterModel clusterModel(long from, long to, ModelCompletenessRequiremen
     final Timer.Context ctx = _clusterModelCreationTimer.time();
     try {
       // Create the racks and brokers.
-      for (Node node : kafkaCluster.nodes()) {
+      // Shuffle nodes before getting their capacity from the capacity resolver.
```
becketqin: We should probably elaborate a little on what does "estimate" mean. The use case was essentially that if the capacity of one of the nodes is not available (due to some 3rd party service issue), the capacity resolver may want to use the capacity of a peer node as the capacity for that node. In order to make this work, we need to handle the case that the first node is problematic so the capacity resolver does not have the chance to get the capacity for the other nodes. Shuffling the nodes order will help in that case as the problematic node is unlikely to always be the first node in the list.

PR #95, opened 2017-12-20
Add Replica Distribution Goal with Configurable Balance Threshold.

efeg:
* This goal enables user-configurable balance threshold for the replica count --`replica.count.balance.threshold`.
For each broker, if replica count is above or below the average replica count more than the percentage set by the configured threshold value, the broker is considered as unbalanced.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
```diff
@@ -5,65 +5,133 @@
...
-   * @return True if action is acceptable by this goal, false otherwise.
+   * @return {@link ActionAcceptance#ACCEPT} if the action is acceptable by this goal,
+   * {@link ActionAcceptance#REPLICA_REJECT} otherwise.
    */
   @Override
-  public boolean isActionAcceptable(BalancingAction action, ClusterModel clusterModel) {
-    int numLocalReplicas = clusterModel.broker(action.sourceBrokerId()).replicas().size();
-    int numRemoteReplicas = clusterModel.broker(action.destinationBrokerId()).replicas().size();
+  public ActionAcceptance actionAcceptance(BalancingAction action, ClusterModel clusterModel) {
+    if (action.balancingAction() == ActionType.REPLICA_SWAP) {
```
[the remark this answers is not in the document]
efeg: Good catch!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -127,9 +174,19 @@ public ModelCompletenessRequirements clusterModelCompletenessRequirements() {
...
-      return true;
+      return action.balancingAction() != REPLICA_SWAP;
+    }
+
+    if (action.balancingAction() == REPLICA_SWAP) {
+      Replica destinationReplica = destinationBroker.replica(action.destinationTopicPartition());
+      double sourceUtilizationDelta = destinationReplica.load().expectedUtilizationFor(resource())
+                                      - sourceReplica.load().expectedUtilizationFor(resource());
+
+      return !(sourceUtilizationDelta == 0) && isLoadStillWithinLimitsAfterSwap(clusterModel, sourceReplica,
```
becketqin: sourceUtilizationDelta != 0?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -127,9 +174,19 @@ public ModelCompletenessRequirements clusterModelCompletenessRequirements() {
...
     Replica sourceReplica = clusterModel.broker(action.sourceBrokerId()).replica(action.topicPartition());
-    // If the source broker is dead and currently self healing dead brokers only, then the action must be executed.
+    // If the source broker is dead and currently self healing dead brokers only, unless it is replica swap, the action
+    // must be executed.
     if (!sourceReplica.broker().isAlive() && _selfHealingDeadBrokersOnly) {
-      return true;
+      return action.balancingAction() != REPLICA_SWAP;
+    }
+
+    if (action.balancingAction() == REPLICA_SWAP) {
```
becketqin: Can we use `switch` here as well? Or put the rest of the code in the `else` statement.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -229,15 +288,12 @@ protected void rebalanceForBroker(Broker broker,
                                     ClusterModel clusterModel,
                                     Set<Goal> optimizedGoals,
                                     Set<String> excludedTopics)
-      throws AnalysisInputException, ModelInputException {
+      throws ModelInputException {
```
becketqin: We agreed that we will remove ModelInputException.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -229,15 +288,12 @@ protected void rebalanceForBroker(Broker broker,
...
     boolean requireLessLoad = !isLoadUnderBalanceUpperLimit(clusterModel, broker);
     boolean requireMoreLoad = !isLoadAboveBalanceLowerLimit(clusterModel, broker);
     if (broker.isAlive() && !requireMoreLoad && !requireLessLoad) {
       // return if the broker is already under limit.
       return;
-    } else if (!clusterModel.newBrokers().isEmpty() && requireMoreLoad && !broker.isNew()) {
```
becketqin: Why are we removing this?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -246,7 +302,7 @@ protected void rebalanceForBroker(Broker broker,
     }
 
     // First try leadership movement
-    if (resource() == Resource.NW_OUT || resource() == Resource.CPU) {
+    if ((resource() == Resource.NW_OUT || resource() == Resource.CPU)) {
```
becketqin: Redundant parenthesis

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -321,22 +389,180 @@ protected boolean rebalanceByMovingLoadIn(Broker broker,
...
+      return result == 0 ? r1.topicPartition().toString().compareTo(r2.topicPartition().toString()) : result;
     });
+
+    sourceReplicas.addAll(resource() == Resource.NW_OUT ? broker.leaderReplicas() : broker.replicas());
+
+    // Sort the replicas initially to avoid sorting it every time.
+    PriorityQueue<CandidateBroker> candidateBrokerPQ = new PriorityQueue<>();
+    for (Broker candidate : clusterModel.healthyBrokersUnderThreshold(resource(), balanceUpperThreshold(clusterModel))
+                                        .stream().filter(b -> !b.replicas().isEmpty()).collect(Collectors.toSet())) {
+      // Get candidate replicas on candidate broker to try swapping with -- sorted in the order of trial.
```
becketqin: what does order of trial mean? ascending? descending?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -321,22 +389,180 @@ protected boolean rebalanceByMovingLoadIn(Broker broker,
     return true;
   }
 
-  protected boolean rebalanceByMovingLoadOut(Broker broker,
+  private SortedSet<Replica> replicasToSwapWith(Broker broker,
```
becketqin: This methods needs more comments. what does isSwapIn mean?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -321,22 +389,180 @@ protected boolean rebalanceByMovingLoadIn(Broker broker,
...
+                                                Set<String> excludedTopics,
+                                                double swapLoadLimit,
+                                                boolean isSwapIn) {
+    SortedSet<Replica> candidateReplicasToSwapWith = new TreeSet<>((r1, r2) -> {
+      int result = isSwapIn ? Double.compare(r2.load().expectedUtilizationFor(resource()), r1.load().expectedUtilizationFor(resource()))
+                            : Double.compare(r1.load().expectedUtilizationFor(resource()), r2.load().expectedUtilizationFor(resource()));
+      return result == 0 ? r1.topicPartition().toString().compareTo(r2.topicPartition().toString()) : result;
+    });
+    candidateReplicasToSwapWith.addAll(
+        resource() == Resource.NW_OUT
```
becketqin: This is probably the time to use `if...else...` rather than the seemingly short question mark.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -321,22 +389,180 @@ protected boolean rebalanceByMovingLoadIn(Broker broker,
...
+
+      Replica swappedInReplica = null;
+      Replica swappedOutReplica = null;
+      for (Replica sourceReplica : sourceReplicas) {
+        if (excludedTopics.contains(sourceReplica.topicPartition().topic())) {
+          continue;
+        }
+
+        // Try swapping the source with the candidate replicas. Get the swapped in replica if successful, null otherwise.
+        Replica swappedIn = maybeApplySwapAction(clusterModel, sourceReplica, candidateReplicasToSwapWith, optimizedGoals);
```
becketqin: It seems we may "over-swap" with this logic? i.e. say the ideal utilization is 20, lower limit is 10, upper limit is 30 and current broker is 40, we may directly go to 10 instead of going 20?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
```diff
@@ -200,17 +199,24 @@ public Load leadershipLoadForNwResources() {
...
+
+  /**
+   * Sort replicas that needs to be sorted to balance the given resource by descending order of resource cost.
+   * If resource is outbound network traffic, only leaders in the broker are sorted. Otherwise all replicas in
+   * the broker are sorted.
+   *
+   * @param resource Type of the resource.
+   * @return A list of sorted replicas chosen from the replicas residing in the given broker.
+   */
+  public List<Replica> sortedReplicas(Resource resource, boolean isReverse) {
```
becketqin: JavaDoc missing for isReverse.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -65,22 +70,64 @@ public ResourceDistributionGoal() {
...
+      double sourceUtilizationDelta = destinationReplica.load().expectedUtilizationFor(resource())
+                                      - sourceReplica.load().expectedUtilizationFor(resource());
+
+      if (sourceUtilizationDelta == 0) {
+        // No change in terms of load.
+        return ACCEPT;
+      }
+      // Check if the source or destination broker is within the balance limit before applying a swap that could
+      // potentially make them unbalanced -- i.e. never make a balanced broker unbalanced.
+      boolean isCurrentlyWithinLimit = sourceUtilizationDelta > 0
```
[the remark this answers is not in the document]
efeg: This is indeed expected. Added more comments to clarify.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -65,22 +70,64 @@ public ResourceDistributionGoal() {
...
+                                          || isLoadUnderBalanceUpperLimit(clusterModel, sourceReplica.broker()))
+                                       : (isLoadAboveBalanceLowerLimit(clusterModel, sourceReplica.broker())
+                                          || isLoadUnderBalanceUpperLimit(clusterModel, destinationBroker));
+
+      if (isCurrentlyWithinLimit) {
+        // Ensure that the resource utilization on brokers do not go out of limits after the swap.
+        return isLoadStillWithinLimitsAfterSwap(clusterModel, sourceReplica, destinationReplica) ? ACCEPT : REPLICA_REJECT;
+      }
+      // Ensure that the swap does not increase the utilization difference between brokers.
+      return isSelfSatisfiedAfterSwap(sourceReplica, destinationReplica) ? ACCEPT : REPLICA_REJECT;
```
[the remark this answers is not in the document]
efeg: Not really. Using only this method would potentially make a balanced broker unbalanced -- i.e. it is possible to decrease the load difference between brokers while (1) pulling one, which was previously above the lower balance limit, under the lower balance limit or (2) pushing one, which was previously under the upper balance limit, above the upper balance limit.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: Because while it is not possible to use `rebalanceByMovingLoadIn` to get more load to an old broker in case there are new brokers, it is potentially possible to swap load using `rebalanceBySwappingLoadIn` between an old broker and a new broker if the replica, which currently resides in the new broker, was originally located in the old broker.
Note that 
- `rebalanceByMovingLoadIn` (via `maybeApplyBalancingAction` > `getEligibleBrokers`) will already disallow moving replicas to an old broker if there are new brokers. After this feedback, to make this cutoff earlier, I copied the logic from within `getEligibleBrokers` to `rebalanceByMovingLoadIn`.
- `rebalanceBySwappingLoadIn` will allow swapping only if the conditions in the paragraph above are satisfied.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: This goal aims to keep the load on the broker between 10 and 30; hence, there is no logic that explicitly prefers going directly to 20 rather than 10 -- i.e. 20 is not considered to be more balanced compared to 10.

Also note that going to 10 (or 20) does not necessarily mean that the broker will preserve this as its final load at the end of the rebalance. During this balancing process, the balancing actions may change the load on a balanced broker _as long as they are not making the balance worse_. This means that a broker whose load was 10 at the end of optimizing the resource distribution goal, may have a load of 20 at the end of the optimization of succeeding goals. Also, note critically that the load on the broker can never go from 20 to 10 once it is balanced. Hence, it would arguably be even more preferable to have the initial goal balance not to set the load to 20 directly due to the additional constraints that it would bring to the succeeding goals -- i.e. cannot make the balance worse.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: I believe we agreed to remove `AnalysisInputException` -- not the `ModelInputException`.
But it makes sense to remove it as well.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -127,14 +184,27 @@ public ModelCompletenessRequirements clusterModelCompletenessRequirements() {
...
-    //Check that current destination would not become more unbalanced.
-    return isLoadUnderBalanceUpperLimitAfterChange(clusterModel, sourceReplica.load(), destinationBroker, ADD) &&
-        isLoadAboveBalanceLowerLimitAfterChange(clusterModel, sourceReplica.load(), sourceReplica.broker(), REMOVE);
+    switch (action.balancingAction()) {
+      case REPLICA_SWAP:
+        Replica destinationReplica = destinationBroker.replica(action.destinationTopicPartition());
+        double sourceUtilizationDelta = destinationReplica.load().expectedUtilizationFor(resource())
+                                        - sourceReplica.load().expectedUtilizationFor(resource());
+
+        return sourceUtilizationDelta != 0 && isLoadStillWithinLimitsAfterSwap(sourceReplica, destinationReplica);
```
becketqin: Not sure if this is right. If the upper limit is 30, lower limit is 10. source broker has utilization of 40 and destination broker has utilization of 3, when source delta is -5, are we going to reject the proposal?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -261,58 +331,75 @@ protected void rebalanceForBroker(Broker broker,
...
+    }
+
+    PriorityQueue<CandidateBroker> candidateBrokerPQ = new PriorityQueue<>();
+    // Sort the replicas initially to avoid sorting it every time.
+
+    double clusterUtilization = clusterModel.load().expectedUtilizationFor(resource()) / clusterModel.capacityFor(resource());
+    for (Broker candidate : clusterModel.healthyBrokers()) {
+      // Get candidate replicas on candidate broker to try moving load from -- sorted in the order of trial (descending load).
+      if (utilizationPercentage(candidate) > clusterUtilization) {
+        SortedSet<Replica> replicasToMoveIn = sortedCandidateReplicas(candidate, excludedTopics, 0, true);
```
becketqin: It is a little awkward that sometimes the boolean means isAscending and sometimes it means isDescending. Can we make them consistent?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -261,58 +331,75 @@ protected void rebalanceForBroker(Broker broker,
...
         Broker b = maybeApplyBalancingAction(clusterModel, replica, Collections.singletonList(broker), actionType, optimizedGoals);
         // Only need to check status if the action is taken. This will also handle the case that the source broker
         // has nothing to move in. In that case we will never reenqueue that source broker.
         if (b != null) {
-          if (isLoadAboveBalanceLowerLimit(clusterModel, broker)) {
+          if (isLoadAboveBalanceLowerLimit(broker)) {
             return false;
           }
+          // Identify the replica that the source broker has received.
+          receivedReplicas.add(replica);
```
becketqin: Why do we need to keep a set for this the received replicas? Can we remove the replicas immediately?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -261,58 +331,75 @@ protected void rebalanceForBroker(Broker broker,
...
+
           // If the source broker has a lower utilization than the next broker in the eligible broker in the queue,
           // we reenqueue the source broker and switch to the next broker.
-          if (!eligibleBrokers.isEmpty() &&
-              utilizationPercentage(sourceBroker) < utilizationPercentage(eligibleBrokers.peek())) {
-            eligibleBrokers.add(sourceBroker);
+          if (!candidateBrokerPQ.isEmpty() &&
+              utilizationPercentage(cb.broker()) < utilizationPercentage(candidateBrokerPQ.peek().broker())) {
+            if (actionType == REPLICA_MOVEMENT) {
+              candidateReplicasToReceive.removeAll(receivedReplicas);
```
becketqin: We are remove elements from set while iterating over it. Will this cause `ConcurrentModificationException`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -321,22 +408,210 @@ protected boolean rebalanceByMovingLoadIn(Broker broker,
...
+    });
+
+    // If the resource is NW_OUT, candidate replicas consider only the leaders -- i.e. only the leaders have NW_OUT load,
+    // otherwise all replicas on broker are considered.
+    Set<Replica> coveredReplicas = resource() == Resource.NW_OUT ? broker.leaderReplicas() : broker.replicas();
+
+    // The given load limit determines the lower cutoff in descending order, upper cutoff in ascending order.
+    if (isDescending) {
+      candidateReplicas.addAll(coveredReplicas.stream()
+                                              .filter(l -> (!excludedTopics.contains(l.topicPartition().topic()))
```
becketqin: We are using excludedTopics directly instead of shouldExclude(). Is this expected?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
```diff
@@ -240,6 +312,35 @@ private boolean legitMove(Replica replica, Broker destBroker, ActionType actionT
     return false;
   }
 
+  private SortedSet<Replica> getEligibleReplicas(ClusterModel clusterModel,
```
becketqin: getEligibleReplicasForSwap

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -462,11 +727,139 @@ private boolean isLoadUnderBalanceUpperLimitAfterChange(ClusterModel clusterMode
...
+  private boolean isGettingMoreBalanced(Replica sourceReplica, double sourceUtilizationDelta, double destinationBrokerUtilization) {
+    double sourceBrokerUtilization = sourceReplica.broker().load().expectedUtilizationFor(resource());
+
+    double prevDiff = sourceBrokerUtilization - destinationBrokerUtilization;
+    double nextDiff = prevDiff + (2 * sourceUtilizationDelta);
+
+    return Math.abs(nextDiff) < Math.abs(prevDiff);
+  }
+
+  private boolean isLoadStillWithinLimitsAfterSwap(Replica sourceReplica, Replica destinationReplica) {
```
becketqin: This can likely be merged with `isHostLoadStillWithinLimitsAfterSwap ` by using lambda.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -462,11 +727,139 @@ private boolean isLoadUnderBalanceUpperLimitAfterChange(ClusterModel clusterMode
...
+
+    return isBrokerAboveLowerLimit;
+  }
+
   /**
    * @param clusterModel the cluster topology and load.
    * @return the utilization upper threshold in percent for the {@link #resource()}
    */
-  protected double balanceUpperThreshold(ClusterModel clusterModel) {
+  private double balanceUpperThreshold(ClusterModel clusterModel) {
```
becketqin: Any reason to make these changes?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: We do _not_ reject the example above -- this method accepts it.
`isLoadStillWithinLimitsAfterSwap` does the same task as `isLoadUnderBalanceUpperLimitAfterChange` and `isLoadAboveBalanceLowerLimitAfterChange` for (1) the source and (2) destination brokers. In other words, it checks whether the load is above the upper limit for the broker after it would receive more load and checks whether the load is under the lower limit after the broker would gave away its load.

Would the change `isLoadStillWithinLimitsAfterSwap` --> `!isSwapBoostingLoadLimitViolation` make this functionality more clear?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: The loop is iterating over `candidateReplicasToReceive`. Removing replicas from this collection immediately would cause `ConcurrentModificationException`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: Added two other protected getters -- without computation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
```diff
@@ -52,32 +54,46 @@ public CapacityGoal() {
...
+   * {@link ActionAcceptance#REPLICA_REJECT} otherwise.
    */
   @Override
-  public boolean isActionAcceptable(BalancingAction action, ClusterModel clusterModel) {
+  public ActionAcceptance actionAcceptance(BalancingAction action, ClusterModel clusterModel) {
     Replica sourceReplica = clusterModel.broker(action.sourceBrokerId()).replica(action.topicPartition());
     Broker destinationBroker = clusterModel.broker(action.destinationBrokerId());
 
-    return isMovementAcceptableForCapacity(resource(), sourceReplica, destinationBroker);
+    if (action.balancingAction() == ActionType.REPLICA_SWAP) {
```
becketqin: Can we use switch statement here? It would be less error prone to make sure are not missing any case.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
```diff
@@ -349,32 +364,49 @@ private boolean isUtilizationOverLimit(Broker broker,
...
+   * goal, false otherwise
+   */
+  private boolean isSwapAcceptableForCapacity(Replica sourceReplica, Replica destinationReplica) {
+    double sourceReplicaUtilization = sourceReplica.load().expectedUtilizationFor(resource());
+    double destinationReplicaUtilization = destinationReplica.load().expectedUtilizationFor(resource());
+
+    double sourceUtilizationDelta = destinationReplicaUtilization - sourceReplicaUtilization;
+    return sourceUtilizationDelta > 0 ? isUtilizationUnderLimitAfterAddingLoad(sourceReplica.broker(), sourceUtilizationDelta)
+                                      : isUtilizationUnderLimitAfterAddingLoad(destinationReplica.broker(),
+                                                                                0 - sourceUtilizationDelta);
```
becketqin: No need to have 0.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderBytesInDistributionGoal.java
```diff
@@ -50,33 +52,56 @@ public LeaderBytesInDistributionGoal() {
...
   @Override
-  public boolean isActionAcceptable(BalancingAction action, ClusterModel clusterModel) {
+  public ActionAcceptance actionAcceptance(BalancingAction action, ClusterModel clusterModel) {
     Replica sourceReplica = clusterModel.broker(action.sourceBrokerId()).replica(action.topicPartition());
     Broker destinationBroker = clusterModel.broker(action.destinationBrokerId());
 
     initMeanLeaderBytesIn(clusterModel);
 
-    if (!sourceReplica.isLeader()) {
+    if (!sourceReplica.isLeader() && (action.balancingAction() != ActionType.REPLICA_SWAP
```
becketqin: Can we write the logic here using `switch(action)`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareGoal.java
```diff
@@ -71,11 +85,23 @@ public boolean isActionAcceptable(BalancingAction action, ClusterModel clusterMo
       // Remove brokers in partition broker racks except the brokers in replica broker rack.
       for (Broker broker : partitionBrokers) {
         if (broker.rack().brokers().contains(destinationBroker)) {
-          return false;
+          return BROKER_REJECT;
+        }
+      }
+      if (action.balancingAction() == ActionType.REPLICA_SWAP) {
+        // Destination broker cannot be in a rack that violates rack awareness.
```
becketqin: The logic here is a little confusing. If source and destination broker are in the same rack, we should allow swap instead of returning REPLICA_REJECT, right?

efeg: Simplified the logic.

PR #99, opened 2018-01-05
Excluded topics should not prevent brokers from being removed.

becketqin:


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -49,8 +49,6 @@
   // Flag to indicate whether the self healing failed to relocate all replicas away from dead brokers in its initial
   // attempt and currently omitting the resource balance limit to relocate remaining replicas.
   private boolean _selfHealingDeadBrokersOnly;
-  private Set<Integer> _brokerIdsAboveBalanceUpperLimit;
```
efeg: PR https://github.com/linkedin/cruise-control/pull/95 has a different fix for the same issue.
I think this one looks good, too!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -168,20 +161,28 @@ protected void initGoalState(ClusterModel clusterModel, Set<String> excludedTopi
    */
   @Override
   protected void updateGoalState(ClusterModel clusterModel, Set<String> excludedTopics) throws AnalysisInputException {
+    Set<Integer> brokerIdsAboveBalanceUpperLimit = new HashSet<>();
+    Set<Integer> brokerIdsUnderBalanceLowerLimit = new HashSet<>();
     // Log broker Ids over balancing limit.
     // While proposals exclude the excludedTopics, the balance still considers utilization of the excludedTopic replicas.
-    if (!_brokerIdsAboveBalanceUpperLimit.isEmpty()) {
+    for (Broker broker : clusterModel.healthyBrokers()) {
+      if (!isLoadUnderBalanceUpperLimitAfterChange(clusterModel, null, broker, REMOVE)) {
```
efeg: When `isLoadUnderBalanceUpperLimitAfterChange(...)` is called with `null` load, it checks whether the current load is under the balance upper limit or not. Hence, it is a little bit confusing to pass `REMOVE` as the `changeType` argument, which has no effect in the way this function works in this case. Can we maybe pass `null` as the change type to make it more clear that change is irrelevant? Or would a new function that specifically performs this check is more clear? 
-- We use this function the same way in one more place.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -168,20 +161,28 @@ protected void initGoalState(ClusterModel clusterModel, Set<String> excludedTopi
...
+        brokerIdsAboveBalanceUpperLimit.add(broker.id());
+      }
+      if (!isLoadAboveBalanceLowerLimitAfterChange(clusterModel, null, broker, ADD)) {
```
efeg: The issue with the use of a non-null change-type is similar to the one that I mentioned in `isLoadUnderBalanceUpperLimitAfterChange (...)` above.
-- We use this function the same way in one more place.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -259,14 +260,10 @@ protected void rebalanceForBroker(Broker broker,
...
     // Update broker ids over the balance limit for logging purposes.
-    if (requireLessLoad && rebalanceByMovingLoadOut(broker, clusterModel, optimizedGoals,
-                                                    REPLICA_MOVEMENT, excludedTopics)) {
-      _brokerIdsAboveBalanceUpperLimit.add(broker.id());
-      LOG.debug("Failed to balance {} for broker {} with replica and leader movements to reduce load.", resource(), broker.id());
-    } else if (requireMoreLoad && rebalanceByMovingLoadIn(broker, clusterModel, optimizedGoals,
-                                                          REPLICA_MOVEMENT, excludedTopics)) {
-      _brokerIdsUnderBalanceLowerLimit.add(broker.id());
-      LOG.debug("Failed to balance {} for broker {} with replica and leader movements to increase load.", resource(), broker.id());
+    if (requireLessLoad) {
```
efeg: Is the `else` case of these conditions executed when we actually expect it to be executed?
I am having difficulty in simulating a case where the resource is successfully balanced by moving leaders and replicas.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/RandomSelfHealingTest.java
```diff
@@ -83,38 +85,55 @@
...
-          entry.getValue()), balancingConstraint};
+          entry.getValue()), balancingConstraint, Collections.emptySet()};
       params.add(singleDeadSingleSoftParams);
+      Object[] singleDeadSingleSoftParamsWithExcludedTopics = 
+          {singleDeadBroker, Collections.singletonMap(entry.getKey(), entry.getValue()), balancingConstraint, 
+          Collections.singleton("T0")};
+      params.add(singleDeadSingleSoftParamsWithExcludedTopics);
     }
-    props.setProperty(KafkaCruiseControlConfig.MAX_REPLICAS_PER_BROKER_CONFIG, Long.toString(5000L));
+    props.setProperty(KafkaCruiseControlConfig.MAX_REPLICAS_PER_BROKER_CONFIG, Long.toString(5100L));
```
efeg: Was 5K causing test failures?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderBytesInDistributionGoal.java
```diff
@@ -139,7 +139,7 @@ protected void rebalanceForBroker(Broker broker, ClusterModel clusterModel, Set<
 
     List<Replica> leaderReplicasSortedByBytesIn = broker.replicas().stream()
         .filter(r -> r.isLeader())
-        .filter(r -> !excludedTopics.contains(r.topicPartition().topic()))
+        .filter(r -> !shouldExclude(r, excludedTopics))
```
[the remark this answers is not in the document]
becketqin: This is worth discussing. The reason we are also excluding the leaders was because applications such as kafka-monitor needs to have full coverage of all the brokers. So we should not move leaders away from a broker for those topics.

So the meaning for excluded topics is the following:
1. When there is no dead brokers, those topics will not be touched by CC at all, whether for leader movement or replica movement.
2. When there is a dead broker, CC will only make necessary movements for the replicas of those excluded topics that are impacted by the broker failure.

In this particular case, if the original broker is dead, the replica/leader has to be moved anyways, so CC will not exclude those replicas.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/RandomSelfHealingTest.java
becketqin: I have seen the ReplicaCapacityGoal been failing from time to time even without this patch. We need to dig a bit further to see why that is the case. This change did not solve the problem so we don't really need it.

efeg:
LGTM.

PR #100, opened 2018-01-11
Add asynchronous operation support to KafkaCruiseControl

becketqin:
Also added HttpSession support for KafkaCruiseControlServlet.


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManagerTest.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.List;
+import java.util.concurrent.Future;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpSession;
+import org.apache.kafka.common.utils.MockTime;
+import org.apache.kafka.common.utils.Time;
+import org.easymock.EasyMock;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
efeg: Should we explicitly define the imported class rather than using the wildcard?

PR #118, opened 2018-01-30
Address the case that partitions of the same topic can have different number of replicas.

becketqin:


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/kafkaassigner/KafkaAssignerEvenRackAwareGoal.java
```diff
@@ -112,11 +112,10 @@ public boolean optimize(ClusterModel clusterModel, Set<Goal> optimizedGoals, Set
     Map<String, Integer> replicationFactorByTopic = clusterModel.replicationFactorByTopic();
```
efeg: We should delete `replicationFactorByTopic`, because it is not used after the patch.

Issue #138, opened 2018-02-14
"completed bytes" is not accurate in Executor logs and verbose state response


efeg:
The `completed bytes` presented in CC logs and verbose state response while replica moves are in progress is not accurate. This metric does not include the in progress bytes movement. It only considers bytes in pending tasks. Hence, it underestimates the real value.

efeg:
Fixed in PR https://github.com/linkedin/cruise-control/pull/152.

Issue #141, opened 2018-02-27
decommission broker doesn't work


chandradeepak:
I installed cruise control . I was trying to remove a broker but the the partitions were never moved to other brokers .

curl --globoff -X POST http://localhost:9090/kafkacruisecontrol/remove_broker?brokerid=1004&&dryrun=true&throttle_removed_broker=true&kafka_assigner=true&json=true.

i was checking my kafka manager to see if the partitions moved from that broker to other brokers it never happens. 

I was looking at the code is it because the goals are empty. But the documentation didn't mention any thing about goals. I even tried putting some goals but that didn't work. 

can you please suggest what is the correct way to decommission a broker. 

`private boolean addOrRemoveBroker(HttpServletRequest request, HttpServletResponse response, EndPoint endPoint)
      throws Exception {
    List<Integer> brokerIds = new ArrayList<>();
    boolean dryrun;
    DataFrom dataFrom;
    boolean throttleAddedOrRemovedBrokers;
    List<String> goals;
    boolean json = wantJSON(request);
    try {
      String[] brokerIdsString = request.getParameter(BROKER_ID_PARAM).split(",");
      for (String brokerIdString : brokerIdsString) {
        brokerIds.add(Integer.parseInt(brokerIdString));
      }

      dryrun = getDryRun(request);
      goals = getGoals(request);
      dataFrom = getDataFrom(request);

      String throttleBrokerString = endPoint == EndPoint.ADD_BROKER ?
          request.getParameter(THROTTLE_ADDED_BROKER_PARAM) : request.getParameter(THROTTLE_REMOVED_BROKER_PARAM);
      throttleAddedOrRemovedBrokers = throttleBrokerString == null || Boolean.parseBoolean(throttleBrokerString);

    } catch (Exception e) {
      StringWriter sw = new StringWriter();
      e.printStackTrace(new PrintWriter(sw));
      setErrorResponse(response, sw.toString(), e.getMessage(), SC_BAD_REQUEST, json);
      // Close session
      return true;
    }
    GoalsAndRequirements goalsAndRequirements = getGoalsAndRequirements(request, response, goals, dataFrom, false);
    if (goalsAndRequirements == null) {
      return false;
    }
    // Get proposals asynchronously.
    GoalOptimizer.OptimizerResult optimizerResult;
    if (endPoint == EndPoint.ADD_BROKER) {
      optimizerResult =
          getAndMaybeReturnProgress(request, response,
[...]

PR #152, opened 2018-03-03
Clean executionTaskManager upon detecting an ongoing execution not started by CC.

efeg:
* Fixes the inconsistent state upon rebooted CC instance encountered in in issue https://github.com/linkedin/cruise-control/issues/141. Does not set the `ExecutorState` back to the `STOPPING_EXECUTION`, instead clears the execution task manager state.
* Make Executor thread-safe.
* This patch also fixes the Issue https://github.com/linkedin/cruise-control/issues/138.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -113,12 +113,19 @@ public ExecutorState state() {
   }
 
   /**
-   * Kick off the execution.
+   * Pause the load monitor and kick off the execution.
+   *
+   * @param loadMonitor Load monitor.
    */
   public void startExecution(LoadMonitor loadMonitor) {
+    if (loadMonitor != null) {
```
becketqin: If we do not expect the load monitor to be null, can we throw `IllegalArgumentException` and pass a mock object in the unit test?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -381,8 +381,7 @@ public void resumeLoadMonitorActivity() {
    */
   public void executeProposals(Collection<ExecutionProposal> proposals,
                                Collection<Integer> unthrottledBrokers) {
-    // Pause the load monitor before the proposal execution.
-    _loadMonitor.pauseMetricSampling();
+    // Add execution proposals and start execution.
```
becketqin: Looking at the code. If there are two threads calling executeProposals() with two different set of proposals, CC may execute the both sets of proposals. Maybe we should just let the executor expose a method of `executeProposals()` and synchronize on that.

efeg: Made the Executor thread-safe.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregatorState.java
```diff
@@ -27,7 +27,7 @@
 
   MetricSampleAggregatorState(long generation, long windowMs, int completenessCacheSize) {
     super(generation);
-    // Only keep 5 completeness caches.
+    // Only keep as much as _windowStates.size() completeness caches.
```
becketqin: Only keep as many as completenessCacheSize entries.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
```diff
@@ -321,11 +323,82 @@ private void completeTask(ExecutionTask task) {
...
-  public void clear() {
+  public synchronized void clear() {
     _brokersToSkipConcurrencyCheck.clear();
     _inProgressReplicaMovementsByBrokerId.clear();
     _inProgressPartitions.clear();
     _executionTaskPlanner.clear();
     _executionTaskTracker.clear();
   }
+
+  public synchronized ExecutionState getExecutionState() {
```
becketqin: Nit: we have already got many "State". Can we have something like ExecutionTasksSummary here? Also the convention in Kafka, as well as KafkaCruiseControl, is to avoid using "get" in the getter method.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
```diff
@@ -321,11 +323,82 @@ private void completeTask(ExecutionTask task) {
...
+    return new ExecutionState(_executionTaskPlanner.remainingReplicaMovements(),
+                              _executionTaskTracker.tasksInState(ExecutionTask.State.IN_PROGRESS),
+                              _executionTaskTracker.tasksInState(ExecutionTask.State.ABORTING),
+                              _executionTaskTracker.tasksInState(ExecutionTask.State.ABORTED),
+                              _executionTaskTracker.tasksInState(ExecutionTask.State.DEAD),
+                              _executionTaskPlanner.remainingDataToMoveInMB());
+  }
+
+  public static class ExecutionState {
```
becketqin: Would package private be sufficient?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -68,107 +73,140 @@ public Executor(KafkaCruiseControlConfig config, Time time, MetricRegistry dropw
...
-                                      _executionTaskManager.remainingPartitionMovements(),
-                                      _executionTaskManager.inProgressTasks(),
-                                      _executionTaskManager.abortingTasks(),
-                                      _executionTaskManager.abortedTasks(),
-                                      _executionTaskManager.deadTasks(),
-                                      _executionTaskManager.remainingDataToMoveInMB(),
-                                      _finishedDataMovementInMB);
-      default:
-        throw new IllegalStateException("Should never be here!");
+    _stateLock.lock();
```
becketqin: The locking becomes a little ugly here. Given that the main thing we want to protect is essentially the state query consistency, maybe we can just create an executor state cache object. The state cache will only be updated by the ExecutionRunnable atomically. Hence we don't need to worry about the consistency.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
efeg: Renamed: `getExecutionState` -> `getExecutionTasksSummary`.
Since this method does not just return an existing object `x`, but building one within this method, it is named `getX` instead of `x`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -180,58 +182,94 @@ public void run() {
...
 
     /**
-     * Start the actual execution of the proposals.
+     *
+     *
+     * @param expectedState Current expected state.
+     * @param nextState Next state to be in if the current state is the given expected state.
+     * @param moveRunnable If the current state is the given expected state, runnable to execute after state transition.
+     */
+    private void maybeMove(ExecutorState.State expectedState,
```
becketqin: I am wondering if this separate method is necessary. The previous logic seem simpler and easier to understand. Is there reason of having this separate method?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -180,58 +182,94 @@ public void run() {
...
+        // Wait indefinitely for partition movements to finish.
         waitForExecutionTaskToFinish();
         partitionsToMove = _executionTaskManager.remainingPartitionMovements().size();
         long dataToMove = _executionTaskManager.remainingDataToMoveInMB();
-        _numFinishedPartitionMovements =
-            numTotalPartitionMovements - partitionsToMove - _executionTaskManager.inExecutionTasks().size();
-        _finishedDataMovementInMB = totalDataToMoveInMB - dataToMove;
+        Set<ExecutionTask> inExecutionTasks = _executionTaskManager.inExecutionTasks();
+        _numFinishedPartitionMovements = numTotalPartitionMovements - partitionsToMove - inExecutionTasks.size();
+        _finishedDataMovementInMB = totalDataToMoveInMB - dataToMove - inExecutionDataToMoveInMB(inExecutionTasks);
```
becketqin: `inExecutionDataToMoveInMB()` may be invoked for N^2 times which could be a waste. Can we just maintain a variable in `ExecutionTaskManager`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -180,58 +182,94 @@ public void run() {
...
+    /**
+     * Get the in execution data to move in MB.
+     * @param inExecutionTasks Tasks that are either in progress or aborting.
+     * @return the in execution data to move in MB.
+     */
+    private long inExecutionDataToMoveInMB(Set<ExecutionTask> inExecutionTasks) {
+      return inExecutionTasks.stream().mapToLong(task -> task.proposal().dataToMoveInMB()).sum();
+    }
+
     private void moveReplicas() {
```
becketqin: I did not see the `_executorState` get updated when move the partitions. Am I missing something?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -242,27 +280,33 @@ private void moveReplicas() {
...
+                 executionState.remainingPartitionMovements().size(),
+                 executionState.abortingTasks(),
+                 executionState.abortedTasks().size(),
+                 executionState.deadTasks().size(),
+                 executionState.remainingDataToMoveInMB());
       }
     }
 
-    private void moveLeaders() {
+    private void moveLeaderships() {
```
becketqin: Orthogonal to this patch and not urgent. We may want to expose the leadership movement to the users as well. This could be a separate patch, though.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: Executor state is updated in `maybeMove()`.

becketqin: But during the movement, the executor state may change, right? i.e. the in-progress tasks, completed tasks, etc.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: Thanks for the feedback!
Is this ticket relevant to your request -> https://github.com/linkedin/cruise-control/issues/159, or is this some other functionality?

becketqin: It is something different. I was thinking about the leadership movement progress.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -242,31 +272,40 @@ private void moveReplicas() {
...
       }
       if (_executionTaskManager.inProgressTasks().isEmpty()) {
         LOG.info("Partition movements finished.");
-      } else if (_stopRequested) {
-        LOG.info("Partition movements stopped. {} in progress, {} pending, {} aborted, {} dead.",
-                 _executionTaskManager.inProgressTasks().size(),
-                 _executionTaskManager.remainingPartitionMovements().size(),
-                 _executionTaskManager.abortedTasks().size(),
-                 _executionTaskManager.deadTasks().size());
+      } else if (_stopRequested.get()) {
```
becketqin: What if the _executionTaskManager does not have more inProgressTasks, but stopRequested == true? It looks we are going to start to move the leaders in that case?

I think we need to updateOngoingExecutionState in the following places:
1. Within waitForExecutionTaskToFinish();
2. Each time after waitForExecutionTaskToFinish() returns;
3. After moveLeaderships() and moveReplicas();

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: Yes, that's definitely something we should add. I will create an issue about it for tracking purposes.

PR #154, opened 2018-03-06
Add a parameter to enforce a limit on the number of entries in partition load response

efeg:
For example, `partition_load?resource=CPU&entries=10` will return the top 10 entries.


becketqin:
Just a reminder. We also need to update the wiki page.

Issue #136, opened 2018-02-14
Load request occasionally throws java.lang.IllegalStateException: Response is committed


efeg:
A `load` request succeeds, but may occasionally throw `java.lang.IllegalStateException`, which can be seen in logs. This has no adverse effects in terms of the execution of the request -- i.e. it still responds with the accurate load information.
```
ERROR [KafkaCruiseControlServlet] [qtp711540569-235071] [kafka-cruise-control] [] Error processing GET request '/load'
java.lang.IllegalStateException: Response is committed
        at org.eclipse.jetty.server.Request.getSession(Request.java:1523) ~[?:?]
        at org.eclipse.jetty.server.Request.getSession(Request.java:1501) ~[?:?]
        at com.linkedin.kafka.cruisecontrol.servlet.SessionManager.closeSession(SessionManager.java:159) ~[kafka-cruise-control-0.1.83.jar:?]
        at com.linkedin.kafka.cruisecontrol.servlet.KafkaCruiseControlServlet.doGet(KafkaCruiseControlServlet.java:283) ~[kafka-cruise-control-0.1.83.jar:?]
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:687) ~[javax.servlet-api-3.1.0.jar:3.1.0]
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:790) ~[javax.servlet-api-3.1.0.jar:3.1.0]
        at org.xeril.util.servlet.ProxyServlet.service(ProxyServlet.java:61) ~[util-servlet-15.0.7.jar:?]
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:790) ~[javax.servlet-api-3.1.0.jar:3.1.0]
        at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:837) ~[jetty-servlet-9.3.11.v20160721.jar:9.3.11.v20160721]
        at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:583) ~[jetty-servlet-9.3.11.v20160721.jar:9.3.11.v20160721]
        at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:143) ~[jetty-server-9.3.11.v20160721.jar:9.3.11.v20160721]
        at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:548) ~[jetty-security-9.3.11.v20160721.jar:9.3.11.v20160721]
[...]

PR #157, opened 2018-03-09
Fix response is committed error upon closing a session.

efeg:
* Fix response is committed error upon closing a session (Issue https://github.com/linkedin/cruise-control/issues/136).

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
@@ -156,7 +156,8 @@ public class SessionManager {
    * @param request the request to close its session.
    */
   synchronized void closeSession(HttpServletRequest request) {
-    HttpSession session = request.getSession();
+    // Response associated with this request has already been flushed; hence, do not attempt to create a new session.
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -156,7 +156,8 @@ synchronized int numSessions() {
    * @param request the request to close its session.
    */
   synchronized void closeSession(HttpServletRequest request) {
-    HttpSession session = request.getSession();
+    // Response associated with this request has already been flushed; hence, do not attempt to create a new session.
+    HttpSession session = request.getSession(false);
```
becketqin: The reason we are trying to get a session was because we need to remove it from the `_inProgressSessions`.  If the session returns null here we will only remove the session after it expires. An alternative is to store a request to session mapping so we don't need to use getSession(). That said, it would be good to see if we can figure the timing out.

efeg: Tests in local deployment has shown that the `getSession` will return the session even after the flush; hence, `_inProgressSessions` can remove the returned session.

PR #181, opened 2018-04-01
Simplifying the KafkaAssignerDiskUsageDistributionGoal logic.

becketqin:
1. Avoid unnecessary sort of the replicas in the brokers.
2. Use TreeSet instead of implementing binary search on list.
3. Speed up the test by avoid unnecessary sort. After the change the test time reduced from 20 min to 11 min.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
@@ -23,7 +23,6 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
-import java.util.NavigableSet;
 import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
 import java.util.HashSet;
 import java.util.List;
@@ -60,7 +59,6 @@ public class ReplicaDistributionGoal extends AbstractGoal {
   private boolean _selfHealingDeadBrokersOnly;
   private final Set<Integer> _brokerIdsAboveBalanceUpperLimit;
   private final Set<Integer> _brokerIdsUnderBalanceLowerLimit;
-  private NavigableSet<BrokerAndSortedReplicas> _allBrokerAndReplicas;
   private Map<Integer, BrokerAndSortedReplicas> _brokerAndReplicasMap;
   private double _avgReplicasOnHealthyBroker;
   private double _balanceUpperLimit;
@@ -197,16 +195,11 @@ public class ReplicaDistributionGoal extends AbstractGoal {
       LOG.warn("All replicas are excluded from {}.", name());
     }
 
-    _allBrokerAndReplicas = new TreeSet<>((bas1, bas2) -> {
-      int result = Double.compare(bas2.broker().replicas().size(), bas1.broker().replicas().size());
-      return result == 0 ? Integer.compare(bas1.broker().id(), bas2.broker().id()) : result;
-    });
     _brokerAndReplicasMap = new HashMap<>();
 
-    for (Broker healthyBroker : clusterModel.brokers()) {
-      BrokerAndSortedReplicas bas = new BrokerAndSortedReplicas(healthyBroker, healthyBroker.replicaComparator(DISK));
-      _allBrokerAndReplicas.add(bas);
-      _brokerAndReplicasMap.put(healthyBroker.id(), bas);
+    for (Broker broker : clusterModel.brokers()) {
+      BrokerAndSortedReplicas bas = new BrokerAndSortedReplicas(broker, broker.replicaComparator(DISK));
+      _brokerAndReplicasMap.put(broker.id(), bas);
     }
 
     _selfHealingDeadBrokersOnly = false;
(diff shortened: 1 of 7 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/internals/BrokerAndSortedReplicas.java
```diff
@@ -0,0 +1,63 @@
...
+  public BrokerAndSortedReplicas(Broker broker, Comparator<Replica> comparator) {
+    _broker = broker;
+    sortedReplicas = new TreeSet<>((r1, r2) -> {
+      int result = comparator.compare(r1, r2);
+      return result == 0 ? r1.compareTo(r2) : result;
+    });
+    sortedReplicas.addAll(broker.replicas());
+  }
+
+  public BrokerAndSortedReplicas(Broker broker, Set<String> exclucedTopics, Comparator<Replica> comparator) {
```
efeg: `exclucedTopics ` --> `excludedTopics`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
```diff
@@ -54,6 +60,8 @@
   private boolean _selfHealingDeadBrokersOnly;
   private final Set<Integer> _brokerIdsAboveBalanceUpperLimit;
   private final Set<Integer> _brokerIdsUnderBalanceLowerLimit;
+  private NavigableSet<BrokerAndSortedReplicas> _allBrokerAndReplicas;
```
efeg: Is `_allBrokerAndReplicas` used purely for sorting the elements of `_brokerAndReplicasMap`? If so, can we indicate this via a comment? Currently it is not obvious why we don't read this variable.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
```diff
@@ -188,6 +196,19 @@ protected void initGoalState(ClusterModel clusterModel, Set<String> excludedTopi
...
       LOG.warn("All replicas are excluded from {}.", name());
     }
+
+    _allBrokerAndReplicas = new TreeSet<>((bas1, bas2) -> {
+      int result = Double.compare(bas2.broker().replicas().size(), bas1.broker().replicas().size());
+      return result == 0 ? Integer.compare(bas1.broker().id(), bas2.broker().id()) : result;
+    });
+    _brokerAndReplicasMap = new HashMap<>();
+
+    for (Broker healthyBroker : clusterModel.brokers()) {
```
efeg: This can include healthy and dead brokers. Why is it named `healthyBroker`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/kafkaassigner/KafkaAssignerDiskUsageDistributionGoal.java
```diff
@@ -37,8 +44,8 @@
 public class KafkaAssignerDiskUsageDistributionGoal implements Goal {
   private static final Logger LOG = LoggerFactory.getLogger(KafkaAssignerDiskUsageDistributionGoal.class);
   private static final double BALANCE_MARGIN = 0.9;
-  // If broker disk usage differ less than BROKER_EQUALITY_DELTA, consider them as equal -- i.e. no swap.
-  private static final double BROKER_EQUALITY_DELTA = 1.0;
+  // If broker disk usage differ less than USAGE_EQUALITY_DELTA, consider them as equal -- i.e. no swap.
+  private static final double USAGE_EQUALITY_DELTA = 0.0001;
```
efeg: While it is desirable to achieve an even balance, I am worried that this level of requirement will lead to excessive swaps. Do we really need up to 100-bytes evenness in disk usage? I would propose to apply a subset of the following 1) increase `USAGE_EQUALITY_DELTA`, 2) set a timeout, 3) set a limit on the number of iterations.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/kafkaassigner/KafkaAssignerDiskUsageDistributionGoal.java
```diff
@@ -160,39 +172,52 @@ private boolean isOptimized(ClusterModel clusterModel, double upperThreshold, do
...
+                                   BrokerAndSortedReplicas toOptimize,
                                    ClusterModel clusterModel,
                                    double meanDiskUsage,
                                    double lowerThreshold,
                                    double upperThreshold,
                                    Set<String> excludedTopics) {
     LOG.trace("Optimizing broker {}. BrokerDiskUsage = {}, meanDiskUsage = {}",
-              broker, diskUsage(broker), meanDiskUsage);
-    double brokerDiskUsage = diskUsage(broker);
+              toOptimize.broker(), diskUsage(toOptimize.broker()), meanDiskUsage);
```
efeg: Overall in this class, it would be good to ensure that each double is printed up to the level of desired decimal digits.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/internals/BrokerAndSortedReplicas.java
becketqin: Ah, this constructor should be removed. The excluded topics themselves is not sufficient to actually exclude topics, we also need to check if the broker is dead or not. Currently this constructor is not used anyways.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
becketqin: Good point. It looks we don't need this variable at all.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/kafkaassigner/KafkaAssignerDiskUsageDistributionGoal.java
becketqin: The usage is in percentage. So 0.0001 of a 1TB disk would be about 100MB difference.

PR #229, opened 2018-05-15
Fix metric anomaly detection.

efeg:
* Fix the broken date format in anomaly descriptions.
* Add the number of entries in the history to the anomaly description.
* Add metric anomaly upper and lower margin.


cruise-control-core/src/main/java/com/linkedin/cruisecontrol/detector/metricanomaly/PercentileMetricAnomalyFinderConfig.java
```diff
@@ -18,24 +18,51 @@
...
-      "The lower threshold for the metric anomaly detector to identify a decrease in the metric "
-          + "values of a broker as a metric anomaly.";
+      "The lower percentile threshold for the metric anomaly detector to identify a decrease in the metric "
+      + "values of a broker as a metric anomaly. The current metric value is compared against the historical value "
+      + "corresponding to given percentile in the metric history after the application of the lower margin.";
+
+  /**
+   * <code>metric.anomaly.upper.margin</code>
+   */
+  public static final String METRIC_ANOMALY_UPPER_MARGIN_CONFIG = "metric.anomaly.upper.margin";
```
gitlw: The docs becomes hard to understand after adding the margin values. I think it'll be useful to give an example and have the docs reference the example.

PR #232, opened 2018-05-17
Shuffle the order of goals used by precomputation threads

kidkun:
This patch adds the feature of generating a set of goals with different priority order when there are multiple proposal precomputing threads. Each of the threads will pick up one unique priority order and calculate based on that. In this case we can potentially get better optimizing result.


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,57 @@
...
+    balancingConstraint.setCapacityThreshold(TestConstants.MEDIUM_CAPACITY_THRESHOLD);
+
+
+    System.out.println("property applied is " + props);
+    GoalOptimizer goalOptimizer = new GoalOptimizer(new KafkaCruiseControlConfig(balancingConstraint.setProps(props)),
+        null,
+        new SystemTime(),
+        new MetricRegistry());
+    try {
+      Field field = goalOptimizer.getClass().getDeclaredField("_goalByPriorityForPrecomputing");
```
efeg: Let's avoid using reflection. We can add a package private getter for the unit test.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,57 @@
...
+      List<SortedMap<Integer, Goal>> randomizedGoal = (List<SortedMap<Integer, Goal>>) field.get(goalOptimizer);
+      System.out.println("generated goals are" + randomizedGoal);
+      for (int i = 0; i < randomizedGoal.size() - 1; i++) {
+        for (int j = i + 1; j < randomizedGoal.size(); j++) {
+          assertNotEquals(randomizedGoal.get(i), randomizedGoal.get(j));
+        }
+      assertEquals(randomizedGoal.size(), 5);
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
```
efeg: Do we need to print a stacktrace in unit test?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,57 @@
...
+        new SystemTime(),
+        new MetricRegistry());
+    try {
+      Field field = goalOptimizer.getClass().getDeclaredField("_goalByPriorityForPrecomputing");
+      field.setAccessible(true);
+      List<SortedMap<Integer, Goal>> randomizedGoal = (List<SortedMap<Integer, Goal>>) field.get(goalOptimizer);
+      System.out.println("generated goals are" + randomizedGoal);
+      for (int i = 0; i < randomizedGoal.size() - 1; i++) {
+        for (int j = i + 1; j < randomizedGoal.size(); j++) {
+          assertNotEquals(randomizedGoal.get(i), randomizedGoal.get(j));
```
efeg: Currently we seem to be comparing whether objects are equal or not -- i.e. not the content of `SortedMap`s. Even if different sortedmaps would have the same content, they would pass this test.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+      assertEquals(randomizedGoal.size(), 5);
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
```
efeg: The unit test should also check whether the original priority order preserved.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -117,6 +118,37 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
...
+      if (i == 0) {
+        //Guarantee one thread is working on the original goal priority
+        _goalByPriorityForPrecomputing.add(_goalsByPriority);
+        shuffledGoalSet.add(new ArrayList(_goalsByPriority.values()));
+      } else {
+        List<Goal> shuffledGoal = new ArrayList(_goalsByPriority.values());
+        //Ensure the shuffled goal is unique
+        do {
+          Collections.shuffle(shuffledGoal);
+        } while (! shuffledGoalSet.add(shuffledGoal));
```
efeg: This loop might take undeterministic time to complete -- potentially infinite loop! Let's avoid generating the random values one-by-one on the fly and comparing them for randomness. We might want to use a list shuffled in a deterministic order. I would recommend another algorithm for making this shuffle more efficiently.

In general, our goal is to:
1. Ensure that selected permutations include the original order -- e.g. `0`, `1`, ..., `defaultGoal.size() - 1`.
2. Generate (a subset of) permutations of `shuffledGoal`
3. Pick as many as `_numPrecomputingThreads` elements from possible permutations.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+@RunWith(Parameterized.class)
+public class GoalShuffleTest {
+
+  @Parameterized.Parameters
+  public static Collection<Object[]> data() {
+    return Arrays.asList(new Object[][]{{0, 2}, {1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}, {6, 6}, {7, 6}});
+  }
+
+  private int _numPrecomuptingThreadCfg;
```
efeg: `_numPrecomuptingThreadCfg` --> `_numPrecomputingThreadCfg`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,85 @@
...
+  private int _numPrecomuptingThreadExp;
```
efeg: `_numPrecomuptingThreadExp` -> `_numPrecomputingThreadExp`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+
+  public GoalShuffleTest(int threadNumCfg, int threadNumExp) {
+    this._numPrecomuptingThreadCfg = threadNumCfg;
+    this._numPrecomuptingThreadExp = threadNumExp;
```
efeg: let's avoid using `this.` in constructor.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,85 @@
...
+  @Parameterized.Parameters
+  public static Collection<Object[]> data() {
+    return Arrays.asList(new Object[][]{{0, 2}, {1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}, {6, 6}, {7, 6}});
+  }
+
+  private int _numPrecomuptingThreadCfg;
+  private int _numPrecomuptingThreadExp;
+
+
+  public GoalShuffleTest(int threadNumCfg, int threadNumExp) {
```
efeg: We typically use the same argument name as the private variable name  -- i.e. `numPrecomuptingThreadCfg` and `numPrecomuptingThreadExp`. Also let's use the full name for each word -- e.g. not `Cfg` `Config`.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,85 @@
...
+        .add(ReplicaCapacityGoal.class.getName())
+        .add(DiskCapacityGoal.class.getName()).toString());
+    BalancingConstraint balancingConstraint = new BalancingConstraint(new KafkaCruiseControlConfig(props));
+    balancingConstraint.setResourceBalancePercentage(TestConstants.LOW_BALANCE_PERCENTAGE);
+    balancingConstraint.setCapacityThreshold(TestConstants.MEDIUM_CAPACITY_THRESHOLD);
+    GoalOptimizer goalOptimizer = new GoalOptimizer(new KafkaCruiseControlConfig(balancingConstraint.setProps(props)),
+        null,
+        new SystemTime(),
+        new MetricRegistry());
+    List<SortedMap<Integer, Goal>> randomizedGoal = goalOptimizer.getgoalByPriorityForPrecomputing();
```
efeg: `randomizedGoal` --> `_goalByPriorityForPrecomputing`.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+    //Check correct number of goal priority is generated
```
efeg: 1) Let's always use space after `//`.
2) `Check correct number` -> `Check whether the correct number`

-- Note that there are other instances of this use in the same test class.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.runners.Parameterized;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
+
+@RunWith(Parameterized.class)
+public class GoalShuffleTest {
+
+  @Parameterized.Parameters
+  public static Collection<Object[]> data() {
+    return Arrays.asList(new Object[][]{{0, 2}, {1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}, {6, 6}, {7, 6}});
```
efeg: Do we need to test the similar use-case multiple times? -- e.g. is there a difference between testing `{3, 3}`, `{4, 4}`, or `{5, 5}`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
kidkun: I have thought about this when I was writing the code. My concern was we might lose the chance of finding some better goal priority order(or some new findings) if we do this in a deterministic way. 
And in real application scenario, the number of priority combination should easily be much larger than the number of thread we config to do precomputing, so the possibility of a conflict out of randomization will be small and we can generate shuffled pattern in a reasonable time.
I will change the code to generate permutation in deterministic way.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalShuffleTest.java
kidkun: it is just the config is different. they should falls in one group. since there are only 7 combinations and it runs fast. I think there is no harm to do this.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -117,6 +118,37 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
     _proposalComputationTimer = dropwizardMetricRegistry.timer(MetricRegistry.name("GoalOptimizer", "proposal-computation-timer"));
   }
 
+  /**
+   * Generate unique goal priority for each of proposal precomupting thread
+   */
+  private void populateGoalByPriorityForPrecomputing() {
+    Set<List<Goal>> shuffledGoalSet = new HashSet<>();
+    //Need at least one computing thread
+    for (int i = 0; i < numProposalComputingThreads(); i++) {
```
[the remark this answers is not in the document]
kidkun: Hi Efe, I think you are right. if _numPrecomputingThreads equals 0, we should only generate one goalByPriority.
Actually I am confused by the numProposalComputingThreads(). based on its comments,

 At least two computing thread is needed if precomputing is disabled. One thread for submitting and waiting for the proposal computing to finish, another one for compute the proposals.

when precomputing is disabled. in GoalOptimizer.computeBestProposal().

    for (int i = 0; i < (_numPrecomputingThreads > 0 ? _numPrecomputingThreads : 1); i++) {
      futures.add(_proposalPrecomputingExecutor.submit(new ProposalCandidateComputer(_goalByPriorityForPrecomputing.get(i))));
    }

does this mean that the actual thread for submitting and waiting for the proposal computing to finish is the main thread instead of one of thread managed by       _proposalPrecomputingExecutor?

I tried to change numProposalComputingThreads do
  private int numProposalComputingThreads() {
    return _numPrecomputingThreads > 0 ? _numPrecomputingThreads : 1;
  }
and is still able to pass all the tests.
I am going to have this in the PR. Please correct me if I am wrong.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/AnalyzerUtils.java
```diff
@@ -276,5 +276,23 @@ public static double nanToZero(double v) {
       }
   }
 
+  /**
+   * Calculate the factorial of the given non-negative value.
+   *
+   * @param v The value for which the factorial will be calculated.
+   * @return the factorial of the given non-negative value.
+   */
+
```
efeg: Redundant line before the function after the JavaDoc.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -117,6 +117,52 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
...
+  /**
+   * Generate unique goal priority for each of proposal precomputing thread.
+   */
+  private void populateGoalByPriorityForPrecomputing() {
+    List<List<Goal>> shuffledGoalList = new ArrayList<>();
+    // Guarantee one thread is working on the original goal priority.
+    shuffledGoalList.add(new ArrayList<>(_goalsByPriority.values()));
+
+    Integer goalCount = _goalsByPriority.values().size();
+    Integer goalIndex = goalCount - 2;
```
efeg: Why do we subtract `2`? Let's add a comment to clarify that.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -117,6 +117,52 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
...
+    Integer numShuffledGoalToGenerate = numProposalComputingThreads() - 1;
```
efeg: Should it be `numShuffledGoalsToGenerate`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -157,10 +203,9 @@ public void run() {
     }
   }
 
-  // At least two computing thread is needed if precomputing is disabled. One thread for submitting and waiting for
-  // the proposal computing to finish, another one for compute the proposals.
+  // If precomputing is disabled, need one thread to compute the proposals.
   private int numProposalComputingThreads() {
-    return _numPrecomputingThreads > 0 ? _numPrecomputingThreads : 2;
+    return _numPrecomputingThreads > 0 ? _numPrecomputingThreads : 1;
```
efeg: This is a special case for handling if precomputing is disabled: At least two computing thread is needed if precomputing is disabled. One thread for submitting and waiting for the proposal computing to finish, another one for compute the proposals.

Let's roll this function back to the initial version of it. See https://github.com/linkedin/cruise-control/commit/451b378f6d0b6de44cfc7588724a8e83381dbbaf for the relevant patch that adds this fix.

Can we also add a test that checks if `numProposalComputingThreads()` returns `2` if `_numPrecomputingThreads` is `0`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -117,6 +117,52 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
...
 
+  /**
+   * Generate unique goal priority for each of proposal precomputing thread.
+   */
+  private void populateGoalByPriorityForPrecomputing() {
+    List<List<Goal>> shuffledGoalList = new ArrayList<>();
+    // Guarantee one thread is working on the original goal priority.
+    shuffledGoalList.add(new ArrayList<>(_goalsByPriority.values()));
+
+    Integer goalCount = _goalsByPriority.values().size();
```
efeg: What if `goalCount` is `1`? Does the rest of the code handle the case with a single goal configuration?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -117,6 +117,52 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
...
+    Integer goalIndex = goalCount - 2;
+    Integer numShuffledGoalToGenerate = numProposalComputingThreads() - 1;
+    while (numShuffledGoalToGenerate > 0) {
```
efeg: What happens when `numShuffledGoalToGenerate` = `3` and `goalCount` = `2`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -117,6 +117,52 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
     _proposalComputationTimer = dropwizardMetricRegistry.timer(MetricRegistry.name("GoalOptimizer", "proposal-computation-timer"));
   }
 
+  /**
+   * Generate unique goal priority for each of proposal precomputing thread.
+   */
+  private void populateGoalByPriorityForPrecomputing() {
+    List<List<Goal>> shuffledGoalList = new ArrayList<>();
```
efeg: Do we need a list of list of goals or would it suffice to have a set of list of goals?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -117,6 +117,52 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
...
+
+          tmpShuffledGoalList.add(shuffledGoal);
+          numShuffledGoalToGenerate--;
+        }
+      }
+      shuffledGoalList.addAll(tmpShuffledGoalList);
+      goalIndex--;
+    }
+
+  for (List<Goal> shuffledGoal:shuffledGoalList) {
```
efeg: Missing space before and after `:`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -117,6 +117,52 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
...
+    List<List<Goal>> shuffledGoalList = new ArrayList<>();
+    // Guarantee one thread is working on the original goal priority.
+    shuffledGoalList.add(new ArrayList<>(_goalsByPriority.values()));
+
+    Integer goalCount = _goalsByPriority.values().size();
+    Integer goalIndex = goalCount - 2;
+    Integer numShuffledGoalToGenerate = numProposalComputingThreads() - 1;
+    while (numShuffledGoalToGenerate > 0) {
+      List<List<Goal>> tmpShuffledGoalList = new ArrayList<>();
+      for (int i = 0; i < shuffledGoalList.size() && numShuffledGoalToGenerate > 0; i++) {
```
efeg: The algorithm needs more comments to clarify the logic.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
kidkun: Hi Efe, this actually confuses me.
looks to me, the thread for submitting and waiting for the proposal computing to finish is the main thread. so if _numPrecomputingThreads is 0, one more thread is enough?

 from code, where _proposalPrecomputingExecutor.submit is used are in GoalOptimizer.java

line 215
    for (int i = 0; i < (_numPrecomputingThreads > 0 ? _numPrecomputingThreads : 1); i++) {
      futures.add(_proposalPrecomputingExecutor.submit(new ProposalCandidateComputer(_goalByPriorityForPrecomputing.get(i))));
    }

and line 308
              if (numWaitingThread == 0) {
                // No precomputing thread is available, schedule a computing and wait for the cache update.
                _proposalPrecomputingExecutor.submit(this::computeBestProposal);
              }

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
kidkun: in this case, _numPrecomputingThreads will be min(3, 2!)=2, numProposalComputingThreads() will be 2, numShuffledGoalsToGenerate will be 1. and while loop will execute once, the result will be the two possible combination of goal priority.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
kidkun: added comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
kidkun: added comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
kidkun: reverted the change.

Issue #224, opened 2018-05-11
send valid json responses for progress emitted via OperationProgress.java


nareshv:
Right now when there is any async operation going for the user submitted requests, the progress is emitted as plain-text even when json response is requested by user. Make necessary changes to `OperationProgress.java` to emit the progress as json.

nareshv:
Here is now the responses are sent right now.

```
(     1 ms) - (100.0%) PENDING: Operation enqueued, waiting to be executed.
(     0 ms) - (100.0%) WAITING_FOR_CLUSTER_MODEL: The job requires a cluster model and it is waiting to get the cluster model lock.
(  3759 ms) - (100.0%) AGGREGATING_METRICS: Retrieve the metrics of all the partitions from the aggregated metrics.
(  6240 ms) - (0.0%) GENERATING_CLUSTER_MODEL: Generating the cluster model for the cluster.
```

PR #241, opened 2018-06-01
fix JSON version of inprogress response

kidkun:
fix for https://github.com/linkedin/cruise-control/issues/224
return format will be
{name: {description: xxx,
              time: xxx,
              completionPercentage:xxx }
  ...
}
            
also fix the JSON response of GET state

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
@@ -6,7 +6,9 @@ package com.linkedin.kafka.cruisecontrol.async.progress;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import javax.servlet.http.HttpSession;
 
 
@@ -82,6 +84,21 @@ public class OperationProgress {
                               time,  step.completionPercentage() * 100, step.name(), step.description()));
     }
     return sb.toString();
+  }
+
+  public synchronized Map<String, Object> getJsonStructure() {
+    Map<String, Object> progressMap = new HashMap<>();
+    for (int i = 0; i < _steps.size(); i++) {
+      OperationStep step = _steps.get(i);
+      long time = (i == _steps.size() - 1 ? System.currentTimeMillis() : _startTimes.get(i + 1)) - _startTimes.get(i);
+      Map<String, Object> stepProgressMap = new HashMap<>();
+      stepProgressMap.put("step", i);
+      stepProgressMap.put("description", step.description());
+      stepProgressMap.put("time-in-ms", time);
+      stepProgressMap.put("completionPercentage", step.completionPercentage() * 100.0);
+      progressMap.put(step.name(), stepProgressMap);
+    }
+    return progressMap;
   }
 
   private void ensureMutable() {
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
@@ -1268,7 +1268,7 @@ public class KafkaCruiseControlServlet extends HttpServlet {
     try {
       return future.get(_maxBlockMs, TimeUnit.MILLISECONDS);
     } catch (TimeoutException te) {
-      returnProgress(response, future, false);
+      returnProgress(response, future, wantJSON(request));
       return null;
     }
   }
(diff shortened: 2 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
```diff
@@ -84,6 +86,21 @@ public synchronized String toString() {
     return sb.toString();
   }
 
+  public synchronized Map<String, Map<String, String>> getJsonStructure() {
```
efeg: Why do we return a `Map<String, Map<String, String>>` and not `Map<String, Object>`? It might be useful to check the format of the other functions that are performing the similar operation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
```diff
@@ -84,6 +86,21 @@ public synchronized String toString() {
...
+    Map<String, Map<String, String>> progressMap = new LinkedHashMap<>();
+    for (int i = 0; i < _steps.size(); i++) {
+      OperationStep step = _steps.get(i);
+      long time = (i == _steps.size() - 1 ? System.currentTimeMillis() : _startTimes.get(i + 1)) - _startTimes.get(i);
+      Map<String, String> stepProgressMap = new LinkedHashMap<>();
```
efeg: Why do we use `LinkedHashMap` and not a `HashMap` here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
```diff
@@ -84,6 +86,21 @@ public synchronized String toString() {
     return sb.toString();
   }
 
+  public synchronized Map<String, Map<String, String>> getJsonStructure() {
+    Map<String, Map<String, String>> progressMap = new LinkedHashMap<>();
```
efeg: While it is possible to build an ordered `json` string from a `LinkedHashMap` ([source](https://stackoverflow.com/questions/29491281/building-ordered-json-string-from-linkedhashmap)), I believe for the sake of processing the generated json file, it would be more preferable to make the step order more explicit rather than implicit -- i.e. hidden in the returned print order. Hence, I propose that we add a new field for each step that indicates the step number. We can name the field `step` and use the index of `_steps` as its value.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
kidkun: because I also want to main the order of 
description, time, percentage
in each step

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -1289,11 +1289,21 @@ private void setResponseCode(HttpServletResponse response, int code, boolean jso
     response.setHeader("Access-Control-Request-Method", "OPTIONS, GET, POST");
   }
 
+
```
efeg: Redundant spaces.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
```diff
@@ -84,6 +86,22 @@ public synchronized String toString() {
...
+      Map<String, Object> stepProgressMap = new LinkedHashMap<>();
+      stepProgressMap.put("step", i);
+      stepProgressMap.put("description", step.description());
+      stepProgressMap.put("time", String.format("%6d ms", time));
+      stepProgressMap.put("completionPercentage", String.format("%3.1f%%", step.completionPercentage() * 100));
+      progressMap.put(step.name(), stepProgressMap);
+    }
+    return progressMap;
+  }
+
```
efeg: redundant space.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
```diff
@@ -84,6 +86,22 @@ public synchronized String toString() {
     return sb.toString();
   }
 
+  public synchronized Map<String, Object> getJsonStructure() {
+    Map<String, Object> progressMap = new LinkedHashMap<>();
```
efeg: Since we do not depend on the order, can we return `HashMap`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
```diff
@@ -84,6 +86,22 @@ public synchronized String toString() {
...
+    for (int i = 0; i < _steps.size(); i++) {
+      OperationStep step = _steps.get(i);
+      long time = (i == _steps.size() - 1 ? System.currentTimeMillis() : _startTimes.get(i + 1)) - _startTimes.get(i);
+      Map<String, Object> stepProgressMap = new LinkedHashMap<>();
```
efeg: Since we do not depend on the order, can we return `HashMap`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -1308,10 +1308,18 @@ private void setResponseCode(HttpServletResponse response, int code, boolean jso
...
     setResponseCode(response, SC_OK, json);
-    response.setContentLength(progressString.length());
-    response.getOutputStream().write(progressString.getBytes(StandardCharsets.UTF_8));
+    String resp;
+    if (!json) {
+      resp = future.progressString();
+    } else {
+      Gson gson = new GsonBuilder().serializeNulls().serializeSpecialFloatingPointValues().create();
+      Map<String, Object> respMap = future.getJsonStructure();
+      respMap.put("version", JSON_VERSION);
```
efeg: Can we combine the progress information under `progress`? -- i.e.:
```
      Map<String, Object> respMap = new HashMap<>();
      respMap.put("version", JSON_VERSION);
      respMap.put("progress", future.getJsonStructure());
```

PR #242, opened 2018-06-01
Ensure that the temporary states in the load monitor does not block execution.

efeg:
If, for example, the load monitor is in `SAMPLING` state, an execution request fails due to unhandled exception while pausing the load monitor. This patch fixes this issue.


lindong28:
Sorry for the delay. Thanks for the PR. LGTM.

PR #244, opened 2018-06-05
Make refresh metadata requests more robust.

efeg:


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -193,9 +194,8 @@ public void shutdown() {
   /**
    * Get the state of the load monitor.
    */
-  public LoadMonitorState state(OperationProgress operationProgress) {
```
gitlw: What's the reason for adding the clusterAndGeneration parameter for this method and the meetCompletenessRequirements method?

efeg: This method is called from `KafkaCruiseControl#state(OperationProgress)`, which also calls `GoalOptimizer#state(MetadataClient.ClusterAndGeneration)` in the same method. Previously, both `GoalOptimizer#state` and `LoadMonitor#state` methods used to call `_metadataClient.refreshMetadata()` independently. This would potentially mean that each method would poll a broker for metadata update. In particular, `GoalOptimizer#state` would even poll a broker for metadata update for each goal.

This change ensures that rather than having `1 + #numberOfGoals` to a broker, we only have a single call for metadata update.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -80,24 +82,31 @@ public synchronized ClusterAndGeneration refreshMetadata(long timeout) {
       int version = _metadata.requestUpdate();
       long remaining = timeout;
       Cluster beforeUpdate = _metadata.fetch();
-      while (_metadata.version() <= version && remaining > 0) {
+      boolean hasUpdatedMetadata = _metadata.version() > version;
```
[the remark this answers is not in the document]
gitlw: You can differentiate the two cases even without the new variable, i.e. by checking _metadata.version() <= version in the if block after the while loop. But it's a matter of taste, and it's up to you to decide whether you want the new variable.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
gitlw: Talked offline with Efe, and this change will prevent the corner case of refreshing the metadata twice during the metadata TTL expirations.

gitlw:
LGTM

PR #251, opened 2018-06-11
Add parameter to allow/disallow capacity estimation upon cluster model generation.

efeg:
For endpoints that Cruise Control requires generating a cluster model, such as `rebalance`, `add_broker`, `remove_broker`, and `proposals`, Cruise Control will provide an optional parameter to specify whether it is allowed to estimate missing broker capacities or not (default: `allowCapacityEstimation: true`).
In case the estimation for broker capacities is not explicitly allowed, If CC is unable to get a well-formed broker capacity from its external dependencies (i.e. the broker id is missing from the `json` file specifying the broker capacities when the default capacity resolver, `BrokerCapacityConfigFileResolver.java`, is used), it will throw an exception specifying the issue.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
@@ -22,6 +22,7 @@ import com.linkedin.kafka.cruisecontrol.monitor.ModelGeneration;
 import com.linkedin.kafka.cruisecontrol.monitor.MonitorUtils;
 import com.linkedin.kafka.cruisecontrol.monitor.task.LoadMonitorTaskRunner;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -114,30 +115,6 @@ public class GoalOptimizer implements Runnable {
     _proposalComputationTimer = dropwizardMetricRegistry.timer(MetricRegistry.name("GoalOptimizer", "proposal-computation-timer"));
   }
 
-  private Set<List<Goal>> getPermutations(List<Goal> toPermute) {
-    Set<List<Goal>> allPermutations = new HashSet<>();
-    // Handle the case with single goal to permute.
-    if (toPermute.size() == 1) {
-      allPermutations.add(toPermute);
-      return allPermutations;
-    }
-
-    for (int i = 0; i < toPermute.size(); i++) {
-      // Copy the original list and remove the goal that we will prepend to the permutations of the remaining goals.
-      List<Goal> remainingToPermute = new ArrayList<>(toPermute);
-      Goal goal = toPermute.get(i);
-      remainingToPermute.remove(i);
-
-      // Prepend the goal to permutations of the remaining goals.
-      for (List<Goal> permutedRemaining: getPermutations(remainingToPermute)) {
-        permutedRemaining.add(0, goal);
-        allPermutations.add(permutedRemaining);
-      }
-    }
-
-    return allPermutations;
-  }
-
   private void populateGoalByPriorityForPrecomputing() {
     Set<List<Goal>> shuffledGoals = new HashSet<>();
 
@@ -155,7 +132,7 @@ public class GoalOptimizer implements Runnable {
     List<Goal> commonPrefix = new ArrayList<>(_goalsByPriority.values()).subList(0, toIndex);
     List<Goal> suffixToPermute = new ArrayList<>(_goalsByPriority.values()).subList(toIndex, _goalsByPriority.size());
 
-    for (List<Goal> shuffledSuffix: getPermutations(suffixToPermute)) {
+    for (List<Goal> shuffledSuffix: AnalyzerUtils.getPermutations(suffixToPermute)) {
       List<Goal> shuffledGoalList = new ArrayList<>(commonPrefix);
       shuffledGoalList.addAll(shuffledSuffix);
       shuffledGoals.add(shuffledGoalList);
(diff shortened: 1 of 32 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -847,6 +849,12 @@ private boolean wantJSON(HttpServletRequest request) {
     return Boolean.parseBoolean(jsonString);
   }
 
+  private boolean allowCapacityEstimation(HttpServletRequest request) {
```
gitlw: What if the passed string cannot be parsed as a boolean.

efeg: The behavior in this case is determined by the `Boolean#parseBoolean(String)` -- i.e. will be `false` for anything except `True` (not case sensitive).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -324,19 +303,27 @@ public AnalyzerState state(MetadataClient.ClusterAndGeneration clusterAndGenerat
...
     } else if (!_loadMonitor.meetCompletenessRequirements(_requirementsWithAvailableValidWindows)) {
       throw new IllegalStateException("Cannot get proposal because model completeness is not met.");
     }
-    if (!validCachedProposal()) {
-      LOG.debug("Cached best proposal is not usable, Cached generation: {}, current generation: {}. Wait for cache update.",
-                _bestProposal == null ? null : _bestProposal.modelGeneration(),
-                _loadMonitor.clusterModelGeneration());
-      synchronized (_cacheLock) {
+    synchronized (_cacheLock) {
+      if (validCachedProposal() && (allowCapacityEstimation || !_bestProposal.isCapacityEstimated())) {
```
gitlw: It seems if the condition is met, we should return the best proposal directly instead of invalidate the cache.

But even after switching the if and else block, you could still have a mismatched situation where request does not allow estimation, and the re-generated proposal still uses estimation. In that case, the switched code would return a proposal not satisfying the user's estimation requirement.

I'm thinking something like

```
public boolean compatibleWithRequest(allowCapacityEstimation, isCapacityEstimated) {
	return allowCapacityEstimation || !isCapacityEstimated;
}

while (!(validCachedProposal() && compatibleWithRequest(allowCapacityEstimation, _bestProposal.isCapacityEstimated()) )) {
	// we cannot use the current bestProposal, and may need to throw exception or re-generate the bestProposal
	if (!compatibleWithRequest(allowCapacityEstimation, _bestProposal.isCapacityEstimated()) {
		throw exception
	}

	invalidate the cache, and regenerate
}

//by the time we reach here, we know it's safe to use the bestProposal since it's both valid and satisfy the estimation requirement
return bestProposal;
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -481,12 +483,20 @@ public ModelGeneration clusterModelGeneration() {
...
    * @return the cached load, null if the load
    */
-  public ClusterModel.BrokerStats cachedBrokerLoadStats() {
+  public ClusterModel.BrokerStats cachedBrokerLoadStats(boolean allowCapacityEstimation) {
     int clusterGeneration = _metadataClient.refreshMetadata().generation();
     synchronized (this) {
       if (_cachedBrokerLoadGeneration != null
           && clusterGeneration == _cachedBrokerLoadGeneration.clusterGeneration()
           && _partitionMetricSampleAggregator.generation() == _cachedBrokerLoadGeneration.loadGeneration()) {
+        if (allowCapacityEstimation) {
```
gitlw: Should the condition be if (! allowCapacityEstimation) ?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerCapacityInfo.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.linkedin.kafka.cruisecontrol.common.Resource;
+import java.util.Map;
+
+
+public class BrokerCapacityInfo {
+  private final Map<Resource, Double> _capacity;
+  private final String _estimationInfo;
+
+  public BrokerCapacityInfo(Map<Resource, Double> capacity, String estimationInfo) {
```
gitlw: Minor: it might be good to have another constructor that only accepts the capacity, which sets the estimationInfo to "". That can be used more easily by the normal cases.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
efeg: Great catch!
The initial `if` condition is supposed to be the negation of the current condition -- i.e. `(!validCachedProposal() || (!allowCapacityEstimation && _bestProposal.isCapacityEstimated()))`.

I believe, we should avoid the other changes in the while loop (i.e. `while (!(validCachedProposal() && compatibleWithRequest(allowCapacityEstimation, _bestProposal.isCapacityEstimated()) ))`). This is because, in case the capacity retrieval is unable to avoid capacity estimation, we don't want to be stuck in the while loop. When the best proposal is returned by this function, the calling function already checks if the returned proposals involve estimation or not. In other words, if the cached proposals are estimated, and the user request asks for no estimation, we invalidate the cache and get the capacity again. If the returned result is still doing estimation, we throw an exception at a higher level in call hierarchy.

gitlw:
LGTM, thanks!

PR #258, opened 2018-06-20
Add max_load parameter to fetch the peak partition_load.

efeg:


efeg:
Relevant issue: https://github.com/linkedin/cruise-control/issues/259.

Issue #261, opened 2018-06-21
Add a threshold to supress metric anomalies for infitesimal upper percentile violations.


efeg:
If a broker has recently joined the cluster and/or stays without traffic for some time, once it starts receiving traffic (e.g. `topic creation`, `add_broker`, `rebalance`), Cruise Control will generate a metric anomaly due to current metric value.

PR #262, opened 2018-06-21
Add a threshold to supress metric anomalies for infitesimal upper percentile violations.

efeg:
Fixes the issue https://github.com/linkedin/cruise-control/issues/261.


cruise-control-core/src/main/java/com/linkedin/cruisecontrol/detector/metricanomaly/PercentileMetricAnomalyFinder.java
```diff
@@ -34,6 +34,8 @@
  */
 public abstract class PercentileMetricAnomalyFinder<E extends Entity> implements MetricAnomalyFinder<E> {
   private static final Logger LOG = LoggerFactory.getLogger(PercentileMetricAnomalyFinder.class);
+  // Ensure that no metric anomaly is generated unless the upper percentile metric value is a significant metric value.
+  private static final double SIGNIFICANT_METRIC_VALUE_THRESHOLD = 1;
```
kidkun: My only concern is that we hard coding threshold to be 1 here, which contains an assumption of the metrics we are monitoring. It is reasonable with the current metrics we have, but we need to be careful in the future for new metrics which extend from this abstract class.

PR #264, opened 2018-06-22
Refactor servlet code.

efeg:


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -655,62 +469,37 @@ private boolean getClusterLoad(HttpServletRequest request, HttpServletResponse r
...
+      // Close session
+      return true;
+    }
+
+    try {
+      Long startMsValue = KafkaCruiseControlServletUtils.startMs(request);
+      startMs = startMsValue == null ? -1L : startMsValue;
+      Long endMsValue = KafkaCruiseControlServletUtils.endMs(request);
+      endMs = endMsValue == null ? System.currentTimeMillis() : endMsValue;
+      partitionLowerBoundary = KafkaCruiseControlServletUtils.partitionLowerBoundary(request);
```
abhishekmendhekar: I think it'll be good to convert this into a single method where an upper and lower bound is returned.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
```diff
@@ -39,4 +185,160 @@ static String getClientIpAddress(HttpServletRequest request) {
...
+      resourceString = DEFAULT_PARTITION_LOAD_RESOURCE;
+    }
+    return resourceString;
+  }
+
+  static List<String> getGoals(HttpServletRequest request) {
+    boolean isKafkaAssignerMode = getMode(request);
+    List<String> goals;
+    String goalsString = request.getParameter(GOALS_PARAM);
+    // TODO: Handle urlencoded value (%2C instead of ,)
```
abhishekmendhekar: I think we should fix this cause it'll cause a serious problem now.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
```diff
@@ -39,4 +185,160 @@ static String getClientIpAddress(HttpServletRequest request) {
...
+  }
+
+  static int partitionLowerBoundary(HttpServletRequest request) {
+    String partitionString = request.getParameter(PARTITION_PARAM);
+    if (partitionString == null) {
+      return Integer.MIN_VALUE;
+    }
+
+    return !partitionString.contains("-") ? Integer.parseInt(partitionString)
+                                          : Integer.parseInt(partitionString.split("-")[0]);
```
abhishekmendhekar: Should there be a strict handling here?
Example, `value=10-11-12-13` based on the code it will take the lower bound as 10 and the upper bound as 11.
I think we should report this as an error.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -655,62 +469,37 @@ private boolean getClusterLoad(HttpServletRequest request, HttpServletResponse r
...
-    Integer entries;
-    Pattern topic;
+    long startMs;
+    long endMs;
+    Pattern topic = KafkaCruiseControlServletUtils.topic(request);
     int partitionUpperBoundary;
     int partitionLowerBoundary;
-    boolean json = wantJSON(request);
+    boolean json = KafkaCruiseControlServletUtils.wantJSON(request);
+    String resourceString = KafkaCruiseControlServletUtils.resourceString(request);
```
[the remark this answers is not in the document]
efeg: The validation was intentionally kept out of `KafkaCruiseControlServletUtils.resourceString()`, because we would like to let user know about the potentially incorrect `resourceString` s/he provided to the servlet via passing `errorMsg` in `handleParameterParseException`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
efeg: I agree that it is better to piggyback the fix for this legacy issue as part of this patch.

Issue #263, opened 2018-06-21
GET load request with granularity set to replica may make Cruise Control OOM


kidkun:
In a large-scale deployment(with many replicas exist in Kafka cluster),GET /kafkacruisecontrol/load?granularity=replica may make Cruise Control crash. This is due to large number of replica to report and CC get OOM when generating Cluster Model. Even if CC survives, the response is GB in size and is barely readable.  

Given that there is already another endpoint partition_load which returns partition level load information. It might be a good idea to remove the replica granularity parameter in this endpoint and add more parameters in partition_load endpoint to support richer ways of query for finer-granularity load information there.

Issue #223, opened 2018-05-11
add LeaderReplicas to the Json Datastructure of ClusterStats in `ClusterModel.java`


nareshv:
Normal formatted string version of cluster stats has `LeaderReplicas` added to it. Put the same value in Json datastructure as well.

kidkun:
Naresh, 
can you put a concrete example here?since from the code I do see that the JSON response returns leader_replica for each partition.

this is JSON return for each partition.
      recordMap.put(TOPIC, partitionInfo.topic());
      recordMap.put(PARTITION, Integer.toString(partitionInfo.partition()));
      recordMap.put(LEADER, Integer.toString(partitionInfo.leader() == null ? -1 : partitionInfo.leader().id()));
      recordMap.put(REPLICAS, replicas.toString());
      recordMap.put(IN_SYNC, inSyncReplicas.toString());
      recordMap.put(OUT_OF_SYNC, outOfSyncReplicas.toString());

which has the same field as normal formatted string version.

      out.write(String.format("%" + topicNameLength + "s%10s%10s%40s%40s%30s%n",
                              partitionInfo.topic(),
                              partitionInfo.partition(),
                              partitionInfo.leader() == null ? -1 : partitionInfo.leader().id(),
                              replicas,
                              inSyncReplicas,
                              outOfSyncReplicas)
                      .getBytes(StandardCharsets.UTF_8));

nareshv:
Code at https://github.com/linkedin/cruise-control/blob/4f83a21b5634b1037f26664516ca9cc4104e8827/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java#L923 has returned as one of the last params `broker.leaderReplicas().size()` whereas code at https://github.com/linkedin/cruise-control/blob/4f83a21b5634b1037f26664516ca9cc4104e8827/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java#L887 does not have this member in JSON

PR #265, opened 2018-06-22
Fix JSON response for multiple endpoints

kidkun:
1. fix the issue reported in ticket https://github.com/linkedin/cruise-control/issues/223 , add leader replica count per broker information in JSON response

2. fix get bootstrap/train JSON response,  remove traceStack field and rename errorMessage to message in normal response

3. fix kafka_cluster_state JSON response, increase readability of some fields

4. fix get load/partition_load JSON response reported in ticket https://github.com/linkedin/cruise-control/issues/263 . remove granularity parameter for get load request; add topic and partition parameter for get partition_load request. the topic parameter supports filtering partition based on matching a regular expression. the partition parameter supports filtering partition based on specific partition number or a partition number range

5. fix post add_broker/remove_broker/rebalance/demote_broker JSON response, refactor the code to generate response(both JSON and normal format) in a central place

6. add support of json=true parameter to post stop_proposal_execution/pause_sampling/resume_sampling request, return a response indicate CC has successfully accept the request

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
@@ -720,23 +720,21 @@ public class KafkaCruiseControlServlet extends HttpServlet {
     } else {
       Map<String, Object> partitionMap = new HashMap<>();
       List<Object> partitionList = new ArrayList<>();
-      List<String> header = new ArrayList<>(Arrays.asList("topic", "partition", "leader", "followers", "CPU", "DISK", "NW_IN", "NW_OUT"));
       partitionMap.put("version", JSON_VERSION);
-      partitionMap.put("header", header);
       for (Partition p : sortedPartitions) {
         if (++numEntries > entries) {
           break;
         }
         List<Integer> followers = p.followers().stream().map((replica) -> replica.broker().id()).collect(Collectors.toList());
-        List<Object> record = new ArrayList<>();
-        record.add(p.leader().topicPartition().topic());
-        record.add(p.leader().topicPartition().partition());
-        record.add(p.leader().broker().id());
-        record.add(followers);
-        record.add(p.leader().load().expectedUtilizationFor(Resource.CPU, wantMaxLoad));
-        record.add(p.leader().load().expectedUtilizationFor(Resource.DISK, wantMaxLoad));
-        record.add(p.leader().load().expectedUtilizationFor(Resource.NW_IN, wantMaxLoad));
-        record.add(p.leader().load().expectedUtilizationFor(Resource.NW_OUT, wantMaxLoad));
+        Map<String, Object> record = new HashMap<>();
+        record.put("topic", p.leader().topicPartition().topic());
+        record.put("partition",p.leader().topicPartition().partition());
+        record.put("leader",p.leader().broker().id());
+        record.put("followers",followers);
+        record.put("CPU",p.leader().load().expectedUtilizationFor(Resource.CPU, wantMaxLoad));
+        record.put("DISK",p.leader().load().expectedUtilizationFor(Resource.DISK, wantMaxLoad));
+        record.put("NW_IN",p.leader().load().expectedUtilizationFor(Resource.NW_IN, wantMaxLoad));
+        record.put("NW_OUT",p.leader().load().expectedUtilizationFor(Resource.NW_OUT, wantMaxLoad));
         partitionList.add(record);
       }
       partitionMap.put("records", partitionList);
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -692,6 +695,9 @@ private boolean getPartitionLoad(HttpServletRequest request, HttpServletResponse
 
     String entriesString = request.getParameter(ENTRIES_PARAM);
     Integer entries = entriesString == null ? Integer.MAX_VALUE : Integer.parseInt(entriesString);
+    String topicToFilter = request.getParameter(TOPIC_PARAM);
```
abhishekmendhekar: Since these are new parameter added, it'll be good to update `doGet()` documentation to reflect the change in `partition_load` endpoint parameters.

Does the topic partition support regular expressions? I feel it'll be useful to add regular expression support.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -692,6 +695,9 @@ private boolean getPartitionLoad(HttpServletRequest request, HttpServletResponse
...
+    String partitionToFilterString = request.getParameter(PARTITION_PARAM);
```
abhishekmendhekar: Similar to regular expression support for `TOPIC_PARAM`, do you feel it'll be useful to support partition range for `PARTITION_PARAM`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -700,6 +706,10 @@ private boolean getPartitionLoad(HttpServletRequest request, HttpServletResponse
                               "CPU (%)", "DISK (MB)", "NW_IN (KB/s)", "NW_OUT (KB/s)")
                       .getBytes(StandardCharsets.UTF_8));
       for (Partition p : sortedPartitions) {
+        if ((topicToFilter != null && !p.topicPartition().topic().equals(topicToFilter)) ||
```
abhishekmendhekar: I am wondering if `TOPIC_PARAM` is required if `PARTITION_PARAM` is set.
I don't think there is value in just fetching partition 1 for all topics for example. Since the logic topic view is top level to partitions and similarly `TOPIC_PARAM` is required with `PARITION_PARAM`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
kidkun: Thanks for the feedback Abhishek. I am wondering which documentation you are refering to? can you point me to it? I will update it. 

I think the regex support is a good idea. let me add it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
kidkun: in a scenario that client forget to set `TOPIC_PARAM`  and just `PARTITION_PARAM`, I think returning the corresponding partitions for all topic is ok since client  will realize they missed the `TOPIC_PARAM` and either resend the request with the parameter added or they can filter out the information in this response, which is probably of the superset of what they want .

abhishekmendhekar: It is better to return an error saying the topic is a required field when partition param is passed. In Kafka, the topic is the higher level abstraction which is used everywhere and we don't really just give the partition number without the topic anywhere.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
abhishekmendhekar: I was talking about the comments about the function definition. Looks like you already added that. Thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -564,6 +576,27 @@ private void setErrorResponse(HttpServletResponse response,
...
+      int responseCode,
+      boolean json)
+      throws IOException {
+    String resp;
+    if (json) {
+      Map<String, Object> respMap = new HashMap<>();
+      respMap.put("version", JSON_VERSION);
+      respMap.put("Message", message);
+      Gson gson = new Gson();
+      resp = gson.toJson(respMap);
```
abhishekmendhekar: nit: since gson is used just once you can directly call `new Gson().toJson(resoMap)`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -671,6 +680,26 @@ private boolean getPartitionLoad(HttpServletRequest request, HttpServletResponse
...
+        topic = null;
+      }
+      String partitionString = request.getParameter(PARTITION_PARAM);
+      if (partitionString == null) {
+        partitionLowerBoundary = Integer.MIN_VALUE;
+        partitionUpperBoundary = Integer.MAX_VALUE;
+      } else if (!partitionString.contains("-")) {
+        partitionLowerBoundary = Integer.parseInt(partitionString);
+        partitionUpperBoundary = partitionLowerBoundary;
+      } else {
```
abhishekmendhekar: Sorry for not adding this in the previous patch but we can also we can add a comma-separated list of partitions too.

Issue #184, opened 2018-04-03
Make stop_proposal_execution cancel an ongoing analyzer process


efeg:
Currently `stop_proposal_execution` takes effect only when the execution has already started. In other words, if the user sent a `rebalance/add_broker/remove_broker` request, s/he needs to wait until 1) the analyzer generates proposals and 2) executor starts executing those proposals. During this period, any `stop_proposal_execution` request is treated as no-op.

efeg:
@kidkun I don't think this issue has been resolved - was it closed intentionally?

PR #270, opened 2018-06-27
Fix the issue that under some scenario stop_proposal_execution does not take effect

kidkun:
fix for ticket 
https://github.com/linkedin/cruise-control/issues/184
Instead of canceling an ongoing analyzer process, the patch let the analyzer process finish but does not execute the proposal. The client will still get anticipated result of execution(just like set dryrun=true).


kidkun:
close this one for now, plan to have a more comprehensive solution after the UUID feature get brought in(to have a better way to track the ongoing jobs).

PR #294, opened 2018-08-10
Add min_valid_partition_ratio parameter to partition_load endpoint

kidkun:
* For better flexibility of specifying the completeness requirement during cluster model generation,  add min_valid_partition_ratio to partition_load endpoint to set the completeness requirement on the fly.  

* Another change is when this parameter is missing in request, instead of using the hard code value(0.98) use the value set via CC config parameter `min.valid.partition.ratio`

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -358,7 +358,7 @@ public class KafkaCruiseControl {
    * Get the cluster model for a given time window.
    * @param from the start time of the window
    * @param to the end time of the window
-   * @param requirements the model completeness requirement to enforce.
+   * @param minValidPartitionRatio the minimal valid partition ratio requirement of model from request
    * @param operationProgress the progress of the job to report.
    * @param allowCapacityEstimation Allow capacity estimation in cluster model if the requested broker capacity is unavailable.
    * @return the cluster workload model.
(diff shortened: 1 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -358,19 +358,23 @@ public ClusterModel clusterModel(long now,
    * Get the cluster model for a given time window.
    * @param from the start time of the window
    * @param to the end time of the window
-   * @param requirements the model completeness requirement to enforce.
+   * @param minValidPartitionRatio the minimal valid partition ratio requirement of model from request
```
efeg: 1. `minimal` --> `minimum` (the meaning is different).
2. `from request` implies a constraint that this method must be called as part of the user request handling, but it can potentially be used elsewhere. --> would removing this make our intention more clear?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
```diff
@@ -290,6 +292,20 @@ static Pattern topic(HttpServletRequest request) {
...
   }
 
+  static Double minValidPartitionRatio(HttpServletRequest request) {
+    String minValidPartitionRatioString = request.getParameter(MIN_VALID_PARTITION_RATIO_PARAM);
+    if (minValidPartitionRatioString == null) {
+      return null;
+    } else {
+      Double minValidPartitionRatio = Double.parseDouble(minValidPartitionRatioString);
+      if (minValidPartitionRatio > 1.0 || minValidPartitionRatio < 0.0) {
+        throw new IllegalArgumentException("The requested minimal partition ratio must be in range [0.0, 1.0] (Requested: "
```
efeg: 1. `minimal` --> `minimum` (the meaning is different).
2. `[0.0, 1.0]` --> `(0.0, 1.0)` -- as `0.0` or `1.0` are not allowed.

PR #295, opened 2018-08-11
Add sensors to track the session lifetime and request execution time

kidkun:
Add following metrics
1. timer to trace the session avg/max life time
2. meter to trace the rate session manager fails new session creation due to capacity limit(already have 5 ongoing sessions)
3. timer per endpoint to trace the avg execution time for requests routed to each endpoint
4. meter per endpoint to trace the rate each endpoint receives new request

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationFuture.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationFuture.java
@@ -19,12 +19,12 @@ public class OperationFuture<T> extends CompletableFuture<T> {
   private final String _operation;
   private final OperationProgress _operationProgress;
   private volatile Thread _executionThread = null;
-  private long _finishTime;
+  private long _finishTimeNs;
 
   public OperationFuture(String operation) {
     _operation = "'" + operation + "'";
     _operationProgress = new OperationProgress();
-    _finishTime = -1;
+    _finishTimeNs = -1;
   }
 
   @Override
@@ -104,11 +104,18 @@ public class OperationFuture<T> extends CompletableFuture<T> {
      return _operationProgress;
   }
 
-  public void setFinishTime(long finishTime) {
-    _finishTime = finishTime;
+  /**
+   * Record the finish time of this operation, invoked at the end of corresponding {@link OperationRunnable#run()}.
+   * @param finishTimeNs the system time when this operation completes.
+   */
+  public void setFinishTimeNs(long finishTimeNs) {
+    _finishTimeNs = finishTimeNs;
   }
 
-  public long getFinishTime() {
-    return _finishTime;
+  /**
+   * @return the integer representing the finish time of this operation.
+   */
+  public long finishTimeNs() {
+    return _finishTimeNs;
   }
 }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationFuture.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationFuture.java
@@ -19,10 +19,12 @@ public class OperationFuture<T> extends CompletableFuture<T> {
   private final String _operation;
   private final OperationProgress _operationProgress;
   private volatile Thread _executionThread = null;
+  private long _finishTime;
 
   public OperationFuture(String operation) {
     _operation = "'" + operation + "'";
     _operationProgress = new OperationProgress();
+    _finishTime = -1;
   }
 
   @Override
@@ -100,5 +102,13 @@ public class OperationFuture<T> extends CompletableFuture<T> {
    */
   public OperationProgress operationProgress() {
      return _operationProgress;
+  }
+
+  public void setFinishTime(long finishTime) {
+    _finishTime = finishTime;
+  }
+
+  public long getFinishTime() {
+    return _finishTime;
   }
 }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationRunnable.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationRunnable.java
@@ -43,7 +43,8 @@ abstract class OperationRunnable<T> implements Runnable {
           ((Pending) steps.get(0)).done();
         }
         _future.complete(getResult());
-        _future.setFinishTime(System.nanoTime());
+        // If operation completes successfully (i.e with no exception thrown), record the operation finish time.
+        _future.setFinishTimeNs(System.nanoTime());
       }
     } catch (Exception e) {
       LOG.debug("Received exception when trying to execute runnable for \"" + _future.operation() + "\"", e);
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationRunnable.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationRunnable.java
@@ -43,6 +43,7 @@ abstract class OperationRunnable<T> implements Runnable {
           ((Pending) steps.get(0)).done();
         }
         _future.complete(getResult());
+        _future.setFinishTime(System.nanoTime());
       }
     } catch (Exception e) {
       LOG.debug("Received exception when trying to execute runnable for \"" + _future.operation() + "\"", e);
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
@@ -168,8 +168,6 @@ public class KafkaCruiseControlServlet extends HttpServlet {
         } else {
           if (EndPoint.getEndpoint().contains(endPoint)) {
             _requestMeter.get(endPoint).mark();
-          } else {
-            throw new UserRequestException("Invalid URL for GET");
           }
           long requestExecutionStartTime = System.nanoTime();
           switch (endPoint) {
@@ -296,8 +294,6 @@ public class KafkaCruiseControlServlet extends HttpServlet {
         } else {
           if (EndPoint.postEndpoint().contains(endPoint)) {
             _requestMeter.get(endPoint).mark();
-          } else {
-            throw new UserRequestException("Invalid URL for POST");
           }
           long requestExecutionStartTime = System.nanoTime();
           switch (endPoint) {
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
@@ -69,8 +69,8 @@ public class KafkaCruiseControlServlet extends HttpServlet {
   private final SessionManager _sessionManager;
   private final long _maxBlockMs;
   private final ThreadLocal<Integer> _asyncOperationStep;
-  private final HashMap<EndPoint, Meter> _requestMeter = new HashMap<>();
-  private final HashMap<EndPoint, Timer> _requestExecutionTimer = new HashMap<>();
+  private final Map<EndPoint, Meter> _requestMeter = new HashMap<>();
+  private final Map<EndPoint, Timer> _requestExecutionTimer = new HashMap<>();
 
   public KafkaCruiseControlServlet(AsyncKafkaCruiseControl asynckafkaCruiseControl,
                                    long maxBlockMs,
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
@@ -4,7 +4,9 @@
 
 package com.linkedin.kafka.cruisecontrol.servlet;
 
+import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
 import com.linkedin.kafka.cruisecontrol.KafkaClusterState;
 import com.linkedin.kafka.cruisecontrol.common.Resource;
 import com.linkedin.kafka.cruisecontrol.KafkaCruiseControlState;
@@ -67,16 +69,24 @@ public class KafkaCruiseControlServlet extends HttpServlet {
   private final SessionManager _sessionManager;
   private final long _maxBlockMs;
   private final ThreadLocal<Integer> _asyncOperationStep;
+  private final HashMap<EndPoint, Meter> _requestMeter = new HashMap<>();
+  private final HashMap<EndPoint, Timer> _requestExecutionTimer = new HashMap<>();
 
   public KafkaCruiseControlServlet(AsyncKafkaCruiseControl asynckafkaCruiseControl,
                                    long maxBlockMs,
                                    long sessionExpiryMs,
                                    MetricRegistry dropwizardMetricRegistry) {
     _asyncKafkaCruiseControl = asynckafkaCruiseControl;
-    _sessionManager = new SessionManager(5, sessionExpiryMs, Time.SYSTEM, dropwizardMetricRegistry);
+    _sessionManager = new SessionManager(5, sessionExpiryMs, Time.SYSTEM, dropwizardMetricRegistry, _requestExecutionTimer);
     _maxBlockMs = maxBlockMs;
     _asyncOperationStep = new ThreadLocal<>();
     _asyncOperationStep.set(0);
+
+    for (EndPoint endpoint : EndPoint.values()) {
+      _requestMeter.put(endpoint, dropwizardMetricRegistry.meter(MetricRegistry.name("KafkaCruiseControlServelet", endpoint.name() + "-request-meter")));
+      _requestExecutionTimer.put(endpoint, dropwizardMetricRegistry.timer(MetricRegistry.name("KafkaCruiseControlServelet",
+                                  endpoint.name() + "-request-execution-timer")));
+    }
   }
 
   @Override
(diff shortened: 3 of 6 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationFuture.java
```diff
@@ -101,4 +103,12 @@ public String progressString() {
   public OperationProgress operationProgress() {
      return _operationProgress;
   }
+
+  public void setFinishTime(long finishTime) {
+    _finishTime = finishTime;
+  }
+
+  public long getFinishTime() {
```
efeg: In getters, we typically do not use `get` in the method name -- i.e. `getFinishTime` -> `finishTime`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationFuture.java
```diff
@@ -101,4 +103,12 @@ public String progressString() {
   public OperationProgress operationProgress() {
      return _operationProgress;
   }
+
+  public void setFinishTime(long finishTime) {
+    _finishTime = finishTime;
+  }
+
```
efeg: Missing JavaDoc -- i.e. is this an absolute time or a duration?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationFuture.java
```diff
@@ -101,4 +103,12 @@ public String progressString() {
   public OperationProgress operationProgress() {
      return _operationProgress;
   }
+
```
efeg: Missing JavaDoc for public method.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationFuture.java
```diff
@@ -19,10 +19,12 @@
   private final String _operation;
   private final OperationProgress _operationProgress;
   private volatile Thread _executionThread = null;
+  private long _finishTime;
```
efeg: The unit of `finishTime` is not clear. Is it in milliseconds, nanoseconds, seconds, ...? Let's update accordingly -- e.g. `_finishTime` -> `_finishTimeNs`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationRunnable.java
```diff
@@ -43,6 +43,7 @@ public void run() {
           ((Pending) steps.get(0)).done();
         }
         _future.complete(getResult());
+        _future.setFinishTime(System.nanoTime());
```
efeg: Can we clarify whether we are sure that the future is complete here -- i.e. we don't set the finish time prematurely?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -67,16 +69,24 @@
   private final SessionManager _sessionManager;
   private final long _maxBlockMs;
   private final ThreadLocal<Integer> _asyncOperationStep;
+  private final HashMap<EndPoint, Meter> _requestMeter = new HashMap<>();
+  private final HashMap<EndPoint, Timer> _requestExecutionTimer = new HashMap<>();
```
efeg: 1. `private final HashMap<EndPoint, Meter>` -> `private final Map<EndPoint, Meter> `
2. `private final HashMap<EndPoint, Timer>` -> `private final Map<EndPoint, Timer>`
3. let's initialize the private variables in the constructor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -49,6 +52,9 @@
       Executors.newSingleThreadScheduledExecutor(new KafkaCruiseControlThreadFactory("SessionCleaner",
                                                                                      true,
                                                                                      null));
+  private final Timer _sessionLifetimeTimer;
+  private final Meter _sessionCreationFailureMeter;
+  private final HashMap<EndPoint, Timer> _requestExecutionTimer;
```
efeg: `private final HashMap` -> `private final Map`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationFuture.java
kidkun: this is an absolute time. the runnable associate with the future will record the finish time once it finish the task.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/OperationRunnable.java
kidkun: I think so, this is the only place where we set the finish time and we set it at the end of run method, which guarantees that it is set only when this method finishes without error. 

Since all the async endpoint task is encapsulated in an subclass of OperationRunnable by overriding its getResult(), which is before the setFinishTime(), we are sure the system time we get at setFinishTime is after all the operation is done.

There is one thing to notice, the endpoint execution time we measure here does not include the execution time of executor for add/remove/demote/rebalance with dryrun=false. this is because the tasks submitted to executor asynchronously. we return the response and close the session once the task is submitted to executor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -67,16 +69,24 @@
...
                                    MetricRegistry dropwizardMetricRegistry) {
     _asyncKafkaCruiseControl = asynckafkaCruiseControl;
-    _sessionManager = new SessionManager(5, sessionExpiryMs, Time.SYSTEM, dropwizardMetricRegistry);
+    _sessionManager = new SessionManager(5, sessionExpiryMs, Time.SYSTEM, dropwizardMetricRegistry, _requestExecutionTimer);
     _maxBlockMs = maxBlockMs;
     _asyncOperationStep = new ThreadLocal<>();
     _asyncOperationStep.set(0);
+
+    for (EndPoint endpoint : EndPoint.values()) {
+      _requestMeter.put(endpoint, dropwizardMetricRegistry.meter(MetricRegistry.name("KafkaCruiseControlServelet", endpoint.name() + "-request-meter")));
```
efeg: Typo: `KafkaCruiseControlServelet` --> `KafkaCruiseControlServlet`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -67,16 +69,24 @@
...
+      _requestExecutionTimer.put(endpoint, dropwizardMetricRegistry.timer(MetricRegistry.name("KafkaCruiseControlServelet",
```
efeg: Typo: `KafkaCruiseControlServelet` --> `KafkaCruiseControlServlet`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -67,16 +69,24 @@
...
                                    long sessionExpiryMs,
                                    MetricRegistry dropwizardMetricRegistry) {
     _asyncKafkaCruiseControl = asynckafkaCruiseControl;
-    _sessionManager = new SessionManager(5, sessionExpiryMs, Time.SYSTEM, dropwizardMetricRegistry);
+    _sessionManager = new SessionManager(5, sessionExpiryMs, Time.SYSTEM, dropwizardMetricRegistry, _requestExecutionTimer);
     _maxBlockMs = maxBlockMs;
     _asyncOperationStep = new ThreadLocal<>();
     _asyncOperationStep.set(0);
+
+    for (EndPoint endpoint : EndPoint.values()) {
```
efeg: Let's add 
1. `private static final List<EndPoint> CACHED_VALUES = Collections.unmodifiableList(Arrays.asList(values()));`, and 
2. the following:
```
  /**
   * Use this instead of values() because values() creates a new array each time.
   * @return enumerated values in the same order as values()
   */
  public static List<Endpoint> cachedValues() {
    return CACHED_VALUES;
  }
```
 to the `Endpoint.java` and call this method rather than the values.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
efeg: We use `-rate`, not `-meter`, while naming the `Meter`s in our code base -- let's keep the naming consistent.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -57,12 +63,16 @@
...
+  SessionManager(int capacity, long sessionExpiryMs, Time time, MetricRegistry dropwizardMetricRegistry, Map<EndPoint, Timer> requestExecutionTimer) {
     _capacity = capacity;
     _sessionExpiryMs = sessionExpiryMs;
     _time = time;
     _inProgressSessions = new HashMap<>();
     _sessionCleaner.scheduleAtFixedRate(new ExpiredSessionCleaner(), 0, 5, TimeUnit.SECONDS);
+    _requestExecutionTimer = requestExecutionTimer;
+    // Metrics registration
+    _sessionLifetimeTimer = dropwizardMetricRegistry.timer(MetricRegistry.name("SessionManager", "session-lifetime-timer"));
+    _sessionCreationFailureMeter = dropwizardMetricRegistry.meter(MetricRegistry.name("SessionManager", "session-creation-failure-meter"));
```
efeg: We use `-rate`, not `-meter`, while naming the `Meter`s in our code base -- let's keep the naming consistent.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -155,7 +166,7 @@ synchronized int numSessions() {
    * Close the session for the given request.
    * @param request the request to close its session.
    */
-  synchronized void closeSession(HttpServletRequest request) {
+  synchronized void closeSession(HttpServletRequest request, boolean hasError) {
```
efeg: Can we update the `JavaDoc` to explain what `hasError` is?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -196,11 +216,15 @@ synchronized void expireOldSessions() {
     private final String _requestUrl;
     private final Map<String, String[]> _requestParameters;
     private final List<OperationFuture> _operationFuture;
+    private final long _requestTime;
```
efeg: Is it `_requestTimeNs`? Can we update the name and corresponding getter, comments, etc?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -184,7 +199,12 @@ synchronized void expireOldSessions() {
         LOG.info("Expiring session {}.", session);
         iter.remove();
         session.invalidate();
-        info.lastFuture().cancel(true);
+        _sessionLifetimeTimer.update(System.nanoTime() - info.requestTime(), TimeUnit.NANOSECONDS);
+        if (info.lastFuture().isDone() && info.executionTime() > 0) {
+          _requestExecutionTimer.get(info.endPoint()).update(info.executionTime(), TimeUnit.NANOSECONDS);
+        } else {
```
efeg: Isn't this a behavior change? Shouldn't `info.lastFuture().cancel(true);` be out of the else statement?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
efeg: Isn't this actually `_requestStartTimeNs`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -165,6 +176,10 @@ synchronized void closeSession(HttpServletRequest request) {
     if (info != null && info.lastFuture().isDone()) {
       LOG.info("Closing session {}", session);
       session.invalidate();
+      _sessionLifetimeTimer.update(System.nanoTime() - info.requestTime(), TimeUnit.NANOSECONDS);
+      if (!hasError && info.executionTime() > 0) {
```
efeg: If we actually check whether there is error or not (i.e. `hasError`), shouldn't we call the `_requestExecutionTimer` as `_successfulRequestExecutionTimer`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -156,35 +166,44 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
                                            endPoint, userParams.toString());
           setErrorResponse(response, "", errorResp, SC_BAD_REQUEST, wantJSON(request));
         } else {
+          if (EndPoint.getEndpoint().contains(endPoint)) {
+            _requestMeter.get(endPoint).mark();
+          } else {
+            throw new UserRequestException("Invalid URL for GET");
```
efeg: Why do we need this check here? Can't we move `_requestMeter.get(endPoint).mark();` within `switch-case` to avoid duplicate code here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -275,33 +294,42 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
                                            endPoint, userParams.toString());
           setErrorResponse(response, "", errorResp, SC_BAD_REQUEST, wantJSON(request));
         } else {
+          if (EndPoint.postEndpoint().contains(endPoint)) {
+            _requestMeter.get(endPoint).mark();
+          } else {
+            throw new UserRequestException("Invalid URL for POST");
```
efeg: Why do we need this check here? Can't we move `_requestMeter.get(endPoint).mark();` within `switch-case` to avoid duplicate code here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -25,6 +27,7 @@
 import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import static com.linkedin.kafka.cruisecontrol.servlet.KafkaCruiseControlServletUtils.*;
```
efeg: Let's avoid wildcard imports.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
kidkun: sessionInfo is a private static inner class of sessionManager and this variable is a final variable. I feel that the getter/setter/comment maybe unnecessary.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
kidkun: If the `future.isDone()` returns True, then `future.cancel()` does not have any effect. I think it is safe to call `cancel()` only in else branch.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
kidkun: Actually that is my original way of implementing and I feel it is more verbose, since we need to have the same `_requestMeter.get(endPoint).mark()` in each switch branch. I did think about removing else here since it looks redundant with the default branch of switch statement. but it actually helps use to jump out of execution and return a little bit earlier.

efeg: Thanks for the clarification. How about (1) keeping the `if` part, and (2) removing just the `else` here? Wouldn't switch case already throws an exception if the endpoint is invalid?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
efeg: Thanks for the clarification. How about (1) keeping the `if` part, and (2) removing just the `else` here? Wouldn't switch case already throws an exception if the endpoint is invalid?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
efeg: Minor style comment: on a second look, I feel like we can completely avoid this static import using `KafkaCruiseControlServletUtils.endPoint()` in just a single line -- feel free to use an approach that you would prefer.

efeg:
Thanks for the update! Left a couple more comments.

PR #301, opened 2018-08-17
Improve the log produced in metric sampling

kidkun:
Improve the log4j log in several places:
* distinguish broker sample generation failure due to missing topic metric or broker metric
* distinguish partition fetch task(normal sampling) and broker fetch task(for training purpose)
* let log capture more information to ease debugging

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
@@ -542,9 +542,9 @@ public class CruiseControlMetricsProcessor {
       boolean result = ((double) missingTopics.size() / topicsInBroker.size()) <= MAX_ALLOWED_MISSING_TOPIC_METRIC_PERCENT
           && ((double) missingPartitions.get() / cluster.partitionsForNode(brokerId).size() <= MAX_ALLOWED_MISSING_PARTITION_METRIC_PERCENT);
       if (!result) {
-        LOG.trace("Broker {} has {}/{} missing topics metrics and {}/{} missing partition metrics. Missing topics: {}. Missing partitions: {}.",
-                 brokerId, missingTopics.size(), topicsInBroker.size(), missingPartitions.get(), cluster.partitionsForNode(brokerId).size(),
-                 missingTopics, missingPartitions);
+        LOG.warn("Broker {} has {}/{} missing topics metrics and {}/{} missing partition metrics. Missing topics: {}.", brokerId,
+            missingTopics.size(), topicsInBroker.size(), missingPartitions.get(), cluster.partitionsForNode(brokerId).size(), missingTopics);
+        LOG.trace("Missing partitions: {}.", missingPartitions);
       }
       return result;
     }


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -327,6 +327,7 @@ public void run() {
           try {
             ConsumerRecords<byte[], byte[]> consumerRecords = _consumer.poll(10);
             if (consumerRecords == SHUTDOWN_RECORDS) {
+              LOG.warn("Metric loader received empty records");
```
efeg: This should be `trace` -- i.e. we expect to receive empty records in case no new data is available.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
```diff
@@ -535,9 +541,9 @@ private boolean enoughTopicPartitionMetrics(Cluster cluster, int brokerId) {
       boolean result = ((double) missingTopics.size() / topicsInBroker.size()) <= MAX_ALLOWED_MISSING_TOPIC_METRIC_PERCENT
           && ((double) missingPartitions.get() / cluster.partitionsForNode(brokerId).size() <= MAX_ALLOWED_MISSING_PARTITION_METRIC_PERCENT);
       if (!result) {
-        LOG.warn("Broker {} has {} missing topics metrics and {} missing partition metrics.",
-                 brokerId, missingTopics, missingPartitions);
-        LOG.trace("Missing topics: {}", missingTopics);
+        LOG.warn("Broker {} has {}/{} missing topics metrics and {}/{} missing partition metrics. Missing topics: {}. Missing partitions: {}.",
+                 brokerId, missingTopics.size(), topicsInBroker.size(), missingPartitions.get(), cluster.partitionsForNode(brokerId).size(),
+                 missingTopics, missingPartitions);
```
efeg: Let's maybe move the missing partitions to a separate `trace` level log -- i.e. `LOG.trace("Missing partitions: {}", missingPartitions);`. The reason is that we have so many partitions that we do not want to keep in the `warn` log.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
```diff
@@ -535,9 +541,9 @@ private boolean enoughTopicPartitionMetrics(Cluster cluster, int brokerId) {
       boolean result = ((double) missingTopics.size() / topicsInBroker.size()) <= MAX_ALLOWED_MISSING_TOPIC_METRIC_PERCENT
```
efeg: Can we update the JavaDoc of this method to refer to `MAX_ALLOWED_MISSING_TOPIC_METRIC_PERCENT` and `MAX_ALLOWED_MISSING_PARTITION_METRIC_PERCENT` rather than the hardcoded `99%` mentioned in `...collected more than 99% of the topic level and partition level metrics...`?
Can we also make these two static variables a configuration -- with default values of `0.01`?

kidkun: Updated the javadoc. 

Currently it is hard to add these two configs to` KafkaCruiseControlConfig` and let `CruiseControlMetricsProcessor` pick up the  value from `cruisecontrol.properties` because `CruiseControlMetricsProcessor` is a static member in `CruiseControlMetricsReporterSampler`. To do that we need lazy initialization of this static member which I find unnecessary. 

I am thinking of making `KafkaCruiseControlConfig` a singleton and can be used every where. I will include this in a config patch so we can discuss there.

PR #305, opened 2018-08-23
Add skip_hard_goal_check parameter to add_broker/remove_broker/rebalance endpoint

kidkun:
Add a new config `hard.goals` which specifies all the hard goals currently supported. At the same time, add a new parameter `skip_hard_goal_check` to `add_broker/remove_broker/rebalance` endpoints. If this parameter is true and `goals` parameter misses any hard goals,  the missing goals will be added to consideration when generating executing proposal.

some notes about this patch:
1. `demote_broker` endpoint does not have this parameter and the PLE proposal generated via this endpoint will not take hard goals into consideration.
2.`proposal` endpoint does not have this parameter and does not enforce the hard goal  when generating the proposal.
3. anomaly detector's fix will automatically consider all the hard goals.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -128,17 +128,18 @@ private boolean isKafkaAssignerMode(List<String> goals) {
...
    * @param requirements The cluster model completeness requirements.
-   * @param operationProgress the progress to report.
+   * @param operationProgress The progress to report.
    * @param allowCapacityEstimation Allow capacity estimation in cluster model if the requested broker capacity is unavailable.
    * @param concurrentPartitionMovements The maximum number of concurrent partition movements per broker
    *                                     (if null, use num.concurrent.partition.movements.per.broker).
    * @param concurrentLeaderMovements The maximum number of concurrent leader movements
    *                                  (if null, use num.concurrent.leader.movements).
-   * @return the optimization result.
+   * @param skipHardGoalCheck Whether check and add all hard goals to {@code goals}.
```
efeg: If we rephrase it as follows, would it be more clear?
`Whether check and add all hard goals to {@code goals}.` -> `True if the provided {@code goals} do not have to contain all hard goals, false otherwise`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -197,19 +199,20 @@ public static void sanityCheckCapacityEstimation(boolean allowCapacityEstimation
...
+   * @param throttleAddedBrokers Whether throttle the brokers that are being added.
+   * @param goals The goals to be met when adding the brokers. When empty all goals will be used.
    * @param requirements The cluster model completeness requirements.
    * @param operationProgress The progress of the job to update.
    * @param allowCapacityEstimation Allow capacity estimation in cluster model if the requested broker capacity is unavailable.
    * @param concurrentPartitionMovements The maximum number of concurrent partition movements per broker
    *                                     (if null, use num.concurrent.partition.movements.per.broker).
    * @param concurrentLeaderMovements The maximum number of concurrent leader movements
    *                                  (if null, use num.concurrent.leader.movements).
+   * @param skipHardGoalCheck Whether check and add all hard goals to {@code goals}.
```
efeg: If we rephrase it as follows, would it be more clear?
`Whether check and add all hard goals to {@code goals}.` -> `True if the provided {@code goals} do not have to contain all hard goals, false otherwise`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -247,26 +251,28 @@ public static void sanityCheckCapacityEstimation(boolean allowCapacityEstimation
...
-   * @param operationProgress the progress of the job to report.
+   * @param operationProgress The progress of the job to report.
    * @param allowCapacityEstimation Allow capacity estimation in cluster model if the requested broker capacity is unavailable.
    * @param concurrentPartitionMovements The maximum number of concurrent partition movements per broker
    *                                     (if null, use num.concurrent.partition.movements.per.broker).
    * @param concurrentLeaderMovements The maximum number of concurrent leader movements
    *                                  (if null, use num.concurrent.leader.movements).
-   * @return the optimization result.
-   * @throws KafkaCruiseControlException when the rebalance encounter errors.
+   * @param skipHardGoalCheck Whether check and add all hard goals to {@code goals}.
```
efeg: If we rephrase it as follows, would it be more clear?
`Whether check and add all hard goals to {@code goals}.` -> `True if the provided {@code goals} do not have to contain all hard goals, false otherwise`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -458,19 +464,21 @@ public void resumeLoadMonitorActivity() {
...
   /**
    * Optimize a cluster workload model.
-   * @param goals a list of goals to optimize. When empty all goals will be used.
-   * @param requirements the model completeness requirements to enforce when generating the proposals.
-   * @param operationProgress the progress of the job to report.
+   * @param goals A list of goals to optimize. When empty all goals will be used.
+   * @param requirements The model completeness requirements to enforce when generating the proposals.
+   * @param operationProgress The progress of the job to report.
    * @param allowCapacityEstimation Allow capacity estimation in cluster model if the requested broker capacity is unavailable.
+   * @param skipHardGoalCheck Whether check and add all hard goals to {@code goals}.
```
efeg: If we rephrase it as follows, would it be more clear?
`Whether check and add all hard goals to {@code goals}.` -> `True if the provided {@code goals} do not have to contain all hard goals, false otherwise`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -586,19 +594,32 @@ private ModelCompletenessRequirements modelCompletenessRequirements(Collection<G
...
     MetadataClient.ClusterAndGeneration clusterAndGeneration = _loadMonitor.refreshClusterAndGeneration();
     return goals.stream().allMatch(g -> _loadMonitor.meetCompletenessRequirements(
         clusterAndGeneration, g.clusterModelCompletenessRequirements()));
   }
 
   /**
    * Get a goals by priority based on the goal list.
+   *
+   * @param goals A list of goals to optimize.
+   * @param skipHardGoalCheck Whether check and add all hard goals to {@code goals}.
```
efeg: If we rephrase it as follows, would it be more clear?
`Whether check and add all hard goals to {@code goals}.` -> `True if the provided {@code goals} do not have to contain all hard goals, false otherwise`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -586,19 +594,32 @@ private ModelCompletenessRequirements modelCompletenessRequirements(Collection<G
...
+    Collection<Goal> goals = goalsByPriority(goalNames, false).values();
     MetadataClient.ClusterAndGeneration clusterAndGeneration = _loadMonitor.refreshClusterAndGeneration();
     return goals.stream().allMatch(g -> _loadMonitor.meetCompletenessRequirements(
         clusterAndGeneration, g.clusterModelCompletenessRequirements()));
   }
 
   /**
    * Get a goals by priority based on the goal list.
+   *
+   * @param goals A list of goals to optimize.
```
efeg: In the context of this method, are the given goals `optimized` in this method -- i.e. do we use this function to optimize the given goals as it is indicated in the description of this parameter?

Would it be more clear to just say `A list of goals`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -586,19 +594,32 @@ private ModelCompletenessRequirements modelCompletenessRequirements(Collection<G
...
       return AnalyzerUtils.getGoalMapByPriority(_config);
     }
+    if (!skipHardGoalCheck) {
+      List<String> hardGoals = _config.getList(KafkaCruiseControlConfig.HARD_GOALS_CONFIG);
+      for (String goalName : goals) {
+        if (!hardGoals.contains(goalName)) {
+          hardGoals.add(goalName);
+        }
+      }
+      goals = hardGoals;
```
efeg: I think we should avoid silently overwriting the user parameters. The role of the `skipHardGoalCheck ` should be: `If the user does not explicitly ask to skip the hard goal check, then if CC identifies that a hard goal is missing from the user provided goals, then throw an exception.`

config/cruisecontrol.properties
```diff
@@ -88,9 +88,12 @@ capacity.config.file=config/capacity.json
 # The list of goals to optimize the Kafka cluster for with pre-computed proposals
 default.goals=com.linkedin.kafka.cruisecontrol.analyzer.goals.RackAwareGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.ReplicaCapacityGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.DiskCapacityGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.NetworkInboundCapacityGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.NetworkOutboundCapacityGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.CpuCapacityGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.ReplicaDistributionGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.PotentialNwOutGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.DiskUsageDistributionGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.NetworkInboundUsageDistributionGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.NetworkOutboundUsageDistributionGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.CpuUsageDistributionGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.TopicReplicaDistributionGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.LeaderBytesInDistributionGoal
 
-# The list of goals that is supported.
+# The list of goals that is supported
```
efeg: `The list of goals that is supported` -> `The list of supported goals`

config/cruisecontrol.properties
```diff
@@ -88,9 +88,12 @@ capacity.config.file=config/capacity.json
...
 
+# The list of hard goals that is supported
```
efeg: `The list of hard goals that is supported` -> `The list of supported hard goals`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -586,19 +594,24 @@ private ModelCompletenessRequirements modelCompletenessRequirements(Collection<G
...
 
   /**
    * Get a goals by priority based on the goal list.
+   *
+   * @param goals A list of goals.
+   * @param skipHardGoalCheck True if the provided {@code goals} do not have to contain all hard goals, false otherwise.
+   * @return A map of goal priority to goal.
    */
-  private Map<Integer, Goal> goalsByPriority(List<String> goals) {
+  private Map<Integer, Goal> goalsByPriority(List<String> goals, boolean skipHardGoalCheck) {
```
efeg: Since in some cases we hard-code the value for `skipHardGoalCheck` when calling this function, I am wondering whether it would make the code more clear if we keep the functionality of this function the same as before, and create a separate function for sanity checking the existence of hard goals in the given goals.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -247,26 +251,28 @@ public static void sanityCheckCapacityEstimation(boolean allowCapacityEstimation
...
                                                  boolean dryRun,
                                                  ModelCompletenessRequirements requirements,
                                                  OperationProgress operationProgress,
                                                  boolean allowCapacityEstimation,
                                                  Integer concurrentPartitionMovements,
-                                                 Integer concurrentLeaderMovements) throws KafkaCruiseControlException {
-    GoalOptimizer.OptimizerResult result = getOptimizationProposals(goals, requirements, operationProgress, allowCapacityEstimation);
+                                                 Integer concurrentLeaderMovements,
+                                                 boolean skipHardGoalCheck) throws KafkaCruiseControlException {
+    GoalOptimizer.OptimizerResult result = getOptimizationProposals(goals, requirements, operationProgress, allowCapacityEstimation, skipHardGoalCheck);
```
efeg: If the goals is only the `PreferredLeaderElectionGoal`, should we by default skip the hard goal check?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
```diff
@@ -351,8 +354,11 @@ static String urlDecode(String s) throws UnsupportedEncodingException {
     goals = goalsString == null ? new ArrayList<>() : Arrays.asList(goalsString.split(","));
     goals.removeIf(String::isEmpty);
 
-    // If KafkaAssigner mode is enabled and goals are not specified, it overrides the goal list.
-    if (goals.isEmpty() && isKafkaAssignerMode) {
+    // KafkaAssigner mode is assumed to use two KafkaAssigner goals, if client specifies goals in request, throw exception.
+    if (isKafkaAssignerMode) {
+      if (!goals.isEmpty()) {
+        throw new UserRequestException("Kafka assigner mode does not support explicitly-specifying goals in request.");
```
efeg: `explicitly-specifying` -> `explicitly specifying`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -444,6 +444,14 @@
   private static final String GOALS_DOC = "A list of case insensitive goals in the order of priority. The high "
       + "priority goals will be executed first.";
 
+  /**
+   * <code>hard.goals</code>
+   */
+  public static final String HARD_GOALS_CONFIG = "hard.goals";
+  private static final String HARD_GOALS_DOC = "A list of case insensitive hard goals in the order of priority. Hard goals"
```
efeg: `in the order of priority` -> does the order matter here?

efeg: Missing space after `goals`, before quotation.

kidkun: It does not matter here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
kidkun: I feel implementing in that way there is a implicit assumption here. 

Given that we provide `skip_hard_goal_check` parameter to rebalance endpoint, not  specializing `PreferredLeaderElectionGoal` make  `skip_hard_goal_check`'s meaning more clear to user. If user really want to do a PLE and forget set `skip_hard_goal_check` to `true`, he/she will get an exception (which clear explain the reason) and he/she just recall with the parameter, it is not a big inconvenience.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
kidkun: current logic is populate the `goal list` with `default goal list` if it is null, then do the check(which is guarantee to pass).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -611,4 +632,20 @@ public boolean meetCompletenessRequirements(List<String> goalNames) {
...
   }
+
+  /**
+   * Check whether all hard goals are included in provided goal list.
+   *
+   * @param goals A list of goals.
+   * @param skipHardGoalCheck True if hard goal checking is not needed.
+   */
+  private void goalListSanityCheck(List<String> goals, boolean skipHardGoalCheck) {
+    // If goals is null, skip hard goal check since all default goals will be picked to calculate proposal.
```
efeg: If default goals config is null, then all goals will be picked up.
Maybe we can remove this comment to make this function self-contained -- i.e. does not depend on how another function works. We can update the javadoc to mention what this function does if `goals` is null.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -611,4 +632,20 @@ public boolean meetCompletenessRequirements(List<String> goalNames) {
...
+    if (goals != null && !skipHardGoalCheck) {
```
efeg: `goals != null` --> `!(goals == null || goals.isEmpty())`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -247,26 +253,34 @@ public static void sanityCheckCapacityEstimation(boolean allowCapacityEstimation
...
                                                  OperationProgress operationProgress,
                                                  boolean allowCapacityEstimation,
                                                  Integer concurrentPartitionMovements,
-                                                 Integer concurrentLeaderMovements) throws KafkaCruiseControlException {
-    GoalOptimizer.OptimizerResult result = getOptimizationProposals(goals, requirements, operationProgress, allowCapacityEstimation);
+                                                 Integer concurrentLeaderMovements,
+                                                 boolean skipHardGoalCheck) throws KafkaCruiseControlException {
+    GoalOptimizer.OptimizerResult result;
+    //If it is a PLE request, skip hard goal check.
+    if (goals != null && goals.size() == 1 && goals.get(0).equals(PreferredLeaderElectionGoal.class.getName())) {
```
efeg: Shouldn't it be `PreferredLeaderElectionGoal.class.getName()` -> `PreferredLeaderElectionGoal.class.getSimpleName()`?

efeg: Would moving this sanity check inside `goalListSanityCheck` make the code flow more clear here?

PR #307, opened 2018-08-25
Fix acceptance of window with zero valid entity as valid window in MetricSampleAggregator

kidkun:
When `_minValidEntityRatio` in `AggregationOptions` is set to zero, `MetricSampleAggregator` will take window with zero valid entity as valid window in its `completeness()` and `aggregate()` method. At the same time, valid entity are collect as **entities which are valid in all valid windows**. This result in a corner scenario that if `_minValidEntityRatio` is set to zero and there happens to be one or more window with zero valid entity,  there will be zero valid entity in `MetricSampleAggregationResult`, which is used in `ClusterModel`.

One Cruise Control client visible case will be that client make a request to `partition_load` endpoint with parameter `min_valid_partition_ratio` set to 0. If in one window in Cruise Control's sample history has zero valid entity( can happen if Kafka cluster down for a while in the past and reporter does not report the metrics), the user will see no partition load at all in response.

This patch deals with with this corner case.

### DIFF
--- a/cruise-control-core/src/test/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregatorTest.java
+++ b/cruise-control-core/src/test/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregatorTest.java
@@ -230,10 +230,11 @@ public class MetricSampleAggregatorTest {
                                  AggregationOptions.Granularity.ENTITY, true);
 
     MetricSampleCompleteness<String, IntegerEntity> completeness = aggregator.completeness(-1, Long.MAX_VALUE, options);
-    assertEquals(20, completeness.validWindowIndexes().size());
-    assertEquals(1, completeness.validEntities().size());
+    assertEquals(17, completeness.validWindowIndexes().size());
+    assertEquals(2, completeness.validEntities().size());
     assertTrue(completeness.validEntities().contains(ENTITY1));
-    assertTrue(completeness.validEntityGroups().isEmpty());
+    assertTrue(completeness.validEntities().contains(ENTITY3));
+    assertTrue(completeness.validEntityGroups().contains("g2"));
     assertCompletenessByWindowIndex(completeness);
   }
 
--- a/cruise-control-core/src/test/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregatorTest.java
+++ b/cruise-control-core/src/test/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregatorTest.java
@@ -234,7 +234,7 @@ public class MetricSampleAggregatorTest {
     assertEquals(2, completeness.validEntities().size());
     assertTrue(completeness.validEntities().contains(ENTITY1));
     assertTrue(completeness.validEntities().contains(ENTITY3));
-    assertTrue(completeness.validEntityGroups().contains("g2"));
+    assertTrue(completeness.validEntityGroups().contains(ENTITY3.group()));
     assertCompletenessByWindowIndex(completeness);
   }
 
--- a/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
+++ b/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
@@ -116,8 +116,10 @@ public class KafkaPartitionMetricSampleAggregatorTest {
     Map<PartitionEntity, ValuesAndExtrapolations> aggregateResult =
         metricSampleAggregator.aggregate(clusterAndGeneration(cluster), Long.MAX_VALUE, new OperationProgress())
                               .valuesAndExtrapolations();
-    assertEquals(1, aggregateResult.size(), 0);
-    assertEquals(NUM_WINDOWS, aggregateResult.get(PE).windows().size(), 0);
+    // Partition "topic-0" should be valid in all NUM_WINDOW windows and Partition "topic1-0" should not since
+    // there is no sample for it.
+    assertEquals(1, aggregateResult.size());
+    assertEquals(NUM_WINDOWS, aggregateResult.get(PE).windows().size());
 
     ModelCompletenessRequirements requirements =
         new ModelCompletenessRequirements(1, 0.0, true);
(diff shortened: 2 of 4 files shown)


cruise-control-core/src/test/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregatorTest.java
```diff
@@ -230,10 +230,11 @@ public void testAggregationOption4() {
...
 
     MetricSampleCompleteness<String, IntegerEntity> completeness = aggregator.completeness(-1, Long.MAX_VALUE, options);
-    assertEquals(20, completeness.validWindowIndexes().size());
-    assertEquals(1, completeness.validEntities().size());
+    assertEquals(17, completeness.validWindowIndexes().size());
+    assertEquals(2, completeness.validEntities().size());
     assertTrue(completeness.validEntities().contains(ENTITY1));
-    assertTrue(completeness.validEntityGroups().isEmpty());
+    assertTrue(completeness.validEntities().contains(ENTITY3));
+    assertTrue(completeness.validEntityGroups().contains("g2"));
```
efeg: Should we avoid hardcoding `g2` and use `ENTITY3.group()`?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
```diff
@@ -109,15 +109,15 @@ public void testAggregateWithUpdatedCluster() throws NotEnoughValidWindowsExcept
...
+    TopicPartition tp1 = new TopicPartition(TOPIC + "1", 0);
     Cluster cluster = getCluster(Arrays.asList(TP, tp1));
     metadata.update(cluster, Collections.emptySet(), 1);
 
     Map<PartitionEntity, ValuesAndExtrapolations> aggregateResult =
         metricSampleAggregator.aggregate(clusterAndGeneration(cluster), Long.MAX_VALUE, new OperationProgress())
                               .valuesAndExtrapolations();
-    assertTrue("tp1 should not be included because recent metric window does not include all topics",
-               aggregateResult.isEmpty());
+    assertEquals(1, aggregateResult.size(), 0);
```
efeg: What code change led to this change, can we add a comment to clarify the expected behavior?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
```diff
@@ -109,15 +109,15 @@ public void testAggregateWithUpdatedCluster() throws NotEnoughValidWindowsExcept
 
     populateSampleAggregator(NUM_WINDOWS + 1, MIN_SAMPLES_PER_WINDOW, metricSampleAggregator);
 
-    TopicPartition tp1 = new TopicPartition(TOPIC, 1);
+    TopicPartition tp1 = new TopicPartition(TOPIC + "1", 0);
```
efeg: Why did we change the topic and partition names?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
```diff
@@ -257,7 +257,7 @@ public void testTooManyFlaws() throws NotEnoughValidWindowsException {
           metricSampleAggregator.aggregate(clusterAndGeneration(metadata.fetch()),
                                            NUM_WINDOWS * WINDOW_MS,
                                            new OperationProgress());
-      assertTrue(result.valuesAndExtrapolations().isEmpty());
+      assertEquals(NUM_WINDOWS - 3, result.valuesAndExtrapolations().get(PE).windows().size(), 0);
```
efeg: Why are we using the `float` comparator for the `assert`s added in this patch?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
```diff
@@ -182,7 +182,7 @@ public void testFallbackToAvgAvailable() throws NotEnoughValidWindowsException {
         metricSampleAggregator.aggregate(clusterAndGeneration(metadata.fetch()),
                                          NUM_WINDOWS * WINDOW_MS,
                                          new OperationProgress());
-    assertTrue(result.valuesAndExtrapolations().isEmpty());
+    assertEquals(NUM_WINDOWS - 2, result.valuesAndExtrapolations().get(PE).windows().size(), 0);
```
efeg: Would it make the code more clear if we add a comment to clarify why do we expect these values?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
kidkun: The scenario here is that there is a partition "topic-0" which has enough samples  and another partition("topic-1"->"topic1-0") which has none. 

If we do not change the topic name of second partition, after the code change 
https://github.com/linkedin/cruise-control/blob/6f133ece636dd8228b9f1895e73ceedbb67e6a91/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/WindowState.java#L70
we will get a `NotEnoughValidWindowsException`. 

This is because if tow partitions has the same  topic name("topic") the invalidity of one partition will invalid the other one(the aggregating granularity of `PartitionEntity` is `ENTITY_GROUP`). Before the code change, this is OK, given that a window with zero valid partition is taken as a valid window, so there will be NUM_WINDOWS valid with zero valid partition sample in them. But after the code change, all these window become invalid.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
kidkun: Changed to `int` comparator.

kidkun: I added a comment about what we expect in this test.  The changes in expectation is because window with zero valid partition sample is no longer taken as valid window(window 1,2,3 in this test). And valid partition in aggregate result is partition which has valid sample in all valid windows( an intersection of valid partition set from all valid partition).  After the exclusion of window 1,2,3, partition `topic-0` now is a valid partition in aggregate result.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
kidkun: Add a comment. The detailed explanation is same as above.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/aggregator/KafkaPartitionMetricSampleAggregatorTest.java
kidkun: Add a comment about expectation. Detailed explanation above.

PR #308, opened 2018-08-29
Add several configs to KafkaCruiseControlConfig and improve its usage in reflective created object.

kidkun:
Remove the config of pluggable component from centralized config definition in KafkaCruiseControlConfig so that user can have their own class implementation and customized config of these pluggable component and plug in to use with Cruise Control.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
@@ -42,8 +42,8 @@ public abstract class AbstractGoal implements Goal {
   private boolean _finished;
   protected boolean _succeeded = true;
   protected BalancingConstraint _balancingConstraint;
-  protected int _numWindows = 1;
-  protected double _minMonitoredPartitionPercentage = 0.995;
+  protected int _numWindows;
+  protected double _minMonitoredPartitionPercentage;
 
   /**
    * Constructor of Abstract Goal class sets the _finished flag to false to signal that the goal requirements have not
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
@@ -55,17 +55,10 @@ public abstract class AbstractGoal implements Goal {
 
   @Override
   public void configure(Map<String, ?> configs) {
-    _balancingConstraint = new BalancingConstraint(new KafkaCruiseControlConfig(configs, false));
-    String numWindowsString = (String) configs.get(KafkaCruiseControlConfig.NUM_PARTITION_METRICS_WINDOWS_CONFIG);
-    if (numWindowsString != null && !numWindowsString.isEmpty()) {
-      _numWindows = Integer.parseInt(numWindowsString);
-    }
-    String minMonitoredPartitionPercentageString =
-        (String) configs.get(KafkaCruiseControlConfig.MIN_VALID_PARTITION_RATIO_CONFIG);
-    if (minMonitoredPartitionPercentageString != null
-        && !minMonitoredPartitionPercentageString.isEmpty()) {
-      _minMonitoredPartitionPercentage = Double.parseDouble(minMonitoredPartitionPercentageString);
-    }
+    KafkaCruiseControlConfig parsedConfig = new KafkaCruiseControlConfig(configs, false);
+    _balancingConstraint = new BalancingConstraint(parsedConfig);
+    _numWindows = parsedConfig.getInt(KafkaCruiseControlConfig.NUM_PARTITION_METRICS_WINDOWS_CONFIG);
+    _minMonitoredPartitionPercentage = parsedConfig.getDouble(KafkaCruiseControlConfig.MIN_VALID_PARTITION_RATIO_CONFIG);
   }
 
   @Override
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
@@ -65,22 +65,24 @@ public class KafkaSampleStore implements SampleStore {
   protected ExecutorService _metricProcessorExecutor;
   protected String _partitionMetricSampleStoreTopic;
   protected String _brokerMetricSampleStoreTopic;
+  protected Integer _metricSampleStoreTopicReplicationFactor;
   protected volatile double _loadingProgress;
   protected Producer<byte[], byte[]> _producer;
   protected volatile boolean _shutdown = false;
 
+  public static final String PARTITION_METRIC_SAMPLE_STORE_TOPIC_CONFIG = "partition.metric.sample.store.topic";
+  public static final String BROKER_METRIC_SAMPLE_STORE_TOPIC_CONFIG = "broker.metric.sample.store.topic";
+  public static final String NUM_SAMPLE_LOADING_THREADS_CONFIG = "num.sample.loading.threads";
+  public static final String NUM_METRIC_SAMPLE_STORE_TOPIC_REPLICATION_FACTOR_CONFIG = "num.metric.sample.store.topic.replication.factor";
 
   @Override
   public void configure(Map<String, ?> config) {
-    _partitionMetricSampleStoreTopic = (String) config.get(KafkaCruiseControlConfig.PARTITION_METRIC_SAMPLE_STORE_TOPIC_CONFIG);
-    _brokerMetricSampleStoreTopic = (String) config.get(KafkaCruiseControlConfig.BROKER_METRIC_SAMPLE_STORE_TOPIC_CONFIG);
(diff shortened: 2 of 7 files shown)


becketqin:
@kidkun Thanks for the patch. I took a quick glance at the patch and have some concerns. It looks that the configurations moved to KafkaCruiseControlConfig are configurations for default implementation of the pluggbles (e.g. KafkaSampleStore). Personally I think it makes more sense to keep those configs in the specific implementation instead of putting them into the global configuration.

If we put those configurations into the global config, when user plug in another implementation, those global config will become unused and can easily cause confusions like "should I still set values for them". So IMO the global configuration should only have the configurations that are used by the framework itself regardless of the pluggins.

I know that currently we have some configurations there in the global config that only make sense to certain goals. It is a legacy issue and I actually think we should move them out to the corresponding goals.

kidkun:
Hey Becket, thanks for the comment.  I find you suggestion make sense, actually I ignore the fact that some component are pluggable and put everything in CruiseControlConfig does harm our long goal of generalizing Cruise Control.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
```diff
@@ -55,17 +55,10 @@ public AbstractGoal() {
...
-    String minMonitoredPartitionPercentageString =
-        (String) configs.get(KafkaCruiseControlConfig.MIN_VALID_PARTITION_RATIO_CONFIG);
-    if (minMonitoredPartitionPercentageString != null
-        && !minMonitoredPartitionPercentageString.isEmpty()) {
-      _minMonitoredPartitionPercentage = Double.parseDouble(minMonitoredPartitionPercentageString);
-    }
+    KafkaCruiseControlConfig parsedConfig = new KafkaCruiseControlConfig(configs, false);
+    _balancingConstraint = new BalancingConstraint(parsedConfig);
+    _numWindows = parsedConfig.getInt(KafkaCruiseControlConfig.NUM_PARTITION_METRICS_WINDOWS_CONFIG);
+    _minMonitoredPartitionPercentage = parsedConfig.getDouble(KafkaCruiseControlConfig.MIN_VALID_PARTITION_RATIO_CONFIG);
```
efeg: Do we still need to initialize `_numWindows` and `_minMonitoredPartitionPercentage` (see below) if they are guaranteed to be configured here?
```
  protected int _numWindows = 1;
  protected double _minMonitoredPartitionPercentage = 0.995;
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -65,22 +65,24 @@
...
   protected String _brokerMetricSampleStoreTopic;
+  protected Integer _metricSampleStoreTopicReplicationFactor;
   protected volatile double _loadingProgress;
   protected Producer<byte[], byte[]> _producer;
   protected volatile boolean _shutdown = false;
 
+  public static final String PARTITION_METRIC_SAMPLE_STORE_TOPIC_CONFIG = "partition.metric.sample.store.topic";
+  public static final String BROKER_METRIC_SAMPLE_STORE_TOPIC_CONFIG = "broker.metric.sample.store.topic";
+  public static final String NUM_SAMPLE_LOADING_THREADS_CONFIG = "num.sample.loading.threads";
+  public static final String NUM_METRIC_SAMPLE_STORE_TOPIC_REPLICATION_FACTOR_CONFIG = "num.metric.sample.store.topic.replication.factor";
```
efeg: Let's ensure that the functionality of each config copied to here is clear. Can we make it clear by adding the relevant config and its document as a JavaDoc for this class? -- e.g. for `NUM_SAMPLE_LOADING_THREADS_CONFIG` add `{@link #NUM_SAMPLE_LOADING_THREADS_CONFIG}: The config for the number of Kafka sample store consumer threads.` to the `JavaDoc` of this class.

efeg: It looks like we lost the default values for the configs copied from `KafkaCruiseControlConfig.java`. Can we ensure that the default values are specified here as well? -- e.g. for `PARTITION_METRIC_SAMPLE_STORE_TOPIC_CONFIG`, the default value is "__KafkaCruiseControlPartitionMetricSamples".

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -65,22 +65,24 @@
   protected ExecutorService _metricProcessorExecutor;
   protected String _partitionMetricSampleStoreTopic;
   protected String _brokerMetricSampleStoreTopic;
+  protected Integer _metricSampleStoreTopicReplicationFactor;
```
efeg: `_metricSampleStoreTopicReplicationFactor` -> `_sampleStoreTopicReplicationFactor`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/ReadOnlyKafkaSampleStore.java
```diff
@@ -21,17 +21,14 @@
...
-    _brokerMetricSampleStoreTopic = (String) config.get(KafkaCruiseControlConfig.BROKER_METRIC_SAMPLE_STORE_TOPIC_CONFIG);
-    if (_partitionMetricSampleStoreTopic == null
-        || _brokerMetricSampleStoreTopic == null
-        || _partitionMetricSampleStoreTopic.isEmpty()
-        || _brokerMetricSampleStoreTopic.isEmpty()) {
-      throw new IllegalArgumentException("The sample store topic names must be configured.");
-    }
-    String numProcessingThreadsString = (String) config.get(KafkaCruiseControlConfig.NUM_SAMPLE_LOADING_THREADS_CONFIG);
+    _partitionMetricSampleStoreTopic = KafkaCruiseControlUtils.getRequiredConfig(config, PARTITION_METRIC_SAMPLE_STORE_TOPIC_CONFIG);
+    _brokerMetricSampleStoreTopic = KafkaCruiseControlUtils.getRequiredConfig(config, BROKER_METRIC_SAMPLE_STORE_TOPIC_CONFIG);
```
efeg: Should we ensure that if the user did not provide the config, we use the default value for it (as we used to do) -- e.g. `DEFAULT_BROKER_METRIC_SAMPLE_STORE_TOPIC`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -149,7 +151,8 @@ private void ensureTopicsCreated(Map<String, ?> config) {
         throw new IllegalStateException(String.format("Kafka cluster has no alive brokers. (zookeeper.connect = %s",
                                                       config.get(KafkaCruiseControlConfig.ZOOKEEPER_CONNECT_CONFIG)));
       }
-      int replicationFactor = Math.min(2, numberOfBrokersInCluster);
+      int replicationFactor = _metricSampleStoreTopicReplicationFactor == null ? Math.min(2, numberOfBrokersInCluster) :
```
efeg: Should we avoid hardcoding values in code? -- e.g. `2` --> can we move it to a variable `DEFAULT_...`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
kidkun: for `PARTITION_METRIC_SAMPLE_STORE_TOPIC_CONFIG` and `BROKER_METRIC_SAMPLE_STORE_TOPIC_CONFIG` we use `KafkaCruiseControlUtils.getRequiredConfig` which will throw exception if user does not explicitly specify the topic name while they are using this implementation of `SampleStore`. I feel this is reasonable. 

for `NUM_SAMPLE_LOADING_THREADS_CONFIG` we have the default value(set to 8). 

for `NUM_METRIC_SAMPLE_STORE_TOPIC_REPLICATION_FACTOR_CONFIG`, it is set to `Math.min(2, numberOfBrokersInCluster)`.

I update the JavaDoc of this class to mention these are required configs. Do you think that is clear enough?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/ReadOnlyKafkaSampleStore.java
kidkun: I think this is reasonable. If we auto create some topic that user are unaware of, they might be surprised to find out later and have some unwanted operation(like delete the topic...)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
efeg: Sure, looks good to me!
-- let's make sure that we mention this behavior change in commit message.

PR #311, opened 2018-08-31
Correctly display 'Cluster load after rebalance'

mgrubent:
Previously, the 'Cluster load after rebalance' prints literal '%n%n' in the response from `cruise-control`
```
STD:{cpu:       0.083 networkInbound:    5949.215 networkOutbound:    3737.438 disk:  218015.239 potentialNwOut:    2972.991 replicas:281.9897012875318 topicReplicas:0.38622891576356444%n%nCluster load after rebalance:%n
```

This is likely because String.format() is not used for this particular string, so '%n' does not resolve into a literal newline character.

Thus, either use String.format() on the existing `"%n%nCluster load after rebalance:%n"`, which seems wasteful, or directly use a literal newline character in this line, so that the response from `cruise-control` displays correctly.


mgrubent:
Actually moving to a literal `\n` will only work for linux boxes

PR #324, opened 2018-09-12
Improve the detection of wrong broker ID in decommission broker request.

kidkun:
Currently we check the user provided broker ID after cluster model is generated, which is a expensive process. The patch moves this check before cluster model generation, thus fail faster for broker ids missing from the cluster in remove broker request.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -39,6 +39,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.utils.SystemTime;
 import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
@@ -163,6 +164,11 @@ public class KafkaCruiseControl {
     ModelCompletenessRequirements modelCompletenessRequirements =
         modelCompletenessRequirements(goalsByPriority.values()).weaker(requirements);
     try (AutoCloseable ignored = _loadMonitor.acquireForModelGeneration(operationProgress)) {
+      Cluster cluster = _loadMonitor.refreshClusterAndGeneration().cluster();
+      List<Integer> invalidBrokerIds = brokerIds.stream().filter(id -> cluster.nodeById(id) == null).collect(Collectors.toList());
+      if (!invalidBrokerIds.isEmpty()) {
+        throw new IllegalArgumentException(String.format("Broker %s does not exist.", invalidBrokerIds));
+      }
       ClusterModel clusterModel = _loadMonitor.clusterModel(_time.milliseconds(), modelCompletenessRequirements,
                                                             operationProgress);
       brokerIds.forEach(id -> clusterModel.setBrokerState(id, Broker.State.DEAD));
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -234,6 +234,7 @@ public class KafkaCruiseControl {
                                                   boolean skipHardGoalCheck,
                                                   Pattern excludedTopics) throws KafkaCruiseControlException {
     try (AutoCloseable ignored = _loadMonitor.acquireForModelGeneration(operationProgress)) {
+      sanityCheckBrokerPresence(brokerIds);
       sanityCheckHardGoalPresence(goals, skipHardGoalCheck);
       Map<Integer, Goal> goalsByPriority = goalsByPriority(goals);
       ModelCompletenessRequirements modelCompletenessRequirements =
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -162,6 +164,11 @@ private boolean isKafkaAssignerMode(List<String> goals) {
     ModelCompletenessRequirements modelCompletenessRequirements =
         modelCompletenessRequirements(goalsByPriority.values()).weaker(requirements);
     try (AutoCloseable ignored = _loadMonitor.acquireForModelGeneration(operationProgress)) {
+      Cluster cluster = _loadMonitor.refreshClusterAndGeneration().cluster();
```
efeg: Can we move this check to a util function -- e.g. `sanityCheckBrokerPresence(Set<Integer> brokers)`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -38,6 +38,8 @@
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
```
efeg: Overall, it seems like we can use the same sanity check for `demote` operation, as well.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -831,6 +831,9 @@ private boolean addOrRemoveBroker(HttpServletRequest request, HttpServletRespons
       concurrentLeaderMovements = concurrentMovementsPerBroker(request, false);
       skipHardGoalCheck = skipHardGoalCheck(request);
       excludedTopics = excludedTopics(request);
+      if (brokerIds.isEmpty()) {
```
efeg: It looks like whenever an endpoint supports `brokerIds`, it is a required parameter.
We would either (1) make this sanity check in `KafkaCruiseControlServletUtils#brokerIds(HttpServletRequest)`, or (2) ensure that all required sanity checks are performed after retrieval of brokerIds. Note that the same sanity check is missing in `#demoteBroker(HttpServletRequest, HttpServletResponse, EndPoint)`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -666,4 +669,16 @@ private void sanityCheckHardGoalPresence(List<String> goals, boolean skipHardGoa
       }
     }
   }
+
+  /**
+   * Sanity check whether the provided brokers to be decommissioned/demoted exist in cluster or not.
```
efeg: Does this method care whether the provided brokers will be decommissioned or demoted?

`Sanity check whether the provided brokers to be decommissioned/demoted exist in cluster or not.` -> `Sanity check whether the provided brokers exist in cluster or not.`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -320,6 +322,7 @@ public static void sanityCheckCapacityEstimation(boolean allowCapacityEstimation
       throws KafkaCruiseControlException {
     PreferredLeaderElectionGoal goal = new PreferredLeaderElectionGoal();
     try (AutoCloseable ignored = _loadMonitor.acquireForModelGeneration(operationProgress)) {
+      sanityCheckBrokerPresence(brokerIds);
```
efeg: Can we move `sanityCheckBrokerPresence(brokerIds);` out of the try-with-resources statement?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -163,6 +164,7 @@ private boolean isKafkaAssignerMode(List<String> goals) {
     ModelCompletenessRequirements modelCompletenessRequirements =
         modelCompletenessRequirements(goalsByPriority.values()).weaker(requirements);
     try (AutoCloseable ignored = _loadMonitor.acquireForModelGeneration(operationProgress)) {
+      sanityCheckBrokerPresence(brokerIds);
```
efeg: Can we move `sanityCheckBrokerPresence(brokerIds);` out of the try-with-resources statement?

kidkun: Since we are calling `_loadMonitor.refreshClusterAndGeneration()` in this method, I think the lock is needed.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
efeg: Sorry if what I refer to `sanity check` here was not clear. I mean the following sanity check regarding whether `brokerIds` is `empty` or not.
```
if (brokerIds.isEmpty()) {
    throw new IllegalArgumentException("Target broker ID is not provided.");
}
```
Would it be possible to do ^this sanity check in (1) only in `KafkaCruiseControlServletUtils#brokerIds(HttpServletRequest)` or (2) both in `#demoteBroker(HttpServletRequest, HttpServletResponse, EndPoint)` and `addOrRemoveBroker(HttpServletRequest, HttpServletResponse, EndPoint)`?

efeg:
Thanks for the updates! Left one more comment to a previous issue.

PR #336, opened 2018-09-20
Fix the bug in MetricSampleAggregator sample window rolling out

kidkun:
There is bug in `MetricSampleAggregator` that the maximal number window can be rolled out is ceilinged by maximal number of window kept in memory. This cause issue in scenario where sampling is resumed after long time pause and more window needs to be rolled out. In case `SamplingTask` will keep failing and Cruise Control is hang(unable to execute any operation).

### DIFF
--- a/cruise-control-core/src/test/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregatorTest.java
+++ b/cruise-control-core/src/test/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregatorTest.java
@@ -161,6 +161,25 @@ public class MetricSampleAggregatorTest {
   }
 
   @Test
+  public void testAddSamplesWithLargeInterval() {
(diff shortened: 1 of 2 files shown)


cruise-control-core/src/test/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregatorTest.java
```diff
@@ -160,6 +160,25 @@ public void testAvailableWindows() {
     }
   }
 
+  @Test
+  public void testAddSamplesWithLargeInterval() {
```
efeg: Would it be possible to add comments to clarify what we test here? -- I believe we had an offline discussion, but I haven't left a feedback here about it.

kidkun: Ah, sorry I forget that...added the explanation fo the new test.

Issue #355, opened 2018-10-08
Should not receive "Missing hard goals" exception if a non-existing goal is specified in the goals parameter


efeg:
If I input a goal name that is not supported by Cruise Control, I receive `Missing hard goals` exception. Instead I expect to receive an exception that says the goal is not supported -- e.g:

```
$curl -vv -X POST 'localhost:2540/kafkacruisecontrol/rebalance?dryrun=false&goals=NonExistingGoal'
```
gives the following:
```
java.lang.IllegalArgumentException: Missing hard goals [NetworkInboundCapacityGoal, DiskCapacityGoal, RackAwareGoal, NetworkOutboundCapacityGoal, CpuCapacityGoal, ReplicaCapacityGoal] in provided goal list [NonExistingGoal].
```

PR #356, opened 2018-10-08
Catch the exception if a non-existing goal is specified in the parameter.

efeg:
Fixes the issue https://github.com/linkedin/cruise-control/issues/355.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -675,9 +675,12 @@ public class KafkaCruiseControl {
    * @param supportedGoals Supported goals.
    */
   private void sanityCheckNonExistingGoal(List<String> goals, Map<String, Goal> supportedGoals) {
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -671,6 +669,17 @@ private void sanityCheckHardGoalPresence(List<String> goals, boolean skipHardGoa
...
+   * Sanity check whether the given goals exist in the given supported goals.
+   * @param goals A list of goals.
+   * @param supportedGoals Supported goals.
+   */
+  private void sanityCheckNonExistingGoal(List<String> goals, Map<String, Goal> supportedGoals) {
+    goals.stream().filter(goalName -> supportedGoals.get(goalName) == null).forEach(goalName -> {
+      throw new IllegalArgumentException("Goal: " + goalName + " does not exist. Supported: " + supportedGoals.keySet());
+    });
+  }
+
```
kidkun: Here does the exception throws when first non existent goal is found? I feel it would be better to return full list of invalid goals.

PR #359, opened 2018-10-10
Improve sample loading efficiency in Cruise Control booting up process.

kidkun:
Several bug fix and performance improvement in this path.
1. fix the bug in creating the broker metric sample topic(wrong config get picked up, result in wrong retention time of the topic)
2. fix the bug in loading partition/broker sample metric from corresponding topic(seek to the wrong beginning offset, result in doubling the Cruise Control boot up time)
3. improve the loading logic to reduce the contention between thread(to acquire the some lock) and burning CPU cycle(due to too short poll time)
4. improve the logging in sample loading process,  to help debugging if things goes wrong

### DIFF
--- a/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
+++ b/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
@@ -92,8 +92,9 @@ public class MetricSampleAggregator<G, E extends Entity<G>> extends LongGenerati
   protected final int _minSamplesPerWindow;
   protected final int _numWindowsToKeep;
   protected final long _windowMs;
+  protected final long _monitoringPeriodMs;
   protected final MetricDef _metricDef;
-  protected String _metricSampleType;
+  protected SampleType _sampleType;
 
   private volatile long _currentWindowIndex;
   private volatile long _oldestWindowIndex;
@@ -118,6 +119,7 @@ public class MetricSampleAggregator<G, E extends Entity<G>> extends LongGenerati
     _rawMetrics = new ConcurrentHashMap<>();
     _numWindows = numWindows;
     _windowMs = windowMs;
+    _monitoringPeriodMs = _numWindows * _windowMs;
     // We keep one more window for the active window.
     _numWindowsToKeep = _numWindows + 1;
     _minSamplesPerWindow = minSamplesPerWindow;
@@ -470,18 +472,19 @@ public class MetricSampleAggregator<G, E extends Entity<G>> extends LongGenerati
           // The first possible window index is actually 1 instead of 0.
           _oldestWindowIndex = Math.max(1, index - _numWindows);
           int numOldWindowIndexesToReset = (int) Math.min(_numWindowsToKeep, _oldestWindowIndex - prevOldestWindowIndex);
-          int numAbandonedSample = 0;
+          int numAbandonedSamples = 0;
           // Reset all the data starting from previous oldest window. After this point the old samples cannot get
           // into the raw metric values. We only need to reset the index if the new index is at least _numWindows;
           if (numOldWindowIndexesToReset > 0) {
-            numAbandonedSample = resetIndexes(prevOldestWindowIndex, numOldWindowIndexesToReset);
+            numAbandonedSamples = resetIndexes(prevOldestWindowIndex, numOldWindowIndexesToReset);
           }
           // Set the generation of the old current window.
           _aggregatorState.updateWindowGeneration(_currentWindowIndex, generation());
           // Lastly update current window.
           _currentWindowIndex = index;
-          LOG.info("{}Aggregator rolled out {} new windows, reset {} windows, current window range [{}, {}], abandon {} samples.", _metricSampleType,
-                   numWindowsToRollOut, numOldWindowIndexesToReset, _oldestWindowIndex * _windowMs, _currentWindowIndex * _windowMs, numAbandonedSample);
+          LOG.info("{} Aggregator rolled out {} new windows, reset {} windows, current window range [{}, {}], abandon {} samples.",
+                    _sampleType, numWindowsToRollOut, numOldWindowIndexesToReset, _oldestWindowIndex * _windowMs,
+                    _currentWindowIndex * _windowMs, numAbandonedSamples);
           return true;
         }
       } finally {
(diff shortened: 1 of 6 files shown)


cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/RawMetricValues.java
```diff
@@ -191,15 +191,17 @@ public synchronized void resetWindowIndexes(long startingWindowIndex, int numWin
         || inValidRange(startingWindowIndex + numWindowIndexesToReset - 1)) {
       throw new IllegalStateException("Should never reset a window index that is in the valid range");
     }
-    LOG.debug("Resetting window index [{}, {}]", startingWindowIndex,
-              startingWindowIndex + numWindowIndexesToReset - 1);
     // We are not resetting all the data here. The data will be interpreted to 0 if count is 0.
+    long numAbandonedSample = 0;
     for (long i = startingWindowIndex; i < startingWindowIndex + numWindowIndexesToReset; i++) {
       int index = arrayIndex(i);
+      numAbandonedSample += _counts[index];
```
efeg: `_counts[index]` is a `short`, whose max value is `(2^15)-1`.`long` has a max value of `(2^63)-1`. If each `_counts[index]` has the max value, this can handle about `(2^48)` entries in `_counts`. Do we need a `long` or would an `int` (i.e. supports max value of `(2^31)-1`) be sufficient -- `64K` entries `_counts` in the worst case?

kidkun: I feel `int` should be enough here. the number of sample we abandon is number of window to reset * `number of sample per window`, I do not expect this to be a huge number.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -336,7 +336,7 @@ public void run() {
         }
         while (!sampleLoadingFinished(endOffsets)) {
           try {
-            ConsumerRecords<byte[], byte[]> consumerRecords = _consumer.poll(10);
+            ConsumerRecords<byte[], byte[]> consumerRecords = _consumer.poll(1000);
```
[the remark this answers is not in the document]
kidkun: I feel this does not needs to be configured by user.  since here we seek to a previous offset and consumes to the current end offset, which guarantees that there will always be data back, we only needs to set the poll time large enough to prevent thread from burning CPU cycle.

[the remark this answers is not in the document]
efeg: I agree that the user should not configure it -- i.e. moving the hardcoded value to a variable is just for clarity.
```
private static final long SAMPLE_POLL_TIMEOUT = 1000L;
...
ConsumerRecords<byte[], byte[]> consumerRecords = _consumer.poll(SAMPLE_POLL_TIMEOUT);
```

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
```diff
@@ -567,4 +574,18 @@ private long windowIndex(long time) {
...
+  /**
+   * Get the length of time aggregator keeps sample in memory.
+   * It is used to determine the time length to look back(from current time) when loading the historical samples.
+   * @return length of time.
+   */
+  public long monitoringPeriodMs() {
+    return _monitoringPeriodMs;
+  }
+
+  protected enum SampleType {
```
efeg: Would it be helpful to add a `JavaDoc` here?

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
```diff
@@ -567,4 +574,18 @@ private long windowIndex(long time) {
   private E identity(E entity) {
     return _identityEntityMap.computeIfAbsent(entity, e -> entity);
   }
+
+  /**
+   * Get the length of time aggregator keeps sample in memory.
```
efeg: `sample` - > `samples`

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
```diff
@@ -567,4 +574,18 @@ private long windowIndex(long time) {
...
+   * It is used to determine the time length to look back(from current time) when loading the historical samples.
```
efeg: `It is used to determine the time length to look back(from current time) when loading the historical samples.` -> This seems to be one of many potential ways of using this method -- i.e. we can use it for other purposes in the future.
Would we avoid specifying our particular use-case in the JavaDoc of this generic method?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -384,6 +390,7 @@ public void run() {
             }
           }
         }
+        LOG.info("Metric loader finished loading samples.");
```
efeg: There are multiple metric loaders (i.e. as many as the number of consumers) . Can we add an identifier here to say something like `5 out of 8 metric loaders finished loading samples.`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -393,12 +400,37 @@ private boolean sampleLoadingFinished(Map<TopicPartition, Long> endOffsets) {
...
-          LOG.trace("Partition {} is still lagging. Current position: {}, LEO: {}", entry.getKey(),
+          LOG.debug("Partition {} is still lagging. Current position: {}, LEO: {}", entry.getKey(),
                     position, entry.getValue());
           return false;
         }
       }
       return true;
     }
+
+    private void prepareConsumerOffset() {
```
efeg: Would it be possible to add comments to this method to explain why we need preparation here? -- i.e. what exactly do we do in this method / what is the benefit of it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
kidkun: When this message is printed, the LOG4J does have the information of which  metric loader  thread(format like pool-`total_thread_number`-thread-`current_thread_number`) prepending to this message, so we can distinguish the threads and a `grep` command can help us locate which thread is actual lagging. I feel to print message like `5 out of 8` we need to pass more args to Runnable and needs a atomic variable sync between worker thread, which is kinda too much for just logging/debugging purpose.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -408,6 +408,11 @@ private boolean sampleLoadingFinished(Map<TopicPartition, Long> endOffsets) {
       return true;
     }
 
+    /**
+     * Config the sample loading consumers to consume from proper starting offsets. The sample store Kafka topic may contains datas
```
efeg: `contain data` -> `contain data`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -408,6 +408,11 @@ private boolean sampleLoadingFinished(Map<TopicPartition, Long> endOffsets) {
...
+     * which are too old for {@link com.linkedin.cruisecontrol.monitor.sampling.aggregator.MetricSampleAggregator} to keep in memory,
+     * to prevent loading these stale datas, manually seek the consumers' staring offset to the offset at proper timestamp.
```
efeg: `datas` -> `data`

Issue #275, opened 2018-07-19
Allow bind address to be specified


pmbuko:
I would like to be able to specify which address cruise-control binds to. I'd like to use localhost and then implement a reverse proxy for authentication. This could also be useful for choosing which of multiple network interfaces to bind to.

nareshv:
completed as per [PR 202](https://github.com/linkedin/cruise-control/pull/202)

PR #364, opened 2018-10-11
Allow bind address to be specified (#275)

nareshv:
By default the application listens on loopback address `(127.0.0.1)` and the users can override this behaviour by passing a hostname or ip address to listen on as a last argument to the server.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlMain.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlMain.java
@@ -23,6 +23,18 @@ import java.net.InetSocketAddress;
  */
 public class KafkaCruiseControlMain {
   private static final String METRIC_DOMAIN = "kafka.cruisecontrol";
+  // Default API protocol exposed by the webserver
+  private static final String DEFAULT_API_PROTOCOL = "http";
+  // IP Address on which the service is made available by default
+  private static final String DEFAULT_BIND_ADDRESS = "127.0.0.1";
+  // Port Number on which the service listens by default
+  private static final int DEFAULT_BIND_PORT = 9090;
+  // API Prefix (from server root) under which all resources are made available
+  private static final String DEFAULT_API_PATH_PREFIX = "/kafkacruisecontrol/*";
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlMain.java
```diff
@@ -47,6 +48,14 @@ public static void main(String[] args) throws Exception {
         printErrorMessageAndDie();
       }
     }
+    // Future ssl bits
+    String protocol = "http";
```
efeg: Nit: For clarity, would it be possible to move the hardcoded value to a variable? -- i.e. 
```
private static final String DEFAULT_PROTOCOL = "http";
...
String hostname = DEFAULT_PROTOCOL;
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlMain.java
```diff
@@ -47,6 +48,14 @@ public static void main(String[] args) throws Exception {
...
+
+    // For security reasons listen on the loopback address by default
+    String hostname = "127.0.0.1";
```
efeg: Nit: For clarity, would it be possible to move the hardcoded value to a variable? -- i.e. 
```
private static final String DEFAULT_HOSTNAME = "127.0.0.1";
...
String hostname = DEFAULT_HOSTNAME;
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlMain.java
nareshv: I was following the approach taken for `port` variable. But, lemme fix all the relevant ones as well.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlMain.java
nareshv: I was following the approach taken for `port` variable. But, lemme fix all the relevant ones as well.

PR #367, opened 2018-10-17
Add per broker swap timeout for resource distribution goals.

efeg:
This patch adds a hard-coded timeout for the per broker swap operations in resource distribution goals.
Future patches will remove this hardcoded value and make it configurable.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
@@ -489,7 +489,12 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
       candidateBrokerPQ.add(candidateBroker);
     }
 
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -574,7 +590,7 @@ private boolean rebalanceBySwappingLoadIn(Broker broker,
       candidateBrokerPQ.add(candidateBroker);
     }
 
-    while (!candidateBrokerPQ.isEmpty()) {
+    while (!candidateBrokerPQ.isEmpty() && remainingPerBrokerSwapTimeMs(swapStartTimeMs) > 0) {
```
kidkun: Could we move `remainingPerBrokerSwapTimeMs(swapStartTimeMs)` inside the loop and make it in an `if` statement. and if the condition holds, we do the same `LOG.debug("Swap load in timeout for broker {}.", broker.id());`. I feel it would help in we want to debug the optimization logic in the future in that under all scenario a message will get printed if we have optimization failed due to timeout.

PR #369, opened 2018-10-17
Make sample store topic's partition count configurable.

kidkun:


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -59,6 +59,16 @@
  *   set to {@link #DEFAULT_NUM_SAMPLE_LOADING_THREADS}.</li>
  *   <li>{@link #SAMPLE_STORE_TOPIC_REPLICATION_FACTOR_CONFIG}: The config for the replication factor of Kafka sample store topics,
  *   default value is set to {@link #DEFAULT_SAMPLE_STORE_TOPIC_REPLICATION_FACTOR}.</li>
+ *   <li>{@link #PARTITION_SAMPLE_STORE_TOPIC_PARTITION_COUNT_CONFIG}: The config for the number of partition for Kafka partition sample store
```
efeg: Can we also 
(1) make `MIN_SAMPLE_TOPIC_RETENTION_TIME_MS` configurable, and
(2) set its default value in a `DEFAULT_...` variable?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -59,6 +59,16 @@
...
+ *    topic, default value is set to {@link #DEFAULT_PARTITION_SAMPLE_STORE_TOPIC_PARTITION_COUNT}.</li>
+ *   <li>{@link #BROKER_SAMPLE_STORE_TOPIC_PARTITION_COUNT_CONFIG}: The config for the number of partition for Kafka broker sample store topic
+ *
+ *
+ *
+ *
+ *
+ *   ,
```
efeg: Redundant spaces in JavaDoc -- i.e.
```
 *
 *
 *
 *
 *
 *   ,
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -88,14 +102,22 @@
...
   @Override
   public void configure(Map<String, ?> config) {
     _partitionMetricSampleStoreTopic = KafkaCruiseControlUtils.getRequiredConfig(config, PARTITION_METRIC_SAMPLE_STORE_TOPIC_CONFIG);
     _brokerMetricSampleStoreTopic = KafkaCruiseControlUtils.getRequiredConfig(config, BROKER_METRIC_SAMPLE_STORE_TOPIC_CONFIG);
     String metricSampleStoreTopicReplicationFactorString = (String) config.get(SAMPLE_STORE_TOPIC_REPLICATION_FACTOR_CONFIG);
     _sampleStoreTopicReplicationFactor = metricSampleStoreTopicReplicationFactorString == null || metricSampleStoreTopicReplicationFactorString.isEmpty()
-                               ? null : Integer.parseInt(metricSampleStoreTopicReplicationFactorString);
+                                         ? null : Integer.parseInt(metricSampleStoreTopicReplicationFactorString);
+    String partitionSampleStoreTopicPartitionCountString = (String) config.get(PARTITION_SAMPLE_STORE_TOPIC_PARTITION_COUNT_CONFIG);
+    _partitionSampleStoreTopicPartitionCount = partitionSampleStoreTopicPartitionCountString == null || partitionSampleStoreTopicPartitionCountString.isEmpty()
```
efeg: Broken indentation for ternary operation -- would the following change help with the indentation and avoiding too long time?
```
    _partitionSampleStoreTopicPartitionCount = partitionSampleStoreTopicPartitionCountString == null 
                                               || partitionSampleStoreTopicPartitionCountString.isEmpty()
                                               ? DEFAULT_PARTITION_SAMPLE_STORE_TOPIC_PARTITION_COUNT
                                               : Integer.parseInt(partitionSampleStoreTopicPartitionCountString);
    String brokerSampleStoreTopicPartitionCountString = (String) config.get(BROKER_SAMPLE_STORE_TOPIC_PARTITION_COUNT_CONFIG);
    _brokerSampleStoreTopicPartitionCount = brokerSampleStoreTopicPartitionCountString == null 
                                            || brokerSampleStoreTopicPartitionCountString.isEmpty()
                                            ? DEFAULT_BROKER_SAMPLE_STORE_TOPIC_PARTITION_COUNT 
                                            : Integer.parseInt(brokerSampleStoreTopicPartitionCountString);
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,20 +192,20 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
       ensureTopicCreated(zkUtils, topics.keySet(), _brokerMetricSampleStoreTopic, brokerSampleRetentionMs,
-                         replicationFactor);
+                         replicationFactor, _brokerSampleStoreTopicPartitionCount);
     } finally {
       KafkaCruiseControlUtils.closeZkUtilsWithTimeout(zkUtils, 10000);
     }
   }
 
-  private void ensureTopicCreated(ZkUtils zkUtils, Set<String> allTopics, String topic, long retentionMs, int replicationFactor) {
+  private void ensureTopicCreated(ZkUtils zkUtils, Set<String> allTopics, String topic, long retentionMs, int replicationFactor, int partitionCount) {
```
efeg: Just noticed that if the topic has not been created before, then we create it with the given props. However, if the topics were already there, then the `replication factor` and `partition count` are ignored.
Would we ensure that these configs also apply to the created topics?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
kidkun: This is weird...

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -88,14 +108,38 @@
...
+                                               : Integer.parseInt(partitionSampleStoreTopicPartitionCountString);
+    String brokerSampleStoreTopicPartitionCountString = (String) config.get(BROKER_SAMPLE_STORE_TOPIC_PARTITION_COUNT_CONFIG);
+    _brokerSampleStoreTopicPartitionCount = brokerSampleStoreTopicPartitionCountString == null
+                                            || brokerSampleStoreTopicPartitionCountString.isEmpty()
+                                            ? DEFAULT_BROKER_SAMPLE_STORE_TOPIC_PARTITION_COUNT
+                                            : Integer.parseInt(brokerSampleStoreTopicPartitionCountString);
+    String minPartitionSampleStoreTopicRetentionTimeMsString = (String) config.get(MIN_PARTITION_SAMPLE_STORE_TOPIC_RETENTION_TIME_MS_CONFIG);
+    _minPartitionSampleStoreTopicRetentionTimeMs = minPartitionSampleStoreTopicRetentionTimeMsString == null
+                                                  || minPartitionSampleStoreTopicRetentionTimeMsString.isEmpty()
+                                                  ? DEFAULT_MIN_PARTITION_SAMPLE_STORE_TOPIC_RETENTION_TIME_MS :
```
efeg: Nit for code-style consistency: colon (i.e. `:`) should be on the next line.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -88,14 +108,38 @@
...
+                                                  Integer.parseInt(minPartitionSampleStoreTopicRetentionTimeMsString);
+    String minBrokerSampleStoreTopicRetentionTimeMsString = (String) config.get(MIN_BROKER_SAMPLE_STORE_TOPIC_RETENTION_TIME_MS_CONFIG);
+    _minBrokerSampleStoreTopicRetentionTimeMs = minBrokerSampleStoreTopicRetentionTimeMsString == null
+                                               || minBrokerSampleStoreTopicRetentionTimeMsString.isEmpty()
+                                               ? DEFAULT_MIN_BROKER_SAMPLE_STORE_TOPIC_RETENTION_TIME_MS :
```
efeg: Nit for code-style consistency: colon (i.e. `:`) should be on the next line.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +214,57 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
       ensureTopicCreated(zkUtils, topics.keySet(), _brokerMetricSampleStoreTopic, brokerSampleRetentionMs,
-                         replicationFactor);
+                         replicationFactor, _brokerSampleStoreTopicPartitionCount);
     } finally {
       KafkaCruiseControlUtils.closeZkUtilsWithTimeout(zkUtils, 10000);
     }
   }
 
-  private void ensureTopicCreated(ZkUtils zkUtils, Set<String> allTopics, String topic, long retentionMs, int replicationFactor) {
+  private void increaseTopicReplicaFactor(ZkUtils zkUtils, MetadataResponse.TopicMetadata topicMetadata, int replicationFactor,
```
efeg: Bad indentation.
```
  private void increaseTopicReplicaFactor(ZkUtils zkUtils,
                                          MetadataResponse.TopicMetadata topicMetadata,
                                          int replicationFactor,
                                          String topic,
                                          Properties props) {
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +214,57 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+      String topic, Properties props) {
+    List<Integer> brokers = new ArrayList<>();
+    JavaConversions.seqAsJavaList(AdminUtils.getBrokerMetadatas(zkUtils, RackAwareMode.Safe$.MODULE$, Option.empty()))
+    .stream().forEach(bm -> brokers.add(bm.id()));
```
efeg: Bad indentation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +214,57 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+      // Make sure the current replicas are in new replica list.
+      pm.replicas().forEach(node -> newAssignedReplica.add(node.id()));
+      // Add new replica to partition in round-robin way.
+      while (newAssignedReplica.size() < replicationFactor) {
+        if (!newAssignedReplica.contains(brokers.get(cursor))) {
+          newAssignedReplica.add(brokers.get(cursor));
+        }
+        cursor = (cursor + 1) % brokers.size();
+      }
+      ((scala.collection.mutable.HashMap<Object, Seq<Object>>) newReplicaAssignment).put(pm.partition(),
```
efeg: Would the following tip work:
`Cast may be removed by changing the type of 'newReplicaAssignment' to 'HashMap<Object, Seq<Object>>'`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +214,57 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+          JavaConverters.asScalaIteratorConverter(newAssignedReplica.iterator()).asScala().toSeq());
```
efeg: bad indentation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +214,57 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
     Properties props = new Properties();
     props.setProperty(LogConfig.RetentionMsProp(), Long.toString(retentionMs));
     props.setProperty(LogConfig.CleanupPolicyProp(), DEFAULT_CLEANUP_POLICY);
     if (!allTopics.contains(topic)) {
-      AdminUtils.createTopic(zkUtils, topic, 32, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
+      AdminUtils.createTopic(zkUtils, topic, partitionCount, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
     } else {
       AdminUtils.changeTopicConfig(zkUtils, topic, props);
+      MetadataResponse.TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(topic, zkUtils);
+      if (replicationFactor > topicMetadata.partitionMetadata().get(0).replicas().size()) {
```
efeg: The replication factor of partitions of the same topic might be inconsistent due to unclean interruption of replica reassignment -- i.e. deletion of `reassign_partitions` znode. I suspect that we might need to check all partitions of the topic -- not just the one in index `0`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +214,60 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
     Properties props = new Properties();
     props.setProperty(LogConfig.RetentionMsProp(), Long.toString(retentionMs));
     props.setProperty(LogConfig.CleanupPolicyProp(), DEFAULT_CLEANUP_POLICY);
     if (!allTopics.contains(topic)) {
-      AdminUtils.createTopic(zkUtils, topic, 32, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
+      AdminUtils.createTopic(zkUtils, topic, partitionCount, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
     } else {
       AdminUtils.changeTopicConfig(zkUtils, topic, props);
+      MetadataResponse.TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(topic, zkUtils);
+      int currentReplicaFactor = topicMetadata.partitionMetadata().stream().mapToInt(pm -> pm.replicas().size()).max().orElse(0);
```
efeg: Would you clarify why `.orElse(0)` is the right action here? I suspect that throwing a `IllegalStateException` in case `max` is empty.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +215,65 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
     Properties props = new Properties();
     props.setProperty(LogConfig.RetentionMsProp(), Long.toString(retentionMs));
     props.setProperty(LogConfig.CleanupPolicyProp(), DEFAULT_CLEANUP_POLICY);
     if (!allTopics.contains(topic)) {
-      AdminUtils.createTopic(zkUtils, topic, 32, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
+      AdminUtils.createTopic(zkUtils, topic, partitionCount, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
     } else {
       AdminUtils.changeTopicConfig(zkUtils, topic, props);
+      MetadataResponse.TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(topic, zkUtils);
+      OptionalInt currentReplicaFactor = topicMetadata.partitionMetadata().stream().mapToInt(pm -> pm.replicas().size()).max();
```
efeg: Two points here:
(1) I think we should get the `min()`, not the `max()` here. For example,  if I would like to set the replication factor to 3 and if this topic has 2 partitions -- one with RF 2 one with RF 3, we should increase the replication factor to 3 for the partition which has an RF 2. Currently we are ignoring such partitions.

(2) Also I suspect that we need to ensure that there is not an ongoing replica move for any partition of this topic. Because if there is a replica move, then we can see that the number of replicas for the partition is more than its actual replication factor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +215,65 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+      if (!currentReplicaFactor.isPresent()) {
+        throw new IllegalStateException("Kafka topic " + topic + " seems does not have any partition from latest metadata : " + topicMetadata + ".");
+      } else if (currentReplicaFactor.getAsInt() <= 0) {
```
efeg: I think the second check is redundant -- i.e. we are checking the size of `replicas` list in `org.apache.kafka.common.requests.MetadataResponse#PartitionMetadata` here. A partition does not (i..e cannot) exist without a replica.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +215,65 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
     props.setProperty(LogConfig.CleanupPolicyProp(), DEFAULT_CLEANUP_POLICY);
     if (!allTopics.contains(topic)) {
-      AdminUtils.createTopic(zkUtils, topic, 32, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
+      AdminUtils.createTopic(zkUtils, topic, partitionCount, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
     } else {
       AdminUtils.changeTopicConfig(zkUtils, topic, props);
+      MetadataResponse.TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(topic, zkUtils);
+      OptionalInt currentReplicaFactor = topicMetadata.partitionMetadata().stream().mapToInt(pm -> pm.replicas().size()).max();
+      if (!currentReplicaFactor.isPresent()) {
+        throw new IllegalStateException("Kafka topic " + topic + " seems does not have any partition from latest metadata : " + topicMetadata + ".");
```
efeg: Let's avoid using space before `:`.
Would the following change make our message more simple:

`throw new IllegalStateException("Kafka topic " + topic + " seems does not have any partition from latest metadata : " + topicMetadata + ".");` -> `throw new IllegalStateException("Kafka topic " + topic + " has no partition (Metadata: " + topicMetadata + ").");`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
efeg: Could you clarify what `currentReplicaFactor` means? Did you intend to say `currentReplicationFactor` or should this say `maxReplicationFactor`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +215,65 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+      throw new RuntimeException("Unable to increase topic " + topic + " replica factor to " + replicationFactor +
+                                 " since there are only " + brokers.size() + " brokers in the cluster.");
+    }
+    scala.collection.mutable.Map<Object, Seq<Object>> newReplicaAssignment = new scala.collection.mutable.HashMap<>();
+    int cursor = 0;
+    for (MetadataResponse.PartitionMetadata pm : topicMetadata.partitionMetadata()) {
+      List<Object> newAssignedReplica = new ArrayList<>();
+      // Make sure the current replicas are in new replica list.
+      pm.replicas().forEach(node -> newAssignedReplica.add(node.id()));
+      // Add new replica to partition in round-robin way.
```
efeg: I am not sure if this way of adding replicas preserves rack-awareness. Can you elaborate on that?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
kidkun: Good catch. I think replica number inconsistency for partitions of the same topic happens during replica movement due to temporal expand AR/ISR to include new replica, in this scenario the `min` replica number is the real replication factor.  so we should first ensure there is not ongoing operation then check replication factor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
kidkun: good catch, this assignment only make sure replica on different broker.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +217,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+          cursors[rackCursor] = (cursor + 1) % brokersByRack.get(rack).size();
+        }
+        rackCursor = (rackCursor + 1) % racks.size();
+      }
+      newReplicaAssignment.put(pm.partition(), JavaConverters.asScalaIteratorConverter(newAssignedReplica.iterator()).asScala().toSeq());
+    }
+    AdminUtils.createOrUpdateTopicPartitionAssignmentPathInZK(zkUtils, topic, newReplicaAssignment, props, true);
+  }
+
+  private void ensureTopicNotUnderPartitionReassignment(ZkUtils zkUtils, String topic) {
```
efeg: Seems like we would benefit from moving this function to some utils class.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +217,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+    int attempt = 0;
+    while (JavaConversions.asJavaCollection(zkUtils.getPartitionsBeingReassigned().keys())
+          .stream().anyMatch(tp -> tp.topic().equals(topic))) {
+      try {
+        sleep(1000 << attempt);
+      } catch (InterruptedException e) {
+        // Let it go.
+      }
+      if (attempt++ == 10) {
```
efeg: Would the following change help us match the number of attempts with the given limit? `attempt++` -> `++attempt`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +217,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+      }
+      newReplicaAssignment.put(pm.partition(), JavaConverters.asScalaIteratorConverter(newAssignedReplica.iterator()).asScala().toSeq());
+    }
+    AdminUtils.createOrUpdateTopicPartitionAssignmentPathInZK(zkUtils, topic, newReplicaAssignment, props, true);
+  }
+
+  private void ensureTopicNotUnderPartitionReassignment(ZkUtils zkUtils, String topic) {
+    int attempt = 0;
+    while (JavaConversions.asJavaCollection(zkUtils.getPartitionsBeingReassigned().keys())
+          .stream().anyMatch(tp -> tp.topic().equals(topic))) {
```
efeg: bad indentation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +217,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+    Map<Integer, String> rackByBroker = new HashMap<>();
+    for (BrokerMetadata bm:
+         JavaConversions.seqAsJavaList(AdminUtils.getBrokerMetadatas(zkUtils, RackAwareMode.Enforced$.MODULE$, Option.empty()))) {
+      String rack = bm.rack().get();
+      brokersByRack.putIfAbsent(rack, new ArrayList<>());
+      brokersByRack.get(rack).add(bm.id());
+      rackByBroker.put(bm.id(), rack);
+    }
+    if (replicationFactor > brokersByRack.size()) {
+      throw new RuntimeException("Unable to increase topic " + topic + " replica factor to " + replicationFactor +
```
efeg: Nit: For code consistency, can we move `+` on the new line.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +217,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+  private void increaseTopicReplicationFactor(ZkUtils zkUtils,
+                                              MetadataResponse.TopicMetadata topicMetadata,
+                                              int replicationFactor,
+                                              String topic,
+                                              Properties props) {
+    Map<String, List<Integer>> brokersByRack = new HashMap<>();
+    Map<Integer, String> rackByBroker = new HashMap<>();
+    for (BrokerMetadata bm:
+         JavaConversions.seqAsJavaList(AdminUtils.getBrokerMetadatas(zkUtils, RackAwareMode.Enforced$.MODULE$, Option.empty()))) {
+      String rack = bm.rack().get();
```
efeg: Rack might not be specified. In such cases, the `LoadMonitor` uses the `host` info as `rack` info. Can we follow a similar approach here? (Please see https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java#L449)

kidkun: `AdminUtils` does not have any method return host information of broker, so here I use `broker id` if rack info is missing.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java
```diff
@@ -140,4 +143,19 @@ public static int totalNumPartitions(Cluster cluster) {
     }
     return totalNumPartitions;
   }
+
+  public static void ensureTopicNotUnderPartitionReassignment(ZkUtils zkUtils, String topic) {
+    int attempt = 0;
+    while (JavaConversions.asJavaCollection(zkUtils.getPartitionsBeingReassigned().keys()).stream()
```
efeg: Let's maybe discuss the indentation configs offline -- i.e.:
```
    while (JavaConversions.asJavaCollection(zkUtils.getPartitionsBeingReassigned().keys()).stream()
                          .anyMatch(tp -> tp.topic().equals(topic))) {
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -234,14 +234,15 @@ private void increaseTopicReplicationFactor(ZkUtils zkUtils,
...
       rackByBroker.put(bm.id(), rack);
     }
     if (replicationFactor > brokersByRack.size()) {
-      throw new RuntimeException("Unable to increase topic " + topic + " replica factor to " + replicationFactor +
-                                 " since there are only " + brokersByRack.size() + " racks in the cluster.");
+      throw new RuntimeException("Unable to increase topic " + topic + " replica factor to " + replicationFactor
+                                 + " since there are only " + brokersByRack.size() + " racks in the cluster.");
     }
 
     scala.collection.mutable.Map<Object, Seq<Object>> newReplicaAssignment = new scala.collection.mutable.HashMap<>();
```
efeg:
```
List<String> racks = new ArrayList<>();
racks.addAll(brokersByRack.keySet());
```
 -> `List<String> racks = new ArrayList<>(brokersByRack.keySet());`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -234,14 +234,15 @@ private void increaseTopicReplicationFactor(ZkUtils zkUtils,
     Map<Integer, String> rackByBroker = new HashMap<>();
     for (BrokerMetadata bm:
          JavaConversions.seqAsJavaList(AdminUtils.getBrokerMetadatas(zkUtils, RackAwareMode.Enforced$.MODULE$, Option.empty()))) {
-      String rack = bm.rack().get();
+      // If the rack is not specified, we use the broker id info as rack info.
+      String rack = bm.rack().isEmpty() ? String.valueOf(bm.id()) : bm.rack().get();
```
efeg: But then if a host has multiple brokers, we will violate rack awareness. 
Can we either use (1) metadata request as used in `LoadMonitor` or (2) explicitly specify this limitation in the documentation of this method?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +217,82 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+    }
+
+    scala.collection.mutable.Map<Object, Seq<Object>> newReplicaAssignment = new scala.collection.mutable.HashMap<>();
+    List<String> racks = new ArrayList<>();
+    racks.addAll(brokersByRack.keySet());
+    int [] cursors = new int[racks.size()];
+    int rackCursor = 0;
+    for (MetadataResponse.PartitionMetadata pm : topicMetadata.partitionMetadata()) {
+      List<Object> newAssignedReplica = new ArrayList<>();
+      List<String> currentOccupiedRack = new ArrayList<>();
```
efeg: Do we need a `List` or could this be a `Set`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +219,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
     Properties props = new Properties();
     props.setProperty(LogConfig.RetentionMsProp(), Long.toString(retentionMs));
     props.setProperty(LogConfig.CleanupPolicyProp(), DEFAULT_CLEANUP_POLICY);
     if (!allTopics.contains(topic)) {
-      AdminUtils.createTopic(zkUtils, topic, 32, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
+      AdminUtils.createTopic(zkUtils, topic, partitionCount, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
     } else {
+      ensureTopicNotUnderPartitionReassignment(zkUtils, topic);
       AdminUtils.changeTopicConfig(zkUtils, topic, props);
+      MetadataResponse.TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(
```
efeg: Why don't we use `MetadataResponse.TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(topic, zkUtils);` here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +219,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+
+  private void ensureTopicCreated(ZkUtils zkUtils, Set<String> allTopics, String topic, long retentionMs, int replicationFactor, int partitionCount) {
     Properties props = new Properties();
     props.setProperty(LogConfig.RetentionMsProp(), Long.toString(retentionMs));
     props.setProperty(LogConfig.CleanupPolicyProp(), DEFAULT_CLEANUP_POLICY);
     if (!allTopics.contains(topic)) {
-      AdminUtils.createTopic(zkUtils, topic, 32, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
+      AdminUtils.createTopic(zkUtils, topic, partitionCount, replicationFactor, props, RackAwareMode.Safe$.MODULE$);
     } else {
+      ensureTopicNotUnderPartitionReassignment(zkUtils, topic);
```
efeg: If there is an ongoing execution that moves partitions of this topic, Cruise Control will fail to start. I believe we should prioritize availability over correctness here -- i.e. we should not fail to start CC.
One potential solution is to log this as an `ERROR`, not change the configs of the existing topic, but start the Sample Store.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +219,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
       AdminUtils.changeTopicConfig(zkUtils, topic, props);
+      MetadataResponse.TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(
+          JavaConversions.asScalaSet(new HashSet<>(Arrays.asList(topic))),
+          zkUtils,
+          ListenerName.forSecurityProtocol(SecurityProtocol.PLAINTEXT)).head();
+      OptionalInt currentReplicationFactor = topicMetadata.partitionMetadata().stream().mapToInt(pm -> pm.replicas().size()).min();
+      if (!currentReplicationFactor.isPresent()) {
+        throw new IllegalStateException("Kafka topic " + topic + " has no partition (Metadata: " + topicMetadata + ").");
```
efeg: Please refer to my previous comment regarding failing to start due to exceptions for changing configs of the created topics.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +219,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+    for (BrokerMetadata bm:
+         JavaConversions.seqAsJavaList(AdminUtils.getBrokerMetadatas(zkUtils, RackAwareMode.Enforced$.MODULE$, Option.empty()))) {
+      // If the rack is not specified, we use the broker id info as rack info.
+      String rack = bm.rack().isEmpty() ? String.valueOf(bm.id()) : bm.rack().get();
+      brokersByRack.putIfAbsent(rack, new ArrayList<>());
+      brokersByRack.get(rack).add(bm.id());
+      rackByBroker.put(bm.id(), rack);
+    }
+    if (replicationFactor > brokersByRack.size()) {
+      throw new RuntimeException("Unable to increase topic " + topic + " replica factor to " + replicationFactor
```
efeg: Please refer to my previous comment regarding failing to start due to exceptions for changing configs of the created topics.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -170,22 +219,96 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+   * @param props The properties to set for the topic.
+   */
+  private void increaseTopicReplicationFactor(ZkUtils zkUtils,
+                                              MetadataResponse.TopicMetadata topicMetadata,
+                                              int replicationFactor,
+                                              String topic,
+                                              Properties props) {
+    Map<String, List<Integer>> brokersByRack = new HashMap<>();
+    Map<Integer, String> rackByBroker = new HashMap<>();
+    for (BrokerMetadata bm:
```
efeg: Missing space before `:`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
kidkun: This is because in a later version of the Kafka,  method `fetchTopicMetadataFromZk(topic: String, zkUtils: ZkUtils)` is removed from AdminUtils. So to make CC compatible with more Kafka versions, I switch to this method(which is available in more versions).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java
```diff
@@ -140,4 +143,19 @@ public static int totalNumPartitions(Cluster cluster) {
...
+                          .anyMatch(tp -> tp.topic().equals(topic))) {
+      try {
+        sleep(1000 << attempt);
+      } catch (InterruptedException e) {
+        // Let it go.
+      }
+      if (++attempt == 10) {
+        throw new IllegalStateException("Kafka topic " + topic + " has stuck partition reassignment task.");
```
efeg: `stuck` --> `ongoing`

PR #372, opened 2018-10-20
fix incorrect const string

craynic:
broker.metrics.windows.ms => broker.metrics.window.ms
this const string is different from its name, and also different from that in the sample config.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -135,7 +135,7 @@
   /**
    * <code>broker.metrics.windows.ms</code>
```
efeg: Nit: can we also update the relevant comment: `broker.metrics.windows.ms` -> `broker.metrics.window.ms`?

Issue #397, opened 2018-11-09
Executor state would stuck at STARTING_EXECUTION if CC metric sampler is stuck at SAMPLING state


efeg:
Assume the following scenario:
1) The user sends `remove_broker` request.
2) In the meantime, the cluster is in a bad state, so Cruise Control metric fetcher keeps getting `java.util.concurrent.TimeoutException` exceptions while attempting to consume from metrics reporter topic.
3) After getting a `timeout`, the metric fetcher starts the next sampling request and goes back to step 2 -- i.e. stuck between step 2 and 3.
4) The executor in the meantime is stuck in `STARTING_EXECUTION` state due to being unable to perform `_loadMonitor.pauseMetricSampling();` -- because every time it checks the load monitor is in `SAMPLING` state (See the [relevant code](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L324)).

Some mitigation approaches...
* `MetricFetcherManager#fetchSamples()` can only catch `catch (Exception e)` for `TimeoutException`. We should catch and log `TimeoutException` separately (See the [relevant code](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/MetricFetcherManager.java#L256)).
* We should set consumer `timeout` to a value strictly smaller than the sampling period.
* We should not start a consecutive sampling if there is a pending `EXECUTION_STARTED` state for the executor.

PR #409, opened 2018-11-15
Fix for Executor state would stuck at STARTING_EXECUTION if CC metric sampler is stuck at SAMPLING state

alexwang08:
The fix added flag to indicate there's a pending Execution request. The LoadMonitor will not kick off consecutive sampling if it sees the flag is set. 
It's noticed that if MetadataClient.refreshMetadata() timeout (KafkaCruiseControlConfig.METADATA_MAX_AGE_CONFIG) is longer than sampling period for SamplingTask (KafkaCruiseControlConfig.METRIC_SAMPLING_INTERVAL_MS_CONFIG ). We need to make sure the timeout is no longer than the sampling period because the code will anyways throw TimeoutException if execution takes longer than sampling period. There's no need to wait longer than necessary. 

-- Fixes the issue https://github.com/linkedin/cruise-control/issues/397.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
@@ -1010,6 +1010,7 @@ public class KafkaCruiseControlConfig extends AbstractConfig {
           "] to be longer than sampling period [" + samplingPeriodMs + "].");
     }
   }
+
   public KafkaCruiseControlConfig(Map<?, ?> originals) {
     super(CONFIG, originals);
     sanityCheckGoalNames();
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
@@ -558,7 +558,7 @@ public class KafkaCruiseControlConfig extends AbstractConfig {
                 CommonClientConfigs.RECEIVE_BUFFER_DOC)
         .define(RECONNECT_BACKOFF_MS_CONFIG, ConfigDef.Type.LONG, 50L, atLeast(0L), ConfigDef.Importance.LOW,
                 CommonClientConfigs.RECONNECT_BACKOFF_MS_DOC)
-        .define(METADATA_MAX_AGE_CONFIG, ConfigDef.Type.LONG, 50 * 1000, atLeast(0), ConfigDef.Importance.LOW,
+        .define(METADATA_MAX_AGE_CONFIG, ConfigDef.Type.LONG, 55 * 1000, atLeast(0), ConfigDef.Importance.LOW,
                 METADATA_MAX_AGE_DOC)
         .define(CONNECTIONS_MAX_IDLE_MS_CONFIG,
                 ConfigDef.Type.LONG,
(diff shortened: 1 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -332,15 +333,16 @@ public Cluster kafkaCluster() {
    */
   public void pauseMetricSampling() {
     _loadMonitorTaskRunner.pauseSampling();
+    _loadMonitorTaskRunner.setAwaitExecution(true);
```
efeg: `pauseMetricSampling ()` and `resumeMetricSampling ()` are not explicitly called by the `Executor` while starting an execution. Users can also use the REST API to resume/pause sampling. Hence, the assumption regarding `setAwaitExecution(true)` here and `setAwaitExecution(false)` in `resumeMetricSampling()` is incorrect.

A potential mitigation is to change this function's signature to get a `boolean` to indicate whether an execution is starting or not.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -340,7 +340,6 @@ public void pauseMetricSampling() {
   public void resumeMetricSampling() {
     _loadMonitorTaskRunner.resumeSampling();
   }
-
```
efeg: Nit: Let's keep the space between functions.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -262,21 +264,29 @@ public void shutdown() {
   /**
    * Pause the scheduled sampling tasks..
    */
-  public void pauseSampling() {
+  public synchronized void pauseSampling() {
     if (_state.get() != PAUSED && !_state.compareAndSet(RUNNING, PAUSED)) {
       throw new IllegalStateException("Cannot pause the load monitor because it is in " + _state.get() + " state.");
     }
+    _awaitExecution = true;
```
efeg: If we throw an exception before setting this variable (e.g. this will happen when in `SAMPLING` state), the sampler would be unable set `_awaitExecution` to `true`. Also, do we need to set it if we can successfully set state to `PAUSED`?
Hence, should we move this assignment to `true` inside the `if`, just before the exception and reset it if we can successfully pause? -- i.e.

```
    if (_state.get() != PAUSED && !_state.compareAndSet(RUNNING, PAUSED)) {
      _awaitingPauseSampling = true;
      throw new IllegalStateException("Cannot pause the load monitor because it is in " + _state.get() + " state.");
    } else {
      _awaitingPauseSampling = false;
    }
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -45,6 +45,7 @@
 
   private AtomicReference<LoadMonitorTaskRunnerState> _state;
   private volatile double _bootstrapProgress;
+  private volatile boolean _awaitExecution;
```
efeg: `_awaitExecution` -> `_awaitingPauseSampling`.

Note: I feel like keeping track of an "awaiting pause sampling request" is what we actually want. Such a request can come from 2 sources:
1) Executor can send this command upon starting an execution.
2) The user of Cruise Control can send it via a REST API call to `pauseSampling`.

In either case, we would like to ensure that an ongoing sampling process would know about such a pending request.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -262,21 +264,29 @@ public void shutdown() {
...
 
   /**
    * Resume the scheduled sampling tasks.
    */
-  public void resumeSampling() {
+  public synchronized void resumeSampling() {
     if (_state.get() != RUNNING && !_state.compareAndSet(PAUSED, RUNNING)) {
       throw new IllegalStateException("Cannot resume the load monitor because it is in " + _state.get() + " state");
     }
+    _awaitExecution = false;
```
efeg: With the changes we have in `pauseSampling()`, I think we should remove this variable from here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -262,21 +264,29 @@ public void shutdown() {
...
       throw new IllegalStateException("Cannot pause the load monitor because it is in " + _state.get() + " state.");
     }
+    _awaitExecution = true;
   }
 
   /**
    * Resume the scheduled sampling tasks.
    */
-  public void resumeSampling() {
+  public synchronized void resumeSampling() {
```
efeg: This particular method should no longer required to be `synchronized` with the changes in method contents -- i.e. they do not share the variable to be set.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -262,21 +264,29 @@ public void shutdown() {
...
     if (_state.get() != RUNNING && !_state.compareAndSet(PAUSED, RUNNING)) {
       throw new IllegalStateException("Cannot resume the load monitor because it is in " + _state.get() + " state");
     }
+    _awaitExecution = false;
   }
 
+  /**
+   * Allow tasks to know if executor wishes to stop sampling b/c it wants to execute
+   */
+  public Boolean awaitingExecution() {
```
efeg: `public Boolean awaitingExecution()` -> `public boolean awaitingPauseSampling()`

1. Can we use `boolean` in the response? -- i.e. `Boolean` is not needed, because the response could not be `null`. 
2. Can we also make the function name change (see my previous comment)?
3. Can we also update the JavaDoc to reflect the use of this new parameter?

alexwang08: For item 3, what JavaDoc do we need to update?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -993,13 +993,32 @@ private void sanityCheckGoalNames() {
...
+   * {@link KafkaCruiseControlConfig#METRIC_SAMPLING_INTERVAL_MS_CONFIG}.
+   */
+  private void sanityCheckSamplingPeriod() {
+    long samplingPeriodMs = getLong(KafkaCruiseControlConfig.METRIC_SAMPLING_INTERVAL_MS_CONFIG);
+    long metadataTimeoutMs = getLong(KafkaCruiseControlConfig.METADATA_MAX_AGE_CONFIG);
+    if (metadataTimeoutMs >  samplingPeriodMs) {
+      throw new ConfigException("Attempt to set metadata refresh timeout [" + metadataTimeoutMs +
+          "] to be longer than sampling period [" + samplingPeriodMs + "].");
+    }
+  }
```
efeg: Missing space after `}`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -277,6 +282,12 @@ public void resumeSampling() {
     }
   }
 
+  /**
+   * Allow tasks to know if another thread, e.g. executor, is waiting on sampling to pause.
+   */
+  public boolean awaitingPauseSampling() {
+    return _awaitPauseSampling;
```
efeg: Nit: Let's keep the getter name and the variable name consistent -- i.e.
`_awaitPauseSampling` -> `_awaitingPauseSampling`.

PR #410, opened 2018-11-16
Change the time format for time information in Servlet response.

kidkun:
Current returned time information is 12 hour format but does not have `am/pm` information, change to 24 hour format.


efeg:
@kidkun the patch seems to hava conflicts with `master`. Could you push the resolved version?

Issue #411, opened 2018-11-16
Ensure that kafka_cluster_state endpoint shows brokers with no replicas


efeg:
Currently broker information section (first part of response) in `kafka_cluster_state` shows only the brokers with replicas. However, if a broker has just been added to the cluster (i.e. hosting no replica), it won't be visible at this endpoint. Adding this to the response helps with retrieval of broker id for `add_broker` requests.

-- Note that for identifying the brokers to be removed, this does not cause any issue, because a broker for which the user would want to run `remove_broker` should already have replicas on it.

PR #423, opened 2018-11-28
Ensure that kafka_cluster_state endpoint shows brokers with no replicas.

efeg:
Fixes the issue https://github.com/linkedin/cruise-control/issues/411.


alexwang08:
LGTM

PR #428, opened 2018-12-01
Fix for GET load endpoint returning wrong cached broker status.

kidkun:


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BrokerStats.java
```diff
@@ -95,12 +97,24 @@ private String getJSONString() {
   @Override
   protected void discardIrrelevantAndCacheRelevant(CruiseControlParameters parameters) {
     // Cache relevant response.
-    _cachedResponse = parameters.json() ? getJSONString() : toString();
+    _cachedJSONResponse = getJSONString();
+    _cachedPlainTextResponse = toString();
```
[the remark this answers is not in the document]
efeg: Ah I see! I believe there is one more issue to fix here:
1. Consider the scenario where the `allowCapacityEstimation` is `false` in the second call.
2. Then [this line](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java#L496) will throw a `NPE` due to discarding the stats in the first call.

^ To fix this issue without the need for keeping the whole `_brokerStats`, can we add a `boolean` to `BrokerStats` to indicate whether there is any capacity estimation in the cached model or not? This boolean can be initialized upon creation of `_cachedBrokerLoadStats`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BrokerStats.java
```diff
@@ -29,6 +29,8 @@
   private final List<SingleBrokerStats> _brokerStats;
   private final SortedMap<String, BasicStats> _hostStats;
   private int _hostFieldLength;
+  private String _cachedPlainTextResponse = null;
+  private String _cachedJSONResponse = null;
```
efeg: Nit: Can we do the initialization in the constructor?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -490,15 +489,9 @@ public BrokerStats cachedBrokerLoadStats(boolean allowCapacityEstimation) {
...
-          // Ensure that there is no capacity estimation in the cached model.
-          for (SingleBrokerStats singleBrokerStats : _cachedBrokerLoadStats.stats()) {
-            if (singleBrokerStats.isEstimated()) {
-              return null;
-            }
-          }
-        }
+          && _partitionMetricSampleAggregator.generation() == _cachedBrokerLoadGeneration.loadGeneration()
+          && _cachedBrokerLoadStats != null
+          && (allowCapacityEstimation || !_cachedBrokerLoadStats.isBrokerStatsEstimated())) {
```
efeg: Can we perform the checks that do not depend on the metadata update before the metadata update? This helps us avoid a potentially redundant metadata query.

PR #430, opened 2018-12-03
Add logic to handle getting broker log dir information timeout scenario.

kidkun:
In Cruise Control 2.0 support for JBOD, we add the logic to populate logdir information to `kafka_cluster_state` endpoint. However, in a corner case where a broker is dead, getting its logdir from Kafka `AdminClient` will timeout, in current code the TimeOutException directly terminate the execution and the an abrupt response is returned to client. This patch is to enhance this behavior to proper encode the timeout and return to client with other broker/partition information.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/KafkaClusterState.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/KafkaClusterState.java
@@ -308,25 +308,31 @@ public class KafkaClusterState extends AbstractCruiseControlResponse {
   private void populateKafkaBrokerLogDirState(Map<Integer, Set<String>> onlineLogDirsByBrokerId,
                                               Map<Integer, Set<String>> offlineLogDirsByBrokerId,
                                               Set<Integer> brokers)
-      throws ExecutionException, InterruptedException, TimeoutException {
+      throws ExecutionException, InterruptedException {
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/KafkaClusterState.java
```diff
@@ -308,25 +308,31 @@ private void writeKafkaPartitionState(StringBuilder sb, SortedSet<PartitionInfo>
...
-      });
+      try {
+        entry.getValue().get(LOGDIR_RESPONSE_TIMEOUT_MS, TimeUnit.MILLISECONDS).forEach((key, value) -> {
+          if (value.error == Errors.NONE) {
+            onlineLogDirsByBrokerId.get(entry.getKey()).add(key);
+          } else {
+            offlineLogDirsByBrokerId.get(entry.getKey()).add(key);
+          }
+        });
+      } catch (TimeoutException te) {
```
efeg: If it is observed whenever a broker is dead, then I believe we should handle it differently. In the current scenario, for each dead broker we would spend `LOGDIR_RESPONSE_TIMEOUT_MS`, which seems high.

If we observe this whenever a broker is dead, can we check whether `replicaCountByBrokerId` and `offlineReplicaCountByBrokerId` have the same size for a broker before attempting to retrieve the corresponding `logDir` info?

Issue #381, opened 2018-10-30
API: Return the CC version in response


nareshv:
Return the version of the CC thats running either in the Header or as part of the Body.

If we are using HTTP Headers, we can follow this https://tools.ietf.org/html/rfc7231#section-7.4.2

alexwang08:
Kafka has a way to determine version and commitId information which we can use in CC. In Kafka, gradle has a task to create file called "kafka-version.properties", which contains version (an property of Gradle object called Project) and commitId (can be determined from .git). This file is read by AppInfoParser.java in Kafka to load version and commitId properties.

efeg:
Support was added in PR https://github.com/linkedin/cruise-control/pull/439.

PR #439, opened 2018-12-06
Feat: added cruise control version in response header. (#381)

alexwang08:
Used a gradle generated file to provide software version and commit id. Added the result to all HttpServletResponse.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -72,6 +74,21 @@
   private final AnomalyDetector _anomalyDetector;
   private final Time _time;
 
+  private static final String VERSION;
+  private static final String COMMIT_ID;
+
+  // Referenced similar method to get software version in Kafka code.
```
kidkun: can we remove this comment?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/AbstractCruiseControlResponse.java
```diff
@@ -28,6 +29,8 @@ public void writeSuccessResponse(CruiseControlParameters parameters, HttpServlet
     OutputStream out = response.getOutputStream();
     boolean json = parameters.json();
     setResponseCode(response, SC_OK, json, null);
+    response.addHeader("Cruise-Control-Version", KafkaCruiseControl.cruiseControlVersion());
+    response.addHeader("Cruise-Control-Commit_Id", KafkaCruiseControl.cruiseControlCommitId());
```
kidkun: I think you also need to add this header logic to [writeResponseToOutputStream](https://github.com/linkedin/cruise-control/blob/c3d3ece6debc9e11c0207c09d0b32d05caf0141c/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ResponseUtils.java#L59).
It is used to return in-progress and error response.

Issue #440, opened 2018-12-06
Make the broker metric sample creation backwards compatible


efeg:
Currently if the Cruise Control metrics reporter is using an old serialization version, then the receiver will have the `WARN` messages such as `Skip generating broker metric sample for broker 123 because the following metrics are missing [BROKER_PRODUCE_TOTAL_TIME_MS_999TH, BROKER_PRODUCE_TOTAL_TIME_MS_50TH, BROKER_FOLLOWER_FETCH_REQUEST_QUEUE_TIME_MS_50TH, BROKER_FOLLOWER_FETCH_REQUEST_QUEUE_TIME_MS_999TH, BROKER_FOLLOWER_FETCH_TOTAL_TIME_MS_999TH, BROKER_PRODUCE_LOCAL_TIME_MS_50TH, BROKER_CONSUMER_FETCH_LOCAL_TIME_MS_50TH, BROKER_LOG_FLUSH_TIME_MS_999TH, BROKER_CONSUMER_FETCH_TOTAL_TIME_MS_50TH, BROKER_FOLLOWER_FETCH_LOCAL_TIME_MS_999TH, BROKER_CONSUMER_FETCH_REQUEST_QUEUE_TIME_MS_999TH, BROKER_PRODUCE_REQUEST_QUEUE_TIME_MS_999TH, BROKER_CONSUMER_FETCH_LOCAL_TIME_MS_999TH, BROKER_FOLLOWER_FETCH_LOCAL_TIME_MS_50TH, BROKER_FOLLOWER_FETCH_TOTAL_TIME_MS_50TH, BROKER_PRODUCE_LOCAL_TIME_MS_999TH, BROKER_LOG_FLUSH_TIME_MS_50TH, BROKER_CONSUMER_FETCH_TOTAL_TIME_MS_999TH, BROKER_CONSUMER_FETCH_REQUEST_QUEUE_TIME_MS_50TH, BROKER_PRODUCE_REQUEST_QUEUE_TIME_MS_50TH]`.

We can resolve this issue by making broker metric sample creation backwards compatible.

craynic:
What does `old serialization version` mean? Now I have CC and metrics-reporter both in 0.1.21, but I still get this warning.

efeg:
@craynic Would you double check that you followed all steps specified on the [Quick Start - Step-1](https://github.com/linkedin/cruise-control#quick-start) to create the new `cruise-control-metrics-reporter` jar, then copied it to the correct path?

[...]

PR #442, opened 2018-12-07
Make the broker metric sample creation backwards compatible.

efeg:
Resolved the issue https://github.com/linkedin/cruise-control/issues/440.

### DIFF
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/metric/RawMetricType.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/metric/RawMetricType.java
@@ -6,8 +6,10 @@ package com.linkedin.kafka.cruisecontrol.metricsreporter.metric;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 
@@ -93,7 +95,7 @@ public enum RawMetricType {
   BROKER_LOG_FLUSH_TIME_MS_999TH(BROKER, (byte) 62, (byte) 5);
 
   private static final List<RawMetricType> CACHED_VALUES = Arrays.asList(RawMetricType.values());
-  private static final SortedMap<Byte, List<RawMetricType>> BROKER_METRIC_TYPES_DIFF_BY_VERSION = buildBrokerMetricTypesDiffByVersion();
+  private static final SortedMap<Byte, Set<RawMetricType>> BROKER_METRIC_TYPES_DIFF_BY_VERSION = buildBrokerMetricTypesDiffByVersion();
   private static final List<RawMetricType> TOPIC_METRIC_TYPES = buildMetricTypeList(TOPIC);
   private static final List<RawMetricType> PARTITION_METRIC_TYPES = buildMetricTypeList(PARTITION);
   private final byte _id;
@@ -126,11 +128,11 @@ public enum RawMetricType {
     return CACHED_VALUES;
   }
 
-  public static Map<Byte, List<RawMetricType>> brokerMetricTypesDiffByVersion() {
+  public static Map<Byte, Set<RawMetricType>> brokerMetricTypesDiffByVersion() {
     return BROKER_METRIC_TYPES_DIFF_BY_VERSION;
   }
 
-  public static List<RawMetricType> brokerMetricTypesDiffForVersion(byte version) {
+  public static Set<RawMetricType> brokerMetricTypesDiffForVersion(byte version) {
     return BROKER_METRIC_TYPES_DIFF_BY_VERSION.get(version);
   }
 
@@ -154,11 +156,11 @@ public enum RawMetricType {
     BROKER, TOPIC, PARTITION
   }
 
-  private static SortedMap<Byte, List<RawMetricType>> buildBrokerMetricTypesDiffByVersion() {
-    SortedMap<Byte, List<RawMetricType>> buildBrokerMetricTypesDiffByVersion = new TreeMap<>();
+  private static SortedMap<Byte, Set<RawMetricType>> buildBrokerMetricTypesDiffByVersion() {
+    SortedMap<Byte, Set<RawMetricType>> buildBrokerMetricTypesDiffByVersion = new TreeMap<>();
     for (RawMetricType type : RawMetricType.values()) {
       if (type.metricScope() == BROKER) {
-        buildBrokerMetricTypesDiffByVersion.computeIfAbsent(type.supportedVersionSince(), t -> new ArrayList<>()).add(type);
+        buildBrokerMetricTypesDiffByVersion.computeIfAbsent(type.supportedVersionSince(), t -> new HashSet<>()).add(type);
       }
     }
 
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/metric/RawMetricType.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/metric/RawMetricType.java
@@ -7,6 +7,9 @@ package com.linkedin.kafka.cruisecontrol.metricsreporter.metric;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
 
 import static com.linkedin.kafka.cruisecontrol.metricsreporter.metric.RawMetricType.MetricScope.BROKER;
 import static com.linkedin.kafka.cruisecontrol.metricsreporter.metric.RawMetricType.MetricScope.TOPIC;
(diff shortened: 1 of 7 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
```diff
@@ -155,30 +156,32 @@ private int addBrokerMetricSamples(Cluster cluster,
...
-          validSample = false;
-          LOG.warn("Skip generating broker metric sample for broker {} because it does not have %s metrics or "
-                        + "the metrics are inconsistent.", node.id(), rawBrokerMetricType);
-          break;
-        } else {
-          MetricInfo metricInfo = brokerMetricDef.metricInfo(KafkaMetricDef.forRawMetricType(rawBrokerMetricType).name());
-          double metricValue = brokerLoad.brokerMetric(rawBrokerMetricType);
-          brokerMetricSample.record(metricInfo, metricValue);
+      BrokerMetricSample brokerMetricSample = new BrokerMetricSample(node.host(), node.id(), brokerLoad._brokerSampleDeserializationVersion);
+      for (Map.Entry<Byte, List<RawMetricType>> entry : RawMetricType.brokerMetricTypesDiffByVersion().entrySet()) {
```
kidkun: I am wondering whether this check is necessary. I `brokerLoad.allBrokerMetricsAvailable()` is enough to generate a broker metric sample. Won't this check make the V4 reporter's metric always get skipped?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/metric/RawMetricType.java
```diff
@@ -19,86 +22,92 @@
...
   // The multiplier is the number of the partitions in the produce request.
-  ALL_TOPIC_PRODUCE_REQUEST_RATE(BROKER, (byte) 8),
+  ALL_TOPIC_PRODUCE_REQUEST_RATE(BROKER, (byte) 8, (byte) 4),
   // Note that this is different from broker fetch request rate. If one FetchRequest fetches from 3 partitions,
   // it would be counted as one FetchRequest on the broker, but ALL_TOPIC_FETCH_REQUEST would increment by 3.
   // The multiplier is the number of the partitions in the fetch request.
-  ALL_TOPIC_FETCH_REQUEST_RATE(BROKER, (byte) 9),
-  ALL_TOPIC_MESSAGES_IN_PER_SEC(BROKER, (byte) 10),
+  ALL_TOPIC_FETCH_REQUEST_RATE(BROKER, (byte) 9, (byte) 4),
+  ALL_TOPIC_MESSAGES_IN_PER_SEC(BROKER, (byte) 10, (byte) 4),
```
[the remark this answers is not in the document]
efeg: Yes, we indeed can and should do it for topic and partition metrics as well. This patch is specific to broker metrics. Created the issue relevant to the topic and partition metrics (See https://github.com/linkedin/cruise-control/issues/446)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/BrokerMetricSample.java
```diff
@@ -19,23 +19,44 @@
  * The class hosting all the broker level metrics in {@link KafkaMetricDef}.
  */
 public class BrokerMetricSample extends MetricSample<String, BrokerEntity> {
-  private static final byte CURRENT_VERSION = 5;
+  static final byte MIN_SUPPORTED_VERSION = 4;
+  static final byte LATEST_SUPPORTED_VERSION = 5;
+  private final byte _deserializationVersion;
```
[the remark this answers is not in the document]
efeg: Cruise Control always serializes the payload based on its latest known serialization format, and uses the deserialization version header to indicate the fields that have a meaningful value. 

To read a sample, Cruise Control should know if it is a version that its deserialzer supports or not. Here are some cases, where removing this version from the header would cause problems:

1. If the sample store topic keeps samples with `version-X`, and we run a Cruise Control instance, which can support the latest version of `version-(X-1)`, then if we do not know the deserialization version, then we will lose existing data. Note, critically, that losing existing data is different from backwards compatibility.
2. Today we retrieve the bytes to deserialize from a Kafka topic by consuming records from it. This design implicitly fits a single sample to a single record value. However, this is a pluggable component, which can use other means to store/retrieve data. Hence, the bytes to deserialize, which are transmitted from the network, might come with dummy data after the actual sample to be deserialized. Hence, without knowing the actual deserialization version, we are likely to read bad data and cannot distinguish it from corrupted data.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
efeg: The `BrokerLoad#setNullableBrokerMetrics()` will ensure that `v4`'s `nullable` metrics will be populated. Hence, we will not skip the raw metrics. This sanity check is different from the `brokerLoad.allBrokerMetricsAvailable()` in the sense that the former checks for the existence of minimum required broker metrics availability while the latter is about ensuring that each metric either has a value or has correctly been nullified.

I think we can make this more clear by changing `allBrokerMetricsAvailable()`  to `minRequiredBrokerMetricsAvailable()`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/BrokerMetricSample.java
```diff
@@ -103,7 +124,7 @@ public int brokerId() {
   public byte[] toBytes() {
     byte[] hostBytes = (entity().group() != null ? entity().group() : "UNKNOWN").getBytes(StandardCharsets.UTF_8);
     ByteBuffer buffer = ByteBuffer.allocate(457 + hostBytes.length);
-    buffer.put(CURRENT_VERSION);
+    buffer.put(_deserializationVersion);
```
kidkun: if it is `V4`, I think we do not need to write the `V5` metrics to byte stream, since they carry no information and  `readV4()` will ignore anyway.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/BrokerMetricSample.java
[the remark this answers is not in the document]
kidkun: Yes, I agree that `version` is a necessary field in byte stream. My original idea is to fill this field with `LATEST_SUPPORTED_VERSION` instead of  `_deserializationVersion`  since we always write in latest version format. 

But after rethinking about it, I think always writing in latest version format may not be the correct way. 
consider the scenario, we collect a `V4` sample from consuming a period of metric records, if we fill 0 to all `V5` fields and store this sample in `V5` version. later when we read sample store topic(like CC restarts) and get this message, we cannot tell these `v5` fields' 0 value is either because this field is missing or is the actual value(which happens to be 0)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
```diff
@@ -141,44 +142,46 @@ private int addPartitionMetricSamples(Cluster cluster,
...
-          validSample = false;
-          LOG.warn("Skip generating broker metric sample for broker {} because it does not have %s metrics or "
-                        + "the metrics are inconsistent.", node.id(), rawBrokerMetricType);
-          break;
-        } else {
-          MetricInfo metricInfo = brokerMetricDef.metricInfo(KafkaMetricDef.forRawMetricType(rawBrokerMetricType).name());
-          double metricValue = brokerLoad.brokerMetric(rawBrokerMetricType);
-          brokerMetricSample.record(metricInfo, metricValue);
+      BrokerMetricSample brokerMetricSample = new BrokerMetricSample(node.host(), node.id(), brokerLoad._brokerSampleDeserializationVersion);
+      for (Map.Entry<Byte, Set<RawMetricType>> entry : RawMetricType.brokerMetricTypesDiffByVersion().entrySet()) {
```
kidkun: If we reach here, it means `setNullableBrokerMetrics()` is already executed and it already fill all the nullable field, so this check will always pass. is my understanding correct?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/BrokerMetricSample.java
efeg: The reason why we serialize with the latest payload is to avoid maintaining multiple deprecated serialization versions. This cost of writing additional fields will only be relevant to the users having the deprecated broker serde versions. As a side-note, once/if we move towards a better serialization format such as flatbuffers, such nullable fields can implicitly be shrunk in size on the wire.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
efeg: Correct -- as I commented on a previous feedback, this sanity check ensures that each metric either has a value or has correctly been nullified.

Issue #426, opened 2018-11-30
Enable control over how the execution process prioritizes partition/leadership moves


efeg:
Consider the following typical scenario:
1. The system receives a request to `rebalance` the cluster, `add`/`remove`/`demote` broker, or `fix offline replicas`.
2. Its `analyzer` generates `proposals` based on the specified goals.
3. These `proposals` (aka `ExecutionTasks`) are passed to the `executor` to be executed in the actual cluster.

While the `Executor` provides control over how many replicas can go in and out of individual brokers in step 3, the selection of proposals to be executed at a given time is random. Cruise Control should expose multiple pluggable algorithms that customizes the execution order. These algorithms could include:
1. `random`: the current algorithm,
2. `postpone_urp`: deprioritize the move of URPs,
3. `prioritize_small_partitions`: partitions with small disk space move first,
4. `prioritize_large_partitions`: partitions with small disk space move first.

PR #456, opened 2018-12-14
Add ExecutionTaskStrategy interface to support customizing partition movement task execution order

kidkun:
patch for [426](https://github.com/linkedin/cruise-control/issues/426)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractExecutionTaskStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+import com.linkedin.kafka.cruisecontrol.executor.ExecutionTask;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import org.apache.kafka.common.Cluster;
+
+
+public abstract class AbstractExecutionTaskStrategy implements ExecutionTaskStrategy {
```
efeg: Missing JavaDoc ot explain the strategy.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/BaseExecutionTaskStrategy.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright 2018 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.cruisecontrol.executor.strategy;
+
+import com.linkedin.kafka.cruisecontrol.executor.ExecutionTask;
+
+
+public class BaseExecutionTaskStrategy extends AbstractExecutionTaskStrategy {
```
efeg: Missing JavaDoc ot explain the strategy.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PostponeUrpExecutionTaskStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import org.apache.kafka.common.Cluster;
+import org.apache.kafka.common.PartitionInfo;
+
+
+public class PostponeUrpExecutionTaskStrategy implements ExecutionTaskStrategy {
```
efeg: Missing JavaDoc ot explain the strategy.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PrioritizeLargeMovementExecutionTaskStrategy.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2018 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.cruisecontrol.executor.strategy;
+
+import com.linkedin.kafka.cruisecontrol.executor.ExecutionTask;
+
+public class PrioritizeLargeMovementExecutionTaskStrategy extends AbstractExecutionTaskStrategy {
```
efeg: Missing JavaDoc ot explain the strategy.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PrioritizeSmallMovementExecutionTaskStrategy.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2018 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.cruisecontrol.executor.strategy;
+
+import com.linkedin.kafka.cruisecontrol.executor.ExecutionTask;
+
+public class PrioritizeSmallMovementExecutionTaskStrategy extends AbstractExecutionTaskStrategy {
```
efeg: Missing JavaDoc ot explain the strategy.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PostponeUrpExecutionTaskStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private boolean isDataMoveOnUrp(Cluster cluster, ExecutionTask task) {
```
efeg: We have the same logic at multiple places (e.g. `PreferredLeaderElectionGoal`, `KafkaClusterState`). Can we move it to some common util function and use it for such checks?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PostponeUrpExecutionTaskStrategy.java
efeg: There seems to be common code in this class and `AbstractExecutionTaskStrategy` -- would it possible to customize  `AbstractExecutionTaskStrategy` such that it would be possible to extend it?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PostponeUrpExecutionTaskStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+    PartitionInfo partitionInfo = cluster.partition(task.proposal().topicPartition());
+    return partitionInfo.inSyncReplicas().length != partitionInfo.replicas().length;
+  }
+
+  @Override
+  public Map<Integer, SortedSet<ExecutionTask>> applyStrategy(Collection<ExecutionTask> executionTasks, Cluster cluster) {
+    Comparator<ExecutionTask> taskComparator = (ExecutionTask task1, ExecutionTask task2) -> isDataMoveOnUrp(cluster, task1) ?
```
efeg: The comparator logic is a little hard to understand -- would you clarify it with a comment?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PrioritizeSmallMovementExecutionTaskStrategy.java
```diff
@@ -0,0 +1,21 @@
...
+
+  @Override
+  public int compareTask(ExecutionTask task1, ExecutionTask task2) {
```
efeg: I am wondering whether we can support the use of multiple strategies at once. For example, a user might prioritize movement first based on `PostponeUrpExecutionTaskStrategy`, then based on `PrioritizeSmallMovementExecutionTaskStrategy`. We can support this by chaining requested comparators.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -438,6 +439,13 @@
       "movements the executor will take as one batch. This is mainly because the ZNode has a 1 MB size upper limit. And it " +
       "will also reduce the controller burden.";
 
+  /**
+   * <code>execution.task.strategy</code>
+   */
+  public static final String EXECUTION_TASK_STRATEGY_CONFIG = "execution.task.strategy";
```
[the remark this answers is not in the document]
kidkun: for `leadership_movement`, I feel that a customized strategy maybe too much since `leadership_movement` is a resource-cheap operation. And it is unrelated with whether partition is large or not, or is partition under replicated(since in `replica_movement` stage all the URP are recovered).

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -39,13 +43,13 @@
...
+      new ExecutionProposal(new TopicPartition(TOPIC2, 0), 4, 0, Arrays.asList(0, 2), Arrays.asList(2, 1));
   private final ExecutionProposal partitionMovement2 =
-      new ExecutionProposal(new TopicPartition(TOPIC2, 1), 2, 0, Arrays.asList(2, 0), Arrays.asList(1, 2));
+      new ExecutionProposal(new TopicPartition(TOPIC2, 1), 3, 1, Arrays.asList(1, 3), Arrays.asList(3, 2));
   private final ExecutionProposal partitionMovement3 =
-      new ExecutionProposal(new TopicPartition(TOPIC2, 2), 3, 2, Arrays.asList(2, 0), Arrays.asList(1, 0));
+      new ExecutionProposal(new TopicPartition(TOPIC2, 2), 2, 2, Arrays.asList(2, 1), Arrays.asList(1, 3));
   private final ExecutionProposal partitionMovement4 =
-      new ExecutionProposal(new TopicPartition(TOPIC2, 2), 4, 3, Arrays.asList(3, 0), Arrays.asList(0, 2));
+      new ExecutionProposal(new TopicPartition(TOPIC2, 3), 1, 3, Arrays.asList(3, 2), Arrays.asList(2, 0));
```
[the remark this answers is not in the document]
kidkun: I just want to better test `PrioritizeLargeReplicaMovementStrategy` logic, here  I deliberately make the task 1 to 4's partition size decrease, so if we apply `PrioritizeLargeReplicaMovementStrategy` the execution order is backwards(different from using other strategies).

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -204,7 +230,7 @@ public void testClear() {
                                           Collections.<String>emptySet());
 
     planner.addExecutionProposals(proposals, expectedCluster);
-    assertEquals(1, planner.remainingDataToMoveInMB());
+    assertEquals(4, planner.remainingDataToMoveInMB());
```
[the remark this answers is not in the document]
kidkun: No, it is because I change to the partition size of `partitionMovement1`'s partition(`TOPIC2`) from 1 to 4, and the reason for this change is to better test `PrioritizeLargeReplicaMovementStrategy`.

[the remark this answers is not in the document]
kidkun: No, it is because I change to the partition size of partitionMovement1's partition(TOPIC2) from 1 to 4, and the reason for this change is to better test `PrioritizeLargeReplicaMovementStrategy`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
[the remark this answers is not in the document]
efeg: But unless the unclean leader election is enabled, wouldn't the leadership transfer take `URPs` into account?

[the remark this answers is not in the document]
kidkun: That is a good point. After a second thought, I feel in that case we may still not need to postpone leadership change operation for `URP` because when Kafka is handling leadership change, it does not wait URP to recovery, seems it makes no difference either partition is under replicated or not.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -442,9 +442,9 @@
   /**
    * <code>execution.task.strategy</code>
    */
-  public static final String EXECUTION_TASK_STRATEGY_CONFIG = "execution.task.strategy";
-  private static final String EXECUTION_TASK_STRATEGY_DOC = "The strategy used to determine execution order for generated partition"
-      + "movement tasks.";
+  public static final String REPLICA_MOVEMENT_STRATEGIES_CONFIG = "replica.movement.strategies";
+  private static final String REPLICA_MOVEMENT_STRATEGIES_DOC = "A list of strategy used to determine execution order for "
```
efeg: Nit: `strategy` -> `strategies`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
```diff
@@ -51,18 +50,24 @@
...
-      throw new IllegalStateException("Default execution strategy " + defaultExecutionStrategy + " is undefined.");
+    for (String replicaMovementStrategy : replicaMovementStrategies) {
+      try {
+        if (_replicaMovementTaskStrategy == null) {
+          _replicaMovementTaskStrategy = (ReplicaMovementStrategy) Class.forName(replicaMovementStrategy).newInstance();
+        } else {
+          _replicaMovementTaskStrategy = _replicaMovementTaskStrategy.chain((ReplicaMovementStrategy) Class.forName(replicaMovementStrategy).newInstance());
+        }
+      } catch (Exception e) {
+        throw new RuntimeException(e);
```
efeg: Nit: Would it be more clear to add a message in addition to the stack trace? -- i.e. 
```
throw new RuntimeException("Error occurred while setting up the replica movement strategy: " + replicaMovementStrategy + ".", e);
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -442,9 +442,9 @@
   /**
    * <code>execution.task.strategy</code>
    */
-  public static final String EXECUTION_TASK_STRATEGY_CONFIG = "execution.task.strategy";
-  private static final String EXECUTION_TASK_STRATEGY_DOC = "The strategy used to determine execution order for generated partition"
-      + "movement tasks.";
+  public static final String REPLICA_MOVEMENT_STRATEGIES_CONFIG = "replica.movement.strategies";
```
efeg: Can we add a sanity check in this class to ensure that the list received from `REPLICA_MOVEMENT_STRATEGIES_CONFIG` is not empty? -- i.e. similar to the sanity check we have for `DEFAULT_GOALS_CONFIG`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+   * higher execution priority.
+   *
+   * @param cluster The current cluster state.
+   * @return The comparator of task.
+   */
+  public abstract Comparator<ExecutionTask> taskComparator(Cluster cluster);
+
+  @Override
+  public ReplicaMovementStrategy chain(ReplicaMovementStrategy strategy) {
+    if (!(strategy instanceof AbstractReplicaMovementStrategy)) {
```
efeg: Would it be possible to move this sanity check to `KafkaCruiseControlConfig`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+        Comparator<ExecutionTask> comparator2 = ((AbstractReplicaMovementStrategy) strategy).taskComparator(cluster);
+
+        return (task1, task2) -> {
+          int compareResult1 = comparator1.compare(task1, task2);
+          int compareResult2 = comparator2.compare(task1, task2);
+          if (compareResult1 == 0) {
+            return compareResult2;
+          }
+          return compareResult1;
+        };
```
efeg: Nit: Would it be possible to avoid the cost of calculating `compareResult2` if not necessary?
```
        return (task1, task2) -> {
          int compareResult1 = comparator1.compare(task1, task2);
          return compareResult1 == 0 ? comparator2.compare(task1, task2) : compareResult1;
        };
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+
+    for (ExecutionTask task : replicaMovementTasks) {
+      ExecutionProposal proposal = task.proposal();
+
+      // Add the task to source broker's execution plan
+      int sourceBroker = proposal.oldLeader();
+      SortedSet<ExecutionTask> sourceBrokerTaskSet = tasksByBrokerId.computeIfAbsent(sourceBroker,
+                                                                                     k -> new TreeSet<>(taskComparator(cluster)));
+      if (!sourceBrokerTaskSet.add(task)) {
+        throw new IllegalStateException("Replica movement strategy " + this.getClass().getSimpleName() + " is unable to determine order of all tasks.");
```
efeg: Will `this.getClass().getSimpleName()` return `AbstractReplicaMovementStrategy` rather than the class name that implements it?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+            return compareResult2;
+          }
+          return compareResult1;
+        };
+      }
+    };
+  }
+
+  @Override
+  public Map<Integer, SortedSet<ExecutionTask>> applyStrategy(Collection<ExecutionTask> replicaMovementTasks, Cluster cluster) {
```
efeg: Any reason why we don't use a `Set<ExecutionTask>` rather than a `Collection<ExecutionTask>` for `replicaMovementTasks` in the interface?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+    Map<Integer, SortedSet<ExecutionTask>> tasksByBrokerId = new HashMap<>();
+
+    for (ExecutionTask task : replicaMovementTasks) {
+      ExecutionProposal proposal = task.proposal();
+
+      // Add the task to source broker's execution plan
+      int sourceBroker = proposal.oldLeader();
+      SortedSet<ExecutionTask> sourceBrokerTaskSet = tasksByBrokerId.computeIfAbsent(sourceBroker,
+                                                                                     k -> new TreeSet<>(taskComparator(cluster)));
+      if (!sourceBrokerTaskSet.add(task)) {
```
efeg: If `replicaMovementTasks` is a `Set` (please see my previous comment), then this check would be redundant.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+        throw new IllegalStateException("Replica movement strategy " + this.getClass().getSimpleName() + " is unable to determine order of all tasks.");
+      }
+
+      // Add the task to destination brokers' execution plan
+      for (int destinationBroker : proposal.replicasToAdd()) {
+        SortedSet<ExecutionTask> destinationBrokerTaskSet = tasksByBrokerId.computeIfAbsent(destinationBroker,
+                                                                                            k -> new TreeSet<>(taskComparator(cluster)));
+        if (!destinationBrokerTaskSet.add(task)) {
```
efeg: If `replicaMovementTasks` is a `Set` (please see my previous comment), then this check would be redundant.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PostponeUrpReplicaMovementStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+package com.linkedin.kafka.cruisecontrol.executor.strategy;
+
+import com.linkedin.kafka.cruisecontrol.executor.ExecutionTask;
+import java.util.Comparator;
+import org.apache.kafka.common.Cluster;
+
+import static com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils.isPartitionUnderReplicated;
+
+/**
+ * The strategy which tries to first move replicas of partitions which currently are not under replicated.
```
efeg: Nit: `The strategy, which tries to first move replicas of partitions that are currently not under replicated.`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
kidkun: I think this is a constraint only for `sanityCheckReplicaMovementStrategies`(abstract class) not `ReplicaMovementStrategy`( interface). If we move this check to `KafkaCruiseControlConfig` we actually place an unwanted restriction that Cruise Control can only use extend class of the abstract class instead of the interface and jeopardize the generic property.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
kidkun: I think it will return the the implemented class type since `Object.getClass()` returns the runtime class of the object itself.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
kidkun: This check is still needed because the way `SortedSet` works is it use passed-in `Comparator.compare(a, b)==0` instead of item's `equal()` to determine whether to replace an existing item. So it highly depends on the `Comparator` which `AbstractReplicaMovementStrategy` provides.

one example is 
  public void test2() {
    SortedSet<Integer> ss = new TreeSet<>((o1, o2) -> {
      return 0;
    });
    ss.add(1);
    assertFalse(ss.add(2));
  }

efeg: Good point!
But, then what prevents any of our existing strategies, such as `PostponeUrpReplicaMovementStrategy#taskComparator`, to fail this sanity check unless `BaseReplicaMovementStrategy#taskComparator` is also chained in the strategies? Do we have an implicit assumption that the user always provides `BaseReplicaMovementStrategy` as one of the required strategies?

If that is the case, I think we should (1) indeed always use `BaseReplicaMovementStrategy` as the last chained strategy in all cases, and (2) not make it a configuration.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
efeg: The sanity check would be redundant once we ensure the use of `BaseReplicaMovementStrategy` (please see previous discussion).

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -112,14 +116,21 @@ public void testGetPartitionMovementTasks() {
     proposals.add(partitionMovement2);
     proposals.add(partitionMovement3);
     proposals.add(partitionMovement4);
-    ExecutionTaskPlanner planner = new ExecutionTaskPlanner();
+    // Test different execution strategies.
+    ExecutionTaskPlanner basePlanner = new ExecutionTaskPlanner(Collections.singletonList(BaseReplicaMovementStrategy.class.getName()));
```
efeg: If we could ensure that `BaseReplicaMovementStrategy` is not a config, we can skip passing it as a parameter in all cases (please see prev. discussion).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
```diff
@@ -0,0 +1,76 @@
...
+   * Generate a comparator for replica movement task which incorporate the strategy to apply. The "smaller" task will have
+   * higher execution priority.
+   *
+   * @param cluster The current cluster state.
+   * @return The comparator of task.
+   */
+  public abstract Comparator<ExecutionTask> taskComparator(Cluster cluster);
+
+  @Override
+  public ReplicaMovementStrategy chain(ReplicaMovementStrategy strategy) {
```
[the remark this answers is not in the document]
kidkun: The issue here is that the `ReplicaMovementStrategy` is unable to store a `Comparator` as internal field because the `Comparator` needs a `Cluster` in its constructor. we have to create new comparator each time `applyStrategy(Set<ExecutionTask> replicaMovementTasks, Cluster cluster)` is called because comparator need to use the cluster status at that time to do comparison.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
kidkun: My original idea does not have the assumption that `the user always provides BaseReplicaMovementStrategy as one of the required strategies`.  

The user is free to apply whatever rule they want and if they can guarantee that comparator they implement never return 0, they do not need to chain with `BaseReplicaMovementStrategy`(one example can be alphabetic order of `TopicPartition`). And if they miss the guarantee in their implementation, Cruise Control will not misbehave silently because we have this sanity check.

It is true the three strategies we provide , `PostponeUrpReplicaMovementStrategy`, `PrioritizeLargeReplicaMovementStrategy` and `PrioritizeSmallReplicaMovementStrategy` do not have this guarantee and I think it is ok because in the configuration we ask for a list of strategy. so if the client want to do things like `postpone urp replica movement and if two movement are urp, execute the one with smaller task id.` they should explicitly provide 2 strategies in the configure because it is indeed two rules here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -1149,38 +1148,14 @@ private void sanityCheckSamplingPeriod() {
   }
 
   /**
-   * Sanity check for
-   * (1) {@link KafkaCruiseControlConfig#REPLICA_MOVEMENT_STRATEGIES_CONFIG} is non-empty.
-   * (2) If any strategy extends from {@link com.linkedin.kafka.cruisecontrol.executor.strategy.AbstractReplicaMovementStrategy},
-   *     then all the strategies should extend from this abstract class.
+   * Sanity check to ensure that {@link KafkaCruiseControlConfig#REPLICA_MOVEMENT_STRATEGIES_CONFIG} is non-empty.
```
efeg: Maybe we can remove this sanity check as well -- since we have the base strategy

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/AbstractReplicaMovementStrategy.java
```diff
@@ -21,23 +21,14 @@
...
-
   @Override
   public ReplicaMovementStrategy chain(ReplicaMovementStrategy strategy) {
     AbstractReplicaMovementStrategy current = this;
     return new AbstractReplicaMovementStrategy() {
       @Override
       public Comparator<ExecutionTask> taskComparator(Cluster cluster) {
         Comparator<ExecutionTask> comparator1 = current.taskComparator(cluster);
-        Comparator<ExecutionTask> comparator2 = ((AbstractReplicaMovementStrategy) strategy).taskComparator(cluster);
+        Comparator<ExecutionTask> comparator2 = strategy.taskComparator(cluster);
```
efeg: `new AbstractReplicaMovementStrategy` -> `new ReplicaMovementStrategy`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -962,6 +970,11 @@
                 atLeast(1),
                 ConfigDef.Importance.MEDIUM,
                 NUM_CONCURRENT_LEADER_MOVEMENTS_DOC)
+        .define(REPLICA_MOVEMENT_STRATEGIES_CONFIG,
+                ConfigDef.Type.LIST,
+                BaseReplicaMovementStrategy.class.getName(),
```
efeg: Can we remove the default since we apply base by default implicitly.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManagerTest.java
```diff
@@ -41,7 +42,9 @@ private Cluster generateExpectedCluster(ExecutionProposal proposal, TopicPartiti
   @Test
   public void testStateChangeSequences() {
     TopicPartition tp = new TopicPartition("topic", 0);
-    ExecutionTaskManager taskManager = new ExecutionTaskManager(1, 1, new MetricRegistry(), new SystemTime());
+    ExecutionTaskManager taskManager = new ExecutionTaskManager(1, 1,
+                                                                Collections.singletonList(BaseReplicaMovementStrategy.class.getName()),
```
efeg: `singletonList` -> `Collections.emptyList()`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -60,7 +64,7 @@ public void testGetLeaderMovementTasks() {
     proposals.add(leaderMovement2);
     proposals.add(leaderMovement3);
     proposals.add(leaderMovement4);
-    ExecutionTaskPlanner planner = new ExecutionTaskPlanner();
+    ExecutionTaskPlanner planner = new ExecutionTaskPlanner(Collections.singletonList(BaseReplicaMovementStrategy.class.getName()));
```
efeg: Base can be removed?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -183,7 +212,7 @@ public void testClear() {
     List<ExecutionProposal> proposals = new ArrayList<>();
     proposals.add(leaderMovement1);
     proposals.add(partitionMovement1);
-    ExecutionTaskPlanner planner = new ExecutionTaskPlanner();
+    ExecutionTaskPlanner planner = new ExecutionTaskPlanner(Collections.singletonList(BaseReplicaMovementStrategy.class.getName()));
```
efeg: `Base`

Issue #451, opened 2018-12-13
GoalViolationDetector caches unnecessary fixing proposals each time it detects a goal violation


kidkun:
Currently when `GoalViolationDetector` detects a goal violation, it try to calculate the fixing proposal **for the goal** and store it with other goal violation information. This fixing proposal may be unneeded since in either self-healing enabled or client manual fixing scenario, we need the fixing proposal against **all the violated goals** at that time. 
Also note these  fixing proposal for specific goal might be long for large cluster, combined with retention time it will sit in memory, this might pose a heavy memory pressure for Cruise Control.

PR #465, opened 2018-12-19
Improve goal violation anomaly detection and auto-fixing logic.

kidkun:
Patch for issue [455](https://github.com/linkedin/cruise-control/issues/451).
Improvement made to goal violation anomaly detection and auto-fixing logic.

* remove caching unneeded proposal for each goal violation
* distinguish fixable and unfixable goal violation and handle them differently
* filter all the support goal to check based on readiness of each goal

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
@@ -98,40 +98,55 @@ public class GoalViolationDetector implements Runnable {
       return;
     }
 
-    GoalViolations goalViolations = new GoalViolations(_kafkaCruiseControl, _allowCapacityEstimation);
-    boolean newModelNeeded = true;
-    long now = _time.milliseconds();
-    ClusterModel clusterModel = null;
-    for (Map.Entry<Integer, Goal> entry : _goals.entrySet()) {
-      Goal goal = entry.getValue();
-      if (_loadMonitor.meetCompletenessRequirements(goal.clusterModelCompletenessRequirements())) {
-        LOG.debug("Detecting if {} is violated.", entry.getValue().name());
-        // Because the model generation could be slow, We only get new cluster model if needed.
-        if (newModelNeeded) {
-          try (AutoCloseable ignored = _loadMonitor.acquireForModelGeneration(new OperationProgress())) {
-            clusterModel = _loadMonitor.clusterModel(now, goal.clusterModelCompletenessRequirements(), new OperationProgress());
+    AutoCloseable clusterModelSemaphore = null;
+    try {
+      GoalViolations goalViolations = new GoalViolations(_kafkaCruiseControl, _allowCapacityEstimation);
+      long now = _time.milliseconds();
+      boolean newModelNeeded = true;
+      ClusterModel clusterModel = null;
+      for (Map.Entry<Integer, Goal> entry : _goals.entrySet()) {
+        Goal goal = entry.getValue();
+        if (_loadMonitor.meetCompletenessRequirements(goal.clusterModelCompletenessRequirements())) {
+          LOG.debug("Detecting if {} is violated.", entry.getValue().name());
+          // Because the model generation could be slow, We only get new cluster model if needed.
+          if (newModelNeeded) {
+            if (clusterModelSemaphore != null) {
+              clusterModelSemaphore.close();
+            }
+            clusterModelSemaphore = _loadMonitor.acquireForModelGeneration(new OperationProgress());
+            // Make cluster model null before generating a new cluster model so the current one can be GCed.
+            clusterModel = null;
+            clusterModel = _loadMonitor.clusterModel(now,
+                                                     goal.clusterModelCompletenessRequirements(),
+                                                     new OperationProgress());
+            KafkaCruiseControl.sanityCheckCapacityEstimation(_allowCapacityEstimation,
+                                                             clusterModel.capacityEstimationInfoByBrokerId());
             _lastCheckedModelGeneration = clusterModel.generation();
-          } catch (NotEnoughValidWindowsException nevwe) {
-            LOG.debug("Skipping goal violation detection because there are not enough valid windows.");
-            return;
-          } catch (Exception e) {
-            LOG.error("Skipping goal violation detection because of unexpected exception when generating cluster model.", e);
-            return;
           }
-        }
-        try {
           newModelNeeded = optimizeForGoal(clusterModel, goal, goalViolations);
-        } catch (KafkaCruiseControlException kcce) {
-          LOG.warn("Goal violation detector received exception.", kcce);
+        } else {
+          LOG.debug("Skipping goal violation detection for {} because load completeness requirement is not met.", goal);
         }
-      } else {
-        LOG.debug("Skipping goal violation detection for {} because load completeness requirement is not met.", goal);
       }
+      if (!goalViolations.violations().isEmpty()) {
+        _anomalies.add(goalViolations);
+      }
+    } catch (NotEnoughValidWindowsException nevwe) {
+      LOG.debug("Skipping goal violation detection because there are not enough valid windows.");
+    } catch (KafkaCruiseControlException kcce) {
+      LOG.warn("Goal violation detector received exception", kcce);
+    } catch (Exception e) {
+      LOG.error("Unexpected exception", e);
+    } finally {
+      if (clusterModelSemaphore != null) {
+        try {
+          clusterModelSemaphore.close();
+        } catch (Exception e) {
+          LOG.error("Received exception when closing auto closable semaphore", e);
+        }
+      }
+      LOG.debug("Goal violation detection finished.");
     }
-    if (!goalViolations.violations().isEmpty()) {
-      _anomalies.add(goalViolations);
-    }
-    LOG.debug("Goal violation detection finished.");
   }
 
   private Set<String> excludedTopics(ClusterModel clusterModel) {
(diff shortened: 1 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
```diff
@@ -87,68 +86,49 @@ public GoalViolationDetector(KafkaCruiseControlConfig config,
...
-                                                             clusterModel.capacityEstimationInfoByBrokerId());
-          }
-          int priority = entry.getKey();
-          newModelNeeded = optimizeForGoal(clusterModel, priority, goal, goalViolations);
-        } else {
-          LOG.debug("Skipping goal violation detection for {} because load completeness requirement is not met.", goal);
-        }
+      ModelCompletenessRequirements requirements = new ModelCompletenessRequirements(0, 0, false);
+      for (Goal goal: readyGoals) {
+        requirements = requirements.stronger(goal.clusterModelCompletenessRequirements());
```
efeg: What is the reasoning for using combined requirements?

IIUC, contrary to before, due to this stronger requirements, the goal violation detector would fail to detect violations of goals with weak requirements (e.g. `rack awareness`) due to the stronger requirements of some other goals -- such as `CpuUsageDistributionGoal` because the monitored cluster model does not have the required number of valid windows, which has nothing to do with detecting `rack awareness`.

I think we should ensure that we do not let the ability to fix goal violations impact the ability to detect goal violations.

kidkun: My original purpose is to reduce the number of times calculating `ClusterModel`, which is a source-heavy thing. My initial idea is to get a proper requirement and just generate cluster model once, but you are right, even `LoadMonitor` can meet each ready goal, it may not meet the `stronger()` of all goals.

kidkun: And also I do not understand the origin code's logic around `NewModelNeeded`.

kidkun: on a second observation. for all the goals we currently support. 
RackAwareGoal, (requiredNumWindows=1, minMonitoredPartitionPercentage=0.000, includedAllTopics=true)
ReplicaCapacityGoal, (requiredNumWindows=1, minMonitoredPartitionPercentage=0.000, includedAllTopics=true)
DiskCapacityGoal, (requiredNumWindows=1, minMonitoredPartitionPercentage=0.980, includedAllTopics=true)
NetworkInboundCapacityGoal, (requiredNumWindows=1, minMonitoredPartitionPercentage=0.980, includedAllTopics=true)
NetworkOutboundCapacityGoal, (requiredNumWindows=1, minMonitoredPartitionPercentage=0.980, includedAllTopics=true)
CpuCapacityGoal, (requiredNumWindows=1, minMonitoredPartitionPercentage=0.980, includedAllTopics=true)
ReplicaDistributionGoal, (requiredNumWindows=1, minMonitoredPartitionPercentage=0.000, includedAllTopics=true)
PotentialNwOutGoal, (requiredNumWindows=168, minMonitoredPartitionPercentage=0.980, includedAllTopics=false)
DiskUsageDistributionGoal, (requiredNumWindows=1, minMonitoredPartitionPercentage=0.980, includedAllTopics=true)
NetworkInboundUsageDistributionGoal, (requiredNumWindows=168, minMonitoredPartitionPercentage=0.980, includedAllTopics=false)
NetworkOutboundUsageDistributionGoal, (requiredNumWindows=168, minMonitoredPartitionPercentage=0.980, includedAllTopics=false)
CpuUsageDistributionGoal, (requiredNumWindows=168, minMonitoredPartitionPercentage=0.980, includedAllTopics=false)
TopicReplicaDistributionGoal, (requiredNumWindows=1, minMonitoredPartitionPercentage=0.000, includedAllTopics=true)
LeaderBytesInDistributionGoal, (requiredNumWindows=168, minMonitoredPartitionPercentage=0.980, includedAllTopics=false)

there are actually 3 types of complete requirement 
1. window=1, partition_percentage=0, all_topic=true
2.window=1, partition_percentage=0.98, all_topic=true
3.window=1, partition_percentage=0.98, all_topic=false

looks like `stronger()` of all ready goal is guaranteed to meet. we do not have the issue of `impact the ability to detect goal violations`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
```diff
@@ -87,68 +86,49 @@ public GoalViolationDetector(KafkaCruiseControlConfig config,
...
       }
-      if (clusterModel != null) {
+      ClusterModel clusterModel;
+      try (AutoCloseable ignored = _loadMonitor.acquireForModelGeneration(new OperationProgress())) {
+        clusterModel = _loadMonitor.clusterModel(_time.milliseconds(),
+            requirements,
+            new OperationProgress());
         _lastCheckedModelGeneration = clusterModel.generation();
+      }  catch (Exception e) {
```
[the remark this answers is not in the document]
kidkun: because we have filter out the ready goal before calculate cluster model(and use a proper requirement), this exception will not happen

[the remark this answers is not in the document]
efeg: Consider the following scenario:
1. `readyGoals` are determined based on  the latest cluster metadata [here](https://github.com/linkedin/cruise-control/pull/465/files#diff-592d1d65991d45cc83e40b7d30f93eafR101) -- i.e. note that `LoadMonitor#meetCompletenessRequirements(ModelCompletenessRequirements)` makes a call to `MetadataClient#refreshMetadata` to retrieve this information.
2. Then the clusterModel generation [here](https://github.com/linkedin/cruise-control/pull/465/files#diff-592d1d65991d45cc83e40b7d30f93eafR113) calls `LoadMonitor#clusterModel(long, ModelCompletenessRequirements, OperationProgress)`, which calls `LoadMonitor#clusterModel(long, long, ModelCompletenessRequirements, OperationProgress`. In this function, there is a call to `MetadataClient#refreshMetadata` to retrieve the latest cluster metadata -- which can be updated depending on the value of `_metadataTTL`.

-- However unlikely it is, the `NotEnoughValidWindowsException` can still be thrown in this method.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
efeg: Thanks for the analysis! -- here are some thoughts:

1. While I agree that redundant calculations of `clusterModel` should be avoided, I am not entirely sure if the current implementation of this part had redundant calculations of cluster model. Note that the current implementation calculates the cluster model only if the cluster had changes (e.g. `if (newModelNeeded) == true`). If we skip this step and use an old cluster model, then we potentially would have `false positives` or `false negatives`. Please note that `GoalViolationDetector#optimizeForGoal` modifies the passed `clusterModel` if there has been any violations to the goal. Using this modified cluster model does not provide an accurate picture of if a goal has been violated or not.

2. (potential future improvement) In the current implementation, we keep a single `clusterModel`, and update the model when/if necessary. But there are some goals which only deal with the topology of the cluster model. For such goals, we can work with a "skinnier" version of the cluster model, which does not need to perform an load aggregation. This would, I presume, provide improvements in the time it takes to generate a cluster model.

Let me know if I can provide further clarification!

kidkun: I missed the point that `GoalViolationDetector#optimizeForGoal` will modify the `ClusterModel`(no wonder the original code logic looks confusing to me...). then in this case what do you feel about adding a `clone()` method to `ClusterModel`(for the sake of reducing the times of  load aggregation)?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolations.java
```diff
@@ -23,73 +23,67 @@
 public class GoalViolations extends KafkaAnomaly {
   private static final Logger LOG = LoggerFactory.getLogger(GoalViolations.class);
   private final KafkaCruiseControl _kafkaCruiseControl;
-  private final List<Violation> _goalViolations = new ArrayList<>();
+  // The priority order of goals is maintained here.
+  private final List<String> _fixableViolatedGoals;
+  private final List<String> _unfixableViolatedGoals;
```
efeg: Rather than having two independent lists, would it make our implementation more clear to have a single `Map<Boolean, List<String>>` for these two?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolations.java
```diff
@@ -23,73 +23,67 @@
...
+      _unfixableViolatedGoals.add(goalName);
+    }
   }
 
   /**
    * Get all the goal violations.
    */
-  public List<Violation> violations() {
-    return _goalViolations;
+  public Set<String> violations() {
```
efeg: If we use `Map<Boolean, List<String>>` (please see prev. proposal), then would we avoid creating a new `Set` for the response?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolations.java
```diff
@@ -23,73 +23,67 @@
...
-
-    public String goalName() {
-      return _goalName;
-    }
-
-    public Set<ExecutionProposal> executionProposals() {
-      return _executionProposals;
+    if (_unfixableViolatedGoals.isEmpty()) {
+      try {
+        // Fix the fixable goal violations with rebalance operation.
```
efeg: Goals used for (1) detecting the anomalies, and the (2) goals used for fixing them are configured independently. Hence, here we seem to be making the implicit assumption that `_unfixableViolatedGoals` are  subset of goals used for fixing the anomalies. This sounds like a reasonable assumption -- but then we should have a sanity check in (maybe in `KafkaCruiseControlConfig`) to ensure that this is the case by construction.

PR #479, opened 2019-01-11
Ensure that recently demoted brokers do not receive leadership if requested/configured so.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/480.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/OptimizationOptions.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/OptimizationOptions.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2019 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.cruisecontrol;
+
+import java.util.Collections;
+import java.util.Set;
+
+
+/**
+ * A class to indicate options intended to be used during optimization of goals.
+ */
+public class OptimizationOptions {
+  private final Set<String> _excludedTopics;
+  private final Set<Integer> _excludedBrokersForLeadership;
+
+  /**
+   * Default value for {@link #_excludedBrokersForLeadership} is an empty set.
+   */
+  public OptimizationOptions(Set<String> excludedTopics) {
+    this(excludedTopics, Collections.emptySet());
+  }
+
+  public OptimizationOptions(Set<String> excludedTopics, Set<Integer> excludedBrokersForLeadership) {
+    _excludedTopics = excludedTopics;
+    _excludedBrokersForLeadership = excludedBrokersForLeadership;
+  }
+
+  public Set<String> excludedTopics() {
+    return _excludedTopics;
+  }
+
+  public Set<Integer> excludedBrokersForLeadership() {
+    return Collections.unmodifiableSet(_excludedBrokersForLeadership);
+  }
+}
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
@@ -5,6 +5,7 @@
 
 package com.linkedin.kafka.cruisecontrol.analyzer.goals;
 
+import com.linkedin.kafka.cruisecontrol.OptimizationOptions;
 import com.linkedin.kafka.cruisecontrol.analyzer.ActionAcceptance;
 import com.linkedin.kafka.cruisecontrol.analyzer.AnalyzerUtils;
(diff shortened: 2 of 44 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
```diff
@@ -0,0 +1,134 @@
...
+   * Check whether the proposed action is legit. An action is legit if it is:
+   * (1) a replica movement and the destination does not have a replica of the same partition, or
+   * (2) a leadership movement, the replica is a leader and the destination broker has a follower of the same partition.
+   *
+   * @param replica Replica that is affected from the given action type.
+   * @param destinationBroker Destination broker.
+   * @param actionType Action type.
+   * @return True if the move is legit, false otherwise.
+   */
+  static boolean legitMove(Replica replica, Broker destinationBroker, ActionType actionType) {
```
kidkun: can we add a sanity check in this method to double check that no leader is moved into excluded broker. Because the logic to filter out the excluded broker from candidate broker is scatter in code. Through it looks like we add filtering in all the place for now, it is possible in the future that in some new operation-generating code we miss this filtering.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/OptimizationOptions.java
```diff
@@ -0,0 +1,37 @@
+/*
```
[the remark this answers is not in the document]
efeg: Can be moved under `com.linkedin.kafka.cruisecontrol.analyzer`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
```diff
@@ -238,12 +252,14 @@ protected Broker maybeApplyBalancingAction(ClusterModel clusterModel,
...
    * @param optimizedGoals Optimized goals.
+   * @param optimizationOptions Options to take into account during optimization -- e.g. excluded brokers for replica move.
    * @return True the swapped in replica if succeeded, null otherwise.
    */
   Replica maybeApplySwapAction(ClusterModel clusterModel,
                                Replica sourceReplica,
                                SortedSet<Replica> candidateReplicasToSwapWith,
-                               Set<Goal> optimizedGoals) {
+                               Set<Goal> optimizedGoals,
+                               OptimizationOptions optimizationOptions) {
```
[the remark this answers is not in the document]
efeg: Agreed that the relevant logic can be moved inside `getEligibleReplicasForSwap`. Regarding the logic, it should be sufficient to check the swap one way, because candidate replicas are chosen to be among the ones that are eligible to be received by this broker.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
efeg: Adding the proposed sanity check to `legitMove` may have some issues:
1. The check would be incomplete due to goals that do not use `legitMove` -- e.g. `PreferredLeaderElectionGoal`.
2. `legitMove` is a frequently accessed method -- i.e. it is used for each balancing action proposal. Hence, adding an extra logic here leads to a significant overhead.
3. Ideally, our unit tests should catch such violations. To mitigate this issue, I added test cases that we missed before.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
```diff
@@ -131,4 +136,62 @@ static boolean legitMove(Replica replica, Broker destinationBroker, ActionType a
     return actionType == ActionType.LEADERSHIP_MOVEMENT && replica.isLeader()
            && destinationBroker.replica(replica.topicPartition()) != null;
   }
+
+  /**
+   * Get eligible replicas among the given candidate replicas for the proposed swap operation of the source replica.
+   * Invariant: No replica is eligible if the destination is excluded for leadership and the source replica is the leader.
```
kidkun: `destination` -> `candidate broker`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/internals/CandidateBroker.java
```diff
@@ -0,0 +1,92 @@
...
+    return result == 0 && _broker.id() == that._broker.id();
+  }
+
+  @Override
+  public int hashCode() {
+    return _broker.id();
+  }
+
+  @Override
+  public String toString() {
```
kidkun: can we also put the information about what resource this utilization refer to here?

Issue #494, opened 2019-01-22
Ensure that offline brokers do not trigger goal violation detection or its self healing


efeg:
Self healing for "broker failure detector" is already responsible for fixing clusters with failed brokers by removing replicas from such brokers. However, currently the goal violation detector also attempts to detect and fix the cluster with failed brokers by rebalancing the whole cluster with the configured goals.

-- Self healing for goal violation should not be triggered if there are offline brokers.

PR #495, opened 2019-01-22
Ensure that offline brokers do not trigger goal violation detection or its self healing.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/494.


alexwang08:
+LGTM

Issue #506, opened 2019-01-26
Cruise Control fails to bootstrap if a sample store topic has already been created in a non-rack-aware cluster


efeg:
`KafkaSampleStore` implicitly enforces rack-awareness even if the sample store topics have already been created in a non-rack-aware cluster -- e.g. via the initial start of Cruise Control, or manually.

During this time, the cluster state is stuck in the `LOADING` monitor state with the following response:
```
MonitorState: {state: LOADING(0.000% trained), LoadingProgress: NaN%, NumValidWindows: (0/0): (NaN%), NumValidPartitions: 0/510 (0.000%), FlawedPartitions: 0}
ExecutorState: {state: NO_TASK_IN_PROGRESS}
AnalyzerState: {isProposalReady: false, readyGoals: []}
AnomalyDetectorState: {selfHealingEnabled:[], selfHealingDisabled:[METRIC_ANOMALY, GOAL_VIOLATION, BROKER_FAILURE], recentGoalViolations:[], recentBrokerFailures:[], recentMetricAnomalies:[]}
```

The root cause of the issue is the premature sanity check in [KafkaSampleStore#maybeIncreaseTopicReplicationFactor](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java#L263). Before this check, the code should verify whether all replicas of the topic already satisfy the intended replication factor regardless of the number of racks in the cluster.

-- The impact of this issue is an error log for each sample store topic. Otherwise, it does not block Cruise Control from starting and functioning normally:

```
ERROR Skip updating topic KafkaCruiseControlPartitionMetricSamples configuration due to failure:Unable to increase topic KafkaCruiseControlPartitionMetricSamples replica factor to 2 since there are only 1 racks in the cluster.. (com.linkedin.kafka.cruisecontrol.monitor.sampling.KafkaSampleStore)
ERROR Skip updating topic KafkaCruiseControlModelTrainingSamples configuration due to failure:Unable to increase topic KafkaCruiseControlModelTrainingSamples replica factor to 2 since there are only 1 racks in the cluster.. (com.linkedin.kafka.cruisecontrol.monitor.sampling.KafkaSampleStore)
```

PR #508, opened 2019-01-28
Increase replication factor of Cruise Control topics in none rack-aware way if there are not enough racks in the cluster.

kidkun:
fix for [506](https://github.com/linkedin/cruise-control/issues/506)

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
@@ -234,8 +234,13 @@ public class KafkaSampleStore implements SampleStore {
 
   /**
    * Increase the replication factor of a Kafka topic through adding new replicas in a rack-aware, round-robin way.
-   * If Zookeeper does not have rack information about brokers, then rack-aware property is not guaranteed.
-   * In this case it is only guaranteed that new replicas are added to brokers which do not currently host the partition.
+   * There are two scenarios that rack awareness property is not guaranteed.
+   * <ul>
+   *   <li> If Zookeeper does not have rack information about brokers, then it is only guaranteed that new replicas are
+   *   added to brokers which do not currently host the partition.</li>
+   *   <li> If replication factor to set for the topic is larger than number of racks in the cluster, then rack awareness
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -260,9 +265,12 @@ private void maybeIncreaseTopicReplicationFactor(ZkUtils zkUtils,
       brokersByRack.get(rack).add(bm.id());
       rackByBroker.put(bm.id(), rack);
     }
+
+    boolean skipRackAwarenessCheck = false;
```
efeg: Would it be possible to make `skipSampleStoreTopicRackAwarenessCheck` a configuration? We can keep the default as `true` (and throw an exception), but if the user intentionally would like to skip this check, s/he can set this config to skip this check.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -260,9 +265,12 @@ private void maybeIncreaseTopicReplicationFactor(ZkUtils zkUtils,
...
     if (replicationFactor > brokersByRack.size()) {
-      throw new RuntimeException("Unable to increase topic " + topic + " replica factor to " + replicationFactor
-                                 + " since there are only " + brokersByRack.size() + " racks in the cluster.");
+      skipRackAwarenessCheck = true;
+      LOG.warn("Target replica factor for topic " + topic + " is larger than number of racks in cluster, new replica maybe" +
```
efeg: `replica ` -> `replication`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
kidkun: I believe you mean default value to be `false`.

efeg: Correct, I meant to say `false`.

Issue #514, opened 2019-02-02
Ensure that self-healing in ReplicaDistributionGoal is limited to eligible and immigrant replicas.


efeg:
If there are offline replicas, the optimization in this goal should be limited to the offline replicas, and replicas that have been moved away from their original location using the previous goals.

PR #515, opened 2019-02-02
Ensure that self-healing in ReplicaDistributionGoal is limited to eligible and immigrant replicas.

efeg:
Fixes the issue https://github.com/linkedin/cruise-control/issues/514.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/internals/BrokerAndSortedReplicas.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/internals/BrokerAndSortedReplicas.java
@@ -6,8 +6,11 @@
 package com.linkedin.kafka.cruisecontrol.analyzer.goals.internals;
 
 import com.linkedin.kafka.cruisecontrol.model.Broker;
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/internals/BrokerAndSortedReplicas.java
```diff
@@ -36,6 +39,33 @@ public Broker broker() {
...
+  /**
+   * Get sorted replicas to move out of this broker, which are filtered based on the state of the cluster.
+   *
+   * If cluster has dead brokers, but this broker is alive, then replicas to move out are limited to the immigrants.
+   *
+   * @param clusterModel The state of the cluster.
+   * @return sorted replicas to move out of this broker, which are filtered based on the state of the cluster.
+   */
+  public List<Replica> replicasToMoveOut(ClusterModel clusterModel) {
+    List<Replica> replicasToMoveOut = new ArrayList<>(sortedReplicas());
```
kidkun: Instead of copying the whole sorted set and then prune, maybe it is more efficient that we scan the sorted set first and find the end element, then just copy a subset using `SortedSet<E> subSet(E fromElement, E toElement)`

Issue #488, opened 2019-01-18
[Cached rebalance request] Multiple iterations in goal optimization causes the rebalance request to take a long time


efeg:
**Background**: A _cached rebalance request_ is a call to the rebalance endpoint with no `goals`, `excluded_topics`, or `exclude_recently_demoted_brokers` parameters specified in the request. Such a request uses the cached proposals or, if the cache is stale, attempts to cache the proposals generated via the optimization process.

**Issue**: the goal optimization might take several iterations, because the value to be cached might be stale by the time a single cycle of goal optimization is complete. This, in particular, can be observed in clusters with a large number of partitions.

The following conditions make the cached proposals stale and requires a recomputation of them:
* Changes in the topology of the cluster -- i.e. its brokers, topics, partitions, leaders, or their distribution might have changed,
* Changes in the partition load -- e.g. the sampling process collects new samples and rolls out a new window, or

PR #520, opened 2019-02-05
Add acceptable staleness lags for model generation to avoid aggressive invalidation of cached rebalance cache

efeg:
Fixes the issue https://github.com/linkedin/cruise-control/issues/488.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/ModelGeneration.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/ModelGeneration.java
@@ -12,6 +12,9 @@ import java.io.Serializable;
  */
 public class ModelGeneration implements Serializable {
   private static final long serialVersionUID = 1494955635123L;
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/ModelGeneration.java
```diff
@@ -40,6 +43,18 @@ public boolean equals(Object o) {
     return _clusterGeneration == other.clusterGeneration() && _loadGeneration == other.loadGeneration();
   }
 
+  /**
+   * Check whether this model generation is stale. A model generation is stale if it is behind the latest model
+   * generation more than the acceptable staleness limit.
+   *
+   * @param latestModelGeneration The model generation to compare against this model generation for staleness check.
+   * @return True if this model generation is within the acceptable staleness limits, false otherwise.
```
kidkun: here `True` means is beyond acceptable staleness limits, right?

Issue #473, opened 2019-01-07
Expose latest self healing failures due to OptimizationFailureException in anomaly detector state


efeg:
In some clusters, if the analyzer receives a `OptimizationFailureException` because it fails to satisfy some hard goal's requirement during self healing, e.g. `RackAwareGoal`, looks like this is printed in logs, but not exposed elsewhere, causing difficulty to detect such cases.

We should ideally expose it in `state?substates=anomaly_detector`.

PR #527, opened 2019-02-06
Add self-healing state under anomaly detector state.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/473.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
@@ -199,17 +199,17 @@ public class AnomalyDetector {
             // Call the anomaly notifier to see if a fix is desired.
             if (anomaly instanceof BrokerFailures) {
               BrokerFailures brokerFailures = (BrokerFailures) anomaly;
-              _anomalyDetectorState.addAnomaly(AnomalyType.BROKER_FAILURE, brokerFailures);
+              _anomalyDetectorState.addAnomalyDetection(AnomalyType.BROKER_FAILURE, brokerFailures);
               notificationResult = _anomalyNotifier.onBrokerFailure(brokerFailures);
               _brokerFailureRate.mark();
             } else if (anomaly instanceof GoalViolations) {
               GoalViolations goalViolations = (GoalViolations) anomaly;
-              _anomalyDetectorState.addAnomaly(AnomalyType.GOAL_VIOLATION, goalViolations);
+              _anomalyDetectorState.addAnomalyDetection(AnomalyType.GOAL_VIOLATION, goalViolations);
               notificationResult = _anomalyNotifier.onGoalViolation(goalViolations);
               _goalViolationRate.mark();
             } else if (anomaly instanceof KafkaMetricAnomaly) {
               KafkaMetricAnomaly metricAnomaly = (KafkaMetricAnomaly) anomaly;
-              _anomalyDetectorState.addAnomaly(AnomalyType.METRIC_ANOMALY, metricAnomaly);
+              _anomalyDetectorState.addAnomalyDetection(AnomalyType.METRIC_ANOMALY, metricAnomaly);
               notificationResult = _anomalyNotifier.onMetricAnomaly(metricAnomaly);
               _metricAnomalyRate.mark();
             }
@@ -290,10 +290,28 @@ public class AnomalyDetector {
       return false;
     }
 
+    private AnomalyType getAnomalyType(Anomaly anomaly) {
+      if (anomaly instanceof BrokerFailures) {
+        return AnomalyType.BROKER_FAILURE;
+      } else if (anomaly instanceof GoalViolations) {
+        return AnomalyType.GOAL_VIOLATION;
+      } else if (anomaly instanceof KafkaMetricAnomaly) {
+        return AnomalyType.METRIC_ANOMALY;
+      } else {
+        throw new IllegalStateException("Unrecognized type for anomaly " + anomaly);
+      }
+    }
+
     private void fixAnomaly(Anomaly anomaly) throws Exception {
       if (isFixable(anomaly)) {
         LOG.info("Fixing anomaly {}", anomaly);
-        anomaly.fix();
+        boolean startedSuccessfully = false;
+        try {
+          startedSuccessfully = anomaly.fix();
+        } finally {
+          _anomalyDetectorState.addSelfHealing(getAnomalyType(anomaly), startedSuccessfully);
+          LOG.info("Self-healing {}.", startedSuccessfully ? "started successfully" : "failed to start");
+        }
       }
 
       _anomalies.clear();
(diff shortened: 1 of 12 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -50,13 +60,41 @@ protected boolean removeEldestEntry(Map.Entry<Long, Anomaly> eldest) {
...
   }
 
+  /**
+   * Add anomaly information for which self-healing has been triggered.
+   *
+   * @param anomalyType Type of the self-healed anomaly.
+   * @param startedSuccessfully True if self-healing attempt was started successfully, false otherwise -- e.g. a failure
+   * due to an {@link com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException}.
+   */
+  public void addSelfHealing(AnomalyType anomalyType, boolean startedSuccessfully) {
```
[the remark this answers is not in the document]
efeg: Good idea -- thanks for the feedback!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -197,21 +198,28 @@ public void run() {
...
+            switch (anomalyType) {
+              case GOAL_VIOLATION:
+                GoalViolations goalViolations = (GoalViolations) anomaly;
+                _anomalyDetectorState.addAnomalyDetection(anomalyType, goalViolations);
+                notificationResult = _anomalyNotifier.onGoalViolation(goalViolations);
+                _goalViolationRate.mark();
+                break;
+              case BROKER_FAILURE:
+                BrokerFailures brokerFailures = (BrokerFailures) anomaly;
+                _anomalyDetectorState.addAnomalyDetection(anomalyType, brokerFailures);
```
kidkun: for the same broker failure, we may call this multiple time( the first time get a `CHECK_WITH_DELAY`) and each time it will create a new `AnomalyState` and reset the `_detectionMs` to current time(which is not the true anomaly detection time)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyState.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Collections;
+import java.util.List;
+
+
+/**
+ * A class to indicate how an anomaly is handled.
+ */
+public class AnomalyState {
+  private final Anomaly _anomaly;
+  private final long _detectionMs;
```
kidkun: `AnomalyState` is created by `AnomalyDetector` when it poll from the blocking queue and begin to handle the anomaly(it may not handle the anomaly after polling from the queue due to executor is busy), so if we record the anomaly detection time as the time we create `AnomalyState`,  it might be inaccurate.
I feel a better idea will be record anomaly detection time as the time anomaly is enqueue to the blocking queue by `GoalViolationDetector/BrokerFailureDetector/MetricAnomalyDetector`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -247,53 +257,55 @@ public void run() {
...
     private void checkWithDelay(Anomaly anomaly, long delay) {
-      // We only do delayed check for broker failures.
-      if (anomaly instanceof BrokerFailures) {
+      // Anomaly detector does delayed check for broker failures, otherwise it ignores the anomaly.
+      if (getAnomalyType(anomaly) == AnomalyType.BROKER_FAILURE) {
         LOG.debug("Scheduling broker failure detection with delay of {} ms", delay);
         _detectorScheduler.schedule(_brokerFailureDetector::detectBrokerFailures, delay, TimeUnit.MILLISECONDS);
+        _anomalyDetectorState.onAnomalyHandle(anomaly, AnomalyState.Status.CHECK_WITH_DELAY);
+      } else {
+        _anomalyDetectorState.onAnomalyHandle(anomaly, AnomalyState.Status.IGNORED);
```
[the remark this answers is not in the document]
efeg: Good catch! Indeed, the code should update the anomaly detector state regardless of the executor state. (moved the `addAnomalyDetectionState` before the corresponding check).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
efeg: TL;DR -- This is intentional.
For the broker failure detection, we enforce a grace period -- i.e. before taking an action on broker failures, we intentionally wait for a certain period time. So the final mitigation action of an initial broker failure should be `CHECK_WITH_DELAY`, and the following one should be `FIX_STARTED` if the self-healing is enabled for broker failures. Ideally, we would like to preserve this transparency to indicate when wast the first time the failure was detected, and when was the second time the same failure was identified.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyState.java
efeg: I believe you mean "time anomaly is enqueue to the blocking queue" -> "time anomaly is polled from the blocking queue"?
If so, the new logic regarding moving the `addAnomalyDetectionState` before the corresponding check should resolve this.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
kidkun: I see, so `CHECK_WITH_DELAY` will actually create a new anomaly.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyState.java
kidkun: yes, the logic change mitigate this issue, now the time we are recording are late by `dequeue_time-enqueue_time` and I think it is ok.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -172,6 +172,22 @@ public boolean setSelfHealingFor(AnomalyType anomalyType, boolean isSelfHealingE
     return oldSelfHealingEnabled;
   }
 
+  private void addAnomalyDetectionState(AnomalyType anomalyType, Anomaly anomaly) {
+    switch (anomalyType) {
+      case GOAL_VIOLATION:
+        _anomalyDetectorState.addAnomalyDetection(anomalyType, (GoalViolations) anomaly);
```
kidkun: the cast here seems redundant here.

Issue #541, opened 2019-02-09
Feature request: A new task state for failed user tasks


akashvacher:
Currently cruise control assigns 'completed' state to all user tasks, even if they run into a failure scenario. It would be great to have cruise control explicitly mark such user tasks differently to indicate that they did not complete normally.
This will further allow user to make more refined queries for failed tasks to know exactly why it failed.

PR #551, opened 2019-02-13
Mark failed user task's state as CompletedWithError in UserTaskManager.

kidkun:
fix for issue [541](https://github.com/linkedin/cruise-control/issues/541)

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
@@ -262,21 +262,15 @@ public class UserTaskManager implements Closeable {
     return userTaskId;
   }
 
-  private synchronized boolean isActiveUserTasksDone(UUID userTaskId) {
-    UserTaskInfo userTaskInfo = _allUserTaskIdToFutureMap.get(TaskState.ACTIVE).get(userTaskId);
-    if (userTaskInfo == null || userTaskInfo.futures().isEmpty()) {
-      return true;
-    }
-
-    List<OperationFuture> futures = userTaskInfo.futures();
-    return futures.get(futures.size() - 1).isDone();
-  }
-
    synchronized void checkActiveUserTasks() {
     Iterator<Map.Entry<UUID, UserTaskInfo>> iter = _allUserTaskIdToFutureMap.get(TaskState.ACTIVE).entrySet().iterator();
     while (iter.hasNext()) {
       Map.Entry<UUID, UserTaskInfo> entry = iter.next();
-      if (isActiveUserTasksDone(entry.getKey())) {
+      if (entry.getValue().isUserTaskDoneExceptionally()) {
+        LOG.warn("UserTask {} is complete with Exception and removed from active tasks list", entry.getKey());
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -525,6 +519,14 @@ public UserTaskInfo setState(TaskState nextState) {
       _state = nextState;
       return this;
     }
+
+    public boolean isUserTaskDone() {
+      return _futures == null || _futures.isEmpty() || _futures.get(_futures.size() - 1).isDone();
+    }
+
+    public boolean isUserTaskDoneExceptionally() {
+      return _futures != null && !_futures.isEmpty() && _futures.get(_futures.size() - 1).isCompletedExceptionally();
```
efeg: [Applies to this method and `isUserTaskDone(..)`]: Is it ever expected to have a user task with `null` or empty `_futures`? If this is an illegal state, should we have a sanity check upon creating `UserTaskInfo` with this check -- i.e. `illegalArgumentException`? Also (1) do we need to make these methods `public` -- if so, (2) do we need some synchronization to avoid missing cases where the task makes a transition from active to completed/completed_with_error?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -262,21 +262,15 @@ public UUID getUserTaskId(HttpServletRequest httpServletRequest) {
...
-    return futures.get(futures.size() - 1).isDone();
-  }
-
    synchronized void checkActiveUserTasks() {
     Iterator<Map.Entry<UUID, UserTaskInfo>> iter = _allUserTaskIdToFutureMap.get(TaskState.ACTIVE).entrySet().iterator();
     while (iter.hasNext()) {
       Map.Entry<UUID, UserTaskInfo> entry = iter.next();
-      if (isActiveUserTasksDone(entry.getKey())) {
+      if (entry.getValue().isUserTaskDoneExceptionally()) {
+        LOG.warn("UserTask {} is complete with Exception and removed from active tasks list", entry.getKey());
```
efeg: Nit: `complete with Exception ` -> `completed with exception ` (let's fix the `complete` -> `completed` at `else if...` part)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -548,7 +550,8 @@ public void run() {
    */
   public enum TaskState {
     ACTIVE("Active"),
-    COMPLETED("Completed");
+    COMPLETED("Completed"),
+    COMPLETED_WITH_ERROR("CompletedWithError");
```
efeg: Let's also update the `public String toString()` function with this.

efeg: Can we add a comment before `_allUserTaskIdToFutureMap` to indicate that `COMPLETED` is also keeps track of tasks that completed with error?

Issue #554, opened 2019-02-14
Proposal cache should be ignored if there is an ongoing execution


efeg:
If there is an ongoing execution, cache will (potentially) be invalid by the time proposals to be cached are computed; hence, it may redundantly lead to a long compute time for `rebalance` and `proposals` requests.

Moreover, regardless of the cached value, the computed proposals will not be executed if there's an ongoing execution.

PR #557, opened 2019-02-14
Ignore proposal cached when there is ongoing execution.

efeg:
Fixes the issue https://github.com/linkedin/cruise-control/issues/554.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
@@ -209,7 +209,7 @@ public class GoalOptimizer implements Runnable {
         sleepTime = 30000L;
       } else {
         try {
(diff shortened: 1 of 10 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -265,7 +277,10 @@ private void computeBestProposal() {
     }
   }
 
-  private boolean validCachedProposal() {
+  private boolean validCachedProposal() throws KafkaCruiseControlException {
```
alexwang08: semantically, _executor.hasOngoingExecution() means cache is potentially not valid, should this be in a different function?

efeg: Checking cache validity during an ongoing execution is a sanity check failure; hence, I feel it should be a part of this validity check.

PR #558, opened 2019-02-15
Record sync task and associate it with UUID

alexwang08:
Added code to store all synchronous user requests and associated them with UUID. Also refactored some code in session key and request headers. Addresses the issue: #570.


alexwang08:
@kidkun I think your suggested approach is fine, as long as we make the decision to not distinguish sync vs async tasks. Also, we need to pass in a flag to indicate all 'sync' tasks need to execute until completion instead of timing out after 10 seconds and return progress info (as in the current 'async' task case)

kidkun:
@alexwang08 maybe the `sync` flag is also not needed. 
In the above example, if in `AsyncKafkaCruiseControl` Class we define a function like
```
  public OperationFuture bootstrapLoadMonitor(BootstrapParameters parameters) {
    OperationFuture future = new OperationFuture("Bootstrap load monitor.");
    future.complete(super.bootstrapLoadMonitor(parameters));
    return future;
  }
```
Then the servlet thread will block until [original sync method](https://github.com/linkedin/cruise-control/blob/21676a67c81f1f5d826597547b03c3e5630e39f2/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java#L518) finish so later the `future.get(10)` will always immediately return the result.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -143,69 +144,87 @@ private static String httpServletRequestToString(HttpServletRequest request) {
...
+  public List<OperationFuture> getOrCreateUserTask(HttpServletRequest httpServletRequest, HttpServletResponse httpServletResponse,
+      Function<String, OperationFuture> function, int step, boolean asyncComplete) {
     UUID userTaskId = getUserTaskId(httpServletRequest);
-    List<OperationFuture> operationFutures = getFuturesByUserTaskId(userTaskId, httpServletRequest);
+    UserTaskInfo userTaskInfo = getUserTaskByUserTaskId(userTaskId, httpServletRequest);
 
-    if (operationFutures != null) {
+    if (!asyncComplete && userTaskInfo != null) {
+      throw new IllegalArgumentException(
+          String.format("Sync operation should not have %s", USER_TASK_HEADER_NAME));
```
kidkun: Is there a particular reason that sync request should not have UUID?
User can still use UUID to retrieve the previous result, this might be useful for endpoint like `kafka_cluster_state`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -143,69 +144,87 @@ private static String httpServletRequestToString(HttpServletRequest request) {
...
       }
-
-      SessionKey sessionKey = new SessionKey(httpServletRequest);
       userTaskId = _uuidGenerator.randomUUID();
-      LOG.info("Create a new UserTask {} with SessionKey {}", userTaskId, sessionKey);
-      operationFutures = insertFuturesByUserTaskId(userTaskId, function, httpServletRequest);
-
-      synchronized (_sessionKeyToUserTaskIdMap) {
-        _sessionKeyToUserTaskIdMap.put(sessionKey, userTaskId);
+      userTaskInfo = insertFuturesByUserTaskId(userTaskId, function, httpServletRequest);
```
kidkun: Note in method `insertFuturesByUserTaskId `, there is [limit](https://github.com/linkedin/cruise-control/blob/f7e15c4aa6c784129dfd196355c8454e89ede4f3/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java#L326) on number of active tasks(currently is 5).  

Consider the scenario that there some requests get hang in execution and use up the slot,  then user cannot send `user_tasks` request to check because this task insertion will fail. I think we probably need to take special care to `user_tasks` request here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -143,69 +144,87 @@ private static String httpServletRequestToString(HttpServletRequest request) {
...
+      // Only create user task id to session mapping for async request
+      if (asyncComplete) {
+        createSessionKeyMapping(userTaskId, httpServletRequest);
```
kidkun: Same as above, is there a particular reason that sync request should not have a session?
User can still use session+URL to retrieve the previous result

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: It might be more appropriate to rewrite `/kafka_cluster_state` as async request? Since `kafka_cluster_state` takes longer than rest of sync requests. I consider each sync request as one-off and one should not need a UUID to retrieve the previous result.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: Similar to above, if sync requests are meant to return right away, we don't need to keep them in session and hope to retrieve their result using UUID.

alexwang08:
Thanks for the feedback, left some comments.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
```diff
@@ -46,7 +46,7 @@ public void testCreateUserTask() {
                                                           100, new MockTime(), mockUUIDGenerator);
     // test-case: create user-task based on request and get future
     OperationFuture future1 =
-        userTaskManager.getOrCreateUserTask(mockHttpServletRequest1, mockHttpServletResponse, uuid -> future, 0).get(0);
+        userTaskManager.getOrCreateUserTask(mockHttpServletRequest1, mockHttpServletResponse, uuid -> future, 0, true).get(0);
```
efeg: This test class seems to be limited to async tasks -- can we add test cases for sync tasks?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -138,74 +146,105 @@ protected boolean removeEldestEntry(Map.Entry<UUID, UserTaskInfo> eldest) {
...
    * @param step The index of the step that has to be added or fetched.
+   * @param asyncComplete Indicate whether the task is sync or async.
    * @return The list of {@link OperationFuture} for the linked UserTask.
    */
-  public List<OperationFuture> getOrCreateUserTask(HttpServletRequest httpServletRequest,
-                                                   HttpServletResponse httpServletResponse,
-                                                   Function<String, OperationFuture> function,
-                                                   int step) {
+  public List<OperationFuture> getOrCreateUserTask(HttpServletRequest httpServletRequest, HttpServletResponse httpServletResponse,
+      Function<String, OperationFuture> function, int step, boolean asyncComplete) {
```
efeg: Bad indentation -- can we ensure that we avoid making indentation errors please?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -591,4 +643,35 @@ public String toString() {
...
+
+  /**
+   * {@link _allUuidToUserTaskInfoMap} stores active and completed user tasks for history. Some historic user requests are
+   * less important for the user, e.g. {@link EndPoint.USER_TASKS} is not very interesting. To prevent un-interesting but
+   * frequently-called endpoints from flooding out interesting requests, we need to periodically cleanup requests. We can set
+   * per endpoint limit and utilize the {@link UserTaskScanner} to do the cleaning.
+   */
+  private class UserTaskCleanerInfo {
+    private final int _limit;
+    private final LinkedList<UUID> _tasks;
```
efeg: Why would we prefer `LinkedList` over `LinkedHashMap`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
efeg: I agree with @kidkun on this one. Users should be able to retrieve the response of sync requests using UUIDs. 

Actually, I am confused and might be missing a basic point -- how would users retrieve the response of sync requests with the current implementation?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -138,74 +146,105 @@ protected boolean removeEldestEntry(Map.Entry<UUID, UserTaskInfo> eldest) {
...
+   * Create the UserTask reference for sync or async task if it doesn't exist.
    *
    * This method creates references {@link UserTaskInfo} and maps it to {@link HttpSession} from httpServletRequest. The
    * {@link HttpSession} is also used to fetch {@link OperationFuture} for the in-progress/completed UserTask. If the
    * UserTaskID is passed in the httpServletRequest header then that takes precedence over {@link HttpSession} to fetch
-   * the {@link OperationFuture} for the UserTask.
+   * the {@link OperationFuture} for the UserTask. Sync task is not saved in Session map.
    *
-   * @param httpServletRequest the HttpServletRequest to create the UserTaskInfo reference.
+   * @param httpServletRequest the HttpServletRequest to create the UserTaskInfo reference. Sync request should not contain
```
efeg: `Sync request should not contain UUID in the header.` -> then how would users retrieve the response associated with a sync request?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -61,6 +63,7 @@
   private final UUIDGenerator _uuidGenerator;
   private final Map<EndPoint, Timer> _successfulRequestExecutionTimer;
   private final long _completedUserTaskRetentionTimeMs;
+  private final Map<EndPoint, UserTaskCleanerInfo> _userTaskCleanerMap;
```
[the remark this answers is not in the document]
alexwang08: Currently we already have a remove completed task mechanism via setting `_completedUserTaskRetentionTimeMs`. Do we want this semantic categorization be an extension to that? If in-memory retention is 24 hours (or days), it might not save us much to have a per-category retention. If the in-memory retention will potentially be weeks or months, then the memory saving is more significant. Just want to understand the purpose and benefit of semantic categorization more.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: For the current design, we just need a FIFO data structure, so `LinkedHashMap` is not necessary. We could use it for other designs.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -553,14 +603,16 @@ public void run() {
...
+        cleanUserTasks();
       } catch (Throwable t) {
         LOG.warn("Received exception when trying to expire sessions.", t);
       }
     }
   }
 
   /**
-   * Possible state of tasks.
+   * Possible state of tasks. We currently accept {@link TaskState#ACTIVE} and {@link TaskState#COMPLETED}, in the
```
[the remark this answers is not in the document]
alexwang08: Added Todo to this comment. Will address #571 in a future pr.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -138,74 +146,105 @@ protected boolean removeEldestEntry(Map.Entry<UUID, UserTaskInfo> eldest) {
     this(sessionExpiryMs, maxActiveUserTasks, completedUserTaskRetentionTimeMs, maxCachedCompletedUserTasks, time, new UUIDGenerator());
   }
 
+  /**
+   * Sets history limit for different endpoints. For now, only limit number of {@link EndPoint.USER_TASKS} in history to 5 entries.
+    */
+  private void initUserTaskCleanerMap() {
+    _userTaskCleanerMap.put(EndPoint.USER_TASKS, new UserTaskCleanerInfo(5));
```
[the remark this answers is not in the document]
alexwang08: Made the limit to a static field. Will make it configurable in a future patch

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
alexwang08: Added test-case: for sync task, UserTaskManager does not create mapping between request URL and UUID. This is the only difference between sync and async request as far as `UserTaskManager.getOrCreateUserTask` is concerned, so the same test cases will cover sync task.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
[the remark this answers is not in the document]
efeg: > Currently we already have a remove completed task mechanism via setting `_completedUserTaskRetentionTimeMs`. Do we want this semantic categorization be an extension to that?

In the current behavior, we  have one giant pool to put all requests regardless of which endpoint they come from. We would like to shard the pool into different multiple subpools each representing a different category. Each such category has its own (1) time-based and (2) number-of-item-based retention -- i.e. they will not share the same `24 hours` retention or X-number of items limit.

> If in-memory retention is 24 hours (or days), it might not save us much to have a per-category retention. If the in-memory retention will potentially be weeks or months, then the memory saving is more significant.

The reasoning is to apply selective retention to different requests. For example, I may not be very interested in the last time someone queried the `partition_load` endpoint and the corresponding response (I can get the historical (to some extent) partition load now without the need for keeping the historical response). However, I would be more interested in seeing the request and the corresponding proposals of a recent `remove_broker` operation. But if I put everything under the same pool, then (1) due to less interesting calls and `number-of-item-based retention`, I will lose valuable information (they will be ejected from history) (2) due to the same time-based retention, I cannot keep more important requests selectively -- e.g. one might want to keep the response of `rebalance` longer than `kafka_cluster_state`.

> Just want to understand the purpose and benefit of semantic categorization more.

Hope the explanation above helps!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -138,74 +146,105 @@ protected boolean removeEldestEntry(Map.Entry<UUID, UserTaskInfo> eldest) {
     this(sessionExpiryMs, maxActiveUserTasks, completedUserTaskRetentionTimeMs, maxCachedCompletedUserTasks, time, new UUIDGenerator());
   }
 
+  /**
+   * Sets history limit for different endpoints. For now, only limit number of {@link EndPoint.USER_TASKS} in history to 5 entries.
```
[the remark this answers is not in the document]
efeg: I see that we have moved the hardcoded value to `COMPLETED_USER_TASKS_LIMIT` (good!) -- but we still have the comment: `For now, only limit...to {@link #COMPLETED_USER_TASKS_LIMIT} entries`.

Why is it for now? What change do we plan to make here that affects this particular documentation?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -591,4 +643,35 @@ public String toString() {
...
     }
   }
+
+  /**
+   * {@link _allUuidToUserTaskInfoMap} stores active and completed user tasks for history. Some historic user requests are
+   * less important for the user, e.g. {@link EndPoint.USER_TASKS} is not very interesting. To prevent un-interesting but
+   * frequently-called endpoints from flooding out interesting requests, we need to periodically cleanup requests. We can set
+   * per endpoint limit and utilize the {@link UserTaskScanner} to do the cleaning.
+   */
+  private class UserTaskCleanerInfo {
```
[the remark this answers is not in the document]
efeg: @alexwang08 I believe this comment from my previous review was not addressed? To avoid redoing review for the same issue, can we either comment on a feedback to explain why it is not addressed or address the issue?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/EmptySyncRequestResult.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2019 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+package com.linkedin.kafka.cruisecontrol.servlet.response;
+
+import com.linkedin.kafka.cruisecontrol.servlet.parameters.CruiseControlParameters;
+
+public class EmptySyncRequestResult extends AbstractCruiseControlResponse {
```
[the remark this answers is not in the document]
efeg: @alexwang08 Was this class redundant? Why did we remove it? To avoid redoing review for the same issue, can we either comment on a feedback to explain why it is not addressed or address the issue?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
efeg: @alexwang08 Do we allow users to send UUID in header? If not we should let them send UUID in request and associate that UUID with the request.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
```diff
@@ -71,6 +71,22 @@ public void testCreateUserTask() {
     Assert.assertEquals(userTaskHeaderValue.getValue(), testUserTaskId.toString());
     Assert.assertEquals(future, future3);
 
+    EasyMock.reset(mockHttpServletResponse);
+    // test-case: for sync task, UserTaskManager does not create mapping between request URL and UUID.
+    HttpServletRequest mockHttpServletRequest3 = prepareRequest(null, null, "test_sync_request", Collections.emptyMap());
+    userTaskManager.getOrCreateUserTask(mockHttpServletRequest3, mockHttpServletResponse, uuid -> future, 0, false).get(0);
```
efeg: Why do we not store and check whether the response is as expected -- i.e. is there nothing to verify?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
```diff
@@ -71,6 +71,22 @@ public void testCreateUserTask() {
...
+
+    // mockHttpServletRequest3 does not contain User-task-id, so the only place to retrieve non-null user-task-id is _sessionKeyToUserTaskIdMap.
```
efeg: Why do we use inconsistent names such as `user-task-id` and `User-task-id` to represent `UserTaskManager#USER_TASK_HEADER_NAME`? Would it be possible to say something like "If the request header does not contain a valid user task id..."?

efeg: Why do we talk about the current implementation details of a different class's variable in this test case -- i.e. `_sessionKeyToUserTaskIdMap`? When we change the implementation or even the name of `_sessionKeyToUserTaskIdMap`, are we supposed to find this comment and fix it? Or is it acceptable for this comment be stale in that case?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
```diff
@@ -71,6 +71,22 @@ public void testCreateUserTask() {
...
+    // The 2 Assert's below contrast different behavior for sync vs async request. For sync, we won't find session to UUID mapping, so expect null;
+    // For async, we will find session to UUID mapping, so expect non null.
+    UUID savedUUID = userTaskManager.getUserTaskId(mockHttpServletRequest3);
+    Assert.assertEquals(savedUUID, null);
+
+    EasyMock.reset(mockHttpServletResponse);
+    userTaskManager.getOrCreateUserTask(mockHttpServletRequest3, mockHttpServletResponse, uuid -> future, 0, true).get(0);
```
efeg: Why do we not store and check whether the response is as expected -- i.e. is there nothing to verify?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
efeg: Ah sorry -- I meant `LinkedHashSet`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
efeg: I am still confused on this one (and on my previous comment -- copied here for convenience):
```
I agree with @kidkun on this one. Users should be able to retrieve the response of sync requests using UUIDs.

Actually, I am confused and might be missing a basic point -- how would users retrieve the response of sync requests with the current implementation?
```

I would appreciate an explanation on how users can retrieve responses.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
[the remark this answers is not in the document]
alexwang08: If 5 is enough we can keep the number and I'll delete the comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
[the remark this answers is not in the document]
alexwang08: Sorry missed this comment. I was hoping to discuss the semantic categorization further before deleting `UserTaskCleanerInfo` as it's needed for the current implementation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/EmptySyncRequestResult.java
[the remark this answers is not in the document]
alexwang08: The class was needed to pass in to `future.complete()` in my create sync request code in KafkaCruiseControlServlet before #569 got merged. Afterwards it was outdated and thus removed.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: We just need FIFO, not sure we need property in Set

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
alexwang08: We allow user to send UUID in header and associated the UUID to request

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
alexwang08: There is nothing to verify. We are just making sure we don't store `SessionKey` to `UUID` mapping in `_sessionKeyToUserTaskIdMap`.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
alexwang08: `_sessionKeyToUserTaskIdMap` is a member variable of `UserTaskManager`, so I think it's fine to mention it here as the test is trying to verify a implementation detail. Perhaps I can put `{@link #_sessionKeyToUserTaskIdMap}` instead in the comment? So IntelliJ will pickup the refactor. Alternatively I can call it 'session to UUID map (currently called 
_sessionKeyToUserTaskIdMap)' in UserTaskManager so it's more stale proof.

alexwang08: will change that.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
alexwang08: There isn't.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: Changed code to allow users retrieve the response associated with a sync request. Will delete the comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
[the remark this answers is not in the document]
alexwang08: Will having different retention parameters create inconsistency that users need to be aware of? Can the user use request endpoint type filter to find the interesting requests?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -458,14 +479,14 @@ public UserTaskInfo(HttpServletRequest httpServletRequest,
...
 
-    public UserTaskInfo(List<OperationFuture> futures,
-                        String requestUrl,
-                        String clientIdentity,
-                        long startMs,
-                        UUID userTaskId,
-                        Map<String, String[]> queryParams,
-                        EndPoint endPoint,
-                        TaskState state) {
+    UserTaskInfo(List<OperationFuture> futures,
```
efeg: Would you clarify why you make `UserTaskInfo` and its methods `package-private`?

Issue #550, opened 2019-02-13
Suppress goal violation detection while there's an ongoing execution


efeg:
Cruise Control detects goal violations while there are ongoing executions -- it should not.

PR #561, opened 2019-02-15
Suppress goal violation detection while there is an ongoing execution.

efeg:
Fixes the issue https://github.com/linkedin/cruise-control/issues/550.


alexwang08:
+LGTM

Issue #565, opened 2019-02-19
'dryRun' parameter not accepted; 'dryrun' is


mgrubent:
Currently, it appears that `cruise-control` treats the `dryrun` parameter with case-sensitivity.

That is, submitting a request (I've tried the `rebalance` and `remove-broker` endpoints) with `dryRun=([tT]rue|[fF]alse)` always causes dry-run behavior (that is, no action taken).

But, submitting the same requests with `dryrun=[fF]alse` causes `cruise-control` to execute the proposal.

`cruise-control` can choose whether it maintains case-sensitivity, but if it does, its documentation should be updated to change all doctstring instances of `dryRun` to `dryrun`.

PR #566, opened 2019-02-19
Ensure that case insensitive user requests are handled properly.

efeg:
Fixes the issue https://github.com/linkedin/cruise-control/issues/565.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
@@ -297,14 +297,22 @@ public class ParameterUtils {
     return true;
   }
 
+  /**
+   * Returns the case sensitive request parameter name, or <code>null</code> if the parameter does not exist.
+   */
+  private static String caseSensitiveParameterName(HttpServletRequest request, String parameter) {
+    return request.getParameterMap().keySet().stream().filter(parameter::equalsIgnoreCase).findFirst().orElse(null);
+  }
+
   private static boolean getBooleanParam(HttpServletRequest request, String parameter, boolean defaultIfMissing) {
-    String parameterString = request.getParameter(parameter);
-    return parameterString == null ? defaultIfMissing : Boolean.parseBoolean(parameterString);
+    String parameterString = caseSensitiveParameterName(request, parameter);
+    return parameterString == null ? defaultIfMissing : Boolean.parseBoolean(request.getParameter(parameterString));
   }
 
   private static List<String> getListParam(HttpServletRequest request, String parameter) throws UnsupportedEncodingException {
-    String paramString = urlDecode(request.getParameter(parameter));
-    List<String> retList = paramString == null ? new ArrayList<>() : Arrays.asList(paramString.split(","));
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -423,20 +435,22 @@ static Double minValidPartitionRatio(HttpServletRequest request) {
...
 
   static String resourceString(HttpServletRequest request) {
-    String resourceString = request.getParameter(RESOURCE_PARAM);
-    return resourceString == null ? DEFAULT_PARTITION_LOAD_RESOURCE : resourceString;
+    String parameterString = caseSensitiveParameterName(request, RESOURCE_PARAM);
+    return parameterString == null ? DEFAULT_PARTITION_LOAD_RESOURCE : request.getParameter(parameterString);
   }
 
   static String reason(HttpServletRequest request) {
-    String reason = request.getParameter(REASON);
```
kidkun: can we change `REASON` to `REASON_PARAM`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -583,22 +597,22 @@ static Integer concurrentMovements(HttpServletRequest request, boolean isPartiti
...
    */
   static int partitionBoundary(HttpServletRequest request, boolean isUpperBound) {
-    String partitionString = request.getParameter(PARTITION_PARAM);
-    if (partitionString == null) {
+    String parameterString = caseSensitiveParameterName(request, PARTITION_PARAM);
+    if (parameterString == null) {
       return isUpperBound ? Integer.MAX_VALUE : Integer.MIN_VALUE;
-    } else if (!partitionString.contains("-")) {
-      return Integer.parseInt(partitionString);
+    } else if (!parameterString.contains("-")) {
```
kidkun: here should we check `request.getParameter(parameterString)` instead of `parameterString`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -297,14 +297,22 @@ public static boolean hasValidParameters(HttpServletRequest request, HttpServlet
     return true;
   }
 
+  /**
+   * Returns the case sensitive request parameter name, or <code>null</code> if the parameter does not exist.
+   */
+  private static String caseSensitiveParameterName(HttpServletRequest request, String parameter) {
```
alexwang08: should the function be named caseInsensitiveParameterName()? Since we return case insensitive param name. Also, how do we 'not' specify a parameter here?

efeg: @alexwang08 
* I believe we return the case sensitive name here -- do you think there is a bug in the logic?
* I am not sure `how do we 'not' specify a parameter here` means -- do you mean what happens if the parameter name does not exist in the request? If so, it is documented in the JavaDoc

alexwang08:
LGTM

Issue #568, opened 2019-02-21
Add common abstraction for servlet requests.


efeg:
Improve request handling logic on the servlet-side to provide clean abstraction for sync and async requests, cleanup code duplication, and enable development of https://github.com/linkedin/cruise-control/issues/567.

PR #569, opened 2019-02-21
Add common abstraction for servlet requests.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/568.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
@@ -11,6 +11,7 @@ import com.linkedin.kafka.cruisecontrol.servlet.parameters.AddedOrRemovedBrokerP
 import com.linkedin.kafka.cruisecontrol.servlet.parameters.AdminParameters;
 import com.linkedin.kafka.cruisecontrol.servlet.parameters.BootstrapParameters;
 import com.linkedin.kafka.cruisecontrol.servlet.parameters.ClusterLoadParameters;
+import com.linkedin.kafka.cruisecontrol.servlet.parameters.CruiseControlParameters;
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -163,26 +155,23 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
...
-            getUserTaskState(request, response);
-            _successfulRequestExecutionTimer.get(endPoint).update(System.nanoTime() - requestExecutionStartTime, TimeUnit.NANOSECONDS);
+            syncRequest(() -> new UserTasksParameters(request), this::userTaskState,
+                        response, _successfulRequestExecutionTimer.get(endPoint));
             break;
           default:
             throw new UserRequestException("Invalid URL for GET");
         }
       }
     } catch (UserRequestException ure) {
```
kidkun: I am wondering whether this catch is necessary. Most of the URE will be caught and handled [here](https://github.com/linkedin/cruise-control/blob/f7e15c4aa6c784129dfd196355c8454e89ede4f3/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/AbstractParameters.java#L45)
the only one left the above `default` branch in `switch` statement(`"Invalid URL for GET/POST"`) and the following general Exception handler can handle.

efeg: The referred handler (`AbstractParameters#parseParameters(HttpServletResponse)`) is implemented in an abstract class; hence, if another implementation of the same method throws `UserRequestException`, we can miss it. Note that it is an unchecked exception and does not necessarily have to be in the method signature.

While general Exception handler can handle every other exception, in this case we prefer to be specific because the response header can return `SC_BAD_REQUEST` or `SC_INTERNAL_SERVER_ERROR` depending on the type of exception.

Issue #581, opened 2019-03-06
Executor should be able to notify when execution status changes


alexwang08:
When an execution is finished or stopped, executor should be able to notify the user so that the user will not wait unnecessarily.

PR #584, opened 2019-03-07
Add notification interface for execution completion status

alexwang08:
Addresses issue #581. This patch provides an interface for sending notification when Executor's execution has stopped (due to successful completion, user requested stop or Cruise Control requested stop). This patch does not provide an implementation (a Noop notifier is provided as stand in). Different orgs can customize their implementation to use email or messaging to deliver the notification.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -126,6 +135,9 @@ public Executor(KafkaCruiseControlConfig config,
            long removalHistoryRetentionTimeMs) {
     _numExecutionStopped = new AtomicInteger(0);
     _numExecutionStoppedByUser = new AtomicInteger(0);
+    _executionStoppedByUser = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
alexwang08: `_numExecutionStopped` and `_numExecutionStoppedByUser` are not tied to a single execution, and they are used in for gauge sensors in `registerGaugeSensors()`, so maybe not put them as local variable in `ProposalExecutionRunnable`. `_executionStoppedByUser` is a boolean flag and is toggled per execution job to indicate whether the `ProposalExecutionRunnable` is stopped, so I think it could be an instance variable of `Executor`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -848,4 +884,99 @@ private void maybeReexecuteTasks() {
...
+      _endMs = _time.milliseconds();
+      _userTaskInfo = null;
+
+      _startedBy = "UNKNOWN";
+      _operation = "UNKNOWN";
+      _startedByUser = false;
+      if (_userTaskManager != null && _uuid != null) {
+        _userTaskInfo = _userTaskManager.getUserTaskById(_uuid);
+        _actionUuid = _uuid;
+        if (_userTaskInfo != null) {
```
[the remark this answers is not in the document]
alexwang08: You are right about `_userTaskInfo != null`. I think it makes sense to store `UserTaskManager` in executor, the reason being since `Executor` is executing some proposal based on UserTasks, it makes sense for it to be able to lookup the context of the request (contained in UserTaskManager) based on UUID. Passing some additional information also involves expanding existing signature of methods to include parameters,  like `Endpoint`, just to pass the information to `Executor`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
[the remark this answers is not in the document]
kidkun: Sorry I was not clear in original comment. Actually I mean `_executionStoppedByUser` , `_executionException` and `_executionStartMs`, from the code they are only get read when we are sending the notification at the end of execution and they are actually one-time thing(i,e. once execution is done and we send the notification, we do not need these information any more). Therefore I feel it is more appropriate thing to make them [local variable of ProposalExecutionRunnable](https://github.com/linkedin/cruise-control/blob/2b98cdbe4380159b912dcbcc5a4e3082f52c1843/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L433)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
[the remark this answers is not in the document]
kidkun: yes, so we need to first work on [598](https://github.com/linkedin/cruise-control/issues/598)... otherwise we are introducing bug here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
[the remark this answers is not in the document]
alexwang08: Yeah it makes sense to put `_executionException` and `_executionStartMs` in the Runnable. For `_executionStoppedByUser`, since it's used in tandem with `_stopRequested`, I think we should keep it as a field of `Executor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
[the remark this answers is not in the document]
alexwang08: Stored the UserTaskInfo at the start of proposal execution, so the issue is fixed.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -41,6 +43,7 @@
 import static com.linkedin.kafka.cruisecontrol.executor.ExecutionTask.State.ABORTING;
 import static com.linkedin.kafka.cruisecontrol.executor.ExecutionTask.State.DEAD;
 import static com.linkedin.kafka.cruisecontrol.executor.ExecutionTask.State.IN_PROGRESS;
+import static com.linkedin.kafka.cruisecontrol.executor.ExecutorState.State.*;
```
kidkun: Can we avoid wildcard import here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -283,6 +292,7 @@ private synchronized void initProposalExecution(Collection<ExecutionProposal> pr
     setRequestedPartitionMovementConcurrency(requestedPartitionMovementConcurrency);
     setRequestedLeadershipMovementConcurrency(requestedLeadershipMovementConcurrency);
     _uuid = uuid;
+    _executionStoppedByUser.set(false);
```
kidkun: looks redundant here since it is set in `startExecution()`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -434,10 +455,16 @@ public boolean hasOngoingExecution() {
...
+    private final long _executionStartMs;
+    private Throwable _executionException;
+    private UserTaskManager.UserTaskInfo _userTask;
 
     ProposalExecutionRunnable(LoadMonitor loadMonitor, Collection<Integer> demotedBrokers, Collection<Integer> removedBrokers) {
       _loadMonitor = loadMonitor;
       _state = ExecutorState.State.NO_TASK_IN_PROGRESS;
+      _executionStartMs = _time.milliseconds();
+      _executionException = null;
+      _userTask = _userTaskManager == null ? null : _userTaskManager.getUserTaskById(_uuid);
```
kidkun: looks like this does not solve the issue since you are just get the reference not clone the `UserTaskInfo`, in this way we still need fix patch for 598.

Also should we panic(throw exception) here if the _userTaskManager is null(which means we forget to pass in reference to Executor).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -848,4 +883,117 @@ private void maybeReexecuteTasks() {
...
     }
   }
+
+  /**
+   * We finish execution after Step. 2: moving leadership. Checking pending leadership or partition movement ascertains
+   * that execution is really done (instead of quitting due to Exception).
+   * @return true if execution for proposal is done.
+   */
+  private boolean executionSucceeded() {
+    return _executorState.state() == LEADER_MOVEMENT_TASK_IN_PROGRESS &&
```
kidkun: should here be `NO_TASK_IN_PROGRESS`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
alexwang08: Maybe it's fine in this case because we are importing from `ExecutorState.State`, since `.State.*` only has the enums. This avoids having to use `ExecutorState.State.SOME_STATE` everywhere

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
alexwang08: The reference is not garbage collected since we still have a reference here even if we remove it in `UserTaskManager`. 
Panic'ing may be too harsh, since `ProposalExecutionRunnable` is also used in unit tests, if `UserTaskManager` isn't set, we just won't have complete info.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
alexwang08: The state is not set to `NO_TASK_IN_PROGRESS` until we call `clearCompletedExecution()`, `clearCompletedExecution()` also clears `_uuid` which we need. I think the conditions in `executionSucceeded()` is also valid.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
kidkun: yes, you are right. Another question is how do you guarantee visibility across the thread, I think if you want to make sure `_userTask` is in-sync, you need `volatile` or `AtomicReference`.

I still feel panic is the right thing to do, in unit test you can pass in a mock object. I think in your current implementation, `UserTaskManager` is a necessary thing since you try to get `UserTaskInfo` from it and whole purpose of this patch is return these information to user.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
kidkun: I see, then what about `REPLICA_MOVEMENT_TASK_IN_PROGRESS`? If there is replica movement task, shouldn't the state be parked at `REPLICA_MOVEMENT_TASK_IN_PROGRESS`?

kidkun: I see, then what about REPLICA_MOVEMENT_TASK_IN_PROGRESS? If there is replica movement task, shouldn't the state be parked at REPLICA_MOVEMENT_TASK_IN_PROGRESS?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
alexwang08: I'll make it AtomicReference.
When you say panic, do you mean throwing an exception? I think even if we don't have user task information, it's still better to send the email out instead of panic out, since if we send nothing, the user will not know execution has finished, resulting in unnecessary waiting, whereas if we send out the email with UUID, the user can lookup the info, either via /user_tasks (within retention), or logs.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
alexwang08: the executor goes from `Starting_execution` to `REPLICA_MOVEMENT_TASK_IN_PROGRESS` to  `LEADER_MOVEMENT_TASK_IN_PROGRESS` to finishing, in this sequence, so we always go through `LEADER_MOVEMENT_TASK_IN_PROGRESS`.

kidkun: looks like it could park at `STOPPING_EXECUTION` state if it is stopped by user or CC?(see [code](https://github.com/linkedin/cruise-control/blob/2b98cdbe4380159b912dcbcc5a4e3082f52c1843/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L557))

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
alexwang08: actually the set to false is in
```
    if (!ExecutorUtils.partitionsBeingReassigned(_zkUtils).isEmpty()) {
      _executionTaskManager.clear();
      _uuid = null;
      _executionStoppedByUser.set(false);
      // Note that in case there is an ongoing partition reassignment, we do not unpause metric sampling.
      throw new IllegalStateException("There are ongoing partition reassignments.");
    }
```
i.e. error cleanup. So we still need to set it.

kidkun: I feel a more clear way could be move this ` _executionStoppedByUser.set(false)` out of the `if` statement and remove the one in `initProposalExecution()`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: I am very confused.

@alexwang08 When would you expect the `_userTaskManager` to be `null`? If it is just for unit tests, why don't we use mocks in the unit tests? If it indicates that your code logic is wrong, it should fail -- would you clarify under which scenario this is true: `if we don't have user task information, it's still better to send the email out instead of panic out`

@kidkun has this issue been resolved? If not, why did we mark the patch as approved?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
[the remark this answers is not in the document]
efeg: I am really confused -- @alexwang08 @kidkun Am I looking at an outdated patch? Let's address these issues before we approve this patch.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorNotifier.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * For sending notification about executor completion status.
+ */
+@InterfaceStability.Evolving
+public interface ExecutorNotifier extends CruiseControlConfigurable {
+  /**
+   * When an execution completes successfully or is stopped this method should be called
+   * @param payload Information to be sent.
+   */
+  void sendNotification(Executor.ExecutionNotifierPayload payload);
```
efeg: Would it make sense for the `sendNotification` to send `notification` rather than `payload`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -327,6 +327,30 @@ private synchronized void removeOldUserTasks() {
     }
   }
 
+  /**
+   * Allow other subsystems to retrieve {@link UserTaskInfo} using uuid. This method assumes uuid comes from some internal book
+   * keeping state, not an external http request. Thus it does not check request parameters like
+   * {@link UserTaskManager#getUserTaskByUserTaskId} below does.
```
efeg: What does `{@link UserTaskManager#getUserTaskByUserTaskId} below does.` mean? Do we assume the placement of where each function will reside in the code will not change?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -327,6 +327,30 @@ private synchronized void removeOldUserTasks() {
...
+      return null;
+    }
+
+    UUID userTaskId = UUID.fromString(uuid);
+    if (_allUuidToUserTaskInfoMap.get(TaskState.COMPLETED).containsKey(userTaskId)) {
+      return _allUuidToUserTaskInfoMap.get(TaskState.COMPLETED).get(userTaskId);
+    }
+
+    if (_allUuidToUserTaskInfoMap.get(TaskState.ACTIVE).containsKey(userTaskId)) {
+      return _allUuidToUserTaskInfoMap.get(TaskState.ACTIVE).get(userTaskId);
```
efeg: What does `return _allUuidToUserTaskInfoMap.get(TaskState.ACTIVE).get(userTaskId);` return if `userTaskId` does not exist in `_allUuidToUserTaskInfoMap.get(TaskState.ACTIVE)` map? 

Would it possible to improve our implementation using this information?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
kidkun: @efeg  We have a discussion offline, the conclusion is this is expected.  if execution is parked at `STOPPING_EXECUTION`, the execution is not considered successful.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
kidkun: @efeg had a discussion offline, Alex's argument is that in that case we send out the email without information in `UserTaskInfo`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
kidkun: @efeg yes only `LEADER_MOVEMENT_TASK_IN_PROGRESS` is used, Alex's point is that this is an enum, there is no big overhead by using wildcard. But yes I feel it is more elegant not using wildcard.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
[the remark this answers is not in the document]
kidkun: @efeg Alex address the issue by `Stored the UserTaskInfo at the start of proposal execution` and that operation is synchronized, he can get the information he needs.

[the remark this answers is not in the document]
alexwang08: @efeg I made updates after Kun's comments, hence this snippet is outdated

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
alexwang08: @efeg `_userTaskManager` is set in `KafkaCruiseControlServlet's` constructor. My code is right. `_userTaskManager` helps to tell if the execution is started by user, what operation it is, and request parameters etc. These information does not affect correctness, so I didn't set `_userTaskManager` in the test case.
In non-testcase code path, `_userTaskManager` will be set. Hypothetically, even if we don't have `UserTaskInfo`, sending out a notification will at least alert that `Executor` state has changed. The notification will contain UUID of the UserTask, which can be used to look up request (within retention). The notification will also tell when and by what is the execution stopped. These provides more information than panic'ing as suggested, which is why we decided to keep it as is.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
alexwang08: @efeg We discussed Kun's case. Basically, execution state will always go through `REPLICA_MOVEMENT_TASK_IN_PROGRESS` and `LEADER_MOVEMENT_TASK_IN_PROGRESS`, at which point, if the execution finished, the remaining leadership and replica movement is 0. If the task reaches this point, we consider it succeeded. All other paths are considered failed or stopped. Could you provide an example where the condition gives false positive?
I can move it as you suggested, although `_executorNotifier.sendNotification` is a separate action from `clearCompletedExecution` and it feels cleaner to keep them separate.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: I'll remove `below`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: Do you mean before when `Executor` tries to get `_userTaskInfo` but the `_userTaskInfo` has not been inserted into `_allUuidToUserTaskInfoMap.get(TaskState.ACTIVE)`? Or do you mean `userTaskId` isn't found in  `_allUuidToUserTaskInfoMap`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -848,4 +882,117 @@ private void maybeReexecuteTasks() {
...
+    private String _startedBy;      // Indicate origin of the execution, could be User or Cruise Control
+    private String _endedBy;        // An executions can be ended by Cruise Control (e.g. due to Exception) or by User
+    private long _startMs;
+    private long _endMs;
+    private UserTaskManager.UserTaskInfo _userTaskInfo;
+    private boolean _executionSucceeded;
+    private String _operation;      // Maybe UserTask Endpoint or Self-healing action
+    private boolean _startedByUser;
+    private Throwable _exception = null;  // The Exception that ended the execution
+    private String _actionUuid = null;    // UUID associated with execution, could be for UserTask for Self-healing
```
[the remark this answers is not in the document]
alexwang08: a lot of the fields (e.g. `_operation`) are initialized to an value and potentially changed after so can't use `final`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorNotification.java
```diff
@@ -0,0 +1,116 @@
...
+public class ExecutorNotification {
+  private ActionAgent _startedBy;       // Indicate origin of the execution, could be User or Cruise Control
+  private ActionAgent _endedBy;         // An executions can be ended by Cruise Control (due to Exception or completion) or by User
+  private long _startMs;
+  private long _endMs;
+  private UserTaskManager.UserTaskInfo _userTaskInfo;
+  private boolean _executionSucceeded;
+  private String _operation;            // Can be UserTask Endpoint or Self-healing action
+  private Throwable _exception = null;  // The Exception that ended the execution
+  private String _actionUuid = null;    // UUID associated with execution, could be for UserTask for Self-healing
```
kidkun: I see Efe has a previous comment to make the fields in Class final, I think it is a good idea, we only needs to make sure they are assigned once in constructor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -156,6 +164,11 @@ public Executor(KafkaCruiseControlConfig config,
     _stopRequested = new AtomicBoolean(false);
     _hasOngoingExecution = false;
     _uuid = null;
+    _executorNotifier = executorNotifier != null
+                        ? executorNotifier
+                        : config.getConfiguredInstance(KafkaCruiseControlConfig.EXECUTOR_NOTIFIER_CLASS_CONFIG,
+                                                                         ExecutorNotifier.class);
```
kidkun: nit: bad indention.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -434,10 +458,19 @@ public boolean hasOngoingExecution() {
     private ExecutorState.State _state;
     private Set<Integer> _recentlyDemotedBrokers;
     private Set<Integer> _recentlyRemovedBrokers;
+    private final long _executionStartMs;
+    private Throwable _executionException;
+    private UserTaskManager.UserTaskInfo _userTaskInfo;
```
kidkun: can we make `_userTaskInfo` final?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorNotifier.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2019 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.cruisecontrol.executor;
+
+import com.linkedin.cruisecontrol.common.CruiseControlConfigurable;
+import org.apache.kafka.common.annotation.InterfaceStability;
+
+
```
kidkun: nit: redundant empty line.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -848,4 +889,15 @@ private void maybeReexecuteTasks() {
...
     }
   }
+
+  /**
+   * We finish execution after Step. 2: moving leadership. Checking pending leadership or partition movement ascertains
+   * that execution is really done (instead of quitting due to Exception).
+   * @return true if execution for proposal is done.
+   */
+  private boolean executionSucceeded() {
+    return _executorState.state() == LEADER_MOVEMENT_TASK_IN_PROGRESS &&
```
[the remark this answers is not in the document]
alexwang08: If stop is requested, we won't have any exceptions, but may still have pending replica or leadership movement. Will there be a case where executionException != null there's still pending replica or leadership movement?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorNotification.java
alexwang08: okay I think exception for `_operation` field they can all be final.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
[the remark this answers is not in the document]
kidkun: But in that case I think we will not part at `LEADER_MOVEMENT_TASK_IN_PROGRESS` state but `STOPPING_EXECUTION` state.

PR #591, opened 2019-03-13
Clarify the user task status information that UserTaskManager returns to client.

kidkun:


alexwang08:
Should we simply have add an entry in `_allUuidToUserTaskInfoMap.put(TaskState.COMPLETED, completedUserTaskIdToFuturesMap);` too?

kidkun:
@alexwang08  I have thought about that option,  the reason I implemented this way in the first place is to let  task completed either with or without error to share the same map and the eviction rule configed by `completedUserTaskRetentionTimeMs` and `maxCachedCompletedUserTasks` to treat them as a whole.

Issue #595, opened 2019-03-16
Log user request and Cruise Control action


alexwang08:
Currently old user requests are stored in-memory for 24 hours up to 50 entries. After the limit is past, old user requests are lost. It would be good to keep a record of old requests, as well as results of the requests. Also, Cruise Control self healing may start rebalance/remove broker operation as well. It would be good to log those actions. Finally, logging executor action associated with Cruise Control actions (given UUID) will help close the loop about the Cruise Control action results. Together, these logs provide complete view of incoming UserTask/Anomaly -> Cruise Control Action -> Action Result, even after expiry time or Cruise Control instance restart.
We can add logs in 3 places: UserTaskManager, Executor, and AnomalyDetector.

PR #596, opened 2019-03-16
Added log for old user tasks, executor action and self-healing action

alexwang08:
Addresses issue #595.
The new log will be called kafkacruisecontrol-result.log. It's rolled daily.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -41,7 +41,7 @@ public class KafkaCruiseControlUtils {
   private static final Set<String> KAFKA_ASSIGNER_GOALS =
       Collections.unmodifiableSet(new HashSet<>(Arrays.asList(KafkaAssignerEvenRackAwareGoal.class.getSimpleName(),
                                                               KafkaAssignerDiskUsageDistributionGoal.class.getSimpleName())));
-  public static final String OPERATION_LOGGER = "log4j.logger.com.linkedin.kafka.cruisecontrol.CruiseControlOperationLog";
+  public static final String OPERATION_LOGGER = "operationLogger";
 
   private KafkaCruiseControlUtils() {
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -41,6 +41,7 @@ public class KafkaCruiseControlUtils {
   private static final Set<String> KAFKA_ASSIGNER_GOALS =
       Collections.unmodifiableSet(new HashSet<>(Arrays.asList(KafkaAssignerEvenRackAwareGoal.class.getSimpleName(),
                                                               KafkaAssignerDiskUsageDistributionGoal.class.getSimpleName())));
+  public static final String OPERATION_LOGGER = "log4j.logger.com.linkedin.kafka.cruisecontrol.CruiseControlOperationLog";
 
   private KafkaCruiseControlUtils() {
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
@@ -67,7 +67,6 @@ public class AnomalyDetector {
   private final AnomalyDetectorState _anomalyDetectorState;
   // TODO: Make this configurable.
   private final List<String> _selfHealingGoals;
-  // TODO: Upgrade log4j to log4j2 to use async logger.
   private final ExecutorService _anomalyLoggerExecutor;
 
   public AnomalyDetector(KafkaCruiseControlConfig config,
@@ -170,6 +169,7 @@ public class AnomalyDetector {
       LOG.warn("Interrupted while waiting for anomaly detector to shutdown.");
     }
     _brokerFailureDetector.shutdown();
+    _anomalyLoggerExecutor.shutdownNow();
     LOG.info("Anomaly detector shutdown completed.");
   }
 
@@ -313,12 +313,18 @@ public class AnomalyDetector {
 
     private void logSelfHealingOperation(Anomaly anomaly, OptimizationFailureException ofe, boolean selfHealingStarted) {
       if (selfHealingStarted) {
-        if (anomaly instanceof KafkaAnomaly) {
-          OPERATION_LOG.info("[{}] calculation finishes, result:\n{}", anomaly.anomalyId(),
-                             ((KafkaAnomaly) anomaly).optimizationResult(false));
-        } else if (anomaly instanceof KafkaMetricAnomaly) {
-          OPERATION_LOG.info("[{}] calculation finishes, result:\n{}", anomaly.anomalyId(),
-                             ((KafkaMetricAnomaly) anomaly).optimizationResult(false));
+        switch (getAnomalyType(anomaly)) {
+          case GOAL_VIOLATION:
+          case BROKER_FAILURE:
+            OPERATION_LOG.info("[{}] calculation finishes, result:\n{}", anomaly.anomalyId(),
+                               ((KafkaAnomaly) anomaly).optimizationResult(false));
+            break;
+          case METRIC_ANOMALY:
+            OPERATION_LOG.info("[{}] calculation finishes, result:\n{}", anomaly.anomalyId(),
+                               ((KafkaMetricAnomaly) anomaly).optimizationResult(false));
+            break;
+          default:
+            throw new IllegalStateException("Unrecognized anomaly type.");
         }
       } else {
         OPERATION_LOG.info("[{}] calculation fails, exception:\n{}", anomaly.anomalyId(), ofe);
@@ -335,6 +341,7 @@ public class AnomalyDetector {
           _anomalyLoggerExecutor.submit(() -> logSelfHealingOperation(anomaly, null, true));
         } catch (OptimizationFailureException ofe) {
           _anomalyLoggerExecutor.submit(() -> logSelfHealingOperation(anomaly, ofe, false));
+          throw ofe;
         } finally {
           _anomalyDetectorState.onAnomalyHandle(anomaly, startedSuccessfully ? AnomalyState.Status.FIX_STARTED
                                                                              : AnomalyState.Status.FIX_FAILED_TO_START);
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
@@ -15,6 +15,7 @@ import com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyNotificationRes
 import com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyNotifier;
 import com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyType;
 import com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException;
+import com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException;
 import com.linkedin.kafka.cruisecontrol.executor.ExecutorState;
 import com.linkedin.kafka.cruisecontrol.monitor.LoadMonitor;
 import com.linkedin.kafka.cruisecontrol.monitor.task.LoadMonitorTaskRunner;
@@ -32,6 +33,7 @@ import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils.OPERATION_LOGGER;
 import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.getAnomalyType;
 import static com.linkedin.kafka.cruisecontrol.servlet.response.CruiseControlState.SubState.EXECUTOR;
 
@@ -43,7 +45,7 @@ public class AnomalyDetector {
   private static final String METRIC_REGISTRY_NAME = "AnomalyDetector";
   private static final int INIT_JITTER_BOUND = 10000;
   private static final Logger LOG = LoggerFactory.getLogger(AnomalyDetector.class);
-  private static final Logger RESULT_LOG = LoggerFactory.getLogger("com.linkedin.kafka.cruisecontrol.CruiseControlResultLog");
+  private static final Logger OPERATION_LOG = LoggerFactory.getLogger(OPERATION_LOGGER);
   private static final Anomaly SHUTDOWN_ANOMALY = new BrokerFailures(null,
                                                                      Collections.emptyMap(),
                                                                      true,
@@ -65,7 +67,9 @@ public class AnomalyDetector {
   private final AnomalyDetectorState _anomalyDetectorState;
   // TODO: Make this configurable.
   private final List<String> _selfHealingGoals;
-  private final ExecutorService _loggerExecutor;    // TODO upgrade log4j to log4j2 to use async logger
+  // TODO: Upgrade log4j to log4j2 to use async logger.
+  private final ExecutorService _anomalyLoggerExecutor;
+
   public AnomalyDetector(KafkaCruiseControlConfig config,
                          LoadMonitor loadMonitor,
                          KafkaCruiseControl kafkaCruiseControl,
@@ -95,8 +99,8 @@ public class AnomalyDetector {
     // Add anomaly detector state
     int numCachedRecentAnomalyStates = config.getInt(KafkaCruiseControlConfig.NUM_CACHED_RECENT_ANOMALY_STATES_CONFIG);
     _anomalyDetectorState = new AnomalyDetectorState(_anomalyNotifier.selfHealingEnabled(), numCachedRecentAnomalyStates);
-    _loggerExecutor = Executors.newFixedThreadPool(1, new KafkaCruiseControlThreadFactory("AnomalyDetectorLoggerExecutor",
-                                                                                          true, null));
+    _anomalyLoggerExecutor =
+        Executors.newSingleThreadScheduledExecutor(new KafkaCruiseControlThreadFactory("AnomalyLogger", true, null));
   }
 
   /**
@@ -126,8 +130,8 @@ public class AnomalyDetector {
     // Add anomaly detector state
     _anomalyDetectorState = new AnomalyDetectorState(new HashMap<>(AnomalyType.cachedValues().size()), 10);
     _selfHealingGoals = Collections.emptyList();
-    _loggerExecutor = Executors.newFixedThreadPool(1, new KafkaCruiseControlThreadFactory("AnomalyDetectorLoggerExecutor",
-                                                                                          true, null));
+    _anomalyLoggerExecutor =
+        Executors.newSingleThreadScheduledExecutor(new KafkaCruiseControlThreadFactory("AnomalyLogger", true, null));
   }
 
   public void startDetection() {
@@ -307,15 +311,17 @@ public class AnomalyDetector {
       return false;
     }
 
-    private void persistSelfHealingInfo(Anomaly anomaly, boolean selfHealingStarted) {
-      RESULT_LOG.info("Self-healing {} for anomaly: {}.", selfHealingStarted ? "started successfully" : "failed to start", anomaly);
+    private void logSelfHealingOperation(Anomaly anomaly, OptimizationFailureException ofe, boolean selfHealingStarted) {
       if (selfHealingStarted) {
-        // Either KafkaAnomaly or KafkaMetricAnomaly may have optimization result
         if (anomaly instanceof KafkaAnomaly) {
-          RESULT_LOG.info("Optimization Result:%n{}", ((KafkaAnomaly) anomaly).optimizationResult(false));
+          OPERATION_LOG.info("[{}] calculation finishes, result:\n{}", anomaly.anomalyId(),
+                             ((KafkaAnomaly) anomaly).optimizationResult(false));
         } else if (anomaly instanceof KafkaMetricAnomaly) {
-          RESULT_LOG.info("Optimization Result:%n{}", ((KafkaMetricAnomaly) anomaly).optimizationResult(false));
+          OPERATION_LOG.info("[{}] calculation finishes, result:\n{}", anomaly.anomalyId(),
+                             ((KafkaMetricAnomaly) anomaly).optimizationResult(false));
         }
+      } else {
+        OPERATION_LOG.info("[{}] calculation fails, exception:\n{}", anomaly.anomalyId(), ofe);
       }
     }
 
@@ -326,12 +332,13 @@ public class AnomalyDetector {
         boolean startedSuccessfully = false;
         try {
           startedSuccessfully = anomaly.fix();
+          _anomalyLoggerExecutor.submit(() -> logSelfHealingOperation(anomaly, null, true));
+        } catch (OptimizationFailureException ofe) {
+          _anomalyLoggerExecutor.submit(() -> logSelfHealingOperation(anomaly, ofe, false));
         } finally {
           _anomalyDetectorState.onAnomalyHandle(anomaly, startedSuccessfully ? AnomalyState.Status.FIX_STARTED
                                                                              : AnomalyState.Status.FIX_FAILED_TO_START);
           LOG.info("Self-healing {}.", startedSuccessfully ? "started successfully" : "failed to start");
-          boolean isSuccessful = startedSuccessfully;
-          _loggerExecutor.submit(() -> persistSelfHealingInfo(anomaly, isSuccessful));
         }
       }
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
@@ -23,6 +23,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.ScheduledExecutorService;
@@ -42,6 +43,7 @@ public class AnomalyDetector {
   private static final String METRIC_REGISTRY_NAME = "AnomalyDetector";
   private static final int INIT_JITTER_BOUND = 10000;
   private static final Logger LOG = LoggerFactory.getLogger(AnomalyDetector.class);
+  private static final Logger RESULT_LOG = LoggerFactory.getLogger("com.linkedin.kafka.cruisecontrol.CruiseControlResultLog");
   private static final Anomaly SHUTDOWN_ANOMALY = new BrokerFailures(null,
                                                                      Collections.emptyMap(),
                                                                      true,
@@ -63,7 +65,7 @@ public class AnomalyDetector {
   private final AnomalyDetectorState _anomalyDetectorState;
   // TODO: Make this configurable.
   private final List<String> _selfHealingGoals;
-
+  private final ExecutorService _loggerExecutor;    // TODO upgrade log4j to log4j2 to use async logger
   public AnomalyDetector(KafkaCruiseControlConfig config,
                          LoadMonitor loadMonitor,
                          KafkaCruiseControl kafkaCruiseControl,
@@ -93,6 +95,8 @@ public class AnomalyDetector {
     // Add anomaly detector state
     int numCachedRecentAnomalyStates = config.getInt(KafkaCruiseControlConfig.NUM_CACHED_RECENT_ANOMALY_STATES_CONFIG);
     _anomalyDetectorState = new AnomalyDetectorState(_anomalyNotifier.selfHealingEnabled(), numCachedRecentAnomalyStates);
+    _loggerExecutor = Executors.newFixedThreadPool(1, new KafkaCruiseControlThreadFactory("AnomalyDetectorLoggerExecutor",
+                                                                                          true, null));
   }
 
   /**
@@ -122,6 +126,8 @@ public class AnomalyDetector {
     // Add anomaly detector state
     _anomalyDetectorState = new AnomalyDetectorState(new HashMap<>(AnomalyType.cachedValues().size()), 10);
     _selfHealingGoals = Collections.emptyList();
+    _loggerExecutor = Executors.newFixedThreadPool(1, new KafkaCruiseControlThreadFactory("AnomalyDetectorLoggerExecutor",
+                                                                                          true, null));
   }
 
   public void startDetection() {
@@ -301,6 +307,18 @@ public class AnomalyDetector {
       return false;
     }
 
+    private void persistSelfHealingInfo(Anomaly anomaly, boolean selfHealingStarted) {
+      RESULT_LOG.info("Self-healing {} for anomaly: {}.", selfHealingStarted ? "started successfully" : "failed to start", anomaly);
+      if (selfHealingStarted) {
+        // Either KafkaAnomaly or KafkaMetricAnomaly may have optimization result
+        if (anomaly instanceof KafkaAnomaly) {
+          RESULT_LOG.info("Optimization Result:%n{}", ((KafkaAnomaly) anomaly).optimizationResult(false));
+        } else if (anomaly instanceof KafkaMetricAnomaly) {
+          RESULT_LOG.info("Optimization Result:%n{}", ((KafkaMetricAnomaly) anomaly).optimizationResult(false));
+        }
+      }
+    }
+
     private void fixAnomaly(Anomaly anomaly) throws Exception {
       boolean isReadyToFix = isReadyToFix(anomaly);
       if (isReadyToFix) {
@@ -312,6 +330,8 @@ public class AnomalyDetector {
           _anomalyDetectorState.onAnomalyHandle(anomaly, startedSuccessfully ? AnomalyState.Status.FIX_STARTED
                                                                              : AnomalyState.Status.FIX_FAILED_TO_START);
           LOG.info("Self-healing {}.", startedSuccessfully ? "started successfully" : "failed to start");
+          boolean isSuccessful = startedSuccessfully;
+          _loggerExecutor.submit(() -> persistSelfHealingInfo(anomaly, isSuccessful));
         }
       }
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -55,6 +55,7 @@ import static com.linkedin.kafka.cruisecontrol.executor.ExecutionTaskTracker.Exe
  */
 public class Executor {
   private static final Logger LOG = LoggerFactory.getLogger(Executor.class);
+  private static final Logger RESULT_LOG = LoggerFactory.getLogger("com.linkedin.kafka.cruisecontrol.CruiseControlResultLog");
   private static final long EXECUTION_HISTORY_SCANNER_PERIOD_SECONDS = 5;
   private static final long EXECUTION_HISTORY_SCANNER_INITIAL_DELAY_SECONDS = 0;
   // The maximum time to wait for a leader movement to finish. A leader movement will be marked as failed if
@@ -521,6 +522,7 @@ public class Executor {
     private void execute() {
       _state = STARTING_EXECUTION;
       _executorState = ExecutorState.executionStarted(_uuid, _recentlyDemotedBrokers, _recentlyRemovedBrokers);
+      RESULT_LOG.info("Proposal execution:[{}] started.", _uuid);
       try {
         // Pause the metric sampling to avoid the loss of accuracy during execution.
         while (true) {
@@ -578,6 +580,7 @@ public class Executor {
                                                                      _executionStoppedByUser.get(),
                                                                      _executionException, executionSucceeded);
         _executorNotifier.sendNotification(notification);
+        RESULT_LOG.info("Proposal execution with UUID:[{}] ended.", _uuid);
         // Clear completed execution.
         clearCompletedExecution();
       }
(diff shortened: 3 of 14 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -34,6 +37,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils.*;
```
kidkun: Can we avoid using wildcard here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -565,6 +568,43 @@ boolean isUserTaskDoneExceptionally() {
...
+  // Persist user request and result to log. Called when the user task completes and is moved into COMPLETED status
+  private void persistUserTask(UserTaskInfo userTaskInfo) {
+    String userTaskId = userTaskInfo.userTaskId().toString();
+    String endPoint = userTaskInfo.endPoint().toString();
+    String requestUrl = userTaskInfo.requestWithParams();
+    String startDate = KafkaCruiseControlUtils.toDateString(userTaskInfo.startMs(), DATE_FORMAT, TIME_ZONE);
+    List<OperationFuture> futures = userTaskInfo.futures();
+
+    StringBuilder message = new StringBuilder();
+    message.append(String.format("UserTask [%s] for endpoint [%s] with requestUrl [%s] started-at [%s].",
```
kidkun: looks like `requestUrl` already have `endpoint` information.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -565,6 +568,43 @@ boolean isUserTaskDoneExceptionally() {
...
+        if (future.isDone()) {
+          if (userTaskInfo.parameters() != null) {
+            AbstractCruiseControlResponse response = ((AbstractCruiseControlResponse) future.get());
+            String cachedResponse = response.cachedResponse(userTaskInfo.parameters());
+            message.append(String.format("Result for step %d:%n", step++))
+                   .append(cachedResponse)
+                   .append("%n");
+          }
+        } else if (future.isCompletedExceptionally()) {
+          message.append(String.format("UserTask [%s] had exception.%n", userTaskId));
```
kidkun: maybe we can also persist Exception detail here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -295,6 +296,22 @@ private boolean isReadyToFix(Anomaly anomaly) {
       return false;
     }
 
+    private void mayBePersistSelfHealingInfo(Anomaly anomaly, boolean selfHealingStarted) {
+      if (selfHealingStarted) {
+        RESULT_LOG.info(String.format("Started self-healing for [%s] anomaly [%s].",
+                                     getAnomalyType(anomaly).toString(), anomaly.anomalyId()));
```
kidkun: Can we also have anomaly detail here?

alexwang08: What other details besides type of anomaly?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: yeah it's not as searchable because it will look like 'http://hostname:portname/kafkacruisecontrol/endpoint?params', I think it's easier for user to make it more explicit.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
kidkun: I mean the anomaly itself,  like [goal violation anomaly](https://github.com/linkedin/cruise-control/blob/2b98cdbe4380159b912dcbcc5a4e3082f52c1843/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolations.java#L104).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -295,6 +296,22 @@ private boolean isReadyToFix(Anomaly anomaly) {
       return false;
     }
 
+    private void mayBePersistSelfHealingInfo(Anomaly anomaly, boolean selfHealingStarted) {
```
kidkun: this method in `AnomalyDetector` only log successfully started task, but `UserTaskManager`  logs both successfully started and fail to start task(task completed exceptionally), is this intended?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -565,6 +569,48 @@ boolean isUserTaskDoneExceptionally() {
...
+        if (future.isDone() || future.isCompletedExceptionally()) {
+          try {
+            AbstractCruiseControlResponse response = ((AbstractCruiseControlResponse) future.get());
+            if (userTaskInfo.parameters() != null) {
+              String cachedResponse = response.cachedResponse(userTaskInfo.parameters());
+              message.append(String.format("Result for step %d:%n", step++))
+                     .append(cachedResponse)
+                     .append("%n");
+            } else {
+              message.append(String.format("UserTask parameter is null, cannot obtain result."));
```
kidkun: `String.format` is redundant here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
alexwang08: Will add log for failure case

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -295,6 +296,20 @@ private boolean isReadyToFix(Anomaly anomaly) {
...
+                                     selfHealingStarted ? "started" : "not started",
+                                     getAnomalyType(anomaly).toString(), anomaly.anomalyId(), anomaly.toString());
+      RESULT_LOG.info(headerMsg);
+      if (selfHealingStarted && anomaly instanceof KafkaAnomaly) {
+        StringBuilder message = new StringBuilder();
+        message.append("Optimization Result:%n")
+               .append(((KafkaAnomaly) anomaly).optimizationResult(false));  // Use plain text because it's more readable in log
+        RESULT_LOG.info(message.toString());
+      }
+      RESULT_LOG.info(String.format("End of anomaly [%s] message.", anomaly.anomalyId()));
```
efeg: Why do we use `String.format(...)` within `info`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -295,6 +296,20 @@ private boolean isReadyToFix(Anomaly anomaly) {
...
+    private void mayBePersistSelfHealingInfo(Anomaly anomaly, boolean selfHealingStarted) {
+      String headerMsg = String.format("Self-healing [%s] for [%s] anomaly [%s]. Anomaly Description: [%s].",
+                                     selfHealingStarted ? "started" : "not started",
+                                     getAnomalyType(anomaly).toString(), anomaly.anomalyId(), anomaly.toString());
+      RESULT_LOG.info(headerMsg);
+      if (selfHealingStarted && anomaly instanceof KafkaAnomaly) {
+        StringBuilder message = new StringBuilder();
+        message.append("Optimization Result:%n")
+               .append(((KafkaAnomaly) anomaly).optimizationResult(false));  // Use plain text because it's more readable in log
+        RESULT_LOG.info(message.toString());
```
efeg: Why don't we directly pass the message to `.info`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -295,6 +296,20 @@ private boolean isReadyToFix(Anomaly anomaly) {
       return false;
     }
 
+    private void mayBePersistSelfHealingInfo(Anomaly anomaly, boolean selfHealingStarted) {
+      String headerMsg = String.format("Self-healing [%s] for [%s] anomaly [%s]. Anomaly Description: [%s].",
+                                     selfHealingStarted ? "started" : "not started",
+                                     getAnomalyType(anomaly).toString(), anomaly.anomalyId(), anomaly.toString());
+      RESULT_LOG.info(headerMsg);
+      if (selfHealingStarted && anomaly instanceof KafkaAnomaly) {
```
efeg: Why don't we use the `type` of anomaly to classify it? Today, the implementation might be based on an instance of `KafkaAnomaly`, but why do we depend on the class, but not the type?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -463,6 +464,7 @@ public void run() {
     private void execute() {
       _state = ExecutorState.State.STARTING_EXECUTION;
       _executorState = ExecutorState.executionStarted(_uuid, _recentlyDemotedBrokers, _recentlyRemovedBrokers);
+      RESULT_LOG.info(String.format("Proposal execution:[%s] started.", _uuid));
```
efeg: Would it be possible to go over all `RESULT_LOG.info(...)` to ensure that we use lazy eval?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -510,6 +512,7 @@ private void execute() {
         LOG.error("Executor got exception during execution", t);
       } finally {
         _loadMonitor.resumeMetricSampling(String.format("Resumed-By-Cruise-Control-After-Completed-Execution (Date: %s)", currentUtcDate()));
+        RESULT_LOG.info(String.format("Proposal execution:[%s] ended.", _uuid));
```
efeg: The message `Proposal execution:[%s]` is not clear that it is printing UUID. Can we make it clear that this is `UUID`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -361,7 +368,8 @@ synchronized UserTaskInfo getUserTaskByUserTaskId(UUID userTaskId, HttpServletRe
    */
   private synchronized UserTaskInfo insertFuturesByUserTaskId(UUID userTaskId,
                                                               Function<String, OperationFuture> operation,
-                                                              HttpServletRequest httpServletRequest) {
+                                                              HttpServletRequest httpServletRequest,
+                                                              CruiseControlParameters parameters) {
```
efeg: Should we add a JavaDoc for this param?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -590,6 +593,48 @@ boolean isUserTaskDoneExceptionally() {
     }
   }
 
+  // Persist user request and result to log. Called when the user task completes and is moved into COMPLETED status
```
efeg: Is this a JavaDoc format we use?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -414,7 +414,7 @@ private void addOrRemoveBroker(HttpServletRequest request,
       return;
     }
 
-    CruiseControlResponse ccResponse = getAndMaybeReturnProgress(request, response, function.apply(parameters));
+    CruiseControlResponse ccResponse = getAndMaybeReturnProgress(request, response, function.apply(parameters), parameters);
```
efeg: Can we refrain from passing parameters? Can't we extract the parameters for logging using the parameterMap from the request itself?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -159,13 +163,14 @@ protected boolean removeEldestEntry(Map.Entry<UUID, UserTaskInfo> eldest) {
    *                 returns a completed Future.
    * @param step The index of the step that has to be added or fetched.
    * @param isAsyncRequest Indicate whether the task is async or sync.
+   * @param parameters Help {@link UserTaskInfo} determine the response to persist.
```
efeg: A JavaDoc should indicate what the `parameters` is -- i.e. not how it is intended to be used from outside the class. Also, do we really need to use parameters in this function, can't we use the parameter map?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -159,13 +163,14 @@ protected boolean removeEldestEntry(Map.Entry<UUID, UserTaskInfo> eldest) {
...
    * @return The list of {@link OperationFuture} for the linked UserTask.
    */
   public List<OperationFuture> getOrCreateUserTask(HttpServletRequest httpServletRequest,
                                                    HttpServletResponse httpServletResponse,
-                                                   Function<String, OperationFuture> function,
-                                                   int step,
-                                                   boolean isAsyncRequest) {
+                                                   Function<String, OperationFuture> function, int step,
```
efeg: Why do we put `step` to the same line as `function` -- is this our coding convention?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -590,6 +593,48 @@ boolean isUserTaskDoneExceptionally() {
     }
   }
 
+  // Persist user request and result to log. Called when the user task completes and is moved into COMPLETED status
+  private void persistUserTask(UserTaskInfo userTaskInfo) {
```
efeg: This function should be called by a separate thread. We should not be blocking when this expensive method is called.

efeg: Does it persist all user tasks or completed user tasks? If it is the latter, can we make the name more clear to indicate that?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -590,6 +593,48 @@ boolean isUserTaskDoneExceptionally() {
...
+    String userTaskId = userTaskInfo.userTaskId().toString();
+    String endPoint = userTaskInfo.endPoint().toString();
+    String requestUrl = userTaskInfo.requestWithParams();
+    String startDate = KafkaCruiseControlUtils.toDateString(userTaskInfo.startMs(), DATE_FORMAT, TIME_ZONE);
+    List<OperationFuture> futures = userTaskInfo.futures();
+
+    StringBuilder message = new StringBuilder();
+    message.append(String.format("UserTask [%s] for endpoint [%s] with requestUrl [%s] started-at [%s].",
+                                 userTaskId, endPoint, requestUrl, startDate))
+           .append(String.format(" UserTask has %d step(s).", futures.size()));
```
efeg: Can we move this logic to `toString()` of `UserTaskInfo`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -590,6 +593,48 @@ boolean isUserTaskDoneExceptionally() {
...
+    RESULT_LOG.info(message.toString());
+    message = new StringBuilder();
+
+    int step = 1;
+    for (OperationFuture future : futures) {
+      try {
+        if (future.isDone() || future.isCompletedExceptionally()) {
```
efeg: Would this ever be false? Do we use this method in other cases?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -590,6 +593,48 @@ boolean isUserTaskDoneExceptionally() {
...
+          try {
+            AbstractCruiseControlResponse response = ((AbstractCruiseControlResponse) future.get());
```
efeg: What if the response is from a class that does not extend `AbstractCruiseControlResponse `, but implements `CruiseControlResponse` directly?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/AbstractCruiseControlResponse.java
```diff
@@ -46,4 +46,9 @@ public void discardIrrelevantResponse(CruiseControlParameters parameters) {
       }
     }
   }
+
+  public String cachedResponse(CruiseControlParameters parameters) {
```
efeg: Does this imply that we can only support responses that extend `AbstractCruiseControlResponse`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -590,6 +593,48 @@ boolean isUserTaskDoneExceptionally() {
...
+    String endPoint = userTaskInfo.endPoint().toString();
+    String requestUrl = userTaskInfo.requestWithParams();
+    String startDate = KafkaCruiseControlUtils.toDateString(userTaskInfo.startMs(), DATE_FORMAT, TIME_ZONE);
+    List<OperationFuture> futures = userTaskInfo.futures();
+
+    StringBuilder message = new StringBuilder();
+    message.append(String.format("UserTask [%s] for endpoint [%s] with requestUrl [%s] started-at [%s].",
+                                 userTaskId, endPoint, requestUrl, startDate))
+           .append(String.format(" UserTask has %d step(s).", futures.size()));
+    RESULT_LOG.info(message.toString());
```
kidkun: We called `RESULT_LOG.info` several times in this method, can we cache everything in StringBuilder and just flush once?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -590,6 +593,48 @@ boolean isUserTaskDoneExceptionally() {
...
+    message = new StringBuilder();
+
+    int step = 1;
+    for (OperationFuture future : futures) {
+      try {
+        if (future.isDone() || future.isCompletedExceptionally()) {
+          try {
+            AbstractCruiseControlResponse response = ((AbstractCruiseControlResponse) future.get());
+            if (userTaskInfo.parameters() != null) {
```
kidkun: I think `userTaskInfo.parameters() != null` will always be false, can you elaborate a scenario that this will be true?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -295,6 +296,20 @@ private boolean isReadyToFix(Anomaly anomaly) {
       return false;
     }
 
+    private void mayBePersistSelfHealingInfo(Anomaly anomaly, boolean selfHealingStarted) {
```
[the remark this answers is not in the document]
kidkun: In this method we always log the anomaly, `mayBe` in method name might be appropriate.
Also can we reduce the times of calling  `RESULT_LOG.info()` in this method.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: Yeah in UserTaskManagerTest.java, this saves us the effort to mock up `CruiseControlParameters`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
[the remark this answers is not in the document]
alexwang08: removed the third RESULT_LOG.info() call. Do you mean it's appropriate or not appropriate?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -295,6 +296,20 @@ private boolean isReadyToFix(Anomaly anomaly) {
...
+      String headerMsg = String.format("Self-healing [%s] for [%s] anomaly [%s]. Anomaly Description: [%s].",
```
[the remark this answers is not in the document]
alexwang08: I changed `KafkaMetricAnomaly.toString()` method from 
```
  public String toString() {
    return String.format("{%nMetric Anomaly windows: %s description: %s%n}", _windows, _description);
  }
```
to 
  ```
public String toString() {
    return String.format("%s anomaly with id: %s Metric Anomaly windows: %s description: %s",
                         getAnomalyType(this), anomalyId(), _windows, _description);
  }```
Let me know if the new formatting for `Metric Anomaly windows` etc is ok.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
alexwang08: `getAnomalyType()` internally uses `instanceof` to get the type, it would be longer if we call `getAnomalyType(anomaly)` and then match to a type of anomaly. Since `optimizationResult()` method is defined for class `KafkaAnomaly` or `KafkaMetricAnomaly`, makes sense to check class and cast appropriately

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
alexwang08: The abstract method `AbstractCruiseControlResponse.discardIrrelevantAndCacheRelevant(CruiseControlParameters parameters)` where each subclass of `AbstractCruiseControlResponse` implement takes `CruiseControlParameters` type arguments. It's easier compared to changing the signature to add `isJson` and `isVerbose` for all the implementations. I think since it's the user task result that dominates memory usage, storing `CruiseControlParameters` in each `UserTaskInfo` to help with retrieval is okay here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: I can add the `AbstractCruiseControlResponse.cachedResponse()` to method to `CruiseControlResponse` so that all implementations implement the method.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/AbstractCruiseControlResponse.java
alexwang08: so far we don't have responses directly implementing `CruiseControlResponse` interface (except one used in test case), will this change?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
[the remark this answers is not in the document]
kidkun: Sorry I mean inappropriate...I feel we should remove the `maybe`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
kidkun: If `selfHealingStarted` is false here, there is no need to record optimization result. maybe the logic here can be changed to 

> 
      if (selfHealingStarted) {
        if (anomaly instanceof KafkaAnomaly) {
          RESULT_LOG.info("Optimization Result:%n{}", ((KafkaAnomaly) anomaly).optimizationResult(false));
        } else if (anomaly instanceof KafkaMetricAnomaly) {
          RESULT_LOG.info("Optimization Result:%n{}", ((KafkaMetricAnomaly) anomaly).optimizationResult(false));
        }
      }

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -305,6 +321,8 @@ private void fixAnomaly(Anomaly anomaly) throws Exception {
           _anomalyDetectorState.onAnomalyHandle(anomaly, startedSuccessfully ? AnomalyState.Status.FIX_STARTED
                                                                              : AnomalyState.Status.FIX_FAILED_TO_START);
           LOG.info("Self-healing {}.", startedSuccessfully ? "started successfully" : "failed to start");
+          boolean isSuccessful = startedSuccessfully;
```
kidkun: redundant local variable `isSuccessful`

alexwang08: needed because `_loggerExecutor.submit(() -> maybePersistSelfHealingInfo(anomaly, isSuccessful));` needed argument to be final or effectively final

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -581,13 +586,57 @@ public UserTaskInfo setState(TaskState nextState) {
...
+                   .append(cachedResponse)
+                   .append("%n");
+          } else {
+            message.append("UserTask parameter is null, cannot obtain result.");
+          }
+        } catch (Exception e) {
+          // Log Exception from completeExceptionally.
+          message.append(String.format("UserTask [%s] had exception:[%s].%n", userTaskId, e.getMessage()));
+        }
+      } catch (Throwable t) {
```
kidkun: I feel the outer try block is unnecessary.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -581,13 +586,57 @@ public UserTaskInfo setState(TaskState nextState) {
...
+                           futures().size());
+    }
+  }
+
+  // Persist user request and result to log. Called when the user task completes and is moved into COMPLETED status
+  private void persistCompletedUserTask(UserTaskInfo userTaskInfo) {
+    String userTaskId = userTaskInfo.userTaskId().toString();
+    List<OperationFuture> futures = userTaskInfo.futures();
+
+    RESULT_LOG.info(userTaskInfo.toString());
```
kidkun: can we also put this message into `StringBuilder` first?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
kidkun: `isSuccessful` is a primitive type, IIUC it is copy-by-value in function call.

alexwang08: IDE doesn't allow using `startedSuccessfully` directly.

kidkun: talked offline, it is required to pass  in an effectively final variable.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
kidkun: talked offline, this LOG will create new line and it is more friendly to search.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
alexwang08: removed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/AbstractCruiseControlResponse.java
```diff
@@ -53,7 +53,7 @@ public void discardIrrelevantResponse(CruiseControlParameters parameters) {
     }
   }
 
-  String cachedResponse() {
+  public String cachedResponse() {
```
efeg: Is it something to make the part of the public interface?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -159,13 +166,15 @@ protected boolean removeEldestEntry(Map.Entry<UUID, UserTaskInfo> eldest) {
    *                 returns a completed Future.
    * @param step The index of the step that has to be added or fetched.
    * @param isAsyncRequest Indicate whether the task is async or sync.
+   * @param parameters Parsed parameters from http request.
```
efeg: Looks like we are explicitly passing `null` in some codepath to this function. Can we explain what it means to pass `null` to this method?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -32,6 +32,7 @@
   public static final String DATE_FORMAT = "YYYY-MM-dd_HH:mm:ss z";
   public static final String DATE_FORMAT2 = "dd/MM/yyyy HH:mm:ss";
   public static final String TIME_ZONE = "UTC";
+  public static final String OPERATION_LOGGER = "log4j.logger.com.linkedin.kafka.cruisecontrol.CruiseControlOperationLog";
```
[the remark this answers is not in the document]
alexwang08: Yes it's logger name, they refer to the keys specified in `log4j.properties`, hence did not use `com.linkedin...`. A similar case is `log4j.logger.com.linkedin.kafka.cruisecontrol=INFO, kafkaCruiseControlAppender`, which is the normal cruise control logger

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -63,6 +67,8 @@
   private final AnomalyDetectorState _anomalyDetectorState;
   // TODO: Make this configurable.
   private final List<String> _selfHealingGoals;
+  // TODO: Upgrade log4j to log4j2 to use async logger.
```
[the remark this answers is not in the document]
alexwang08: We can move this to log4j.properties file.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -301,13 +311,30 @@ private boolean isReadyToFix(Anomaly anomaly) {
...
     private void fixAnomaly(Anomaly anomaly) throws Exception {
       boolean isReadyToFix = isReadyToFix(anomaly);
       if (isReadyToFix) {
         LOG.info("Fixing anomaly {}", anomaly);
         boolean startedSuccessfully = false;
         try {
           startedSuccessfully = anomaly.fix();
+          _anomalyLoggerExecutor.submit(() -> logSelfHealingOperation(anomaly, null, true));
+        } catch (OptimizationFailureException ofe) {
+          _anomalyLoggerExecutor.submit(() -> logSelfHealingOperation(anomaly, ofe, false));
```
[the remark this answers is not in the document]
alexwang08: because lambda functions expects effective final

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -301,13 +311,30 @@ private boolean isReadyToFix(Anomaly anomaly) {
...
+    }
+
     private void fixAnomaly(Anomaly anomaly) throws Exception {
       boolean isReadyToFix = isReadyToFix(anomaly);
       if (isReadyToFix) {
         LOG.info("Fixing anomaly {}", anomaly);
         boolean startedSuccessfully = false;
         try {
           startedSuccessfully = anomaly.fix();
+          _anomalyLoggerExecutor.submit(() -> logSelfHealingOperation(anomaly, null, true));
```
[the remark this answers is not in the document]
alexwang08: because lambda function expects effective final variable.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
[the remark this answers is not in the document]
kidkun: yes, we are trying to be consistent with [naming convention in log4j.properties](https://github.com/linkedin/cruise-control/blob/master/config/log4j.properties#L25), change `CruiseControlOperationLog` to `operationLogger` is a good idea.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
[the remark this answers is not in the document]
kidkun: Moved to `build.gradle`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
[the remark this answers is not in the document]
kidkun: yes, we get a syntax error saying `variable used in lambda should be effectively final`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -301,13 +311,30 @@ private boolean isReadyToFix(Anomaly anomaly) {
       return false;
     }
 
+    private void logSelfHealingOperation(Anomaly anomaly, OptimizationFailureException ofe, boolean selfHealingStarted) {
+      if (selfHealingStarted) {
+        if (anomaly instanceof KafkaAnomaly) {
+          OPERATION_LOG.info("[{}] calculation finishes, result:\n{}", anomaly.anomalyId(),
```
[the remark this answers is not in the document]
kidkun: IIUC, `%n` only works in formatting string, here we need to use `\` to escape.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/AbstractCruiseControlResponse.java
kidkun: Originally because it is used in other packages, now moved this method to interface.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
kidkun: We only pass `null` in `UserTaskManager`,and that is because we do not do logging in test.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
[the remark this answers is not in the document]
efeg: @alexwang08 @kidkun 
1. Thanks for confirming that this is the logger name,
2. I don't think my initial question is answered. Would you please clarify why you think we need to specify `log4j.logger.` prefix before the logger name? Isn't the logger name supposed to be the string that follows this prefix? In this case isn't it supposed to be `com.linkedin.kafka.cruisecontrol.CruiseControlOperationLog`? I think this is how we use the logger name in all the other loggers, no? Am I missing something? (please see: https://logging.apache.org/log4j/1.2/manual.html)
3. I think this question is not answered -- why do we use hierarchical naming here? Would you please point out the need for it based on our use-case and the original documentation (please see: https://logging.apache.org/log4j/1.2/manual.html)?

In summary, why don't we do the following changes:
1) [in `log4j.properties`]: 
```
log4j.logger.com.linkedin.kafka.cruisecontrol.operationLog=INFO, operationAppender -> log4j.logger.OperationLogger=INFO, operationAppender
```
2) [in `KafkaCruiseControlUtils.java`] 
```
public static final String OPERATION_LOGGER = "log4j.logger.com.linkedin.kafka.cruisecontrol.CruiseControlOperationLog";` -> `public static final String OPERATION_LOGGER = "OperationLogger";
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
efeg: Would it be possible to indicate what passing `null` means in the `JavaDoc`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -41,7 +41,7 @@
   private static final Set<String> KAFKA_ASSIGNER_GOALS =
       Collections.unmodifiableSet(new HashSet<>(Arrays.asList(KafkaAssignerEvenRackAwareGoal.class.getSimpleName(),
                                                               KafkaAssignerDiskUsageDistributionGoal.class.getSimpleName())));
-  public static final String OPERATION_LOGGER = "log4j.logger.com.linkedin.kafka.cruisecontrol.CruiseControlOperationLog";
+  public static final String OPERATION_LOGGER = "operationLogger";
```
efeg: Shouldn't the name of the logger match the name we use in `config/log4j.properties`? In this case `operationLogger` -> `operationLog`?

Issue #599, opened 2019-03-20
Only one sample topic's replication factor can be increased during CC booting up.


kidkun:
During booting up process, Cruise Control has an optional sanity check that whether the two sample topic, `__KafkaCruiseControlPartitionMetricSamples` and `__KafkaCruiseControlModelTrainingSamples`'s replication factor meets the requirement set by config `sample.store.topic.replication.factor`.  

 For each of the topic, we run this [function](https://github.com/linkedin/cruise-control/blob/2b98cdbe4380159b912dcbcc5a4e3082f52c1843/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java#L256), in which we first check that there is no ongoing reassignment operation, then if necessary we trigger an reassignment operation to increase the replication factor. 

There is a scenario that both topic need to increase its RF, after we trigger an operation for the first topic, the second topic's check for no ongoing operation will fail and does not trigger operation.

PR #604, opened 2019-03-21
Fix the bug that only one sample store topic's replication factor can increase in bootup sanity check.

kidkun:
fix for bug described in [599](https://github.com/linkedin/cruise-control/issues/599).

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
@@ -9,6 +9,7 @@ import com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils;
 import com.linkedin.kafka.cruisecontrol.metricsreporter.exception.UnknownVersionException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -233,13 +234,14 @@ public class KafkaSampleStore implements SampleStore {
                          replicationFactor, _partitionSampleStoreTopicPartitionCount);
       ensureTopicCreated(zkUtils, topics.keySet(), _brokerMetricSampleStoreTopic, brokerSampleRetentionMs,
                          replicationFactor, _brokerSampleStoreTopicPartitionCount);
+      maybeIncreaseTopicReplicationFactor(zkUtils, replicationFactor, Arrays.asList(_partitionMetricSampleStoreTopic, _brokerMetricSampleStoreTopic));
     } finally {
       KafkaCruiseControlUtils.closeZkUtilsWithTimeout(zkUtils, ZK_UTILS_CLOSE_TIMEOUT_MS);
     }
   }
 
   /**
-   * Increase the replication factor of a Kafka topic through adding new replicas in a rack-aware, round-robin way.
+   * Increase the replication factor of Kafka topics through adding new replicas in a rack-aware, round-robin way.
    * There are two scenarios that rack awareness property is not guaranteed.
    * <ul>
    *   <li> If Zookeeper does not have rack information about brokers, then it is only guaranteed that new replicas are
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -274,45 +274,51 @@ private void maybeIncreaseTopicReplicationFactor(ZkUtils zkUtils,
...
+      for (MetadataResponse.PartitionMetadata pm : topicMetadata.partitionMetadata()) {
+        if (pm.replicas().size() < replicationFactor) {
+          List<Object> newAssignedReplica = new ArrayList<>();
+          Set<String> currentOccupiedRack = new HashSet<>();
+          // Make sure the current replicas are in new replica list.
+          pm.replicas().forEach(node -> {
+            newAssignedReplica.add(node.id());
+            currentOccupiedRack.add(rackByBroker.get(node.id()));
+          });
+          // Add new replica to partition in rack-aware(if possible), round-robin way.
```
alexwang08: could we use `AdminUtil. assignReplicasToBrokers()` for rack aware assignment?

kidkun: IIUC, `AdminUtil. assignReplicasToBrokers()` does not consider the current replica distribution, but here we do not want to move current replicas, we just want to add new replica in Rack-aware way to satisfy replication factor requirement. 

we actually use this `AdminUtil. assignReplicasToBrokers()` to increase the partition count(if needed) in this Class,  because there we can assign wherever we want.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -249,16 +251,14 @@ private void ensureTopicsCreated(Map<String, ?> config) {
...
+   * @param topics The topics to check.
    */
   private void maybeIncreaseTopicReplicationFactor(ZkUtils zkUtils,
-                                                   MetadataResponse.TopicMetadata topicMetadata,
                                                    int replicationFactor,
-                                                   String topic) {
+                                                   Collection<String> topics) {
     if (!ensureNoPartitionUnderPartitionReassignment(zkUtils)) {
-      LOG.warn("There are ongoing partition reassignments, skip checking replication factor of topic {}.", topic);
+      LOG.warn("There are ongoing partition reassignments, skip checking replication factor of topics {}.", topics);
```
efeg: Will this check prevent RF increase, if there has been a partition count increase during the bootstrap time? -- i.e. this effectively means that CC can increase partition count of topics or RF of topics, but not both.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
```diff
@@ -274,45 +274,51 @@ private void maybeIncreaseTopicReplicationFactor(ZkUtils zkUtils,
...
-        // Add new replica to partition in rack-aware(if possible), round-robin way.
-        while (newAssignedReplica.size() < replicationFactor) {
-          if (!currentOccupiedRack.contains(racks.get(rackCursor)) || currentOccupiedRack.size() == racks.size()) {
-            String rack = racks.get(rackCursor);
-            int cursor = cursors[rackCursor];
-            newAssignedReplica.add(brokersByRack.get(rack).get(cursor));
-            currentOccupiedRack.add(rack);
-            cursors[rackCursor] = (cursor + 1) % brokersByRack.get(rack).size();
+    for (String topic : topics) {
+      MetadataResponse.TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(JavaConversions.asScalaSet(Collections.singleton(topic)),
```
efeg: Is it possible to send a single request for both topics rather than sending individual requests for each topic? -- i.e. why do we need `Collections.singleton(topic))`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
kidkun: For partition count increase, no data is written to `/admin/partition_reassignment` Znode, data is directly write to `/brokers/topics/topic_name` Znode to change AR list.So it will not block topic from increasing replica factor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java
efeg: @alexwang08 has this issue been resolved?

Issue #611, opened 2019-03-26
plaintext response of user_tasks endpoint can be misaligned.


kidkun:
Due to the [PR-551](https://github.com/linkedin/cruise-control/pull/551), now the task status field of task information returned by `user_tasks` endpoint can be double in length, and it breaks the original formatting assumption.

PR #612, opened 2019-03-26
Fix formatting issue in user_tasks endpoint's plaintext response.

kidkun:
Fix for issue [611](https://github.com/linkedin/cruise-control/issues/611)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/UserTaskState.java
```diff
@@ -124,7 +124,7 @@ private String getPlaintext(CruiseControlParameters parameters) {
     int userTaskIdLabelSize = 20;
     int clientAddressLabelSize = 20;
     int startMsLabelSize = 20;
-    int statusLabelSize = 10;
+    int statusLabelSize = 20;
```
efeg: Is it possible to adjust the size dynamically based on the existing `userTaskInfo` -- i.e. in the loop below?

PR #516, opened 2019-02-03
Make Cruise Control to better manage JBOD Kafka cluster.

kidkun:
* Add replica placement over disk information to cluster model. (address [issue 572](https://github.com/linkedin/cruise-control/issues/572))

* Add IntraBrokerDiskCapacityGoal and IntraBrokerDiskUsageDistributionGoal to support rebalance over disks within the broker. (address issue [573](https://github.com/linkedin/cruise-control/issues/573))
 
* Add Executor support to execute Execution task to move replica across disk within the same broker(address issue [577](https://github.com/linkedin/cruise-control/issues/577))

* Add a new parameter,`rebalance_disk` to `proposals` and `rebalance` servelet endpoint to access the new functionality(address issue [586](https://github.com/linkedin/cruise-control/issues/586))

PR #617, opened 2019-03-29
Refactor executor code.

kidkun:
Extract changes in [PR-516](https://github.com/linkedin/cruise-control/pull/516) which are applicable to master branch.This helps to make the code base in two branches more consistent and reduce the works for future cherry-pick between branches.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
@@ -105,9 +105,9 @@ public class ExecutionTaskManager {
    * Returns a list of execution proposal that moves the partitions.
    */
   public synchronized List<ExecutionTask> getReplicaMovementTasks() {
-    Map<Integer, Integer> getBrokersReadyForReplicaMovement = getBrokersReadyForReplicaMovement(_inProgressReplicaMovementsByBrokerId,
-                                                                                                partitionMovementConcurrency());
-    return _executionTaskPlanner.getReplicaMovementTasks(getBrokersReadyForReplicaMovement, _inProgressPartitions);
+    Map<Integer, Integer> brokersReadyForReplicaMovement = brokersReadyForReplicaMovement(_inProgressReplicaMovementsByBrokerId,
+                                                                                          partitionMovementConcurrency());
+    return _executionTaskPlanner.getReplicaMovementTasks(brokersReadyForReplicaMovement, _inProgressPartitions);
 
   }
 
@@ -119,8 +119,8 @@ public class ExecutionTaskManager {
    * @param movementConcurrency The required per-broker replica movement concurrency.
    * @return A map of how many new replica movements can be triggered for each broker.
    */
-  private Map<Integer, Integer> getBrokersReadyForReplicaMovement(Map<Integer, Integer> inProgressReplicaMovementsByBrokerId,
-                                                                  int movementConcurrency) {
+  private Map<Integer, Integer> brokersReadyForReplicaMovement(Map<Integer, Integer> inProgressReplicaMovementsByBrokerId,
+                                                               int movementConcurrency) {
     Map<Integer, Integer> readyBrokers = new HashMap<>(inProgressReplicaMovementsByBrokerId.size());
     for (Map.Entry<Integer, Integer> entry : inProgressReplicaMovementsByBrokerId.entrySet()) {
       // We skip the concurrency level check if caller requested so.
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
@@ -4,9 +4,7 @@
 
 package com.linkedin.kafka.cruisecontrol.executor;
 
-import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
-
 import com.linkedin.kafka.cruisecontrol.executor.strategy.ReplicaMovementStrategy;
 import java.util.Collection;
 import java.util.HashMap;
@@ -18,7 +16,9 @@ import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.utils.Time;
 
-
+import static com.linkedin.kafka.cruisecontrol.executor.ExecutionTask.TaskType;
+import static com.linkedin.kafka.cruisecontrol.executor.ExecutionTask.State;
+import static com.linkedin.kafka.cruisecontrol.executor.ExecutionTaskTracker.ExecutionTasksSummary;
 /**
  * The class that helps track the execution status for the balancing.
  * It does the following things:
@@ -41,32 +41,7 @@ public class ExecutionTaskManager {
   private Integer _requestedPartitionMovementConcurrency;
   private Integer _requestedLeadershipMovementConcurrency;
   private final Set<Integer> _brokersToSkipConcurrencyCheck;
-  private final Time _time;
-  private volatile long _inExecutionDataToMove;
   private boolean _isKafkaAssignerMode;
-
-  private static final String REPLICA_ACTION = "replica-action";
-  private static final String LEADERSHIP_ACTION = "leadership-action";
-  private static final String IN_PROGRESS = "in-progress";
-  private static final String PENDING = "pending";
-  private static final String ABORTING = "aborting";
-  private static final String ABORTED = "aborted";
-  private static final String DEAD = "dead";
-  private static final String ONGOING_EXECUTION = "ongoing-execution";
-  private static final String KAFKA_ASSIGNER_MODE = "kafka_assigner";
-
-  private static final String GAUGE_REPLICA_ACTION_IN_PROGRESS = REPLICA_ACTION + "-" + IN_PROGRESS;
-  private static final String GAUGE_LEADERSHIP_ACTION_IN_PROGRESS = LEADERSHIP_ACTION + "-" + IN_PROGRESS;
-  private static final String GAUGE_REPLICA_ACTION_PENDING = REPLICA_ACTION + "-" + PENDING;
-  private static final String GAUGE_LEADERSHIP_ACTION_PENDING = LEADERSHIP_ACTION + "-" + PENDING;
-  private static final String GAUGE_REPLICA_ACTION_ABORTING = REPLICA_ACTION + "-" + ABORTING;
-  private static final String GAUGE_LEADERSHIP_ACTION_ABORTING = LEADERSHIP_ACTION + "-" + ABORTING;
-  private static final String GAUGE_REPLICA_ACTION_ABORTED = REPLICA_ACTION + "-" + ABORTED;
-  private static final String GAUGE_LEADERSHIP_ACTION_ABORTED = LEADERSHIP_ACTION + "-" + ABORTED;
-  private static final String GAUGE_REPLICA_ACTION_DEAD = REPLICA_ACTION + "-" + DEAD;
-  private static final String GAUGE_LEADERSHIP_ACTION_DEAD = LEADERSHIP_ACTION + "-" + DEAD;
-  private static final String GAUGE_ONGOING_EXECUTION_IN_KAFKA_ASSIGNER_MODE = ONGOING_EXECUTION + "-"  + KAFKA_ASSIGNER_MODE;
-  private static final String GAUGE_ONGOING_EXECUTION_IN_NON_KAFKA_ASSIGNER_MODE = ONGOING_EXECUTION + "-non-"  + KAFKA_ASSIGNER_MODE;
 
   /**
    * The constructor of The Execution task manager.
@@ -86,19 +61,14 @@ public class ExecutionTaskManager {
                               Time time) {
     _inProgressReplicaMovementsByBrokerId = new HashMap<>();
     _inProgressPartitions = new HashSet<>();
-    _executionTaskTracker = new ExecutionTaskTracker();
+    _executionTaskTracker = new ExecutionTaskTracker(dropwizardMetricRegistry, time);
     _executionTaskPlanner = new ExecutionTaskPlanner(replicaMovementStrategies);
     _defaultPartitionMovementConcurrency = defaultPartitionMovementConcurrency;
     _defaultLeadershipMovementConcurrency = defaultLeadershipMovementConcurrency;
     _brokersToSkipConcurrencyCheck = new HashSet<>();
-    _inExecutionDataToMove = 0L;
-    _time = time;
     _isKafkaAssignerMode = false;
     _requestedPartitionMovementConcurrency = null;
     _requestedLeadershipMovementConcurrency = null;
-
-    // Register gauge sensors.
-    registerGaugeSensors(dropwizardMetricRegistry);
   }
 
   /**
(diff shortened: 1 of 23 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
```diff
@@ -360,169 +213,81 @@ public synchronized void markTaskDone(ExecutionTask task) {
...
+    _executionTaskTracker.clear();
+  }
 
-    /**
-     * Returns the remaining data to move in MB.
-     */
-    public long remainingDataToMoveInMB() {
-      return _remainingDataToMoveInMB;
-    }
+  public synchronized ExecutionTasksSummary getExecutionTasksSummary(List<TaskType> taskTypesToGetSnapshot,
```
efeg: I am worried that having an _implicit_ filtering based on (1) task types and (2) `inProgressOnly` may cause confusion.

For example, during a replica movement, it will show that there are `N` pending tasks, which will go down to `0` and then will increase again to `M` for leadership movements. This is a silent behavior change that user have no control over.

Rather than a silent change, I feel that it would make it more obvious to the user if we let them control such a filtering on the executor state response using some filtering parameters (if requested so). Hence, we can preserve the existing default response behavior with additional functionality to filter (if requested).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorState.java
```diff
@@ -143,154 +97,32 @@ public static ExecutorState executionStarted(String uuid,
...
-                                       long finishedDataMovementInMB,
-                                       int maximumConcurrentPartitionMovementsPerBroker,
-                                       int maximumConcurrentLeaderMovements,
-                                       String uuid,
-                                       Set<Integer> recentlyDemotedBrokers,
-                                       Set<Integer> recentlyRemovedBrokers) {
-    return new ExecutorState(State.STOPPING_EXECUTION,
-                             finishedPartitionMovements,
-                             finishedLeadershipMovements,
+  public static ExecutorState operationInProgress(State state,
```
efeg: It seems this public function is intended to accept only a certain set of states, but does not ensure that it is the case -- i.e. what happens if I send `STARTING_EXECUTION` state to it?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -607,30 +605,27 @@ private void moveReplicas() {
...
       LOG.info("Starting {} leadership movements.", numTotalLeadershipMovements);
-      _numFinishedLeadershipMovements = 0;
-      while (!_executionTaskManager.remainingLeadershipMovements().isEmpty() && !_stopRequested.get()) {
+      while (_executionTaskManager.numRemainingLeadershipMovements() != 0 && !_stopRequested.get()) {
         updateOngoingExecutionState();
-        _numFinishedLeadershipMovements += moveLeadershipInBatch();
-        LOG.info("{}/{} ({}%) leadership movements completed.", _numFinishedLeadershipMovements,
-                 numTotalLeadershipMovements, _numFinishedLeadershipMovements * 100 / numTotalLeadershipMovements);
+        moveLeadershipInBatch();
+        int numFinishedLeadershipMovements = _executionTaskManager.numFinishedLeadershipMovements();
```
efeg: It's not clear why we would make this change -- i.e. wouldn't this change receive the same information in a more expensive manner?

Previously we were just using the return value of `moveLeadershipInBatch()`, now we are making an expensive computation to get the same value.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -607,30 +605,27 @@ private void moveReplicas() {
...
-                 + "{} remaining data to move.",
-                 executionTasksSummary.inProgressTasks().size(),
-                 executionTasksSummary.remainingPartitionMovements().size(),
-                 executionTasksSummary.abortingTasks(),
-                 executionTasksSummary.abortedTasks().size(),
-                 executionTasksSummary.deadTasks().size(),
-                 executionTasksSummary.remainingDataToMoveInMB());
-      }
+      ExecutionTasksSummary executionTasksSummary = _executionTaskManager.getExecutionTasksSummary(Collections.emptyList(), false);
+      LOG.info("Partition movements {} with {} tasks {}, {} tasks in-progress, {} tasks aborting, {} tasks aborted, {} tasks dead, {} tasks completed.",
```
efeg: IIUC, this change does not show the affected _leadership_ movements. Note that if `_stopRequested.get()` is true, all tasks are affected -- not only replica movements.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -607,30 +605,27 @@ private void moveReplicas() {
...
+               _stopRequested.get() ? STOPPED : FINISHED,
+               executionTasksSummary.taskStat().get(REPLICA_ACTION).get(ExecutionTask.State.PENDING),
+               _stopRequested.get() ? CANCELLED : PENDING,
+               executionTasksSummary.taskStat().get(REPLICA_ACTION).get(ExecutionTask.State.IN_PROGRESS),
+               executionTasksSummary.taskStat().get(REPLICA_ACTION).get(ExecutionTask.State.ABORTING),
+               executionTasksSummary.taskStat().get(REPLICA_ACTION).get(ExecutionTask.State.ABORTED),
+               executionTasksSummary.taskStat().get(REPLICA_ACTION).get(ExecutionTask.State.DEAD),
+               executionTasksSummary.taskStat().get(REPLICA_ACTION).get(ExecutionTask.State.COMPLETED));
```
efeg: Is it possible to avoid calling `executionTasksSummary.taskStat().get(REPLICA_ACTION)` repetitively?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
```diff
@@ -61,70 +159,138 @@ public void setExecutionMode(boolean isKafkaAssignerMode) {
...
 
-  public int numAbortedLeadershipAction() {
-    return _leaderActionTasks.get(ExecutionTask.State.ABORTED).size();
+  public long finishedDataMovementInMB() {
+    return _finishedDataMovementInMB;
   }
-
-  public int numInProgressReplicaAction() {
-    return _replicaActionTasks.get(ExecutionTask.State.IN_PROGRESS).size();
+  public Set<ExecutionTask> inExecutionTasks(Collection<TaskType> types) {
```
efeg: Nit: Missing new line before `public Set<ExecutionTask> inExecutionTasks(Collection<TaskType> types)`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
```diff
@@ -4,51 +4,149 @@
...
+        dropwizardMetricRegistry.register(MetricRegistry.name(metricName, typeString + "-" + stateString),
+                                          (Gauge<Integer>) () -> _tasksByType.get(type).get(state).size());
+      }
+    }
+    dropwizardMetricRegistry.register(MetricRegistry.name(metricName, GAUGE_ONGOING_EXECUTION_IN_KAFKA_ASSIGNER_MODE),
+                                      (Gauge<Integer>) () -> _isKafkaAssignerMode
+                                                             && !inExecutionTasks(TaskType.cachedValues()).isEmpty() ? 1 : 0);
+    dropwizardMetricRegistry.register(MetricRegistry.name(metricName, GAUGE_ONGOING_EXECUTION_IN_NON_KAFKA_ASSIGNER_MODE),
+                                      (Gauge<Integer>) () -> !_isKafkaAssignerMode
+                                                             && !inExecutionTasks(TaskType.cachedValues()).isEmpty() ? 1 : 0);
```
efeg: Did we verify (e.g. via `jConsole`) that the existing metrics are not renamed/removed after this patch?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
kidkun: The main reason I have a filtering here is `_taskSnapshot` is a large object(if full populated it will contains handle for ExecutionTask objects of all type and state). So I want save some memory here and only store necessary task handles which are need for [response](https://github.com/linkedin/cruise-control/blob/37e77b32bafd288ea81594b9a498b758a71fbd41/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorState.java#L171). 

And whenever returns to client, the length of task list always match the statistic number( means if non-verbose response says there are N pending replica movement task, then in verbose response there will be N task in pending list) because they are generating while [holding ExecutionTaskManager](https://github.com/linkedin/cruise-control/blob/37e77b32bafd288ea81594b9a498b758a71fbd41/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java#L289)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
kidkun: Yes, I checked on a local deployment.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -218,11 +218,9 @@ public void testClear() {
                                           Collections.<String>emptySet());
 
     planner.addExecutionProposals(proposals, expectedCluster, null);
-    assertEquals(4, planner.remainingDataToMoveInMB());
     assertEquals(2, planner.remainingLeadershipMovements().size());
     assertEquals(2, planner.remainingReplicaMovements().size());
     planner.clear();
-    assertEquals(0, planner.remainingDataToMoveInMB());
```
efeg: I am curious on whether we have any new test to verify `remainingDataToMoveInMB` after the refactoring?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasTest.java
```diff
@@ -70,7 +71,7 @@ public void testLazyInitialization() {
     assertEquals("The replicas should be sorted lazily", 0, sr.numReplicas());
     sr.remove(replica);
     assertEquals("The replicas should be sorted lazily", 0, sr.numReplicas());
-    NavigableSet<ReplicaWrapper> sortedReplicas = sr.sortedReplicas();
+    List<Replica> sortedReplicas = sr.sortedReplicas();
```
efeg: Nit: I believe here we were testing `sortedReplicaWrappers` after the rename -- i.e. not `sortedReplicas` -- hence, we do not need to import `List` for this change.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
```diff
@@ -131,54 +101,30 @@ public synchronized int leadershipMovementConcurrency() {
...
-    Map<Integer, Integer> readyBrokers = new HashMap<>();
-    int partitionMovementConcurrency = partitionMovementConcurrency();
-    for (Map.Entry<Integer, Integer> entry : _inProgressReplicaMovementsByBrokerId.entrySet()) {
+    Map<Integer, Integer> readyBrokers = getReadyBrokers(_inProgressReplicaMovementsByBrokerId,
+                                                         partitionMovementConcurrency());
+    return _executionTaskPlanner.getReplicaMovementTasks(readyBrokers, _inProgressPartitions);
+
+  }
+
+  private Map<Integer, Integer> getReadyBrokers(Map<Integer, Integer> inProgressReplicaMovementsByBrokerId,
```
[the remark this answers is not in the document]
kidkun: The reason we factor out this chunk of code is because in 2.0 branch, there will be 2 counterpart method for `getReplicaMovementTasks()`, one for inter-broker movement and one for intra-broker movement,  and this method will be the common logic in two method.  Here I want to make two branches' code structure similar.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
```diff
@@ -4,51 +4,149 @@
...
-   * @param taskState the task state to get the execution task for.
-   * @return the leader action execution task set with the given task state.
+   * Add new tasks to ExecutionTaskTracker to track their execution state.
+   *
+   * @param tasks    New tasks to add.
+   * @param taskType Task type of new tasks.
    */
-  public Set<ExecutionTask> taskForLeaderAction(ExecutionTask.State taskState) {
-    return _replicaActionTasks.get(taskState);
+  public void initializeTask(Collection<ExecutionTask> tasks, TaskType taskType) {
```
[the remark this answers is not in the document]
kidkun: Since [this method](https://github.com/linkedin/cruise-control/blob/90e0e3f4a2e4044718badf9661e90bb18c0ce9a0/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java#L163) returns collections...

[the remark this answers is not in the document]
kidkun: Updated the Javadoc. 
I still keeps the `taskType` argument for this method just to make the implementation simple.
Otherwise we need to add the logic to (1) check either the collection is empty and (2)fetch the one item to determine task type, which will be hard for collection/set(need an iterator to get the first item, then while(iterator.hasNext()) to go over all the items.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
kidkun: It would be as before, i.e when executor is executing tasks, the what user get will be a little bit stale, but since both get the value from [_tasksByType](https://github.com/linkedin/cruise-control/blob/90e0e3f4a2e4044718badf9661e90bb18c0ce9a0/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java#L26), the value would be consistent(it is like snapshot isolation here...).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
```diff
@@ -105,14 +105,22 @@ public synchronized int leadershipMovementConcurrency() {
    * Returns a list of execution proposal that moves the partitions.
    */
   public synchronized List<ExecutionTask> getReplicaMovementTasks() {
-    Map<Integer, Integer> readyBrokers = getReadyBrokers(_inProgressReplicaMovementsByBrokerId,
-                                                         partitionMovementConcurrency());
-    return _executionTaskPlanner.getReplicaMovementTasks(readyBrokers, _inProgressPartitions);
+    Map<Integer, Integer> getBrokersReadyForReplicaMovement = getBrokersReadyForReplicaMovement(_inProgressReplicaMovementsByBrokerId,
```
efeg: `Map<Integer, Integer> getBrokersReadyForReplicaMovement` -> `Map<Integer, Integer> brokersReadyForReplicaMovement`

PR #641, opened 2019-04-05
Distinguish inter-broker and intra-broker replica movement.

kidkun:
This patch rename `replica movement` operation to `inter-broker replica movement` operation, it is a prerequisite to add a new type of operation, `intra-broker replica movement`, as part of JBOD support.

### DIFF
--- a/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
+++ b/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
@@ -200,7 +200,7 @@ public class ExecutorTest extends AbstractKafkaIntegrationTestHarness {
       Thread.sleep(10);
     }
     // Sleep over 180000 (the hard coded timeout)
-    time.sleep(180001);
+    time.sleep(200000);
     // The execution should finish.
     waitUntilExecutionFinishes(executor);
   }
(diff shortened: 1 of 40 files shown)


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -200,7 +200,7 @@ public void testTimeoutLeaderActions() throws InterruptedException {
       Thread.sleep(10);
     }
     // Sleep over 180000 (the hard coded timeout)
-    time.sleep(180001);
+    time.sleep(200000);
```
efeg: Nit: The comment above this line says `Sleep over 180000 (the hard coded timeout)` -- where does this number come from and why do we change it to `200K`?

kidkun: The reason is that I notice this test occasionally fail and the reason is since the startTime field of `ExecutionTask` is [marked](https://github.com/linkedin/cruise-control/blob/52627c4c98f9a473b2971851ecc7cdec7371a514/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTask.java#L144) in [another thread](https://github.com/linkedin/cruise-control/blob/52627c4c98f9a473b2971851ecc7cdec7371a514/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L430), the time marked there is not the same time as `time.milliseconds()` here,  occasionally it is a little bit later, and so I increase the margin here.

efeg: Ah I see -- could you add this comment to the line before this `sleep`?

PR #667, opened 2019-04-17
Add header field Access-Control-Allow-Credentials in GET/POST response e if CORS is enabled.

kidkun:
Per conversation in Slack channel, extra header field is needed for CORS to work.


nareshv:
Thanks - LGTM

Issue #580, opened 2019-03-01
Feature Request: LeaderDistributionGoal


mgrubent:
# Issue
In `kafka` clusters with a high degree of fan-out, there is non-trivial load placed on a broker simply by virtue of it serving as the leader for many partitions.

I would assert that this load does not manifest itself on the existing resource dimensions that `cruise-control` contemplates in its cluster model (e.g. CPU, network, disk), but is rather a form of thread exhaustion and/or lock contention within the `kafka` artifact itself.

This load results in the Request Handler Pool of the affected `kafka` brokers to become completely busy, such that it cannot effectively serve traffic.

Currently, there is no `goal` in `cruise-control` for balancing the leadership distribution across a `kafka cluster`

I concede that a naïve implementation of this goal could be at cross-purposes to the other, existing resource goals, but I would posit that, all else being equal, a tighter leadership distribution makes for a better-balanced `kafka` cluster.

# Proposal
`cruise-control` should implement a `LeaderDistributionGoal`, whose purpose would be to achieve a tighter distribution of leaderships.

It could be included or omitted in the `goals` parameter list as-desired.

PR #668, opened 2019-04-18
Add a new goal to balance leader replica distribution in cluster.

kidkun:
Patch for [issue 580](https://github.com/linkedin/cruise-control/issues/580).


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ExcludedBrokersForLeadershipTest.java
```diff
@@ -179,6 +181,20 @@
...
+    // Test: With single excluded broker, satisfiable cluster, no dead broker (No exception, No proposal for
+    // excluded broker, Expected to look optimized)
+    p.add(params(1, LeaderReplicaDistributionGoal.class, excludeB0, null, unbalanced3(), noDeadBroker, true));
+    // Test: With all brokers excluded, balance not satisfiable, no dead broker (No exception, No proposal
+    // for excluded brokers, Not expected to look optimized)
+    p.add(params(2, LeaderReplicaDistributionGoal.class, excludeAllBrokers, null, unbalanced3(), noDeadBroker, false));
+    // Test: With no brokers excluded, satisfiable cluster, no dead broker (No exception, No proposal
+    // for excluded brokers, Expected to look optimized)
+    p.add(params(3, LeaderReplicaDistributionGoal.class, noExclusion, null, unbalanced3(), noDeadBroker, true));
+
```
efeg: Nit: Can we also have a test with dead broker(s)?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/RandomClusterTest.java
```diff
@@ -75,6 +76,7 @@
                                                     NetworkOutboundCapacityGoal.class.getName(),
                                                     CpuCapacityGoal.class.getName(),
                                                     ReplicaDistributionGoal.class.getName(),
+                                                    LeaderReplicaDistributionGoal.class.getName(),
```
efeg: Is it possible to keep the order consistent with the default goals?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/RandomGoalTest.java
```diff
@@ -72,6 +73,7 @@
                                                        NetworkOutboundCapacityGoal.class.getName(),
                                                        CpuCapacityGoal.class.getName(),
                                                        ReplicaDistributionGoal.class.getName(),
+                                                       LeaderReplicaDistributionGoal.class.getName(),
```
efeg: Is it possible to keep the order consistent with the default goals?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  /**
+   * To avoid churns, we add a balance margin to the user specified rebalance threshold. e.g. when user sets the
+   * threshold to be {@link BalancingConstraint#leaderReplicaBalancePercentage()}, we use
+   * ({@link BalancingConstraint#leaderReplicaBalancePercentage()}-1)*{@link #BALANCE_MARGIN} instead.
+   * @return the rebalance threshold with a margin.
+   */
+  private double balancePercentageWithMargin() {
+    return (_balancingConstraint.leaderReplicaBalancePercentage() - 1) * BALANCE_MARGIN;
```
efeg: Why does this function omit `goalViolationDistributionThresholdMultiplier` -- i.e. isn't this a distribution goal?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  @Override
+  public boolean isHardGoal() {
+    return false;
+  }
+
+  /**
+   * Get brokers that the rebalance process will go over to apply balancing actions to replicas they contain.
+   * Leadership movement is unable to move replica off dead broker, therefore this goal only go over alive brokers.
```
efeg: Is it possible to make this goal consistent with `ReplicaDistributionGoal` and `TopicReplicaDistributionGoal` -- e.g. can we add `_selfHealingDeadBrokersOnly` to move replicas off of dead brokers. It seems this goal fails to fix offline replicas as it is implemented.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,336 @@
...
+      LOG.debug("Failed to sufficiently decrease leader replica count in broker {} with leadership movements. Leader replicas: {}.",
+                broker.id(), broker.leaderReplicas().size());
+    } else if (numLeaderReplicas < _balanceLowerLimit && rebalanceByMovingLeadershipIn(broker, clusterModel, optimizedGoals, optimizationOptions)) {
+      _brokerIdsUnderBalanceLowerLimit.add(broker.id());
+      LOG.debug("Failed to sufficiently increase leader replica count in broker {} with leadership movements. Leader replicas: {}.",
+                broker.id(), broker.leaderReplicas().size());
+    }
+  }
+
+  private boolean rebalanceByMovingLeadershipOut(Broker broker,
```
efeg: Why do we only allow leadership moves? If leadership move doesn't work, can't we allow (1) replica moves and (2) replica swaps?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
kidkun: I think this is get missed in rebase.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
kidkun: I added leader replica move operation

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
kidkun: I added this back. But instead of `move replicas off of dead brokers` this goal can only `move **leader** replica off dead brokers` since this goal is all about leader replica distribution, it does not have enough information to make decision about follower replica movement.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
```diff
@@ -344,6 +354,35 @@ private void numForReplicas(ClusterModel clusterModel) {
...
+    // Average, minimum, and maximum number of leader replicas in brokers.
+    int maxLeaderReplicasInBroker = 0;
+    int minLeaderReplicasInBroker = Integer.MAX_VALUE;
+    int numAliveBrokers = clusterModel.aliveBrokers().size();
+    for (Broker broker : clusterModel.brokers()) {
+      int numLeaderReplicasInBroker = broker.leaderReplicas().size();
+      maxLeaderReplicasInBroker = Math.max(maxLeaderReplicasInBroker, numLeaderReplicasInBroker);
+      minLeaderReplicasInBroker = Math.min(minLeaderReplicasInBroker, numLeaderReplicasInBroker);
+    }
+    double avgLeaderReplicas = ((double) clusterModel.numLeaderReplicas()) / numAliveBrokers;
```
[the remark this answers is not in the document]
kidkun: I feel this need further thinking. 
let's use `ReplicaDistributionGoal` as an example, in this goal's logic, the target average replica on broker is [based on alive brokers](https://github.com/linkedin/cruise-control/blob/bcb8257c5b47b16c3f612e588864aa3ba4993597/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java#L201).
So for a cluster of broker [1,2,3] and each broker host to 2 replicas. let's say broker 3 is dead, `ReplicaDistributionGoal` will self-heal cluster to broker 1,2 have 3 replica each.  but if we use [current way](https://github.com/linkedin/cruise-control/blob/bcb8257c5b47b16c3f612e588864aa3ba4993597/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java#L316) of calculating cluster status, the pre-optimization replicaStas's SDV is 0 and post-optimization is around 0.5. But shouldn't post-optimization cluster status better?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
efeg: Is it possible for the leader of any partition to reside on a dead broker/disk when `RF>1`? Can't we move any replica off of the failed brokers/disks if `_selfHealingDeadBrokersOnly`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,354 @@
...
+                                                                 .thenComparingInt(Broker::id));
+    candidateBrokers.addAll(_selfHealingDeadBrokersOnly ? clusterModel.aliveBrokers()
+                                                        : clusterModel.aliveBrokers()
+                                                                      .stream()
+                                                                      .filter(b -> b.leaderReplicas().size() < _balanceUpperLimit)
+                                                                      .collect(Collectors.toSet()));
+
+    int balanceUpperLimit = broker.isAlive() ? _balanceUpperLimit : 0;
+    Set<String> excludedTopics = optimizationOptions.excludedTopics();
+    for (Replica replica : new HashSet<>(broker.leaderReplicas())) {
```
efeg: This goal does not keep the replicas to move out limited to immigrant replicas in case there are dead brokers -- i.e. replicas whose original broker is not this one. Wouldn't lack of this prioritization cause leadership/replica actions across alive brokers during fixing broker failures?

Don't we have a test case to ensure that the moves are limited to replicas that came from failed brokers -- if not can we add a test case?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,354 @@
...
+      }
+      Broker b = maybeApplyBalancingAction(clusterModel,
+                                           replica,
+                                           candidateBrokers,
+                                           ActionType.INTER_BROKER_REPLICA_MOVEMENT,
+                                           optimizedGoals,
+                                           optimizationOptions);
+      // Only check if we successfully moved something.
+      if (b != null) {
+        if (broker.leaderReplicas().size() <= balanceUpperLimit) {
```
efeg: rather than getting the size of a new UnmodifiableSet every time, is it possible to use a counter to keep track of numLeaders?

efeg: Here we are missing the case where we are self healing a dead broker and trying to remove all replicas regardless of their leadership.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,354 @@
...
+      }
+    }
+    return true;
+  }
+
+  private boolean rebalanceByMovingLeaderReplicasIn(Broker broker,
+                                                   ClusterModel clusterModel,
+                                                   Set<Goal> optimizedGoals,
+                                                   OptimizationOptions optimizationOptions) {
+    PriorityQueue<Broker> eligibleBrokers = new PriorityQueue<>((b1, b2) -> {
```
efeg: for early termination (as in a previous function in this class), can we check:
```
    if (optimizationOptions.excludedBrokersForLeadership().contains(broker.id())) {
      return true;
    }
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,354 @@
...
+      }
+      Broker b = maybeApplyBalancingAction(clusterModel,
+                                           clusterModel.partition(replica.topicPartition()).leader(),
+                                           candidateBrokers,
+                                           ActionType.LEADERSHIP_MOVEMENT,
+                                           optimizedGoals,
+                                           optimizationOptions);
+      // Only check if we successfully moved something.
+      if (b != null) {
+        if (broker.leaderReplicas().size() >= _balanceLowerLimit) {
```
efeg: Nit: Extra space after `>=`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,354 @@
...
+  private void moveReplicasOffDeadBroker(Broker broker,
+                                         ClusterModel clusterModel,
+                                         Set<Goal> optimizedGoals,
+                                         OptimizationOptions optimizationOptions) {
+    // Get the eligible brokers.
+    SortedSet<Broker> candidateBrokers = new TreeSet<>(Comparator.comparingInt((Broker b) -> b.replicas().size())
+                                                                 .thenComparingInt(Broker::id));
+    candidateBrokers.addAll(clusterModel.aliveBrokers());
+
+    Set<String> excludedTopics = optimizationOptions.excludedTopics();
```
efeg: If we intend to use this function for removing replicas off of a dead broker, why would we check `excludedTopics`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,354 @@
...
+    for (Replica replica : new HashSet<>(broker.replicas())) {
+      if (shouldExclude(replica, excludedTopics)) {
```
efeg: Is `shouldExclude(replica, excludedTopics)` ever `true`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,354 @@
...
+    } else if (requireMoreLeaderReplicas
+               && rebalanceByMovingLeadershipIn(broker, clusterModel, optimizedGoals, optimizationOptions)
+               && rebalanceByMovingLeaderReplicasIn(broker, clusterModel, optimizedGoals, optimizationOptions)) {
+      _brokerIdsUnderBalanceLowerLimit.add(broker.id());
+      LOG.debug("Failed to sufficiently increase leader replica count in broker {}. Leader replicas: {}.",
+                broker.id(), broker.leaderReplicas().size());
+    }
+    // If the broker is dead and currently self healing dead brokers only, then generate proposals to move all the
+    // remaining replicas off this broker.
+    if (_selfHealingDeadBrokersOnly && !broker.isAlive() && !broker.replicas().isEmpty()) {
```
efeg: Is `!broker.replicas().isEmpty()` a needed check?

efeg: This flow seems to be contradicting the other goals -- why don't we move this logic to `rebalanceByMovingLeadershipOut`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,354 @@
...
+   * @param optimizationOptions Options to take into account during optimization -- e.g. excluded topics.
+   */
+  @Override
+  protected void rebalanceForBroker(Broker broker,
+                                    ClusterModel clusterModel,
+                                    Set<Goal> optimizedGoals,
+                                    OptimizationOptions optimizationOptions) {
+    LOG.debug("Rebalancing broker {} [limits] lower: {} upper: {}.", broker.id(), _balanceLowerLimit, _balanceUpperLimit);
+    int numLeaderReplicas = broker.leaderReplicas().size();
+    boolean requireLessLeaderReplicas = broker.isAlive() ? numLeaderReplicas > _balanceUpperLimit : numLeaderReplicas > 0;
```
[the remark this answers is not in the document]
kidkun: The case I have in mind is RF=1 partition and ULE is enabled, and I want to move the leader replica to a broker and satisfy threshold limit. Do you think this is a valid case?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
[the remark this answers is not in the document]
kidkun: There is another method, `moveReplicasOffDeadBroker`, which is responsible for moving follower replicas off the dead broker. 

My idea is that in self-healing, this goal does a 2 pass.
1st pass: only consider the leader replicas and try not violate leader replica threshold
2nd pass:  `_selfHealingDeadBrokersOnly` is true, then move all offline replicas off dead broker

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
kidkun: It is done by another method `moveReplicasOffDeadBroker`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
kidkun: Looks this cause a lot of confusion... 
My origin idea is to make `rebalanceByMovingLeadershipOut` easy-to-induction, I will merge the logic into `rebalanceByMovingLeadershipOut`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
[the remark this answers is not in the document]
kidkun: merged the logic of `moveReplicasOffDeadBroker()` into `rebalanceByMovingReplicasOut()`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
kidkun: merged the logic of `moveReplicasOffDeadBroker()` into `rebalanceByMovingReplicasOut()`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionAbstractGoal.java
```diff
@@ -181,18 +186,18 @@ protected void updateGoalState(ClusterModel clusterModel, Set<String> excludedTo
     // Log broker Ids over balancing limit.
     // While proposals exclude the excludedTopics, the balance still considers utilization of the excludedTopic replicas.
     if (!_brokerIdsAboveBalanceUpperLimit.isEmpty()) {
-      LOG.warn("Replicas count on broker ids:{} {} above the balance limit of {} after {}.",
-          _brokerIdsAboveBalanceUpperLimit, (_brokerIdsAboveBalanceUpperLimit.size() > 1) ? "are" : "is",
-          _balanceUpperLimit,
-          (clusterModel.selfHealingEligibleReplicas().isEmpty()) ? "rebalance" : "self-healing");
+      LOG.info("Replicas count on broker ids:{} {} above the balance limit of {} after {}.",
```
efeg: Nit: Can we make this one `debug` rather than `info` -- i.e. because it specifies a soft goal violation in a usual rebalance, and we already show if the goal is `FIXED` or `VIOLATED` after the rebalance. This level of detail might be more appropriate for a debug-level.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionAbstractGoal.java
```diff
@@ -181,18 +186,18 @@ protected void updateGoalState(ClusterModel clusterModel, Set<String> excludedTo
...
+               (clusterModel.selfHealingEligibleReplicas().isEmpty()) ? "rebalance" : "self-healing");
       _brokerIdsAboveBalanceUpperLimit.clear();
       _succeeded = false;
     }
     if (!_brokerIdsUnderBalanceLowerLimit.isEmpty()) {
-      LOG.warn("Replica count on broker ids:{} {} under the balance limit of {} after {}.",
-          _brokerIdsUnderBalanceLowerLimit, (_brokerIdsUnderBalanceLowerLimit.size() > 1) ? "are" : "is",
-          _balanceLowerLimit,
-          (clusterModel.selfHealingEligibleReplicas().isEmpty()) ? "rebalance" : "self-healing");
+      LOG.info("Replica count on broker ids:{} {} under the balance limit of {} after {}.",
```
efeg: Nit: Can we make this one `debug` rather than `info` -- i.e. because it specifies a soft goal violation in a usual rebalance, and we already show if the goal is `FIXED` or `VIOLATED` after the rebalance. This level of detail might be more appropriate for a debug-level.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -257,13 +257,13 @@ protected void updateGoalState(ClusterModel clusterModel, Set<String> excludedTo
       }
     }
     if (!brokerIdsAboveBalanceUpperLimit.isEmpty()) {
-      LOG.warn("Utilization for broker ids:{} {} above the balance limit for:{} after {}.",
+      LOG.info("Utilization for broker ids:{} {} above the balance limit for:{} after {}.",
```
efeg: Nit: Can we make this one `debug` rather than `info` -- i.e. because it specifies a soft goal violation in a usual rebalance, and we already show if the goal is `FIXED` or `VIOLATED` after the rebalance. This level of detail might be more appropriate for a debug-level.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -257,13 +257,13 @@ protected void updateGoalState(ClusterModel clusterModel, Set<String> excludedTo
...
                brokerIdsAboveBalanceUpperLimit, (brokerIdsAboveBalanceUpperLimit.size() > 1) ? "are" : "is", resource(),
                (clusterModel.selfHealingEligibleReplicas().isEmpty()) ? "rebalance" : "self-healing");
       _succeeded = false;
     }
     if (!brokerIdsUnderBalanceLowerLimit.isEmpty()) {
-      LOG.warn("Utilization for broker ids:{} {} under the balance limit for:{} after {}.",
+      LOG.info("Utilization for broker ids:{} {} under the balance limit for:{} after {}.",
```
efeg: Nit: Can we make this one `debug` rather than `info` -- i.e. because it specifies a soft goal violation in a usual rebalance, and we already show if the goal is `FIXED` or `VIOLATED` after the rebalance. This level of detail might be more appropriate for a debug-level.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/TopicReplicaDistributionGoal.java
```diff
@@ -306,7 +306,7 @@ protected void updateGoalState(ClusterModel clusterModel, Set<String> excludedTo
         throw ofe;
       }
       _selfHealingDeadBrokersOnly = true;
-      LOG.warn("Ignoring topic replica balance limit to move replicas from dead brokers to healthy ones.");
+      LOG.info("Ignoring topic replica balance limit to move replicas from dead brokers to healthy ones.");
```
efeg: This is great for `info`-level -- thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -276,7 +276,7 @@ protected void updateGoalState(ClusterModel clusterModel, Set<String> excludedTo
         throw ofe;
       }
       _selfHealingDeadBrokersOnly = true;
-      LOG.warn("Ignoring resource balance limit to move replicas from dead brokers to healthy ones.");
+      LOG.info("Ignoring resource balance limit to move replicas from dead brokers to healthy ones.");
```
efeg: This is great for `info`-level -- thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionAbstractGoal.java
```diff
@@ -204,7 +209,7 @@ protected void updateGoalState(ClusterModel clusterModel, Set<String> excludedTo
         throw ofe;
       }
       _selfHealingDeadBrokersOnly = true;
-      LOG.warn("Ignoring replica balance limit to move replicas from dead brokers to healthy ones.");
+      LOG.info("Ignoring replica balance limit to move replicas from dead brokers to healthy ones.");
```
efeg: This is great for `info`-level -- thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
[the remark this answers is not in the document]
efeg: TL;DR -- No, I don't think we can do anything to resolve `RF=1` failures.

Consider a replica from a partition with RF=1. If the broker that holds the replica dies (e.g. the machine is down/inaccessible), then the partition itself is offline. Hence, there is no available replica from which any system, including Kafka, can retrieve the offline data.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,354 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(LeaderReplicaDistributionGoal.class);
+
+  /**
+   * Constructor for Leader Replica Distribution Goal.
+   */
+  public LeaderReplicaDistributionGoal() {
+  }
+
+  public LeaderReplicaDistributionGoal(BalancingConstraint balancingConstraint) {
+    this();
```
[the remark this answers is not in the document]
efeg: I noticed that we keep the call to `this()`, I don't think we need to call `this()`. -- am I missing something?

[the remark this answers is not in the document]
kidkun: My understanding is that  parametrized constructor does not call default constructor automatically, but here we need to call default constructor since default constructor will call super class's default constructor and we need to initialize the related fields in [ReplicaDistributionAbstractGoal](https://github.com/linkedin/cruise-control/blob/905b9aee243aab05034579a6a7aeb21aa9ef7e3b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionAbstractGoal.java#L42) and [AbstractGoal](https://github.com/linkedin/cruise-control/blob/905b9aee243aab05034579a6a7aeb21aa9ef7e3b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java#L54).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,343 @@
...
+                                                 .thenComparingInt(Broker::id));
+      candidateBrokers.addAll(clusterModel.aliveBrokers()
+                      .stream()
+                      .filter(b -> b.leaderReplicas().size() < _balanceUpperLimit)
+                      .collect(Collectors.toSet()));
+    }
+
+    int balanceUpperLimit = _selfHealingDeadBrokersOnly ? 0 : _balanceUpperLimit;
+    Set<String> excludedTopics = optimizationOptions.excludedTopics();
+    Set<Replica> candidateReplicas = _selfHealingDeadBrokersOnly ? new HashSet<>(broker.replicas())
```
efeg: Nit (`style`): `Set<Replica> candidateReplicas = new HashSet<>(_selfHealingDeadBrokersOnly ? broker.replicas() : broker.leaderReplicas());`

-- It is also perfectly OK to keep it as is, but typically I try to extract common logic when using ternary.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -0,0 +1,343 @@
...
+    List<Broker> candidateBrokers = Collections.singletonList(broker);
+    Set<String> excludedTopics = optimizationOptions.excludedTopics();
+    int numLeaderReplicas = broker.leaderReplicas().size();
+    while (!eligibleBrokers.isEmpty()) {
+      Broker sourceBroker = eligibleBrokers.poll();
+      for (Replica replica : new HashSet<>(sourceBroker.leaderReplicas())) {
+        if (shouldExclude(replica, excludedTopics) || broker.replica(replica.topicPartition()) != null) {
+          continue;
+        }
+        if (!clusterModel.deadBrokers().isEmpty() || !broker.immigrantReplicas().contains(replica)) {
```
efeg: Would you clarify what this check means?
It looks like the current code skips moving leader replicas in if (1) there is a dead broker in the cluster or (2) the leader replica that is intended to be moved to this broker (an currently resides on another broker) is not an immigrant replica in this broker (which I think will always be `false`).

kidkun: Good catch! 
It should be `!clusterModel.deadBrokers().isEmpty() && !sourceBroker.immigrantReplicas().contains(replica)`, which limit the movement to only immigrant replica in self-healing case.

PR #670, opened 2019-04-22
Add intra.broker.goals config to specify supported intra-broker goals.

kidkun:
patch for [issue-671](https://github.com/linkedin/cruise-control/issues/671)

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/AnalyzerUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/AnalyzerUtils.java
@@ -153,6 +153,7 @@ public class AnalyzerUtils {
    */
   public static Map<String, Goal> getCaseInsensitiveGoalsByName(KafkaCruiseControlConfig config) {
     List<Goal> goals = config.getConfiguredInstances(KafkaCruiseControlConfig.GOALS_CONFIG, Goal.class);
+    goals.addAll(config.getConfiguredInstances(KafkaCruiseControlConfig.INTRA_BROKER_GOALS_CONFIG, Goal.class));
     Map<String, Goal> caseInsensitiveGoalsByName = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
     for (Goal goal: goals) {
       caseInsensitiveGoalsByName.put(goal.name(), goal);
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/AnalyzerUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/AnalyzerUtils.java
@@ -153,7 +153,6 @@ public class AnalyzerUtils {
    */
   public static Map<String, Goal> getCaseInsensitiveGoalsByName(KafkaCruiseControlConfig config) {
     List<Goal> goals = config.getConfiguredInstances(KafkaCruiseControlConfig.GOALS_CONFIG, Goal.class);
-    goals.addAll(config.getConfiguredInstances(KafkaCruiseControlConfig.INTRA_BROKER_GOALS_CONFIG, Goal.class));
     Map<String, Goal> caseInsensitiveGoalsByName = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
     for (Goal goal: goals) {
       caseInsensitiveGoalsByName.put(goal.name(), goal);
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -1140,6 +1148,11 @@
                     .add(TopicReplicaDistributionGoal.class.getName()).toString(),
                 ConfigDef.Importance.HIGH,
                 GOALS_DOC)
+        .define(INTRA_BROKER_GOALS_CONFIG,
+                ConfigDef.Type.LIST,
+                "",
```
efeg: Shouldn't this be an empty list rather than an empty string?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -488,6 +488,14 @@
   private static final String GOALS_DOC = "A list of case insensitive goals in the order of priority. The high "
       + "priority goals will be executed first.";
 
+  /**
+   * <code>intra.broker.goals</code>
+   */
+  public static final String INTRA_BROKER_GOALS_CONFIG = "intra.broker.goals";
+  private static final String INTRA_BROKER_GOALS_DOC = "A list of case insensitive intra-broker goals in the order of priority. "
+      + "The high priority goals will be executed first. The intra-broker goals are only relevant if intra-broker operation is "
+      + "supported(i.e. in migrate_to_kafka_2_0 branch), otherwise this list should be empty. ";
```
efeg: `empty. "` -> `empty."` -- i.e. redundant space after period.

efeg: I think it would be better to indicate CC version rather than Github branch name (because we may change the branch name in the future) here -- i.e. maybe we can say this is supported only in CC versions above `2.*`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/AnalyzerUtils.java
```diff
@@ -153,6 +153,7 @@ public static int compare(double d1, double d2, double epsilon) {
    */
   public static Map<String, Goal> getCaseInsensitiveGoalsByName(KafkaCruiseControlConfig config) {
     List<Goal> goals = config.getConfiguredInstances(KafkaCruiseControlConfig.GOALS_CONFIG, Goal.class);
+    goals.addAll(config.getConfiguredInstances(KafkaCruiseControlConfig.INTRA_BROKER_GOALS_CONFIG, Goal.class));
```
efeg: If it is not relevant here (Note that we don't seem to have a sanity check to ensure that this config is not empty), should we refrain from adding these goals in this branch?  -- we may limit this line to `migrate_to_XXX` branch.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -1277,6 +1290,13 @@ private void sanityCheckGoalNames() {
         throw new ConfigException("Attempt to configure goals with case sensitive names.");
       }
     }
+    // Check goal names are case insensitive for intra-broker goals.
+    for (String goalName: getList(KafkaCruiseControlConfig.INTRA_BROKER_GOALS_CONFIG)) {
+      if (!caseInsensitiveGoalNames.add(goalName.replaceAll(".*\\.", ""))) {
+        throw new ConfigException("Attempt to configure intra.broker.goals with case sensitive names.");
```
efeg: Nit: If we intend to
(1) use exact config name, can we use `INTRA_BROKER_GOALS_CONFIG` in a formatted string?
(2) not use exact config name, would it be possible to change `intra.broker.goals` to `intra-broker goals`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
kidkun: I think both are ok, the return result from `Config.getList()` will be empty list. I will change to empty list since it adheres to this config's type.

Issue #681, opened 2019-04-30
Improve performance and accuracy of anomaly detector.


efeg:
* Avoid excessive generation of broker failure anomalies when there are dead brokers not ready for a fix. An anomaly is not ready if it (1) does not meet completeness requirements or (2) load monitor is in an unexpected state.
* Avoid redundant and repeated hard goal presence sanity checks for goals used in anomaly detection.

PR #682, opened 2019-04-30
Improve performance and accuracy of anomaly detector.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/681.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
@@ -61,6 +61,8 @@ public class AnomalyDetector {
   private final Map<AnomalyType, Meter> _anomalyRateByType;
   private final LoadMonitor _loadMonitor;
   private final AnomalyDetectorState _anomalyDetectorState;
+  // TODO: Make this configurable.
+  private final List<String> _selfHealingGoals;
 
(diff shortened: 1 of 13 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -284,8 +288,7 @@ private boolean isReadyToFix(Anomaly anomaly) {
         LOG.info("Skipping {} because load monitor is in {} state.", skipMsg, loadMonitorTaskRunnerState);
         _anomalyDetectorState.onAnomalyHandle(anomaly, AnomalyState.Status.LOAD_MONITOR_NOT_READY);
       } else {
-        boolean meetCompletenessRequirements = _kafkaCruiseControl.meetCompletenessRequirements(Collections.emptyList());
-        if (meetCompletenessRequirements) {
+        if (_kafkaCruiseControl.meetCompletenessRequirementsWithDefaultGoals()) {
```
kidkun: Since we add `_selfHealingGoals` field to `BrokerFailures` and `GoalViolations`, I feel we should not unconditionally check completeness against default goal list here.

 Currently this is ok since we always set `_selfHealingGoals` to empty list and default goal list will be picked up. I think the reason we add `_selfHealingGoals` is because we may want to tune the goals use for self-healing in the future and this implicit assumption will not hold?

Issue #685, opened 2019-05-01
Remove the use of deprecated org.apache.kafka.common.Metric API in CC supporting Kafka versions 1.0.0+


efeg:
As of `1.0.0`, `Metric#value()` is deprecated, and it is recommended to use {@link #metricValue()} instead.

PR #686, opened 2019-05-01
Remove the use of deprecated org.apache.kafka.common.Metric API.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/685.

### DIFF
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/metric/MetricsUtils.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/metric/MetricsUtils.java
@@ -98,7 +98,13 @@ public class MetricsUtils {
    */
   public static CruiseControlMetric toCruiseControlMetric(KafkaMetric kafkaMetric, long now, int brokerId) {
     org.apache.kafka.common.MetricName metricName = kafkaMetric.metricName();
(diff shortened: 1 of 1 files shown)


cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/metric/MetricsUtils.java
```diff
@@ -98,7 +98,13 @@ private MetricsUtils() {
    */
   public static CruiseControlMetric toCruiseControlMetric(KafkaMetric kafkaMetric, long now, int brokerId) {
     org.apache.kafka.common.MetricName metricName = kafkaMetric.metricName();
-    CruiseControlMetric ccm = toCruiseControlMetric(now, brokerId, metricName.name(), metricName.tags(), kafkaMetric.value());
+    if (!(kafkaMetric.metricValue() instanceof Double)) {
```
kidkun: Is this check too strict? I suspect some metric value is not of type of subclass of `Double` but can be cast to `Double` (e.g. Long)

efeg: The current default metrics, metrics reporter, and the monitor (e.g. the monitoring windows) are designed and implemented to handle only the `double` metrics. This change is a no-op for the current version of CC to support the evolving API -- though if we intend to support other types in the upcoming Kafka versions, we would indeed need to update the whole relevant pipeline to do so.

Issue #706, opened 2019-05-14
Ensure that reconnect.backoff.ms configuration applied to metric sampler consumer and sample store producer/consumers.


efeg:
` reconnect.backoff.ms` is a common client config that specifies the time to wait before reconnecting to a given host. This config currently applies only to metadata client, but it should also apply to metric sampler consumer and sample store producer/consumers.

PR #707, opened 2019-05-14
Ensure that reconnect.backoff.ms configuration applied to metric sampler consumer and sample store producer/consumers

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/706.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsReporterSampler.java
```diff
@@ -32,6 +32,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig.*;
```
kidkun: nit: redundant import

Issue #579, opened 2019-02-28
Looking for a kafka-cruise-control-stop.sh it would be useful if you guys add in coming versions


shanrd7:

efeg:
@emilyki please feel free to work on this issue.

emilyki:
@efeg I'll be working on this issue, thank you!

PR #726, opened 2019-05-22
Add kafka-cruise-control-stop.sh.

emilyki:
This pull request addresses the issue #579.

Note: this script assumes that the only running java process is the cruise-control.

This is an example log output that I got from testing the command:
https://gist.github.com/emilyki/1dbbf380646c2079400faeb2c084f2fb

### DIFF
--- a/kafka-cruise-control-stop.sh
+++ b/kafka-cruise-control-stop.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# Copyright 2019 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").
+# See License in the project root for license information.
+
+SIGNAL=${SIGNAL:-TERM}
+PIDS=$(ps ax | grep -i 'cruise-control' | grep java | grep -v grep | awk '{print $1}')
+
+if [ -z "$PIDS" ]; then
+  echo "No cruise-control to stop"
+  exit 1
+else
+  kill -s $SIGNAL $PIDS
+fi


emilyki:
Requesting for review from @efeg and @kidkun . Thanks!

kidkun:
Hey @emilyki  Thanks for the patch!
I feel the assumption of ` only running java process is the cruise-control` is too restrict and may not hold in many scenarios.

Probably a better way to `grep` the java process which has key word `cruise` and `control` to filter out the pid, like the [kafka-server-stop.sh script](https://github.com/apache/kafka/blob/trunk/bin/kafka-server-stop.sh).

emilyki:
Hello @kidkun!
Thanks for your feedback, I agree that the restriction is too strict and doesn't hold for many cases! I have updated (and squashed) the PR with the suggested changes. It seems to work well when I tested the changes locally.  May I ask for your review on the fix? Thanks!

kafka-cruise-control-stop.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").
```
efeg: Nit: `2017` -> `2019`

kafka-cruise-control-stop.sh
```diff
@@ -0,0 +1,13 @@
...
+# See License in the project root for license information.
+
+SIGNAL=${SIGNAL:-TERM}
+PIDS=$(ps ax | grep -i 'cruise' | grep -i 'control' | awk '{print $1}')
```
efeg: Does the following command work? -- i.e. it also ensures that the full phrase `cruise-control` is present, it is a java application and not `grep` itself:

`ps ax | grep -i 'cruise-control' | grep java | grep -v grep | awk '{print $1}'`

emilyki: Yes, this works perfectly! I've tested the command and verified that it works. :)

emilyki:
Thank you @efeg for the feedback, I have updated the pull request based on your suggestions. May I ask for another review on the updated PR? Thank you!

Issue #697, opened 2019-05-09
Filter kafka_cluster_state by given topic parameter


efeg:
Add support for `topic` parameter that takes a regex to filter the response of `kafka_cluster_state` by the given topic pattern.

Will-Lo:
@efeg I'll pick this issue up, thanks!

PR #727, opened 2019-05-27
Filter kafka cluster by topic

Will-Lo:
Addresses issue #697 

Adds topic parameter in `GET /kafka_cluster_state` to filter verbose output based on topic parameter with matching regex

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/KafkaClusterStateParameters.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/KafkaClusterStateParameters.java
@@ -6,6 +6,7 @@ package com.linkedin.kafka.cruisecontrol.servlet.parameters;
 
 import com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig;
 import java.io.UnsupportedEncodingException;
+import java.util.regex.Pattern;
 import javax.servlet.http.HttpServletRequest;
 
 
@@ -19,6 +20,7 @@ import javax.servlet.http.HttpServletRequest;
  */
 public class KafkaClusterStateParameters extends AbstractParameters {
   private boolean _isVerbose;
+  private Pattern _topic;
 
   public KafkaClusterStateParameters(HttpServletRequest request, KafkaCruiseControlConfig config) {
     super(request, config);
@@ -28,9 +30,14 @@ public class KafkaClusterStateParameters extends AbstractParameters {
   protected void initParameters() throws UnsupportedEncodingException {
     super.initParameters();
     _isVerbose = ParameterUtils.isVerbose(_request);
+    _topic = ParameterUtils.topic(_request);
   }
 
   public boolean isVerbose() {
     return _isVerbose;
+  }
+
+  public Pattern topic() {
+    return _topic;
   }
 }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/KafkaClusterStateParameters.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/KafkaClusterStateParameters.java
@@ -15,7 +15,7 @@ import javax.servlet.http.HttpServletRequest;
  *
  * <pre>
  * Retrieve the kafka cluster state.
- *    GET /kafkacruisecontrol/kafka_cluster_state?verbose=[true/false]&amp;json=[true/false]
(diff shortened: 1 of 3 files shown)


kidkun:
@Will-Lo Thanks for the patch!

@efeg I feel there are some confusions in the original description in issue #697.
Looking at the following verbose response from `kafka_cluster_state` endpoint, do we want to filter the replica list only in `Other Partition` field based on this `topic` regex or we want to filter data in all fields( include `Brokers`, `Offline Partitions` etc.)? I feel previous design makes more sense.

image

efeg:
@kidkun @Will-Lo 
I was leaning towards applying the filter to _all replicas_ regardless of their current health status. In other words, the filter will be applied to each of the individual set (if relevant):

* `Offline Partitions`
* `Partitions with Offline Replicas` (relevant to `migrate_to_kafka_2_0` branch)
* `Under Replicated Partitions`
* `Under MinIsr Partitions` (relevant to `migrate_to_kafka_2_0` branch)
* `Other Partitions`

-- @Will-Lo Note that you can ignore the sets relevant to `migrate_to_kafka_2_0` branch -- we will apply the relevant changes while cherry-picking the patch.

Will-Lo:
@efeg I changed the PR to only filter the topics on the partitions you mentioned by only populating the partitions that match the topic on `populateKafkaPartitionState`, which should cover underreplicated, offline, and other partitions. Is that okay?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/KafkaClusterState.java
```diff
@@ -68,19 +71,22 @@ private String getJSONString(CruiseControlParameters parameters) {
   private void populateKafkaPartitionState(Set<PartitionInfo> underReplicatedPartitions,
                                            Set<PartitionInfo> offlinePartitions,
                                            Set<PartitionInfo> otherPartitions,
-                                           boolean verbose) {
+                                           boolean verbose,
+                                           Pattern topicPattern) {
```
kidkun: Can we also update the Javadoc for this new parameter?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/KafkaClusterState.java
```diff
@@ -155,7 +161,7 @@ private void populateKafkaBrokerState(Map<Integer, Integer> leaderCountByBrokerI
    *
    * @param verbose True if verbose, false otherwise.
    */
-  public Map<String, Object> getJsonStructure(boolean verbose) {
+  public Map<String, Object> getJsonStructure(boolean verbose, Pattern topic) {
```
kidkun: Can we also update the Javadoc for this new parameter?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/KafkaClusterState.java
```diff
@@ -52,7 +53,9 @@ public KafkaClusterState(Cluster kafkaCluster, KafkaCruiseControlConfig config)
 
   private String getJSONString(CruiseControlParameters parameters) {
     Gson gson = new Gson();
-    Map<String, Object> jsonStructure = getJsonStructure(((KafkaClusterStateParameters) parameters).isVerbose());
+    boolean isVerbose = ((KafkaClusterStateParameters) parameters).isVerbose();
```
kidkun: nit: in this method and in `getPlaintext` we do cast form `CruiseControlParameters` to `KafkaClusterStateParameters` couple of times, maybe we can save one redundant cast.

Will-Lo:
@kidkun Thanks for the feedback! I addressed the PR accordingly.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/KafkaClusterStateParameters.java
```diff
@@ -14,11 +15,12 @@
  *
  * <pre>
  * Retrieve the kafka cluster state.
- *    GET /kafkacruisecontrol/kafka_cluster_state?verbose=[true/false]&amp;json=[true/false]
+ *    GET /kafkacruisecontrol/kafka_cluster_state?verbose=[true/false]&amp;json=[true/false];topic=[topic]
```
efeg: Nit: `;topic=[topic]` -> `&amp;topic=[topic]`

Issue #435, opened 2018-12-04
Filter partition_load by given broker ids


efeg:
Add the necessary parameter to filter the response of partition_load by `brokerids`.

ian952:
Hi Efe! I can take a look at this issue.

PR #733, opened 2019-06-02
Filter partition_load by given broker ids

ian952:
Addresses issue #435 

Add a new optional `brokerid` parameter to `partition_load` endpoint to filter partitions by their broker id. When `brokerid` is not specified, the behavior is unchanged and all partitions will be returned.

I chose to name the parameter `brokerid` instead of `brokerids` to keep it consistent with other endpoints since they all use `brokerid`. 

Tested the change locally by starting 5 brokers and querying `parition_load` with and without `brokerid`. Verified that the brokers are correctly filtered when `brokerid` is specified.


ian952:
Requesting for review from @efeg and @kidkun . Thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
```diff
@@ -63,6 +65,11 @@ protected void initParameters() throws UnsupportedEncodingException {
     _entries = ParameterUtils.entries(_request);
     _minValidPartitionRatio = ParameterUtils.minValidPartitionRatio(_request);
     _allowCapacityEstimation = ParameterUtils.allowCapacityEstimation(_request);
+    try {
```
kidkun: If the provided broker id is invalid, I feel a better approach would be return an alert to user. 
In current code base, if we do not swallow the Exception here, it would be capture by the [error handling logic in `AbstractParameter` specifically for user parameter issue](https://github.com/linkedin/cruise-control/blob/cde921adb571f5c50f3afdb30770eddc29143c3b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/AbstractParameters.java#L54).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/GetClusterModelInRangeRunnable.java
```diff
@@ -35,7 +39,13 @@ protected PartitionLoadState getResult() throws Exception {
                                                                  _future.operationProgress(),
                                                                  _parameters.allowCapacityEstimation());
     int topicNameLength = clusterModel.topics().stream().mapToInt(String::length).max().orElse(20) + 5;
-    return new PartitionLoadState(clusterModel.replicasSortedByUtilization(_parameters.resource(), _parameters.wantMaxLoad()),
+    List<Partition> partitionList = clusterModel.replicasSortedByUtilization(_parameters.resource(), _parameters.wantMaxLoad());
+    if (_parameters.brokerIds() != null) {
```
kidkun: Note the cluster model generation(at line 36)is a very heavy calculation, normally we want to do some sanity check to ensure that the request is valid before calling that method.

 I think here one extra thing we need to check is the broker ids set in request are indeed valid(i.e. there are brokers with those ids in the cluster), we have a util method to do that called [`SanityCheckBrokerPresence`](https://github.com/linkedin/cruise-control/blob/cde921adb571f5c50f3afdb30770eddc29143c3b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java#L1212) for that purpose.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
ian952: Currently, the endpoint can be used without specifying the broker id. If we do surface the error to the user, I think this would break backward compatibility of the endpoint. Is that okay?

ian952:
@kidkun Sorry about the delay. I have added the broker ID sanity check as suggested and rebased on the latest master. Let me know if there is anything else!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
kidkun: I see,  so this catch makes `brokerid` optional for this endpoint. looks good.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/GetClusterModelInRangeRunnable.java
```diff
@@ -4,11 +4,15 @@
 
 package com.linkedin.kafka.cruisecontrol.async;
 
+import java.util.List;
+import java.util.stream.Collectors;
+
```
kidkun: nit: redundant new line.

ian952:
@kidkun Read for another look!

ian952:
@kidkun Sorry about the slow update. I have updated the endpoint to check if any brokers associated with the partition are in the specified broker id.

Issue #740, opened 2019-06-06
Fix NPE in TopicReplicaDistributionGoal during broker removal / fixing offline replicas


efeg:
This issue is observed if a lower priority goal attempts to move a replica or leadership between healthy brokers during a broker removal or fixing offline replicas.

```
Caused by: java.lang.NullPointerException
        at com.linkedin.kafka.cruisecontrol.analyzer.goals.TopicReplicaDistributionGoal.isReplicaCountUnderBalanceUpperLimitAfterChange(TopicReplicaDistributionGoal.java:164)
        at com.linkedin.kafka.cruisecontrol.analyzer.goals.TopicReplicaDistributionGoal.actionAcceptance(TopicReplicaDistributionGoal.java:153)
        at com.linkedin.kafka.cruisecontrol.analyzer.AnalyzerUtils.isProposalAcceptableForOptimizedGoals(AnalyzerUtils.java:109)
        at com.linkedin.kafka.cruisecontrol.analyzer.goals.AbstractGoal.maybeApplyBalancingAction(AbstractGoal.java:225)
        at com.linkedin.kafka.cruisecontrol.analyzer.goals.LeaderReplicaDistributionGoal.rebalanceByMovingLeaderReplicasIn(LeaderReplicaDistributionGoal.java:301)
        at com.linkedin.kafka.cruisecontrol.analyzer.goals.LeaderReplicaDistributionGoal.rebalanceForBroker(LeaderReplicaDistributionGoal.java:157)
        at com.linkedin.kafka.cruisecontrol.analyzer.goals.AbstractGoal.optimize(AbstractGoal.java:84)
        at com.linkedin.kafka.cruisecontrol.analyzer.GoalOptimizer.optimizations(GoalOptimizer.java:429)
        at com.linkedin.kafka.cruisecontrol.KafkaCruiseControl.getProposals(KafkaCruiseControl.java:934)
        at com.linkedin.kafka.cruisecontrol.KafkaCruiseControl.decommissionBrokers(KafkaCruiseControl.java:238)
        at com.linkedin.kafka.cruisecontrol.async.DecommissionBrokersRunnable.getResult(DecommissionBrokersRunnable.java:67)
        at com.linkedin.kafka.cruisecontrol.async.DecommissionBrokersRunnable.getResult(DecommissionBrokersRunnable.java:23)
        at com.linkedin.kafka.cruisecontrol.async.OperationRunnable.run(OperationRunnable.java:45)
[...]

PR #741, opened 2019-06-06
Fix NPE in TopicReplicaDistributionGoal during broker removal / fixing offline replicas.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/740.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/TopicReplicaDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/TopicReplicaDistributionGoal.java
@@ -58,7 +58,9 @@ public class TopicReplicaDistributionGoal extends AbstractGoal {
 
   private final Map<String, Set<Integer>> _brokerIdsAboveBalanceUpperLimitByTopic;
   private final Map<String, Set<Integer>> _brokerIdsUnderBalanceLowerLimitByTopic;
+  // Must contain only the topics to be rebalanced.
   private final Map<String, Double> _avgTopicReplicasOnAliveBroker;
+  // Must contain all topics to ensure that the lower priority goals work w/o an NPE.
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/TopicReplicaDistributionGoal.java
```diff
@@ -246,11 +248,15 @@ protected void initGoalState(ClusterModel clusterModel, OptimizationOptions opti
...
 
     // Initialize the average replicas on an alive broker.
-    for (String topic : topicsToRebalance) {
+    for (String topic : clusterModel.topics()) {
       int numTopicReplicas = clusterModel.numTopicReplicas(topic);
       _avgTopicReplicasOnAliveBroker.put(topic, (numTopicReplicas / (double) clusterModel.aliveBrokers().size()));
       _balanceUpperLimitByTopic.put(topic, balanceUpperLimit(topic, optimizationOptions));
       _balanceLowerLimitByTopic.put(topic, balanceLowerLimit(topic, optimizationOptions));
+      // Retain only the topics to rebalance in _avgTopicReplicasOnAliveBroker
+      if (!topicsToRebalance.contains(topic)) {
```
kidkun: Instead of add then remove topic from `_avgTopicReplicasOnAliveBroker`, maybe we can move the if at line 253 to not add in the first place?

efeg: We add it initially to ensure that `balanceUpperLimit` and `balanceLowerLimit` can use the corresponding average values.

kidkun: gotcha.

Issue #751, opened 2019-06-08
In-execution user task state is not properly handled by UserTaskManger.


kidkun:
`UserTaskManager` does not properly handle the state transfer of user tasks which have operations for `Executor` to execute.
In ideal world,  such task's state should go through following state:
`active` -> `in-execution` -> `completed`.

however, in current implementation, several issues observed:
1. such task's calculation result is not logged in cruise-control operation log(which is a new log introduced by [PR-596](https://github.com/linkedin/cruise-control/pull/596) to record all the operation execution.)
2. in `user_tasks` endpoint there will be two entry of the same task
3. once the `Executor` finish the execution, the task's state is not updated in `user_tasks` endpoint, i.e.  it is still marked as `InExecution`

Issue #752, opened 2019-06-08
Race condition in logging task's operation result


kidkun:
There is a race condition between `ServletSessionExecutor` thread and `UserTaskLogger` that they may try to cache the result of a completed user task(and[ abandon](https://github.com/linkedin/cruise-control/blob/db3432bcae86a4aded7d42f42046d3972d41a1a6/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/AbstractCruiseControlResponse.java#L49) the original result to reduce memory pressure) at the same time, which will cause one of the thread get `NullPointerException`.

PR #753, opened 2019-06-08
Fix the user task state bookkeeping bug in UserTaskManager.

kidkun:
Fix issue #751 and issue #752.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -482,7 +482,7 @@ public class Executor {
     private Set<Integer> _recentlyRemovedBrokers;
     private final long _executionStartMs;
     private Throwable _executionException;
-    private UserTaskManager.UserTaskInfo _userTaskInfo;
+
 
     ProposalExecutionRunnable(LoadMonitor loadMonitor, Collection<Integer> demotedBrokers, Collection<Integer> removedBrokers) {
       _loadMonitor = loadMonitor;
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -482,23 +482,16 @@ public boolean hasOngoingExecution() {
     private Set<Integer> _recentlyRemovedBrokers;
     private final long _executionStartMs;
     private Throwable _executionException;
-    private final UserTaskManager.UserTaskInfo _userTaskInfo;
+    private UserTaskManager.UserTaskInfo _userTaskInfo;
```
efeg: Looks like `_userTaskInfo` is used only in `ProposalExecutionRunnable#execute()` -- can we make it a local variable in this function rather than making it a field of the class?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -399,6 +399,9 @@ public synchronized UserTaskInfo markTaskInExecution(String uuid) {
 
     if (_uuidToActiveUserTaskInfoMap.containsKey(userTaskId)) {
       _inExecutionUserTaskInfo = _uuidToActiveUserTaskInfoMap.remove(userTaskId).setState(TaskState.IN_EXECUTION);
+      // Normally a user task's operation result is logged when the task's state is transferred from ACTIVE to COMPLETED_WITH_ERROR.
+      // If the user task's state is transferred from ACTIVE directly to IN_EXECUTION, need to log the task's operation result here.
```
efeg: Can we draw the full state machine under the JavaDoc of `TaskState`? Currently it is not clear which state is supposed to go to which final state.

Issue #736, opened 2019-06-03
Ongoing self-healing operation's associated anomaly should always be shown in anomaly detector state.


kidkun:
If `Executor` is executing the operation generated by a detected anomaly, this anomaly should be kept in-memory and its information should be returned to user via `state` endpoint.

PR #756, opened 2019-06-09
Add anomaly under self-healing information to anomaly detector state.

kidkun:
Fix issue #736


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -87,6 +104,9 @@ void onAnomalyHandle(Anomaly anomaly, AnomalyState.Status status) {
     AnomalyState recentAnomalyState = _recentAnomaliesByType.get(anomalyType).get(anomalyId);
     if (recentAnomalyState != null) {
       recentAnomalyState.setStatus(status);
+      if (status == AnomalyState.Status.FIX_STARTED) {
+        _ongoingSelfHealingAnomaly.set(_recentAnomaliesByType.get(anomalyType).remove(anomalyId));
```
efeg: What if `_recentAnomaliesByType.get(anomalyType).remove(anomalyId)` returns `null` due to updating `_recentAnomaliesByType` after https://github.com/linkedin/cruise-control/pull/756/files#diff-a05de16a44ab4fe3089e7064a288b471R104 via `addAnomalyDetection`?

Moreover, after the anomaly is added, but before it is handled it can be purged from `_recentAnomaliesByType`. Looks like we won't have information about such anomalies. Is it possible to ensure that we always have the ongoing anomaly information?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -184,25 +190,38 @@ private static void populateCommonDetails(AnomalyState anomalyState, Map<String,
   }
 
   public Map<String, Object> getJsonStructure() {
-    Map<String, Object> anomalyDetectorState = new HashMap<>(_recentAnomaliesByType.size() + 2);
+    AnomalyState anomalyUnderFixing = _ongoingSelfHealingAnomaly.get();
+    Map<String, Object> anomalyDetectorState = new HashMap<>(_recentAnomaliesByType.size() + (anomalyUnderFixing == null ? 0 : 1));
```
efeg: `anomalyUnderFixing == null ? 0 : 1`-> `anomalyUnderFixing == null ? 2 : 3`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -62,8 +66,21 @@ protected boolean removeEldestEntry(Map.Entry<String, AnomalyState> eldest) {
       });
     }
     _selfHealingEnabled = selfHealingEnabled;
+    _ongoingSelfHealingAnomaly = new AtomicReference<>();
```
efeg: Would you clarify why we use an `AtomicReference` rather than `volatile` here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -575,9 +584,12 @@ private void execute() {
         LOG.error("Executor got exception during execution", t);
         _executionException = t;
       } finally {
-        // If the finished task is triggered by user request, update task status in user task manager.
+        // If the finished task is triggered by user request, update task status in user task manager; if task is triggered
```
efeg: Nit: **is** triggered by user -> **was** triggered by **a** user

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -383,12 +383,12 @@ private synchronized void removeOldUserTasks() {
...
 
   /**
-   * Mark the user task which is currently executed in {@link com.linkedin.kafka.cruisecontrol.executor.Executor}.
+   * Update status for user task which is currently executed in {@link com.linkedin.kafka.cruisecontrol.executor.Executor}.
    *
    * @param uuid UUID associated with the in-execution user task.
    * @return {@link UserTaskInfo} associated with the task in-execution.
    */
-  public synchronized UserTaskInfo markTaskInExecution(String uuid) {
+  public synchronized UserTaskInfo taskExecutionBegan(String uuid) {
```
efeg: Nit: The name change `markTaskInExecution` -> `taskExecutionBegan` makes the functionality of the method not clear (e.g. is it a getter or setter)?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -408,11 +408,11 @@ public synchronized UserTaskInfo markTaskInExecution(String uuid) {
   }
 
   /**
-   * Unmark the in-execution user task once {@link com.linkedin.kafka.cruisecontrol.executor.Executor} has finish the execution.
+   * Update user task status once {@link com.linkedin.kafka.cruisecontrol.executor.Executor} has finished the execution.
    *
    * @param uuid UUID associated with the in-execution user task.
    */
-  public synchronized void markTaskFinishExecution(String uuid) {
+  public synchronized void taskExecutionFinished(String uuid) {
```
efeg: Nit: The name change `markTaskFinishExecution` -> `taskExecutionFinished` makes the functionality of the method not clear (e.g. is it a getter or setter)?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -307,19 +308,26 @@ private void executeAndVerifyNotification(Collection<ExecutionProposal> proposal
 
   private UserTaskManager getMockUserTaskManager() {
     UserTaskManager mockUserTaskManager = EasyMock.mock(UserTaskManager.class);
-    mockUserTaskManager.markTaskFinishExecution(anyString());
-    EasyMock.expect(mockUserTaskManager.markTaskInExecution(anyObject())).andReturn(null).anyTimes();
+    mockUserTaskManager.taskExecutionFinished(anyString());
```
efeg: Rather than passing `anyString()` can we actually verify that the `uuid` input is the one used in the test?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -307,19 +308,26 @@ private void executeAndVerifyNotification(Collection<ExecutionProposal> proposal
...
+    EasyMock.expect(mockUserTaskManager.taskExecutionBegan(anyObject())).andReturn(null).anyTimes();
     EasyMock.replay(mockUserTaskManager);
     return mockUserTaskManager;
   }
 
+  private AnomalyDetector getMockAnomalyDetector() {
+    AnomalyDetector mockAnomalyDetector = EasyMock.mock(AnomalyDetector.class);
+    mockAnomalyDetector.selfHealingFinished(anyString());
```
efeg: Can we use the `anomalyId` used in the test?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -62,8 +66,21 @@ protected boolean removeEldestEntry(Map.Entry<String, AnomalyState> eldest) {
...
+   * Update anomaly status once {@link com.linkedin.kafka.cruisecontrol.executor.Executor} has finished corresponding
+   * self-healing operation.
+   *
+   * @param anomalyId UUID of anomaly which triggered self-healing operation.
+   */
+  public void selfHealingFinished(String anomalyId) {
+    if (!anomalyId.equals(_ongoingSelfHealingAnomaly.get().anomalyId())) {
+      throw new IllegalStateException(String.format("Anomaly %s is not marked as FIX_START state in AnomalyDetector.", anomalyId));
+    }
+    _ongoingSelfHealingAnomaly.set(null);
```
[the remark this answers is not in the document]
kidkun: I think the read is not a concern here since I may get an slightly stale ongoing anomaly but I get a (snapshot-level )consistent value.

I agree that write will be a potential source of bug.My original idea is to use CAS operation of atomic variable, but I cannot do it here since we cannot do compare atomically(need to dereference and compare the anomalyId field).I fall back to `synchronized` due to 
(1) though as you mentioned, current access pattern of `onAnomalyHandle` and `selfHealingFinished` ensures there is no race condition but we cannot guarantee for future
(2)there place is not a heavily contented place, so `synchronized`  will not bring much extra load.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
kidkun: since there are two thread can update this value, I feel `volatile` will have consistency issue. I change back to `synchronized`  due to reason above.

efeg: 1. Excerpt from https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/package-summary.html
```
The memory effects for accesses and updates of atomics generally follow the rules for volatiles, as stated in The Java Language Specification (17.4 Memory Model):

get has the memory effects of reading a volatile variable.
set has the memory effects of writing (assigning) a volatile variable.
```
-- i.e. the typical benefit of `AtomicReference` is CAS. `volatile` has lower overhead; hence, should be preferred against `AtomicReference` when possible.

2. Changing to `synchronized` (from `volatile` or `AtomicReference`) to handle multiple writers seems correct solution to me.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -80,13 +98,16 @@ void addAnomalyDetection(AnomalyType anomalyType, Anomaly anomaly) {
...
+  synchronized void onAnomalyHandle(Anomaly anomaly, AnomalyState.Status status) {
     AnomalyType anomalyType = getAnomalyType(anomaly);
     String anomalyId = anomaly.anomalyId();
 
-    AnomalyState recentAnomalyState = _recentAnomaliesByType.get(anomalyType).get(anomalyId);
+    AnomalyState recentAnomalyState = _recentAnomaliesByType.get(anomalyType).remove(anomalyId);
     if (recentAnomalyState != null) {
       recentAnomalyState.setStatus(status);
+      if (status == AnomalyState.Status.FIX_STARTED) {
+        _ongoingSelfHealingAnomalyState = recentAnomalyState;
```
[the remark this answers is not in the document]
kidkun: looks like we [add the anomaly to _recentAnomaliesByType](_recentAnomaliesByType) and we [remove it from _recentAnomaliesByType and mark it as ongoing anomaly](https://github.com/linkedin/cruise-control/blob/b0c81b29e15a6006211da6d79a17846ab1395aad/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java#L280) are in the same method and it is executed in a [dedicated thread](https://github.com/linkedin/cruise-control/blob/b0c81b29e15a6006211da6d79a17846ab1395aad/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java#L158), I feel in this way as long as the `LinkedHashMap` used in  `_recentAnomaliesByType` has size larger than 1, we will not get null from this [remove](https://github.com/linkedin/cruise-control/blob/b0c81b29e15a6006211da6d79a17846ab1395aad/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java#L105).

However, I think the suggestion of switching to use  ongoing anomaly has a benefit of making response more readable. If we return  `ongoingSelfHealingAnomalyState` in `anomalyState` response, we probably return the same thing in `recentAnomaly` part, so we return the same thing twice and it could be very long since for `GoalViolation` and `BrokerFailure` it has the optimization result.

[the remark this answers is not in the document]
efeg: 1. I agree, it sounds reasonable to assume that `addAnomalyDetection()` and `onAnomalyHandle(...)` will always be handled by a single thread.
2. Switching to use ongoing anomaly sounds good to me!

One final question: Why do we remove the anomaly state from `_recentAnomaliesByType` rather than simply getting its value? -- i.e. why `_recentAnomaliesByType.get(anomalyType).remove(anomalyId);` but not `_recentAnomaliesByType.get(anomalyType).get(anomalyId);`?

I suspect that removing it from the recent anomalies will lead to loss of anomaly state upon completion of the ongoing execution (i.e. anomaly fix)?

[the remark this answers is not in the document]
kidkun: Ah, good catch. Previously since we want to keep ongoing anomaly state in `_ongoingSelfHealingAnomalyState`, it make sense to remove it from `_recentAnomaliesByType`, now we do not need to remove since we switch to keep ongoing anomaly.

Issue #760, opened 2019-06-11
NPE on startup


MattPumphrey:
[2019-06-11 14:31:46,579] INFO Kafka version : 2.0.0 (org.apache.kafka.common.utils.AppInfoParser)
[2019-06-11 14:31:46,579] INFO Kafka commitId : 3402a8361b734732 (org.apache.kafka.common.utils.AppInfoParser)
[2019-06-11 14:31:46,757] INFO Cluster ID: K48tg2h1ShW40q8-99TgvQ (org.apache.kafka.clients.Metadata)
Exception in thread "main" java.lang.NullPointerException
	at com.linkedin.kafka.cruisecontrol.monitor.sampling.CruiseControlMetricsReporterSampler.refreshPartitionAssignment(CruiseControlMetricsReporterSampler.java:186)
	at com.linkedin.kafka.cruisecontrol.monitor.sampling.CruiseControlMetricsReporterSampler.configure(CruiseControlMetricsReporterSampler.java:253)
	at com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig.getConfiguredInstance(KafkaCruiseControlConfig.java:1413)
	at com.linkedin.kafka.cruisecontrol.monitor.sampling.MetricFetcherManager.<init>(MetricFetcherManager.java:129)
	at com.linkedin.kafka.cruisecontrol.monitor.sampling.MetricFetcherManager.<init>(MetricFetcherManager.java:69)
	at com.linkedin.kafka.cruisecontrol.monitor.task.LoadMonitorTaskRunner.<init>(LoadMonitorTaskRunner.java:74)
	at com.linkedin.kafka.cruisecontrol.monitor.LoadMonitor.<init>(LoadMonitor.java:177)
	at com.linkedin.kafka.cruisecontrol.monitor.LoadMonitor.<init>(LoadMonitor.java:127)
	at com.linkedin.kafka.cruisecontrol.KafkaCruiseControl.<init>(KafkaCruiseControl.java:124)
	at com.linkedin.kafka.cruisecontrol.async.AsyncKafkaCruiseControl.<init>(AsyncKafkaCruiseControl.java:75)
	at com.linkedin.kafka.cruisecontrol.KafkaCruiseControlMain.main(KafkaCruiseControlMain.java:71)

[...]

PR #762, opened 2019-06-11
use defaults when there is no explicit user config

linehrr:
this is to fix the config loading NPE bug raised here: https://github.com/linkedin/cruise-control/issues/760
we have defined defaults under `ConfigDef` but when we create class instance by `getConfiguredInstance` we are passing the origins which does not contain defaults as fallback, therefore causing NPE.


kidkun:
@linehrr  The reason why we do not use `value()` to initiate configurable object is because the `value()` will parse whatever in `.properties` and only keeps the the configs which are defined in `KafkaCruiseControlConfig`. 

Since Cruise Control has many pluggable components and these pluggable components may their own configs which is not defined in `KafkaCruiseControlConfig`(defined in their own Class, e.g [KafkaSampleStore](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java#L112)), If we use `value()` when we are reflectively creating corresponding objects, we will miss the their specific configs set in `.properties` file.

linehrr:
@kidkun thanks for explaining, but wouldn't other modules extend `AbstractConfig` and do their own `getConfiguredInstance()` ? that way each module can define its own required configs and defaults. and have proper fallbacks?

kidkun:
@linehrr  I think that is a good idea for some pluggable component Class, actually we have done this in some of them, one example is [PercentileMetricAnomalyFinderConfig](https://github.com/linkedin/cruise-control/blob/master/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/detector/metricanomaly/PercentileMetricAnomalyFinderConfig.java) .

Basically to do that we need to have a companion config Class for each of the pluggable component Class. And since some of the pluggable component object is created ad-hoc( e.g. `Goal`, which is the goals used to calculate proposal and is created every time user send a operation  like `rebalance`) and have only one or two specific config, it would be kind troublesome to always instantiate a companion config object for their instantiation.

For some component Class, like `KafkaSampleStore` here, it probably is the better way to standardize the config default in a config Class . 
For now I have a quick fix patch for issue #760 to unblock the usability. 
Do you want to try to improve this part?

linehrr:
@kidkun I see, this makes total sense. I will revise this PR then, hopefully I will find somewhere in between so we can still have centralized config loading and instance creation but also at the same time use defaults in a better manner so we don't have to do such `null judging` boilerplate code.

linehrr:
@kidkun 
revised, please review.

linehrr:
@kidkun any thoughts on this PR?

kidkun:
@linehrr  Sorry for delay in response.

I think this is a good idea, but to make it work we have more things to do.
If you try to deploy with this patch, the deployment will fail.

The issue is that now configurable components are configured with the union of original configs(`String` -> `String`) and parsed values ((`String` -> `Boolean`/`Integer` etc, whatever we have defined in `KafkaCruiseControlConfig`)).
Since before this change, configurable components are always configured with original values(`String` -> `String`) , it is ok to first cast value to `String` type and do whatever parse within these components(one example is [CruiseControlMetricsReporterSampler](https://github.com/linkedin/cruise-control/blob/a2c0298c33c9e8532cd7963048cd41e39909fd16/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsReporterSampler.java#L208).
But now it has issue since the passed-in config value may not be `String`, casting to `String` may throw `ClassCastException` since we cannot directly cast value of type `Boolean`/`Integer` to `String`.

linehrr:
@kidkun I see now, do you think `parsedConfigs.mapValue(_ => _.toString)` would solve this? I am sorry I am not very familiar with the code base.

kidkun:
@linehrr 
Casting parsed config back to `String` definitely solve the issue, but I feel it is not that elegant, the reason is that later in those pluggable components, we need to recast back to actual value type before using it.

Another way to solve this is to change the `configure()` methods of these pluggable components to make them aware that once it use a config defined in `KafkaCruiseControlConfig`, the value of that config will be whatever type defined in `KafkaCruiseControlConfig`.

 Since the the pluggable components are scatter everywhere and it is tricky to go through all of them, I create patch to your repo, [PR-1](https://github.com/linehrr/cruise-control/pull/1), please take a look. Thanks!

linehrr:
@kidkun that's much better, so you are going to merge this into your master right?

kidkun:
@linehrr Oh, I was planning to create PR to your `NPE_on_config_loading` branch. Looks like I picked the wrong destination and the patch goes to your `master` branch. 
If the patch looks good to you, can you cherry-pick to `NPE_on_config_loading` branch, then I can merge your patch to `linkedin/master`

linehrr:
@kidkun ok done. please take a look.

kidkun:
@linehrr  looks like your local history is messed up( I see some commits from other people in this patch's commit history)
I suggest you to:
1. at local `master` branch, `git pull linkedin master`
2. at local `NPE_on_config_loading` branch, `git rebase -i HEAD~11` to squash all the commits(yours and others) currently listed in this patch to a single commit; then do `git rebase master` and resolve conflicts if any, the do a `git push --force`

linehrr:
@kidkun oh it's because I did the re-sync from upstream/master and resolved some conflicts. that's the reason why you see other ppl's commits. those are the ones got synced into this branch. they won't affect the diff for this patch, since they are actually from upstream/master itself.

linehrr:
@kidkun you know what, when I tried to create another PR from my branch, the diff showed correctly(only 5 files changed and only yours and mine commits), and it redirected me to this existing PR. so Git is considering those are the same PR. if you want I can close this PR and create another one.

Issue #768, opened 2019-06-12
Fix concurrency bug causing reporting broker failures from future


efeg:
Cruise Control Anomaly Detector occasionally reports broker failures from the year `292269055` (i.e. `two hundred ninety two million, two hundred sixty nine thousand and fifty five`) with no information about the failed broker(s) -- i.e. empty failed brokers. Naturally, this is beyond a reasonable `clock skew`, and a broker anomaly without failed broker(s) indicates a bug. In particular, it was noticed that `KafkaCruiseControlUtils.toDateString(Long.MAX_VALUE + 1800000, "dd/MM/yyyy HH:mm:ss", "UTC")` generates a date from this year.

Further investigation revealed that the root cause is the lack of synchronization for `BrokerFailureDetector#_failedBrokers`.

A side effect of this issue is if the generated broker anomaly is not ready to fix (i.e. an anomaly is not ready if it (1) does not meet completeness requirements or (2) load monitor is in an unexpected state), then a new broker failure anomaly detection will be scheduled to run after `_anomalyDetectionIntervalMs` for each such broker anomaly.

PR #769, opened 2019-06-12
Fix concurrency bug causing reporting broker failures from future.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/768.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
@@ -6,12 +6,19 @@ package com.linkedin.kafka.cruisecontrol.detector;
 
 import com.linkedin.cruisecontrol.detector.Anomaly;
 import com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyType;
+import java.util.Collections;
 
 
 /**
  * A util class for Anomaly Detectors.
  */
 public class AnomalyDetectorUtils {
+  public static final Anomaly SHUTDOWN_ANOMALY = new BrokerFailures(null,
+                                                                    null,
+                                                                    true,
+                                                                    true,
+                                                                    true,
+                                                                    Collections.emptyList());
 
   private AnomalyDetectorUtils() {
   }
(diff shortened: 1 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
```diff
@@ -6,12 +6,19 @@
...
 /**
  * A util class for Anomaly Detectors.
  */
 public class AnomalyDetectorUtils {
+  public static final Anomaly SHUTDOWN_ANOMALY = new BrokerFailures(null,
+                                                                    null,
+                                                                    true,
+                                                                    true,
+                                                                    true,
+                                                                    Collections.emptyList());
```
kidkun: Maybe also change this `Collections.emptyList()` to `null`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
```diff
@@ -88,16 +88,22 @@ void startDetection() {
...
+    if (updated) {
+      // persist the updated failed broker list.
+      persistFailedBrokerList();
+    }
     // Report the failures to anomaly detector to handle.
     reportBrokerFailures();
   }
 
-  Map<Integer, Long> failedBrokers() {
+  synchronized void detectBrokerFailures() {
```
kidkun: Is this `synchronized` necessary?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
```diff
@@ -88,16 +88,22 @@ void startDetection() {
...
+    detectBrokerFailures(aliveBrokers());
+  }
+
+  synchronized Map<Integer, Long> failedBrokers() {
```
kidkun: This method is called by `reportBrokerFailures()` , `reportBrokerFailures()` is called by `detectBrokerFailures()` which is already protected by `synchronized`, maybe this `synchronized` is not necessary?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
efeg: This ensures that if concurrent calls to `BrokerFailureDetector#detectBrokerFailures()` wait for `BrokerFailureDetector#detectBrokerFailures(Set<Integer>)`, then the call to `BrokerFailureDetector#aliveBrokers()` will be more fresh.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
efeg: `_failedBrokers` is a shared resource. While today the `failedBrokers()` function is called from within the described synchronized code path, since it is package private, it can as well be called from another code path in the future. Hence, I believe the `synchronized` is required here.

Issue #770, opened 2019-06-14
Make servlet request handling logic pluggable


efeg:
Making the servlet request handling logic pluggable enables users to customize:
* The way requests are handled on endpoints.
* The format and number of `parameters` supported by endpoints.
* The format and content of `response` of endpoints.

An example use-case of making the handling logic pluggable is being able to support functionalities provided by different flavors of Kafka -- e.g. experimental features that have not yet been merged in `Apache Kafka`.

PR #771, opened 2019-06-14
Make servlet request handling logic pluggable (PART-1).

efeg:
This patch is the PART-1 of the work towards addressing https://github.com/linkedin/cruise-control/issues/770.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/RemoveBrokerRequest.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/RemoveBrokerRequest.java
@@ -24,7 +24,7 @@ public class RemoveBrokerRequest extends AbstractAsyncRequest {
 
   @Override
   public RemoveBrokerParameters parameters() {
-    return null;
+    return _parameters;
   }
 
   @Override
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/RemoveBrokerRequest.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/RemoveBrokerRequest.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2019 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.cruisecontrol.servlet.handler.async;
+
+import com.linkedin.kafka.cruisecontrol.async.OperationFuture;
+import com.linkedin.kafka.cruisecontrol.servlet.KafkaCruiseControlServlet;
+import com.linkedin.kafka.cruisecontrol.servlet.parameters.RemoveBrokerParameters;
+
+
+public class RemoveBrokerRequest extends AbstractAsyncRequest {
+  private final RemoveBrokerParameters _parameters;
+
+  public RemoveBrokerRequest(KafkaCruiseControlServlet servlet, RemoveBrokerParameters parameters) {
+    super(servlet);
+    _parameters = parameters;
+  }
+
+  @Override
+  protected OperationFuture handle(String uuid) {
+    return _asyncKafkaCruiseControl.decommissionBrokers(_parameters, uuid);
+  }
+
+  @Override
+  public RemoveBrokerParameters parameters() {
+    return null;
+  }
+
+  @Override
+  public String name() {
+    return RemoveBrokerRequest.class.getSimpleName();
+  }
+}
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/sync/AdminRequest.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/sync/AdminRequest.java
@@ -47,14 +47,14 @@ public class AdminRequest extends AbstractSyncRequest {
       _kafkaCruiseControl.setRequestedInterBrokerPartitionMovementConcurrency(concurrentInterBrokerPartitionMovements);
       ongoingConcurrencyChangeRequest += String.format("Inter-broker partition movement concurrency is set to %d%n",
                                                        concurrentInterBrokerPartitionMovements);
-      LOG.warn("Inter-broker partition movement concurrency is set to: {} by user.", concurrentInterBrokerPartitionMovements);
+      LOG.info("Inter-broker partition movement concurrency is set to: {} by user.", concurrentInterBrokerPartitionMovements);
     }
     // 1.2. Change leadership concurrency.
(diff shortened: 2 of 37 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ProgressResult.java
```diff
@@ -0,0 +1,62 @@
...
+    for (OperationFuture operationFuture: _futures) {
+      sb.append(String.format("%s:%n%s", operationFuture.operation(), operationFuture.progressString()));
+    }
+    return sb.toString();
+  }
+
+  @Override
+  protected void discardIrrelevantAndCacheRelevant(CruiseControlParameters parameters) {
+    // Cache relevant response.
+    _cachedResponse = parameters.json() ? getJSONString() : getPlaintext();
```
kidkun: do we need to cache response for in-progress result, looks this is a one-time thing.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/RemoveBrokerRequest.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  @Override
+  protected OperationFuture handle(String uuid) {
+    return _asyncKafkaCruiseControl.decommissionBrokers(_parameters, uuid);
+  }
+
+  @Override
+  public RemoveBrokerParameters parameters() {
+    return null;
```
kidkun: `null` -> `_parameters`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/sync/AdminRequest.java
```diff
@@ -0,0 +1,141 @@
...
+      ongoingConcurrencyChangeRequest += String.format("Inter-broker partition movement concurrency is set to %d%n",
+                                                       concurrentInterBrokerPartitionMovements);
+      LOG.warn("Inter-broker partition movement concurrency is set to: {} by user.", concurrentInterBrokerPartitionMovements);
+    }
+    // 1.2. Change leadership concurrency.
+    Integer concurrentLeaderMovements = _parameters.concurrentLeaderMovements();
+    if (concurrentLeaderMovements != null) {
+      _kafkaCruiseControl.setRequestedLeadershipMovementConcurrency(concurrentLeaderMovements);
+      ongoingConcurrencyChangeRequest += String.format("Leadership movement concurrency is set to %d%n", concurrentLeaderMovements);
+      LOG.warn("Leadership movement concurrency is set to: {} by user.", concurrentLeaderMovements);
```
kidkun: `warn` -> `info`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/sync/AdminRequest.java
```diff
@@ -0,0 +1,141 @@
...
+        sb.append(String.format("Dropped recently removed brokers (requested: %s after-dropping: %s).%n",
+                                brokersToDropFromRecentlyRemoved, recentlyRemovedBrokers));
+        LOG.warn("Recently removed brokers are dropped by user (requested: {} after-dropping: {}).",
+                 brokersToDropFromRecentlyRemoved, recentlyRemovedBrokers);
+      }
+    }
+
+    Set<Integer> brokersToDropFromRecentlyDemoted = _parameters.dropRecentlyDemotedBrokers();
+    if (!brokersToDropFromRecentlyDemoted.isEmpty()) {
+      if (!_kafkaCruiseControl.dropRecentBrokers(brokersToDropFromRecentlyRemoved, false)) {
```
kidkun: `brokersToDropFromRecentlyRemoved` -> `brokersToDropFromRecentlyRemoved`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/sync/AdminRequest.java
efeg: All admin request logs were intentionally set to `warn` level as bad `admin` requests have the power to overload / kill / harm the Kafka cluster.

-- That being said, I don't have a strong opinion on this. Do you think we should change all / some logs in `Admin` request?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ProgressResult.java
efeg: 1. `AbstractAsyncRequest#getResponse(HttpServletRequest, HttpServletResponse)` returns `ProgressResult`.
2. Then `AbstractRequest#handle(HttpServletRequest, HttpServletResponse)` calls `CruiseControlResponse#writeSuccessResponse(CruiseControlParameters, HttpServletResponse)` on the response.
3. Then `AbstractCruiseControlResponse#writeSuccessResponse(CruiseControlParameters, HttpServletResponse)` uses the `_cachedResponse`.

Note that after returning the response, it is not stored in memory. Am I missing something?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/sync/AdminRequest.java
efeg: Ah yes it should be `...` -> `brokersToDropFromRecentlyDemoted`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ProgressResult.java
kidkun: Yeah, since it is thrown anyway, I think this really does not matter much.
My original opinion is that since `ProgressResult#writeSuccessResponse(CruiseControlParameters, HttpServletResponse)` runs once anyway, there is no need to cache the response string in memory since( they will not be retrieved)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/sync/AdminRequest.java
kidkun: My experience in debugging is that when I look for suspicious things in log, I always grep for `WARN`/`ERROR`/`Exception` first. I feel `WARN` still means things are bad(to some extent, but not as bad as `Error`...).

Maybe we can keep `None of the user-requested brokers to drop ({}) are in the recently demoted broker set ({}).` as `warn` and change other to `info`?

efeg: Sure, that makes sense -- I will make the changes.

Issue #759, opened 2019-06-11
add API to get long-term running average of metrics


radai-rosenblatt:
it would be useful to have a long term average for the size and msg/byte rates of topic partitions.
a 7-day average would be ideal for neutralizing diurnal and weekly cycle effects.

PR #777, opened 2019-06-18
Add explicit parameter to force retrieval of avg utilization of partitions

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/759 by adding `avg_load` parameter to `partition_load` endpoint.

-- Note that to get the average over a given time frame (e.g. the past 7 days), users can set `start` and `end` parameters with the relevant timestamps, accordingly.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
```diff
@@ -53,6 +54,10 @@ protected void initParameters() throws UnsupportedEncodingException {
     }
 
     _wantMaxLoad = ParameterUtils.wantMaxLoad(_request);
+    _wantAvgLoad = ParameterUtils.wantAvgLoad(_request);
+    if (_wantMaxLoad && _wantAvgLoad) {
+      throw new UserRequestException("Configurations to ask for max and avg load are mutually exclusive to each other.");
```
kidkun: nit : This is `parameter` rather than `configuration`?

Issue #723, opened 2019-05-21
Add metrics to evaluate the impact of self-healing


efeg:
Add the following metrics (to the `anomaly_detector` substate) to enrich the understanding regarding the impact of enabling self-healing process -- e.g. reacts faster, stays healthy for a longer period of time.

`MTBA`: Mean time between anomalies for a specific anomaly type.
```
MTBA = (total-operating-time-of-CC)/(number-of-anomalies-of-certain-type)
```

`MTTR`: Mean time to take action to fix an anomaly type.
```
MTTR = (time-during-which-an-anomaly-is-ongoing)/(number-of relevant-admin-op or self-healing-action-to-fix-anomaly)
```

`Self-healing ratio`: The ratio for which the self-healing is enabled for a particular anomaly type.
```
Self-healing ratio = (duration-self-healing-is-enabled) / (total-operating-time-of-CC)
```

efeg:
Looks like it makes more sense to have a single `MTTR` (i.e. not per anomaly type) to reflect the time until taking any action that starts a proposal execution to address the ongoing anomalies. If the action is taken by anomaly detector, it addresses all hanging anomalies, if the user starts the action then s/he is expected/assumed to take corresponding actions to fix ongoing issues. Hence:
```
Mean time to start fix: Mean time between the detection of earliest ongoing anomaly and start of a proposal execution. This metric covers executions started both by anomaly detector and as a result of on-demand requests.
```

PR #786, opened 2019-06-20
Add self healing enabled ratio metric under anomaly_detector substate of state response.

efeg:
Adds `self healing enabled ratio metric` specified at issue https://github.com/linkedin/cruise-control/issues/723.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
@@ -67,7 +67,7 @@ public class AnomalyDetectorState {
     _selfHealingEnabledRatio = null;
   }
 
-  void setSelfHealingEnabledRatio(Map<AnomalyType, Float> selfHealingEnabledRatio) {
+  public void setSelfHealingEnabledRatio(Map<AnomalyType, Float> selfHealingEnabledRatio) {
     if (selfHealingEnabledRatio == null) {
       throw new IllegalArgumentException("Attempt to set selfHealingEnabledRatio with null.");
     }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
@@ -39,6 +39,7 @@ public class AnomalyDetectorState {
   private static final String DESCRIPTION = "description";
   private static final String SELF_HEALING_ENABLED = "selfHealingEnabled";
   private static final String SELF_HEALING_DISABLED = "selfHealingDisabled";
+  private static final String SELF_HEALING_ENABLED_RATIO = "selfHealingEnabledRatio";
   private static final String RECENT_GOAL_VIOLATIONS = "recentGoalViolations";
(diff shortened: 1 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -62,6 +64,19 @@ protected boolean removeEldestEntry(Map.Entry<String, AnomalyState> eldest) {
...
+  void setSelfHealingEnabledRatio(Map<AnomalyType, Float> selfHealingEnabledRatio) {
+    if (selfHealingEnabledRatio == null) {
+      throw new IllegalArgumentException("Attempt to set selfHealingEnabledRatio with null.");
+    }
+    _selfHealingEnabledRatio = new HashMap<>(selfHealingEnabledRatio.size());
+    selfHealingEnabledRatio.forEach((key, value) -> _selfHealingEnabledRatio.put(key.name(), value));
+  }
+
+  private Map<String, Float> selfHealingEnabledRatio() {
+    return _selfHealingEnabledRatio == null ? Collections.emptyMap() : _selfHealingEnabledRatio;
```
kidkun: looks like `_selfHealingEnabledRatio == null` will always be `false` since AnomalyDetector [always set ratio before turn AnomalyDetectorState](https://github.com/linkedin/cruise-control/blob/6d8d98f90dab4ec473bc6b8f868b4d1b8bceb340/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java#L187)?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/notifier/AnomalyNotifier.java
```diff
@@ -57,4 +57,17 @@
...
+   *
+   * @return The ratio during which the self-healing is enabled over the total operating time for each anomaly type.
+   */
+  Map<AnomalyType, Float> selfHealingEnabledRatio();
+
+  /**
+   * @param nowMs Current time in ms.
+   * @return Uptime until now in ms.
+   */
+  long uptimeMs(long nowMs);
```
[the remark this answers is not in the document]
efeg: The function intentionally returns uptime until the given `nowMs`, which is the current time at the beginning of some process.

Passing the time as an argument rather than computing it in the function enables users to provide a deterministic computation (i.e. not a couple `ms` off) in the calling logic. Please see `SelfHealingNotifier#selfHealingEnabledRatio(...)` for an instance of this use-case, where `nowMs` is used in a loop with a common current time and `uptime` based on this current time returned by this function.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
efeg: I think we should be careful to distinguish (1) the existing implementation (how the functions are currently accessed) and (2) supported use-cases (how the functions can be accessed) -- i.e. please see [defensive programming](https://en.wikipedia.org/wiki/Defensive_programming).

This check provides protection against an `NPE` when/if there is an attempt to access the public interfaces (1) `getJsonStructure()` or (2) `toString()` without setting self healing enabled ratio. A potential improvement that I see is to make `setSelfHealingEnabledRatio` public rather than package private to provide it with the same access constraints as  `getJsonStructure()` and `toString()`.

Issue #713, opened 2019-05-15
Cruise Control use goals to determine the location of new replicas when increasing topic's replication factor


kidkun:
As part of [feature request 590](https://github.com/linkedin/cruise-control/issues/590) and follow up of [PR-710](https://github.com/linkedin/cruise-control/pull/710), Cruise Control should be smarter in allocating new replicas when it is trying to increase the topic's replication factor. 

It can take all or part of supported goals into consideration when determining the location of new replicas. This will avoid new replica generation violates some goals and trigger a redundant rebalance operation.

Issue #722, opened 2019-05-21
Size of data to move is not properly populated for operations generated for topic replication factor change


kidkun:
There is a known issue for [PR-710](https://github.com/linkedin/cruise-control/pull/710) that Cruise Control is unable to determine how many data will be moved for the proposals it generates( and passes to `Executor`) to change topic replication factor. 

So when these operations are executed by `Executor`, in `state` endpoint user will see message like `xxx inter-broker replicas movement are ongoing and 0MB of data will be move`, where xxx is accurate value.

This issue is expected to be address together with [issue 713](https://github.com/linkedin/cruise-control/issues/713) when we generate a cluster model and use goals to determine the operation.

PR #789, opened 2019-06-21
Honor goals when increasing topic replication factor.

kidkun:
Fix issue #713  and issue #722

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -22,6 +22,7 @@ import com.linkedin.kafka.cruisecontrol.executor.ExecutorState;
 import com.linkedin.kafka.cruisecontrol.executor.strategy.ReplicaMovementStrategy;
 import com.linkedin.kafka.cruisecontrol.model.Broker;
 import com.linkedin.kafka.cruisecontrol.model.ClusterModel;
+import com.linkedin.kafka.cruisecontrol.model.Load;
 import com.linkedin.kafka.cruisecontrol.model.ModelParameters;
 import com.linkedin.kafka.cruisecontrol.model.ModelUtils;
 import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
@@ -29,6 +30,7 @@ import com.linkedin.kafka.cruisecontrol.monitor.LoadMonitor;
 import com.linkedin.kafka.cruisecontrol.monitor.MonitorUtils;
 import com.linkedin.kafka.cruisecontrol.monitor.metricdefinition.KafkaMetricDef;
 import com.linkedin.kafka.cruisecontrol.servlet.UserTaskManager;
+import com.linkedin.kafka.cruisecontrol.servlet.parameters.ParameterUtils;
 import com.linkedin.kafka.cruisecontrol.servlet.response.CruiseControlState;
 import com.linkedin.kafka.cruisecontrol.servlet.response.stats.BrokerStats;
 import java.io.InputStream;
@@ -57,8 +59,8 @@ import org.slf4j.LoggerFactory;
 
 import static com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils.isKafkaAssignerMode;
 import static com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils.sanityCheckNonExistingGoal;
+import static com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils.sanityCheckNoOfflineReplica;
 import static com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils.shouldRefreshClusterAndGeneration;
-import static com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils.offlineReplicasForPartition;
 import static com.linkedin.kafka.cruisecontrol.servlet.response.CruiseControlState.SubState.*;
 
 
@@ -180,7 +182,7 @@ public class KafkaCruiseControl {
    *                                      these brokers (if empty, no explicit filter is enforced -- cannot be null).
    * @return The optimization result.
    *
-   * @throws KafkaCruiseControlException when any exception occurred during the decommission process.
+   * @throws KafkaCruiseControlException When any exception occurred during the decommission process.
    */
   public GoalOptimizer.OptimizerResult decommissionBrokers(Set<Integer> removedBrokers,
                                                            boolean dryRun,
@@ -419,6 +421,7 @@ public class KafkaCruiseControl {
    * @param uuid UUID of the execution.
    * @param excludeRecentlyDemotedBrokers Exclude recently demoted brokers from proposal generation for leadership transfer.
    * @return the optimization result.
+   * @throws KafkaCruiseControlException When any exception occurred during the broker demotion.
    */
   public GoalOptimizer.OptimizerResult demoteBrokers(Set<Integer> brokerIds,
                                                      boolean dryRun,
@@ -478,7 +481,7 @@ public class KafkaCruiseControl {
    * @param operationProgress the progress of the job to report.
    * @param allowCapacityEstimation Allow capacity estimation in cluster model if the requested broker capacity is unavailable.
    * @return the cluster workload model.
-   * @throws KafkaCruiseControlException when the cluster model generation encounter errors.
+   * @throws KafkaCruiseControlException When the cluster model generation encounter errors.
    */
   public ClusterModel clusterModel(long now,
                                    ModelCompletenessRequirements requirements,
@@ -504,7 +507,7 @@ public class KafkaCruiseControl {
    * @param operationProgress the progress of the job to report.
    * @param allowCapacityEstimation Allow capacity estimation in cluster model if the requested broker capacity is unavailable.
    * @return the cluster workload model.
-   * @throws KafkaCruiseControlException when the cluster model generation encounter errors.
+   * @throws KafkaCruiseControlException When the cluster model generation encounter errors.
    */
   public ClusterModel clusterModel(long from,
                                    long to,
@@ -694,7 +697,7 @@ public class KafkaCruiseControl {
    * @param requestedDestinationBrokerIds Explicitly requested destination broker Ids to limit the replica movement to
    *                                      these brokers (if empty, no explicit filter is enforced -- cannot be null).
    * @return The optimization result.
-   * @throws KafkaCruiseControlException
+   * @throws KafkaCruiseControlException If anything goes wrong in optimization proposal calculation.
    */
   public GoalOptimizer.OptimizerResult getProposals(List<String> goals,
                                                     ModelCompletenessRequirements requirements,
@@ -760,11 +763,7 @@ public class KafkaCruiseControl {
       sanityCheckBrokerPresence(requestedDestinationBrokerIds);
     }
     synchronized (this) {
-      ExecutorState executorState = null;
-      if (excludeRecentlyDemotedBrokers || excludeRecentlyRemovedBrokers) {
-        executorState = state(operationProgress, Collections.singleton(EXECUTOR)).executorState();
-      }
-
+      ExecutorState executorState = _executor.state();
       Set<Integer> excludedBrokersForLeadership = excludeRecentlyDemotedBrokers ? executorState.recentlyDemotedBrokers()
                                                                                 : Collections.emptySet();
 
@@ -778,7 +777,9 @@ public class KafkaCruiseControl {
                                           excludedBrokersForLeadership,
                                           excludedBrokersForReplicaMove,
                                           isTriggeredByGoalViolation,
-                                          requestedDestinationBrokerIds);
+                                          requestedDestinationBrokerIds,
+                                          null,
+                                          false);
     }
   }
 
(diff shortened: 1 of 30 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -886,33 +886,83 @@ public synchronized void userTriggeredStopExecution() {
...
+                                                                boolean allowCapacityEstimation,
+                                                                Integer concurrentInterBrokerPartitionMovements,
+                                                                Integer concurrentLeaderMovements,
+                                                                boolean skipHardGoalCheck,
+                                                                ReplicaMovementStrategy replicaMovementStrategy,
+                                                                boolean excludeRecentlyDemotedBrokers,
+                                                                boolean excludeRecentlyRemovedBrokers,
+                                                                boolean dryRun,
+                                                                String uuid)
+      throws KafkaCruiseControlException {
```
efeg: Nit: Is it possible to move some logic from within this function to smaller function(s)? This function seems quite monolithic.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -774,7 +771,10 @@ private boolean ignoreProposalCache(List<String> goals,
                                           excludedBrokersForLeadership,
                                           excludedBrokersForReplicaMove,
                                           isTriggeredByGoalViolation,
-                                          requestedDestinationBrokerIds);
+                                          requestedDestinationBrokerIds,
+                                          null,
+                                          null,
+                                          false);
```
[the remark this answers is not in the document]
kidkun: I agree, I will create a new issue #795  to track it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/AnalyzerUtils.java
```diff
@@ -48,21 +48,26 @@ private AnalyzerUtils() {
...
    * @param optimizedClusterModel The optimized cluster model.
+   * @param skipReplicationFactorChangeCheck Whether skip sanity check of topic partition's replication factor change before
+   *                                         and after optimization.
    * @return The diff represented by the set of balancing proposals to move from initial to final distribution.
    */
   public static Set<ExecutionProposal> getDiff(Map<TopicPartition, List<Integer>> initialReplicaDistribution,
                                                Map<TopicPartition, Integer> initialLeaderDistribution,
-                                               ClusterModel optimizedClusterModel) {
+                                               ClusterModel optimizedClusterModel,
+                                               boolean skipReplicationFactorChangeCheck) {
```
efeg: Nit: It seems `skipReplicationFactorChangeCheck` is irrelevant to the majority of the existing users of this function. Can we keep the old one and create a new method with this extra parameter to be used by only the relevant callers?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -376,6 +379,9 @@ public OptimizerResult optimizations(ClusterModel clusterModel,
    * @param isTriggeredByGoalViolation True if optimization of goals is triggered by goal violation, false otherwise.
    * @param requestedDestinationBrokerIds Explicitly requested destination broker Ids to limit the replica movement to
    *                                      these brokers (if empty, no explicit filter is enforced -- cannot be null).
+   * @param initReplicaDistributionForProposalGeneration The initial replica distribution of the cluster.
+   * @param initLeaderDistributionForProposalGeneration The initial leader replica distribution of the cluster.
```
efeg: `initReplicaDistributionForProposalGeneration` and `initLeaderDistributionForProposalGeneration` need more explanation in JavaDoc. It should answer questions such as when is it relevant and what its value should be when it is not relevant (i.e. `null`)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -376,6 +379,9 @@ public OptimizerResult optimizations(ClusterModel clusterModel,
...
+   * @param onlyMoveImmigrantReplicas Whether restrict replica movement only to immigrant replicas or not.
```
efeg: Do we need `onlyMoveImmigrantReplicas`?
Shouldn't it be inferred from the fact that wither `initReplicaDistributionForProposalGeneration != null` or `initLeaderDistributionForProposalGeneration != null`? Are there cases that conflict with this assumption?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -612,6 +629,12 @@ public boolean isCapacityEstimated() {
       return _optimizationOptions.excludedBrokersForReplicaMove();
     }
 
+    public Set<String> proposalTopics() {
```
efeg: Missing JavaDoc for public function.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -449,7 +460,13 @@ public OptimizerResult optimizations(ClusterModel clusterModel,
       LOG.trace("Broker level stats after optimization: {}%n", clusterModel.brokerStats(null));
     }
 
-    Set<ExecutionProposal> proposals = AnalyzerUtils.getDiff(initReplicaDistribution, initLeaderDistribution, clusterModel);
+    Set<ExecutionProposal> proposals =
+        AnalyzerUtils.getDiff(initReplicaDistributionForProposalGeneration != null ? initReplicaDistributionForProposalGeneration
```
efeg: The difference between `initReplicaDistributionForProposalGeneration` and `initReplicaDistribution` is confusing. Can we name them more clear (ideally the parameter name)?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationOptions.java
```diff
@@ -17,6 +17,8 @@
   private final Set<Integer> _excludedBrokersForReplicaMove;
   private final boolean _isTriggeredByGoalViolation;
   private final Set<Integer> _requestedDestinationBrokerIds;
+  // _onlyMoveImmigrantReplicas is set to true only when modifying topic's replication factor.
```
efeg: If that is the case, should we call this variable `isChangingReplicationFactor` and then explain what changing replication factor implies?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
```diff
@@ -173,10 +173,13 @@ protected void initGoalState(ClusterModel clusterModel, OptimizationOptions opti
                         _balancingConstraint.capacityThreshold(resource())));
     }
     clusterModel.trackSortedReplicas(sortName(),
+                                     optimizationOptions.onlyMoveImmigrantReplicas() ? ReplicaSortFunctionFactory.selectImmigrants() :
+                                                                                       null,
```
efeg: This change requires knowledge about what it means to set the value of the selection function to `null`, which can only be clear when the code is traced until `SortedReplicas#add(Replica)` -- i.e. note that the JavaDoc of `trackSortedReplicas(String, Function<Replica, Boolean>, Function<Replica, Integer>, Function<Replica, Double>)` does not specify what it means to pass `null`.

Can we either (1) keep the old way unless it is an RF increase, or (2) update the relevant JavaDoc for the expected behavior when `null` is passed?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -889,33 +889,141 @@ public synchronized void userTriggeredStopExecution() {
...
+                                               initLeaderDistribution,
+                                               true);
+      }
+    } catch (KafkaCruiseControlException kcce) {
+      throw kcce;
+    } catch (Exception e) {
+      throw new KafkaCruiseControlException(e);
+    }
+
+    if (!dryRun) {
```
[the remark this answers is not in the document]
kidkun: Good catch!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
kidkun: Currently the assumption hold, but in the future we may have another scenario that needs to specify `onlyMoveImmigrantReplicas` to be `true`.
Fundamentally since `initReplicaDistributionForProposalGeneration`/`initLeaderDistributionForProposalGeneration` and `onlyMoveImmigrantReplicas` serving two different purpose(one is for proposal diff, one is an requirement for proposal calculation), I prefer not tightly coupled them.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationOptions.java
kidkun: My concerns is this may be changed, i.e some other operations we want to add have the same restriction.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
kidkun: Add more comments in JavaDoc to explain what it is used for.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
kidkun: Update the Javadoc of `trackSortedReplicas(String, Function<Replica, Boolean>, Function<Replica, Integer>, Function<Replica, Double>)`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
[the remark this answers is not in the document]
efeg: Thanks for creating the issue.
I agree that the cached proposal computation requires special attention -- handling this at a separate issue sounds good to me!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationOptions.java
efeg: Then can we move this assertion from here to `KafkaCruiseControl#updateTopicConfiguration(...)`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -449,7 +462,12 @@ public OptimizerResult optimizations(ClusterModel clusterModel,
...
+                                                                                   : initReplicaDistribution,
+                              initLeaderDistribution,
+                              clusterModel,
+                              true);
```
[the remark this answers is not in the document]
kidkun: we already checked [here](https://github.com/linkedin/cruise-control/blob/ef195916f9d90f3dde428253ce03641f497ff03c/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java#L443) that for each goal, we do not mess up replication factor, so I think here we can safely skip the sanity check.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -714,6 +716,57 @@ public Replica createReplica(String rackId, int brokerId, TopicPartition tp, int
...
+    // Update partition info.
+    Partition partition = _partitionsByTopicPartition.get(topicPartition);
+    partition.deleteReplica(brokerId);
+    _replicationFactorByTopic.put(topicPartition.topic(), partition.replicas().size());
+  }
+
+  /**
+   * Refresh the maximal topic replication factor statistic.
+   */
+  public void refreshClusterMaxReplicationFactor() {
```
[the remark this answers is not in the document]
kidkun: I have considered that way. There is a corner case to consider. 
say we are deleting a replica to make  topic partition's RF from 4 to 3, and the old `_maxReplicationFactor` is 4. In this case, we do not know we should update `_maxReplicationFactor` to 3 or not since we do not know if there are other topic partition still have RF=4. so if we want to do it each time we are deleting a replica, we need to run this heavy global scan.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -714,6 +716,57 @@ public Replica createReplica(String rackId, int brokerId, TopicPartition tp, int
...
+    partition.addFollower(replica, partition.followers().size());
+
+    // Update replication factor statistics in cluster.
+    int replicationFactor = Math.max(_replicationFactorByTopic.get(topicPartition.topic()), partition.replicas().size());
+    _replicationFactorByTopic.put(topicPartition.topic(), replicationFactor);
+    _maxReplicationFactor = Math.max(_maxReplicationFactor, replicationFactor);
+
+    // Populate the load for imaginary replica.
+    replica.setMetricValues(partition.leader().load().loadByWindows(), partition.leader().load().windows());
+    replica.makeFollower(true);
```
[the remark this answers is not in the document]
kidkun: This method is used to modify the load for this new replica, in the first line we clone the load from the leader replica(and we need to clone from leader replica to cover the case of increasing RF from 1 to 2), so we need to modify the load to reflect follower replica. 
Maybe I should not modify the existing method which originally serves different purpose here, I will create a dedicated method to get follower load from leader.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/common/DeterministicCluster.java
```diff
@@ -268,16 +268,16 @@ public static ClusterModel smallClusterModel(Map<Resource, Double> brokerCapacit
 
     // Create snapshots and push them to the cluster.
     List<Long> windows = Collections.singletonList(1L);
-    cluster.setReplicaLoad(RACK_BY_BROKER.get(0).toString(), 0, pInfoT10, createLoad(100.0, 100.0, 130.0, 75.0), windows);
+    cluster.setReplicaLoad(RACK_BY_BROKER.get(0).toString(), 0, pInfoT10, createLoad(20.0, 100.0, 130.0, 75.0), windows);
```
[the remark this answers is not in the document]
kidkun: When I was writing new test, I found that in this model the replica's CPU utilization is too high and CpuCapacityGoal can hardly be used(e.g. one leader replica consume 100 cpu utilization and already exceed the hard limit of 80, which cause CpuCapacityGoal throw exception), I decrease the replica cpu utilization to make the scenario more practical.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ExcludedBrokersForLeadershipTest.java
```diff
@@ -90,11 +86,23 @@
...
+    // ============ReplicaCapacityGoal============
+    // Test: With single excluded broker, satisfiable cluster, no dead brokers (No exception, No proposal
+    // for excluded broker, Expected to look optimized)
+    p.add(params(0, ReplicaCapacityGoal.class, excludeB1, null, unbalanced(), noDeadBroker, true));
+    // Test: With single excluded broker, satisfiable cluster, one dead brokers (No exception, No proposal
+    // for excluded broker, Expected to look optimized)
+    p.add(params(1, ReplicaCapacityGoal.class, excludeB1, null, unbalanced(), deadBroker0, true));
+    // Test: With all brokers excluded, no dead brokers, not satisfiable (No exception, No proposal
+    // for excluded broker, Expected to look optimized)
+    p.add(params(2, ReplicaCapacityGoal.class, excludeAllBrokers, null, unbalanced(), noDeadBroker, true));
```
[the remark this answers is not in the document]
kidkun: This is similar with the case of CpuCapacityGoal, since we originally set the per-broker replica count to 1, this goal can hardly be satisfied and always throw exception.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -889,33 +889,135 @@ public synchronized void userTriggeredStopExecution() {
...
+   *     Further optimize new replica's location with provided {@link Goal} list.
+   *   </li>
+   * </ol>
+   *
    * If partition's current replication factor is larger than target replication factor, remove one or more follower replicas
-   * from the partition.
+   * from the partition. Replicas are removed following the reverse order of position in partition's replica list.
    *
-   * @param cluster The metadata of the cluster.
+   * @param topicPattern The name pattern of topics to apply the change.
```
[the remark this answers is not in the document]
kidkun: I think we should terminate early in this case.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -714,6 +716,57 @@ public Replica createReplica(String rackId, int brokerId, TopicPartition tp, int
...
+  /**
+   * Add an imaginary replica to the cluster.The replica does not correspond to any existing replica in the cluster, but
+   * a replica we are going to add to the cluster. The imaginary replica's original broker will not be any existing broker
+   * so that it will be treated as an immigrant replica for whatever broker it is assigned to and grant goals greatest
+   * freedom to allocate to an existing broker.
+   *
+   * @param topicPartition The topic partition of the imaginary replica.
+   * @param brokerId       Id of broker which the imaginary replica is assigned to.
+   */
+  public void addReplica(TopicPartition topicPartition, int brokerId) {
```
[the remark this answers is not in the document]
kidkun: I think I already added the replica to `rack`/`host` [here](https://github.com/linkedin/cruise-control/blob/ef195916f9d90f3dde428253ce03641f497ff03c/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java#L290). Note `rack.addReplica()` will call `host.addReplica()` and `host.addReplica()` will call  `broker. addReplica()`, in this way we add the load to all the place.

The reason why I do not modify `createReplica` to support the usage here is because here we need the logic from both [createReplica](createReplica) and [setReplicaLoad](setReplicaLoad), maybe we can first merge the logic of these two method, then modify the merged to method to support our use case here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Replica.java
```diff
@@ -124,11 +131,12 @@ void clearLoad() {
    * (2) Clear and get the outbound network load associated with leadership from the given replica.
    * (3) Clear and get the CPU leadership load associated with leadership from the given replica.
    *
+   * @param skipLeadershipCheck Whether check replica currently is leader replica or not before applying the change.
    * @return Removed leadership load by windows -- i.e. outbound network and fraction of CPU load by windows.
    */
-  AggregatedMetricValues makeFollower() {
+  AggregatedMetricValues makeFollower(boolean skipLeadershipCheck) {
```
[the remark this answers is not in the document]
kidkun: I will create a dedicated method to serve my purpose.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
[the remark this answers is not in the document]
efeg: Got it!
Nit: Can we add this explanation as a comment?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareGoal.java
```diff
@@ -235,11 +235,13 @@ protected void rebalanceForBroker(Broker broker,
...
     LOG.debug("balancing broker {}, optimized goals = {}", broker, optimizedGoals);
     Set<String> excludedTopics = optimizationOptions.excludedTopics();
+    boolean onlyMoveImmigrantReplicas = optimizationOptions.onlyMoveImmigrantReplicas();
     // Satisfy rack awareness requirement.
     SortedSet<Replica> replicas = new TreeSet<>(broker.replicas());
     for (Replica replica : replicas) {
       if ((broker.isAlive() && satisfiedRackAwareness(replica, clusterModel))
-          || shouldExclude(replica, excludedTopics)) {
+          || shouldExclude(replica, excludedTopics)
+          || (onlyMoveImmigrantReplicas && !replica.isImmigrant())) {
```
efeg: If a cluster is already violating a hard goal (e.g. rack awareness here), will this not allow movement of replicas between alive brokers?

`OPTION-1`: Should hard goals still allow replica moves across alive brokers (but as usual, they should prioritize immigrant replicas)?

`OPTION-2`: Otherwise, can we move `(onlyMoveImmigrantReplicas && !replica.isImmigrant())` logic inside `shouldExclude(...)`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
```diff
@@ -166,8 +166,8 @@ protected void rebalanceForBroker(Broker broker,
     } else if (!clusterModel.newBrokers().isEmpty() && requireMoreReplicas && !broker.isNew()) {
       // return if we have new brokers and the current broker is not a new broker but requires more load.
       return;
-    } else if (!clusterModel.deadBrokers().isEmpty() && requireLessReplicas && broker.isAlive()
-        && broker.immigrantReplicas().isEmpty()) {
+    } else if (((!clusterModel.deadBrokers().isEmpty() && broker.isAlive()) || optimizationOptions.onlyMoveImmigrantReplicas())
+               && requireLessReplicas && broker.immigrantReplicas().isEmpty()) {
```
efeg: The comment in `else if` seems outdated.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
```diff
@@ -204,8 +204,9 @@ private boolean rebalanceByMovingReplicasOut(Broker broker,
 
     // Get the replicas to rebalance. Replicas are sorted from smallest to largest disk usage.
     List<Replica> replicasToMove = broker.trackedSortedReplicas(name()).sortedReplicas();
-    // If cluster has dead brokers, but this broker is alive, then limit moving replica to immigrant replica.
-    if (!clusterModel.deadBrokers().isEmpty() && broker.isAlive()) {
+    // If (1) cluster has dead brokers, but this broker is alive, or (2) optimization option requests so,
+    // then limit moving replica to immigrant replica.
+    if ((!clusterModel.deadBrokers().isEmpty() && broker.isAlive()) || optimizationOptions.onlyMoveImmigrantReplicas()) {
```
efeg: Should `initGoalState` add the following `selectionFunc` as parameter as in capacity and resource distribution goals?

```
optimizationOptions.onlyMoveImmigrantReplicas() ? ReplicaSortFunctionFactory.selectImmigrants() : null,
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
efeg: Nit (code style): Can we move `:` to the line that `null` resides in?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/TopicReplicaDistributionGoal.java
```diff
@@ -425,7 +427,7 @@ private boolean rebalanceByMovingReplicasOut(Broker broker,
 
     int numReplicasOfTopicInBroker = broker.numReplicasOfTopicInBroker(topic);
 
-    for (Replica replica : replicasToMoveOut(clusterModel, broker, topic)) {
+    for (Replica replica : replicasToMoveOut(clusterModel, broker, topic, optimizationOptions)) {
```
efeg: As a general pattern, we seem to double filter replicas (once to select them, once again to check whether `shouldExclude(...)`) to move in/out. I am not sure if this is necessary / optimal under different cases.

Can we create an issue to investigate whether we should/can merge this logic cleanly, and evaluate the impact on optimization performance?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ReplicationFactorChangeTest.java
```diff
@@ -0,0 +1,265 @@
...
+import static com.linkedin.kafka.cruisecontrol.common.DeterministicCluster.mediumClusterModel;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.fail;
+
+/**
+ * Unit test for testing modifying topic replication factor with different requirement under fixed cluster properties.
+ */
+@RunWith(Parameterized.class)
+public class ReplicationFactorChangeTest {
```
efeg: Can this test also check the case in which the hard goal is violated, and it is not possible to fix the hard goal by just moving immigrant replicas -- i.e. what is the expected behavior in this case?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -714,6 +742,26 @@ public Replica createReplica(String rackId, int brokerId, TopicPartition tp, int
...
+   * Delete a replica from cluster.
+   * @param topicPartition Topic partition of the replica to be removed.
+   * @param brokerId Id of the broker hosting the replica.
+   */
+  public void deleteReplica(TopicPartition topicPartition, int brokerId) {
+    removeReplica(brokerId, topicPartition);
+    // Update partition info.
+    Partition partition = _partitionsByTopicPartition.get(topicPartition);
+    partition.deleteReplica(brokerId);
+    _replicationFactorByTopic.put(topicPartition.topic(), partition.replicas().size());
```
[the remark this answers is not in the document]
kidkun: It is intentional, and there is an implicit assumption that `deleteReplica` will be applied across all partition of the topic and in the end the replication factor will be consistent, I should document this assumption.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -714,6 +742,26 @@ public Replica createReplica(String rackId, int brokerId, TopicPartition tp, int
...
+  }
+
+  /**
+   * Refresh the maximum topic replication factor statistic.
+   */
+  public void refreshClusterMaxReplicationFactor() {
+    _maxReplicationFactor =  _replicationFactorByTopic.values().stream().max(Integer::compareTo).orElse(0);
```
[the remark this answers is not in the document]
kidkun: In that case it would be no topic in the cluster, I think it is ok here. It is not the responsibility of this method to ensure a cluster is empty or not.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareGoal.java
kidkun: I prefer moving `(onlyMoveImmigrantReplicas && !replica.isImmigrant())` logic inside `shouldExclude(...)`.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ReplicationFactorChangeTest.java
kidkun: It is actually covered in current test(In current test, there are[ test case ](https://github.com/linkedin/cruise-control/blob/6003be795a46ff7338555b2518d8c6ed5b106aa5/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ReplicationFactorChangeTest.java#L103)where `RackAwareGoal` and `ReplicaCapacityGoal` cannot be satisfied if we increase RF to 3) , in this case we expect that `OptimizationFailureException` get thrown and no replication factor will happen.

I think thi

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/TopicReplicaDistributionGoal.java
kidkun: Agree, create PR #802 to track it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -321,7 +323,7 @@ protected void rebalanceForBroker(Broker broker,
         // return if the broker is already within limits.
         return;
       }
-      moveImmigrantsOnly = !clusterModel.deadBrokers().isEmpty();
+      moveImmigrantsOnly = !clusterModel.deadBrokers().isEmpty() || optimizationOptions.onlyMoveImmigrantReplicas();
```
efeg: I am not sure if our assumption holds: how did we conclude that an overloaded broker (e.g. in terms of disk utilization) would never have immigrant replicas when there is a dead broker?

Let's say I have 2 goals
1. `RackAwareGoal`
2. `DiskUsageDistributionGoal`
3 brokers: `B0` dead, `B1`, `B2` alive
let's assume `B2` has no replicas and receives no replicas during optimization of `RackAwareGoal`
-- i.e. `B1` has 1 immigrant replica and 9 non-immigrant replica with high load
how does `B1` stop giving away non-immigrant replicas to `B2`?

Is it possible to add a test case (if missing) for each goal to ensure that they handle it as expected?

kidkun: It is a valid concern and I am able to reproduce this issue locally with a deterministic test.
There is actually a blind spot in our current self-healing test. I have a patch to fix this issue and also add more test coverage in PR #810

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaCapacityGoal.java
```diff
@@ -240,6 +241,10 @@ protected void rebalanceForBroker(Broker broker,
         continue;
       }
 
+      if (onlyMoveImmigrantReplicas && !replica.isImmigrant()) {
```
efeg: What if the (original) broker is dead and contains replicas? Then if `onlyMoveImmigrantReplicas` is true, this goal will be unable to move offline replicas away from it.

I feel that we can make the code cleaner and more concise if we can centralize the exclusion logic -- e.g. via a centralized `shouldExclude(Replica, OptimizationOptions)`.

kidkun: I think the logic in this patch so far is not considering moving offline replicas.
What this means is that if we have offline replicas in the cluster and specify to use hard goals when increasing RF,  the request will fail with `OptimizationException`.

To centralize the filtering logic, we can address that in PR #810

efeg: If the expectation is to not run when there are offline replicas, can we fail early? -- i.e. we have the cluster model before starting the optimization.

Issue #792, opened 2019-06-21
Ensure that self.healing.goals can be configured with fully-qualified class names


efeg:
`self.healing.goals` enables users to specify the list of goals to be used for self-healing relevant anomalies. If empty, CC uses the `default.goals` for self healing. 

Currently `self.healing.goals` requires a list of goal names. However, all the other `goal` configs get the fully-qualified class names for the goals. This inconsistency can confuse the user.

PR #793, opened 2019-06-21
Ensure that self.healing.goals can be configured with fully-qualified class names.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/792.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -251,7 +252,8 @@ public static void sanityCheckCapacityEstimation(boolean allowCapacityEstimation
    * @param brokerIds The broker ids.
    * @param dryRun Whether it is a dry run or not.
    * @param throttleAddedBrokers Whether throttle the brokers that are being added.
-   * @param goals The goals to be met when adding the brokers. When empty all goals will be used.
+   * @param goals The goal names (i.e. each matching {@link Goal#name()}) to be met when adding the brokers.
+   *              When empty all goals will be used.
```
kidkun: Same Javadoc update needs to be applied to `rebalance()` and `ignoreProposalCache()` method of this Class.

PR #798, opened 2019-06-25
Modularize cli

mgrubent:
Since it is expected that users of this client may modify `cccli.py` for their own convenience, this PR attempts to modularize the `cccli.py` interface to the core cruise-control python client.

### DIFF
--- a/cruise-control-client/cruisecontrolclient/client/cccli.py
+++ b/cruise-control-client/cruisecontrolclient/client/cccli.py
@@ -6,8 +6,8 @@
 # To be able to easily parse command-line arguments
 import argparse
 
-# To be able to make more precise type hints
-from typing import Dict, Tuple, Type  # noqa
+# To be able to easily pass around the available endpoints and parameters
+from cruisecontrolclient.client.ExecutionContext import ExecutionContext
 
 # Convenience functions for displaying the retrieved Response
 from cruisecontrolclient.client.Display import display_response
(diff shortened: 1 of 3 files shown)


cruise-control-client/cruisecontrolclient/client/cccli.py
```diff
@@ -301,26 +141,67 @@ def query_cruise_control(args):
...
+def get_responder(endpoint: Endpoint.AbstractEndpoint,
+                  url: str) -> AbstractResponder:
     # Handle instantiating the correct Responder
     if endpoint.http_method == "GET":
         json_responder = JSONDisplayingResponderGet(url)
     elif endpoint.http_method == "POST":
         json_responder = JSONDisplayingResponderPost(url)
     else:
-        raise ValueError("Unexpected http_method {} in endpoint".format(endpoint.http_method))
+        raise ValueError(f"Unexpected http_method {endpoint.http_method} in endpoint")
```
efeg: In case relevant for the other parts of the client (not for the `get_responder`), CC supports `HTTP OPTIONS` method, as well (see the [servlet code](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java#L141)).

mgrubent: It is relevant; `ccdo` does not yet provide `cors` support

PR #804, opened 2019-06-29
Add anomaly metrics: (1) meanTimeBetweenAnomalies (2) meanTimeToStartFix, and (3) numSelfHealingStarted.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/723.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
@@ -174,7 +174,7 @@ public class AnomalyDetector {
 
   /**
    * Shutdown the anomaly detector.
-   * Note that if a fix is being started as shutdown is requested, shutdown will wait until the fix is initiated.Update
+   * Note that if a fix is being started as shutdown is requested, shutdown will wait until the fix is initiated.
    */
   public void shutdown() {
     LOG.info("Shutting down anomaly detector.");
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
@@ -8,7 +8,6 @@ import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.linkedin.cruisecontrol.detector.Anomaly;
 import com.linkedin.kafka.cruisecontrol.KafkaCruiseControl;
-import com.linkedin.kafka.cruisecontrol.async.progress.OperationProgress;
 import com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig;
 import com.linkedin.kafka.cruisecontrol.common.KafkaCruiseControlThreadFactory;
 import com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyNotificationResult;
@@ -29,6 +28,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import org.apache.kafka.common.utils.SystemTime;
 import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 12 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -160,7 +174,7 @@ public void startDetection() {
 
   /**
    * Shutdown the anomaly detector.
-   * Note that if a fix is being started as shutdown is requested, shutdown will wait until the fix is initiated.
+   * Note that if a fix is being started as shutdown is requested, shutdown will wait until the fix is initiated.Update
```
stanislavkozlovski: nit: typo

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -183,11 +197,43 @@ public void shutdown() {
...
     return _anomalyDetectorState;
   }
 
+  /**
+   * If there is an ongoing anomaly, clear the earliest detection time to indicate start of an ongoing fix.
+   */
+  public synchronized void maybeClearOngoingAnomalyDetectionTimeMs() {
+    if (_ongoingAnomalyDetectionTimeMs != NO_ONGOING_ANOMALY_FLAG) {
+      double elapsed = _time.milliseconds() - _ongoingAnomalyDetectionTimeMs;
+      _ongoingAnomalyDurationSumForAverageMs += elapsed;
```
kidkun: IIUC, `_ongoingAnomalyDurationSumForAverageMs` and `_ongoingAnomalyCount` are used to calculate `meanTimeToStartFix`(though a division),   but their value are updated in two places. `_ongoingAnomalyCount` is updated when an anomaly is detected; `_ongoingAnomalyDurationSumForAverageMs` is updated when any execution starts.

There are two scenario I think that may generate confusing `meanTimeToStartFix` metrics.
(1)  t0: a goal violation is detected 
       t1 : self-healing disabled/failed 
      t2 : user send a request to increase RF of a topic 
      in this case, looks like CC will take RF increasing operation as a self-healing operation, and `meanTimeToStartFix` will be `t2 - t0`

(2) t0: a goal violation detected
     t1: self-healing succeed 
     t2: another goal violation detected 
     t3: self-healing failed
     in this case, `meanTimeToStartFix` will be `(t1 - t0)/2`, shouldn't it be `(t1 - t0)`?

efeg: Thanks for the input!

The first scenario:
* Please see the JavaDoc for the `AnomalyMetrics` class, which defines `mean time to start fix` as follows:
```
Mean time between the detection of earliest ongoing anomaly and start of a proposal execution. This metric covers executions started both by anomaly detector and as a result of on-demand requests.
```
This means that any execution (e.g. a `rebalance`, `add`/`remove`/`demote` broker operation, `PLE`, or `RF increase` that start replica movements/addition/deletion or leadership transfer) is considered as an action towards a fix. If it is (1) a self-healing action, then the designated mitigation process is already well-defined a priori, (2) an on-demand request, then the cluster operator intends to perform an admin operation to address a certain issue -- e.g. cluster expansion, replacement of bad hardware, alleviating the load from selected set of brokers. Note that all such admin operations that start an execution use a list of goals to keep the cluster in the desired state (e.g. no offline replicas, satisfy goals). Regardless, even if the state after the fix still violates a goal or cluster operator does not know what s/he is doing, then this will trigger another anomaly and cause a rise in the mean time to start a fix.

The second scenario:
* Good point! It should indeed be `(t1 - t0)` (will fix it). That brings up the question of capturing "how long has the latest unfixed/unfixable anomaly there". I will also add a new metric to represent that value.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyMetrics.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyType;
+import java.util.Map;
+
+
+public class AnomalyMetrics {
+  private final Map<AnomalyType, Double> _meanTimeBetweenAnomalies;
+  private final double _meanTimeToStartFix;
+  private final long _numSelfHealingStarted;
+  private final long _ongoingAnomalyDurationMs;
```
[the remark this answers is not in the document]
efeg: I feel that if an `ongoing` anomaly implies that it is `unfixed`. Moreover, the corresponding JavaDoc already provides further details.
```
@param ongoingAnomalyDurationMs The duration of the ongoing (unfixed/unfixable) anomaly if there is any, 0 otherwise.
```
Hence, for brevity, I would prefer to avoid the name change.

PR #806, opened 2019-07-01
[WIKI] Mention self.healing.enabled configuration in docs and better clarify default behavior

stanislavkozlovski:
The documentation was misleading because it did not mention the self.healing.enabled config and was false, because self-healing is [disabled by default](https://github.com/linkedin/cruise-control/blob/29bd28b0fb6156a9ea1c240f80a820324fd52781/config/cruisecontrol.properties#L255).

cc @efeg @kidkun


stanislavkozlovski:
@efeg could you also update the Wiki with this change?
https://github.com/linkedin/cruise-control/wiki/Configurations/

efeg:
> @efeg could you also update the Wiki with this change?
> https://github.com/linkedin/cruise-control/wiki/Configurations/

Thanks for the reminder -- updated the wiki.

Issue #802, opened 2019-06-27
potential optimization in selecting right replicas to operate on for each goal


kidkun:
Currently we have 2 mechanisms to filter out the right replica for replica movement/leadership transfer consideration. 
(1) [AbstractGoal#shouldExclude()](https://github.com/linkedin/cruise-control/blob/29bd28b0fb6156a9ea1c240f80a820324fd52781/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java#L115)
(2) [ClusterModel#trackSortedReplicas](https://github.com/linkedin/cruise-control/blob/29bd28b0fb6156a9ea1c240f80a820324fd52781/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java#L522)

There maybe room for improvement on efficiency if we can centralize the filtering.

PR #810, opened 2019-07-06
Optimize replica filtering logic in goals.

kidkun:
Fix issue #802


kidkun:
This patch is out of date. will create a new one.

Issue #819, opened 2019-07-09
NPE when using NoopNotifier and querying state endpoint


efeg:
Since `NoopNotifier#selfHealingEnabledRatio()` returns `null` response, this will cause an NPE in `AnomalyDetector#anomalyDetectorState()`.

Issue #818, opened 2019-07-09
NPE upon requesting anomaly_detector state when self healing is enabled by default


efeg:
If CC is deployed when self healing is enabled by default, the response of `state` will throw an `NPE`.
```
Received exception when trying to execute runnable for "Get state"
java.lang.NullPointerException: null
        at com.linkedin.kafka.cruisecontrol.detector.notifier.SelfHealingNotifier.selfHealingEnabledRatio(SelfHealingNotifier.java:166)
        at com.linkedin.kafka.cruisecontrol.detector.AnomalyDetector.anomalyDetectorState(AnomalyDetector.java:207)
        at com.linkedin.kafka.cruisecontrol.KafkaCruiseControl.state(KafkaCruiseControl.java:1205)
        at com.linkedin.kafka.cruisecontrol.async.GetStateRunnable.getResult(GetStateRunnable.java:28)
        at com.linkedin.kafka.cruisecontrol.async.GetStateRunnable.getResult(GetStateRunnable.java:16)
        at com.linkedin.kafka.cruisecontrol.async.OperationRunnable.run(OperationRunnable.java:45) 
```

PR #820, opened 2019-07-09
Fix NPE upon retrieving the response of anomaly detector.

efeg:
Addresses the issues https://github.com/linkedin/cruise-control/issues/818 and https://github.com/linkedin/cruise-control/issues/819.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/notifier/SelfHealingNotifier.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/notifier/SelfHealingNotifier.java
@@ -152,7 +152,8 @@ public class SelfHealingNotifier implements AnomalyNotifier {
       long enabledTimeMs = _selfHealingEnabledHistoricalDurationMs.get(anomalyType);
       if (_selfHealingEnabled.get(anomalyType)) {
         // Add current duration during which the self-healing is enabled.
-        enabledTimeMs += nowMs - _selfHealingStateChangeTimeMs.get(false).get(anomalyType);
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/notifier/SelfHealingNotifier.java
```diff
@@ -152,7 +152,8 @@ private void updateSelfHealingStateChange(AnomalyType anomalyType, Boolean oldVa
       long enabledTimeMs = _selfHealingEnabledHistoricalDurationMs.get(anomalyType);
       if (_selfHealingEnabled.get(anomalyType)) {
         // Add current duration during which the self-healing is enabled.
-        enabledTimeMs += nowMs - _selfHealingStateChangeTimeMs.get(false).get(anomalyType);
+        Long currentEnabledSelfHealingStartTime = _selfHealingStateChangeTimeMs.get(false).get(anomalyType);
+        enabledTimeMs += nowMs - (currentEnabledSelfHealingStartTime == null ? 0L : currentEnabledSelfHealingStartTime);
```
kidkun: I think here should be `_notifierStartTimeMs` instead of `0L`.
`currentEnabledSelfHealingStartTime == null` happens only when CC is running with self-healing on by default, so the base time to calculate the time self-healing  being on should be CC start time(i.e. `_notifierStartTimeMs`).

efeg: Correct -- nice catch!

PR #821, opened 2019-07-09
Add project description, add command-line entrypoint

mgrubent:
Functionally, this commit introduces the ability for direct command-line access to `cccli.py` through the use of the [`scripts` keyword argument](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-scripts-keyword-argument).

This commit also updates the Python Client README to direct users to `pip` for acquiring the `cruise-control-client` package, rather than cloning into this repo directly.

Finally, this commit adds a `long_description`, so that the [PyPI project page](https://pypi.org/project/cruise-control-client/) can have a description.


mgrubent:
Just kidding `scripts` is an abandoned method of specifying a script entry point

Issue #800, opened 2019-06-25
Ensure that the user receives a UserRequestException upon providing negative review id


efeg:
`review_ids` and `review_id` parameters should return a `UserRequestException` if the provided argument(s) are negative.

PR #831, opened 2019-07-10
Add sanity check for negative review ids.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/800.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
@@ -669,10 +669,24 @@ public class ParameterUtils {
   }
 
   /**
+   * @param values Integer values
+   * @return A set of negative integer values contained in the given set.
+   */
+  private static Set<Integer> getNegatives(Set<Integer> values) {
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -668,11 +668,25 @@ public static int entries(HttpServletRequest request) {
     return parameterString == null ? Integer.MAX_VALUE : Integer.parseInt(request.getParameter(parameterString));
   }
 
+  /**
+   * @param values Integer values
+   * @return A set of negative integer values contained in the given set.
+   */
+  private static Set<Integer> getNegatives(Set<Integer> values) {
+    return values.stream().filter(v -> v < 0).collect(Collectors.toCollection(() -> new HashSet<>(values.size())));
```
kidkun: Just curious, why using `Collectors.toCollection(() -> new HashSet<>(values.size()))` instead of `Collectors.toSet()` here? 
Is it because it preallocate the capacity of the set?

efeg: Yes, this provides a minor memory optimization if the given set is small -- which is the expectation in its current use-cases.

Issue #809, opened 2019-07-06
Distribution goals may move more than necessary replicas in self-healing scenario


kidkun:
As brought up in one discussion of PR #789 , when there are offline replicas in the cluster(due to broker failure or disk failure), the self-healing operation generate by Cruise Control( with any distribution goal provided) may generate replica movement operations between alive brokers, which will increase the amount of data to move and time to execute self-healing operations.

One example case is like this
Broker 1: on rack 1, host replica T1P0R0(replica 0 of of topic 1 partition 0), T1P1R0, T1P2R0, T2P0R0, T2P1R0, T2P2R0
Broker 2: on rack 2, host replica T1P0R1, T1P3R0,T2P0R1, T2P3R0
Broker 3: on rack 3, host replica T1R1R1, T2R1R1
Broker 4: on rack 4, host replica T1P2R1, T2P2R1
Broker 5: on rack 5, host replica T1P3R1, T2P3R1

Now Broker 5 dies and we try to fix offline replicas with `RackAwareGoal` and `ReplicaDistributionGoal`, under currently implementation 
(1) `RackAwareGoal` will move T1P3R1, T2P3R1 to Broker 1 since it is on different rack and has smallest broker id.
(2) Now broker 1 has 8 replicas and `ReplicaDistributionGoal` will move 4 replicas to broker 3,4 to make each broker host 4 replicas.

Here we should only move two offline replicas away from Broker 5, but we actually execute 4 replica movement operations, which is not needed for self-healing purpose.

PR #834, opened 2019-07-12
Fix the bug that distribution goals may move more than necessary replicas in self-healing scenario.

kidkun:
Fix issue #809


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationVerifier.java
```diff
@@ -122,18 +127,34 @@ static boolean executeGoalsFor(BalancingConstraint constraint,
                                                     new SystemTime(),
                                                     new MetricRegistry(),
                                                     EasyMock.mock(Executor.class));
-    GoalOptimizer.OptimizerResult optimizerResult = goalOptimizer.optimizations(clusterModel,
-                                                                                goalByPriority,
-                                                                                new OperationProgress());
+    // First optimize with all hard goals if there is any.
+    List<Goal> hardGoals =  goalByPriority.stream().filter(Goal::isHardGoal).collect(Collectors.toList());
+    GoalOptimizer.OptimizerResult optimizerResult1 = hardGoals.isEmpty() ? null
```
efeg: `optimizerResult1` -> `optimizationResultWithHardGoals`?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationVerifier.java
```diff
@@ -122,18 +127,34 @@ static boolean executeGoalsFor(BalancingConstraint constraint,
...
+                                                                         : goalOptimizer.optimizations(clusterModel,
+                                                                                                       hardGoals,
+                                                                                                       new OperationProgress());
+
+    // If provided goal list contains any soft goals, then optimize with all goals.
+    // In this iteration, all the hard goals will not generate any new proposal since they are already satisfied in previous
+    // iteration (otherwise an exception has already been thrown). The optimize result of this iteration will only contains
+    // proposals generated by soft goals.
+    List<Goal> softGoals = goalByPriority.stream().filter(g -> !g.isHardGoal()).collect(Collectors.toList());
+    GoalOptimizer.OptimizerResult optimizerResult2 = softGoals.isEmpty() ? null
```
efeg: `optimizerResult1` -> `optimizationResultWithSoftGoals`?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationVerifier.java
```diff
@@ -122,18 +127,34 @@ static boolean executeGoalsFor(BalancingConstraint constraint,
...
+    List<Goal> hardGoals =  goalByPriority.stream().filter(Goal::isHardGoal).collect(Collectors.toList());
+    GoalOptimizer.OptimizerResult optimizerResult1 = hardGoals.isEmpty() ? null
+                                                                         : goalOptimizer.optimizations(clusterModel,
+                                                                                                       hardGoals,
+                                                                                                       new OperationProgress());
+
+    // If provided goal list contains any soft goals, then optimize with all goals.
+    // In this iteration, all the hard goals will not generate any new proposal since they are already satisfied in previous
+    // iteration (otherwise an exception has already been thrown). The optimize result of this iteration will only contains
+    // proposals generated by soft goals.
```
efeg: The explanation is correct, but this change might introduce potential issues:
1. We no longer test the optimization process with a mix of soft and hard goals.
2. In particular, we no longer test the cases, where soft goals have higher priority compared to hard goals.
-- 
Does it make sense to:
(1) preserve the existing test cases with the original list of goals and
(2) introduce "selected" (i.e. targeted) test cases for 
    2.1. testing each soft goal in a cluster state such as the one exemplified in the corresponding ticket of this PR? -- i.e. a case that the goal needs to move non-immigrant replica(s) to achieve balance. But, for correctness, it should not move.
    2.2. testing each hard goal in a cluster state such that the goal needs to move non-immigrant replica(s) to achieve balance. And, for correctness, it indeed moves replicas.

kidkun: Good point, I am adding a new switch to `OptimizationVerifier` to specify wether optimize with provided goals as a whole or optimize in a two-pass way(first optimize with all hard goals, then optimize with all soft goals).

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationVerifier.java
```diff
@@ -93,12 +95,38 @@ static boolean executeGoalsFor(BalancingConstraint constraint,
...
+  }
+
+  /**
+   * Execute given goals in the given cluster enforcing the given constraint. Return pass / fail status of a test.
+   * A test fails if:
+   * 1) Rebalance: During the optimization process, optimization of a goal leads to a worse cluster state (in terms of
+   * the requirements of the same goal) than the cluster state just before starting the optimization.
+   * 2) Self Healing: There are replicas on dead brokers after self healing or soft goal(if any) moves non-immigrant
+   * online replicas.
+   * 3) Adding a new broker causes the replicas to move among old brokers.
```
efeg: Nit: Looks like the following JavaDoc is repeated in each `#executeGoalsFor`:
```
   * Execute given goals in the given cluster enforcing the given constraint. Return pass / fail status of a test.
   * A test fails if:
   * 1) Rebalance: During the optimization process, optimization of a goal leads to a worse cluster state (in terms of
   * the requirements of the same goal) than the cluster state just before starting the optimization.
   * 2) Self Healing: There are replicas on dead brokers after self healing or soft goal(if any) moves non-immigrant
   * online replicas.
   * 3) Adding a new broker causes the replicas to move among old brokers.
```
For conciseness, should we refer to this JavaDoc from the other two `#executeGoalsFor` for the excerpt above? -- i.e. we can say something like:
```
See {@link #executeGoalsFor(BalancingConstraint, ClusterModel, List, Collection, List, boolean)}.
```

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/DeterministicSelfHealingTest.java
```diff
@@ -0,0 +1,129 @@
...
+                                                    TopicReplicaDistributionGoal.class.getName());
+
+    Properties props = KafkaCruiseControlUnitTestUtils.getKafkaCruiseControlProperties();
+    props.setProperty(KafkaCruiseControlConfig.MAX_REPLICAS_PER_BROKER_CONFIG, Long.toString(5L));
+    BalancingConstraint balancingConstraint = new BalancingConstraint(new KafkaCruiseControlConfig(props));
+    balancingConstraint.setCapacityThreshold(TestConstants.LOW_CAPACITY_THRESHOLD);
+    balancingConstraint.setResourceBalancePercentage(TestConstants.LOW_BALANCE_PERCENTAGE);
+    List<OptimizationVerifier.Verification> verifications = Arrays.asList(DEAD_BROKERS, REGRESSION);
+
+    // ----------##TEST 1: SINGLE GOAL.
```
efeg: Don't we already have the same tests with a single goal in `RandomSelfHealingTest`? If they are the same, can we remove the test with single goal and keep only the `RACK AWARE GOAL + ONE OTHER GOAL` part?
Finally, can we rename this test class to a more descriptive name -- e.g. `DeterministicSelfHealingTest` -> `SelfHealingWithOptimizedGoal`?

Issue #839, opened 2019-07-17
Anomaly metrics in plaintext response of state endpoint is unreadable.


kidkun:
`meanTimeBetweenAnomalies`, `meanTimeToStartFix` and `ongoingAnomalyDurationMs` are returned in unit of ms and the former two does not specify time unit, which cause confusion for user. 
One example
`metrics:{meanTimeBetweenAnomalies={DISK_FAILURE=0.0, BROKER_FAILURE=0.0, METRIC_ANOMALY=0.0, GOAL_VIOLATION=1.7889666313204174E-4}, meanTimeToStartFix=627302.6, numSelfHealingStarted=2, ongoingAnomalyDurationMs=0}`

PR #840, opened 2019-07-17
Make anomaly metrics readable.

kidkun:
Fix issue #839

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -37,6 +37,10 @@ public class KafkaCruiseControlUtils {
   public static final String DATE_FORMAT = "YYYY-MM-dd_HH:mm:ss z";
   public static final String DATE_FORMAT2 = "dd/MM/yyyy HH:mm:ss";
   public static final String TIME_ZONE = "UTC";
+  private static final int SEC_TO_MS = 1000;
+  private static final int MIN_TO_MS = SEC_TO_MS * 60;
+  private static final int HOUR_TO_MS = MIN_TO_MS * 60;
+  private static final int DAY_TO_MS = HOUR_TO_MS * 24;
   private static final Set<String> KAFKA_ASSIGNER_GOALS =
       Collections.unmodifiableSet(new HashSet<>(Arrays.asList(KafkaAssignerEvenRackAwareGoal.class.getSimpleName(),
                                                               KafkaAssignerDiskUsageDistributionGoal.class.getSimpleName())));
@@ -87,33 +91,33 @@ public class KafkaCruiseControlUtils {
    * @param durationMs Duration in milliseconds
    * @return String representation of duration
    */
-  public static String toDurationString(double durationMs) {
+  public static String toPrettyDuration(double durationMs) {
+    if (durationMs < 0) {
+      throw new IllegalArgumentException(String.format("Duration cannot be negative value, get %f", durationMs));
+    }
+
     // If the duration is less than one second, represent in milliseconds.
-    if (durationMs < 1000) {
+    if (durationMs < SEC_TO_MS) {
       return String.format("%.2f milliseconds", durationMs);
     }
 
-    durationMs = durationMs / 1000;
     // If the duration is less than one minute, represent in seconds.
-    if (durationMs < 60) {
-      return String.format("%.2f seconds", durationMs);
+    if (durationMs < MIN_TO_MS) {
+      return String.format("%.2f seconds", durationMs / SEC_TO_MS);
     }
 
-    durationMs = durationMs / 60;
     // If the duration is less than one hour, represent in minutes.
-    if (durationMs < 60) {
-      return String.format("%.2f minutes", durationMs);
+    if (durationMs < HOUR_TO_MS) {
+      return String.format("%.2f minutes", durationMs / MIN_TO_MS);
     }
 
-    durationMs = durationMs / 60;
     // If the duration is less than one day, represent in hours.
-    if (durationMs < 24) {
-      return String.format("%.2f hours", durationMs);
+    if (durationMs < DAY_TO_MS) {
+      return String.format("%.2f hours", durationMs / HOUR_TO_MS);
     }
 
     // Represent in days.
-    durationMs = durationMs / 24;
-    return String.format("%.2f days", durationMs);
+    return String.format("%.2f days", durationMs / DAY_TO_MS);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyMetrics.java
```diff
@@ -58,4 +63,14 @@ public long numSelfHealingStarted() {
   public long ongoingAnomalyDurationMs() {
     return _ongoingAnomalyDurationMs;
   }
+
+  @Override
+  public String toString() {
+    return String.format("{meanTimeBetweenAnomalies:{%s:%s, %s:%s, %s:%s}, "
+                         + "meanTimeToStartFix:%s, numSelfHealingStarted:%d, ongoingAnomalyDuration=%s}",
```
efeg: Nit: Can we update the following in (1) all hard coded Strings (e.g. including the exception messages), (2) variable names here, (3) variable names used while creating `AnomalyMetrics` object?

1. `meanTimeBetweenAnomalies` -> `meanTimeBetweenAnomaliesMs`
2. `meanTimeToStartFix` -> `meanTimeToStartFixMs`
3. `ongoingAnomalyDuration` -> `ongoingAnomalyDurationMs`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyMetrics.java
```diff
@@ -58,4 +63,14 @@ public long numSelfHealingStarted() {
...
+                         GOAL_VIOLATION, toDurationString(_meanTimeBetweenAnomalies.get(GOAL_VIOLATION)),
+                         BROKER_FAILURE, toDurationString(_meanTimeBetweenAnomalies.get(BROKER_FAILURE)),
+                         METRIC_ANOMALY, toDurationString(_meanTimeBetweenAnomalies.get(METRIC_ANOMALY)),
```
efeg: These may throw an exception if `_meanTimeBetweenAnomalies.get(...)` is `null`. Can we add a sanity check in constructor?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -100,8 +100,8 @@ public void setMetrics(AnomalyMetrics metrics) {
    */
   Map<String, Object> metrics() {
     Map<String, Object> metrics = new HashMap<>(4);
-    metrics.put(MEAN_TIME_BETWEEN_ANOMALIES, _metrics.meanTimeBetweenAnomalies());
-    metrics.put(MEAN_TIME_TO_START_FIX, _metrics.meanTimeToStartFix());
+    metrics.put(MEAN_TIME_BETWEEN_ANOMALIES_MS, _metrics.meanTimeBetweenAnomalies());
```
efeg: I believe currently `_metrics.meanTimeBetweenAnomalies()` returns mean time per second due to the following line in `AnomalyDetector#anomalyDetectorState()`:
```
meanTimeBetweenAnomalies.put(anomalyType, _anomalyRateByType.get(anomalyType).getMeanRate());
```

We should make the naming and value consistent -- i.e. can we ensure that `meanTimeBetweenAnomalies` map is initialized with ms values rather than sec?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -82,6 +82,40 @@ public static String toDateString(long time, String dateFormat, String timeZone)
     return formatter.format(new Date(time));
   }
 
+  /**
+   * Format the duration from double to human readable string.
+   * @param durationMs Duration in milliseconds
+   * @return String representation of duration
+   */
+  public static String toDurationString(double durationMs) {
```
efeg: Nit: For consistency with `CruiseControlUtils#toPrettyTime`, should we rename this function to `toPrettyDuration`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -82,6 +82,40 @@ public static String toDateString(long time, String dateFormat, String timeZone)
...
+    // If the duration is less than one second, represent in milliseconds.
```
efeg: Let's add a sanity check to guarantee that the duration is non-negative.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -82,6 +82,40 @@ public static String toDateString(long time, String dateFormat, String timeZone)
...
+    if (durationMs < 1000) {
+      return String.format("%.2f milliseconds", durationMs);
+    }
+
+    durationMs = durationMs / 1000;
```
efeg: Looks like the meaning of `durationMs` changes as we go along the code due to the reassingments. Can we instead represent all comparison constants in `ms` and then just divide the ms value to the relevant conversion value? -- For example, 1 day is `86400000 milliseconds`

Also can we move the constant times to static variables?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyMetrics.java
kidkun: But here we actually cast the duration to second/minute/hour/day based on duration and we have print the unit after the value.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
kidkun: good catch!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -198,10 +198,10 @@ public void shutdown() {
   }
 
   public synchronized AnomalyDetectorState anomalyDetectorState() {
-    // Retrieve mean time between anomalies
+    // Retrieve mean time between anomalies, record the time in ms.
     Map<AnomalyType, Double> meanTimeBetweenAnomalies = new HashMap<>(AnomalyType.cachedValues().size());
     for (AnomalyType anomalyType : AnomalyType.cachedValues()) {
-      meanTimeBetweenAnomalies.put(anomalyType, _anomalyRateByType.get(anomalyType).getMeanRate());
+      meanTimeBetweenAnomalies.put(anomalyType, _anomalyRateByType.get(anomalyType).getMeanRate() * 1000);
```
efeg: Nit: Can we Move `1000` to a `static` var?

PR #843, opened 2019-07-17
Make request handling logic pluggable.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/770.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/AdminParameters.java
```diff
@@ -32,13 +32,12 @@
   private Integer _concurrentInterBrokerPartitionMovements;
   private Integer _concurrentLeaderMovements;
   private Integer _reviewId;
-  private final boolean _twoStepVerificationEnabled;
+  private boolean _twoStepVerificationEnabled;
```
kidkun: (Not introduce by this patch)this field is one-time use(in `initParameters()`), looks like we do not need to store it in a field, just like what `RebalanceParameters` does. also apply to `PauseResumeParameters`/`StopProposalParameters`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/AddBrokerRequest.java
```diff
@@ -5,16 +5,17 @@
...
 
 
 public class AddBrokerRequest extends AbstractAsyncRequest {
-  private final AddBrokerParameters _parameters;
+  private AddBrokerParameters _parameters;
 
-  public AddBrokerRequest(KafkaCruiseControlServlet servlet, AddBrokerParameters parameters) {
-    super(servlet);
-    _parameters = parameters;
+  public AddBrokerRequest() {
```
kidkun: maybe we do not need to explicitly specify default constructor here(apply to other `Request` class)?

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/servlet/EndPoint.java
```diff
@@ -0,0 +1,16 @@
...
+
+package com.linkedin.cruisecontrol.servlet;
+
+/**
+ * An enum that lists all supported endpoints by Cruise Control.
+ */
+public interface EndPoint {
+
+  /**
+   * @return The endpoint type, which might be relevant for the application logic -- e.g. selective response caching.
```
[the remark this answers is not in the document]
efeg: Made it more verbose.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/AddBrokerParameters.java
```diff
@@ -28,8 +28,8 @@
 public class AddBrokerParameters extends AddedOrRemovedBrokerParameters {
   private boolean _throttleAddedBrokers;
 
-  public AddBrokerParameters(HttpServletRequest request, KafkaCruiseControlConfig config) {
-    super(request, config);
+  public AddBrokerParameters() {
```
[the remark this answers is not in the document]
efeg: To create an instance of such subclasses using `KafkaCruiseControlConfig#getConfiguredInstance(String, Class<T>, Map<String, Object>`, or any other methods that call `Utils#newInstance(Class<T>)`, the class needs to explicitly specify a public no-argument constructor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/AddBrokerRequest.java
efeg: To create an instance of such subclasses using `KafkaCruiseControlConfig#getConfiguredInstance(String, Class<T>, Map<String, Object>`, or any other methods that call `Utils#newInstance(Class<T>)`, the class needs to explicitly specify a public no-argument constructor.

PR #848, opened 2019-07-19
allow some timeouts to be overridable from runtime instead of hard coded

linehrr:
we have encountered some timeouts in prod, and found out that they are hard coded and impossible to override. like `LOGDIR_RESPONSE_TIMEOUT_MS`. 
therefore this PR is to allow overrides by env vars.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -60,6 +62,66 @@ private KafkaCruiseControlUtils() {
 
   }
 
+  public static int getZkSessionTimeout() {
```
kidkun: I feel in general retrieving value from environment may have some potential issues.

(1) environment value is out-of-control of cruise control, and potentially make cruise control mangle with other application running in the same environment(if they are using the same env variable), also make CC tightly-coupled with OS(deploying CC in another env without migrating variable may result in inconsistent behaviour)
(2) it lacks some sanity check, i.e. what if the these timeout is mistakenly set to negative value

Another question I have is in this implementation, looks like each time we want these timeout value, we will do a  retrieval from environment, is this level of dynamic necessary(i.e. do we want to dynamically change these value when CC is running)?

I think it is a good idea to be able to config these timeout by user instead of hard-coding. 
Maybe we can expose them in [KafkaCruiseControlConfig](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java) and config them through property files.

linehrr: @kidkun I actually prefer your way, however the reason we did it in env vars are simply because it's easy to implement, and we need this change in prod real quick as we are getting timeouts constantly. by looking at the code, you can see those util methods are all static, and those classes that are holding those values are also static, so it makes it hard to load from config, since static got inited first in JVM before anything else. 
we can look into moving those settings into proper configs, but that might require bigger code refactor.

linehrr:
close this PR, will make another one with proper config loading.

Issue #842, opened 2019-07-17
Why we set replicationThrottle to null in the fix method of Anomaly?


Wennn:
In my view, the value replicationThrottle works only when accepting the HTTP request, and the values passed by the fix method are all null. So why we do not use the value from config instead of null?
If I misunderstand it, please point out my mistake. 
Thx a lot.

kidkun:
hey @Wennn , I think this is a good point! We should pick up the config default value(just like `InterBrokerPartitionMovementConcurrency` and `LeadershipMovementConcurrency`).
Are you interested in creating a PR to address it?

Wennn:
ok, I will do it.
thx for reply:)

PR #851, opened 2019-07-20
use config value for replicaThrottle in the fix method of Anomaly

Wennn:
for issue #842 

If default.replication.throttle is not set in config file, null is the default value as defined in KafkaCruiseControlConfig(pr #748 ).


Wennn:
ping~ @kidkun

Issue #860, opened 2019-07-26
Add support for BrokerCapacityInfo to keep number of CPU cores


efeg:
Cruise Control does not keep track of the number of CPU cores that brokers have. Hence, the `percentage` capacity would become insufficient in making decisions regarding load distribution in clusters with heterogeneous CPUs.

As a prerequisite of handling heterogeneous CPUs, broker capacities should reflect the number of CPU cores used by brokers.

PR #861, opened 2019-07-26
Add support for BrokerCapacityInfo to keep number of CPU cores.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/860.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerCapacityConfigFileResolver.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerCapacityConfigFileResolver.java
@@ -15,6 +15,7 @@ import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Paths;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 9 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerCapacityConfigFileResolver.java
```diff
@@ -140,8 +183,12 @@ private static boolean isJBOD(Map<Resource, Object> brokerCapacity) {
     return brokerCapacity.get(Resource.DISK) instanceof Map;
   }
 
+  private static boolean hasNumCores(Map<Resource, Object> brokerCapacity) {
```
kidkun: This method does not check what the name suggests, it is based on an assumption that if it is a map, the map contains only one key and the key is `num.cores`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerCapacityConfigFileResolver.java
```diff
@@ -156,12 +203,15 @@ private static boolean isJBOD(Map<Resource, Object> brokerCapacity) {
             totalDiskCapacity += Double.parseDouble(diskEntry.getValue());
           }
           totalCapacity.put(resource, totalDiskCapacity);
+        } else if (hasNumCores && resource == Resource.CPU) {
+          totalCapacity.put(resource, DEFAULT_CPU_CAPACITY_WITH_CORES);
```
kidkun: (This maybe correct, just to confirm) so in this patch, if `num.cores` is specified, the cpu capacity is `100` instead of `100 * num_core`.

efeg: Yes, this is the intended behavior.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerCapacityConfigFileResolver.java
efeg: Added `getUserSpecifiedNumCores(Map<Resource, Object>)` to make the logic more clear.

`... the map contains only one key ...`: This, I suspect, would make the sanity check stricter than intended -- i.e. the map can have other entries without violating our requirements.

Issue #699, opened 2019-05-10
Feature request: cruise-control should be able to provide capacity of a broker via /load endpoint


kidkun:
Currently, cruise control provides brokers' utilization of resource(`CPU`/`DISK`/`NW_IN`/`NW_OUT`) via the `/load` endpoint, it is good to make CC report the capacity of these resource so the user get a view of how heavy the broker is loaded.

A straightforward way is to add more columns in response(`DISK_CAPACITY`, `CPU_CAPACITY`, `NW_IN_CAPACITY`, `NW_OUT_CAPACITY`), but doing so have two drawbacks.
1. it is not back-compatible, which may not be an issue in plaintext format, but it will break anything built upon CC(e.g CCFE) which depends on the JSON format.
2. calculate utilization is a resource-heavy calculation(and take long time for large cluster), if user only want to get capacity info, it would be waste of both computing resource and user's time

A better approach is to add a new parameter, like `capacity`, to `load` endpoint, by default it is set to `false` and CC return the same thing as of today; if this is set to `true`, CC can only return broker capacity information.

jingyli:
@kidkun Hi Kun, I can take a look into this issue and pick it up, thanks!

jingyli:
@kidkun Hi Kun, I am in the middle of implementing this issue and would like to ask for a clarification on the code structure. 

Should we create a new class dedicated to only fetch the capacity related stats for each broker? Or should we extend the existing `BrokerStats` class to support this functionality?

kidkun:
@jingyli  Sorry for delay in response, I think we can just updates `BrokerStats`, otherwise it needs larger scope of refactor of servlet code(if one endpoint can potentially return response of different class type).

efeg:
Closing the issue as it has been addressed in https://github.com/linkedin/cruise-control/pull/868.

PR #868, opened 2019-07-30
Provide capacity stats for a broker.

jingyli:
Addresses issue #699 

Added `capacity` parameter to `load` so CC can dynamically decide which stats to return (utilization vs. capacity).

Since we are trying to provide a new set of stats, I think the least intrusive way is to have `BrokerStats` as an abstract class and extend it to different concrete classes for different stat types. Right now we have `BrokerUtilizationStats` (which is what used to be in `BrokerStats`) and `BrokerCapacityStats` (the class added for this issue).

Tested the change locally by starting 5 brokers and accessing `/load` endpoint with `capacity` set to true and false. Verified that when `capacity == true`, we obtain a table where the column headers are `HOST`, `BROKER`, `DISK_CAPACITY(MB)`, `CPU_CAPACITY(MB)`, `NW_IN_CAPACITY(KB/s)`, `NW_OUT_CAPACITY(KB/s)`. 

This is a larger patch as it involves both refactoring of existing codebase and implementing new classes to support this issue. Requesting review from @efeg and @kidkun . Thanks!


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/SingleBrokerCapacityStats.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package com.linkedin.kafka.cruisecontrol.servlet.response.stats;
+
+import java.util.Map;
+
+
+public class SingleBrokerCapacityStats {
+  private static final String HOST = "Host";
+  private static final String BROKER = "Broker";
```
kidkun: we have the same macro in `SingleBrokerStats`, also there are some same method(`host()`,`id()`), we probably need a base class here to hold the common things here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/SingleBrokerStats.java
```diff
@@ -53,8 +53,8 @@ public boolean isEstimated() {
    * Return an object that can be further used
    * to encode into JSON
    */
-  public Map<String, Object> getJSONStructure() {
-    Map<String, Object> entry = _basicStats.getJSONStructure();
+  public Map<String, Object> getJsonStructure() {
```
kidkun: now we have `SingleBrokerCapacityStats`, looks like this can be renamed as `SingleBrokerUtilizationStats`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BrokerCapacityStats.java
```diff
@@ -0,0 +1,77 @@
...
+    SingleBrokerCapacityStats singleBrokerCapacityStats = 
+        new SingleBrokerCapacityStats(host, id, isEstimated, diskCap, cpuCap, nwInCap, nwOutCap);
+    _brokerCapacityStats.add(singleBrokerCapacityStats);
+    _isBrokerStatsEstimated = _isBrokerStatsEstimated || isEstimated;
+    _hostFieldLength = Math.max(_hostFieldLength, host.length());
+  }
+
+  @Override
+  protected void discardIrrelevantResponse() {
+    // Discard irrelevant response.
```
kidkun: miss `super.discardIrrelevantResponse()`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BrokerCapacityStats.java
```diff
@@ -0,0 +1,77 @@
...
+    for (SingleBrokerCapacityStats stats : _brokerCapacityStats) {
+      sb.append(String.format("%" + _hostFieldLength + "s,%14d,%24.3f,%24.3f,%24.3f,%24.3f%n",
+                              stats.host(),
+                              stats.id(),
+                              stats.basicCapacityStats().diskCapacity(),
+                              stats.basicCapacityStats().cpuCapacity(),
+                              stats.basicCapacityStats().bytesInCapacity(),
+                              stats.basicCapacityStats().bytesOutCapacity()));
+    }
+
```
kidkun: Looks like we are missing host capacity information here. 
Note the distinction between host and broker matter in scenario there are multiple kafka broker process running on the same VM/server, in this case , each Kafka process is a broker and the VM/server is a host.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/SingleBrokerCapacityStats.java
```diff
@@ -0,0 +1,51 @@
...
+  private final String _host;
+  private final int _id;
+  private final BasicCapacityStats _basicCapacityStats;
+  private final boolean _isEstimated;
+
+  SingleBrokerCapacityStats(String host, int id, boolean isEstimated, double diskCap, double cpuCap, double nwInCap, double nwOutCap) {
+    _host = host;
+    _id = id;
+    _basicCapacityStats = new BasicCapacityStats(diskCap, cpuCap, nwInCap, nwOutCap);
+    _isEstimated = isEstimated;
```
kidkun: Can we also populate `rack` information here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/GetClusterModelInRangeRunnable.java
```diff
@@ -33,7 +34,7 @@
   @Override
   protected PartitionLoadState getResult() throws Exception {
     _kafkaCruiseControl.sanityCheckBrokerPresence(_parameters.brokerIds());
-
+    sleep(10000);
```
efeg: Unintentional `sleep`?

Will-Lo:
@kidkun @efeg Sorry this took so long, it's ready for review!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -1136,15 +1138,15 @@ public void sanityCheck() {
...
    */
   public BrokerStats brokerStats(KafkaCruiseControlConfig config) {
-    BrokerStats brokerStats = new BrokerStats(config);
+    BrokerUtilizationStats brokerUtilizationStats = new BrokerUtilizationStats(config);
     brokers().forEach(broker -> {
       double leaderBytesInRate = broker.leadershipLoadForNwResources().expectedUtilizationFor(Resource.NW_IN);
       double cpuUsagePercent = UNIT_INTERVAL_TO_PERCENTAGE * broker.load().expectedUtilizationFor(Resource.CPU)
                                / broker.capacityFor(Resource.CPU);
-      brokerStats.addSingleBrokerStats(broker.host().name(),
+      brokerUtilizationStats.addSingleBrokerUtilizationStats(broker.host().name(),
```
kidkun: nit: bad indention.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BasicCapacityStats.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package com.linkedin.kafka.cruisecontrol.servlet.response.stats;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+class BasicCapacityStats {
+    private static final String DISK_CAPACITY = "DiskCapacity";
```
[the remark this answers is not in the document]
Will-Lo: Sounds good! I'll make all the fields returned by the stats classes in camelcase for consistency

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -611,6 +613,29 @@ private double getMonitoredPartitionsPercentage() {
...
+  public BrokerCapacityStats brokerCapacityStats(KafkaCruiseControlConfig config) {
+    MetadataClient.ClusterAndGeneration clusterAndGeneration = _metadataClient.refreshMetadata();
+    List<Node> clusterNodes = new ArrayList<>(clusterAndGeneration.cluster().nodes());
+    BrokerCapacityStats brokerCapacityStats = new BrokerCapacityStats(config);
+    for (Node node : clusterNodes) {
+      String rack = getRackHandleNull(node);
+      BrokerCapacityInfo brokerCapacityInfo = _brokerCapacityConfigResolver.capacityForBroker(rack, node.host(), node.id());
+      Map<Resource, Double> capacities = brokerCapacityInfo.capacity();
+      brokerCapacityStats.addSingleBrokerCapacityStats(node.host(), 
+                                               node.id(),
```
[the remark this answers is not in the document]
Will-Lo: Ah my editor seems to also not convert tabs to spaces, I'll get this fixed soon

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ClusterLoadParameters.java
```diff
@@ -28,7 +28,7 @@
  * <pre>
  * Get the cluster load
  *    GET /kafkacruisecontrol/load?start=[START_TIMESTAMP]&amp;end=[END_TIMESTAMP]&amp;time=[END_TIMESTAMP]&amp;allow_capacity_estimation=[true/false]
- *    &amp;json=[true/false]
+ *    &amp;json=[true/false];&capacity=[true/false]
```
kidkun: `;&`->`&amp;`

Also can we rename this parameter to `capacity_only`? I had a offline discussion with @efeg , we want to change the default behavior of `load` endpoint (i.e. when `capacity_only` = `false`) to return both broker capacity and utilization for different resources ( I think we can make that change to a follow-up patch). In that case, `capacity_only` is more accurate about the purpose of this parameter.

Will-Lo: Sounds good! I think `capacity_only` makes more sense as well

Will-Lo: I'll rename the other parameter items to reflect this change as well

efeg:
@Will-Lo Are there any remaining feedback to address for this PR? (cc @kidkun)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BrokerCapacityStats.java
Will-Lo: the parent `BrokerStats` defined method for `discardIrrelevantResponse` has a parameter for CC parameters. The regular `discardIrrelevantResponse()` is initialized as abstract so `super.discardIrrelevantResponse()` should be unnecessary.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BasicCapacityStats.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final String CPU_CAPACITY = "CpuCapacity";
+    private static final String NW_IN_CAPACITY = "NwInCapacity";
+    private static final String NW_OUT_CAPACITY = "NwOutCapacity";
+    private double _diskCapacity;
+    private double _cpuCapacity;
+    private double _bytesInCapacity;
+    private double _bytesOutCapacity;
+
+  BasicCapacityStats(double diskCapacity, double cpuCapacity, double nwInCapacity, double nwOutCapacity) {
+    _diskCapacity = diskCapacity < 0.0 ? 0.0 : diskCapacity;
```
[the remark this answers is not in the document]
Will-Lo: In this case I'll set `_diskCapacity = diskCapacity < 0.0 ? Broker.DEAD_BROKER_CAPCITY : diskCapacity;`
Would that be sufficient?

Will-Lo:
@efeg I'll take a look over all of the comments and test again since it's been a while and let you know once it's ready

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BrokerCapacityStats.java
Will-Lo: I realized the way how I did this was pretty bad, it had 2 methods where `discardIrrelevantResponse()` was called from another function, `discardIrrelevantAndCacheRelevant()`, and was overloaded by the children classes but it led to more confusion. I modified the function to just be defined in the children classes so the logic is more centralized.

Will-Lo:
@kidkun @efeg I addressed the feedback, could you take a look at the PR again? Thank you!

Issue #493, opened 2019-01-19
Programmatically tell whether cruise-control has returned a async request


efeg:
-- thanks @mgrubent for the feature request.

Currently, human operators ask cruise-control for information about Kafka clusters (usually via `curl`), then view the response to see if cruise-control has returned a valid response.

Humans know that a response like the following means that cruise-control is still computing its response.

```
Get broker stats:
(     2 ms) - (100.0%) PENDING: Operation enqueued, waiting to be executed.
(     0 ms) - (100.0%) WAITING_FOR_CLUSTER_MODEL: The job requires a cluster model and it is waiting to get the cluster model lock.
[...]
```

And humans know that a response like the following means that cruise-control has returned its final response

```
                           HOST            DISK(MB)/_(%)_         CPU(%)       LEADER_NW_IN(KB/s)     FOLLOWER_NW_IN(KB/s)        NW_OUT(KB/s)       PNW_OUT(KB/s)    LEADERS/REPLICAS
[...]
```

However, unless users use the `json=true` in their request and check for the expected field in the response, it is non-trivial to say that the response is `final`. Headers do not currently provide this information -- but it might be useful to have a flag to indicate this "finality".

nareshv:
HTTP-202 header can be used for such responses. In [UI](https://github.com/linkedin/cruise-control-ui) we have written code to parse the responses (text/plain, json) and arrive at the right response type. But, having this feature helps the communication between clients & cc.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

emilyki:
# Async Response Design Doc

## Abstract

Design doc that presents a feature to indicate asynchronous response from Cruise Control.

## Motivation

Currently, it is impossible to tell programmatically whether cruise-control is computing the response or not. Example response of such a state can only be recognized, as it looks like the following:

    Get broker stats:
    (     2 ms) - (100.0%) PENDING: Operation enqueued, waiting to be executed.
    (     0 ms) - (100.0%) WAITING_FOR_CLUSTER_MODEL: The job requires a cluster model and it is waiting to get the cluster model lock.

This leads to difficulty in automating clients that require listening and reacting to the cruise-control's response. Moreover, there has been an increase in demand for this feature recently - [cruise-control/pull/783#discussion_r295982202](https://github.com/linkedin/cruise-control/pull/783#discussion_r295982202).

## Goals

[...]

PR #876, opened 2019-07-31
Support in progress response for async requests

emilyki:
This pull request addresses the issue #493.

The change addresses support for the `HTTP 202` response (which indicates that a request was accepted for processing, but was not completed) for Cruise Control asynchronous requests.

This is an example log output from testing the 202 response:
test

### DIFF
--- a/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/servlet/response/CruiseControlResponse.java
+++ b/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/servlet/response/CruiseControlResponse.java
@@ -27,6 +27,15 @@ public interface CruiseControlResponse {
    */
   void writeSuccessResponse(CruiseControlParameters parameters, HttpServletResponse response) throws IOException;
 
+
+  /**
+   * Write in progress response with the given parameters to the provided HTTP response.
+   *
+   * @param parameters Parameters of the HTTP request of user.
+   * @param response HTTP response to return to user.
+   */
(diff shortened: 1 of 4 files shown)


emilyki:
It seems like the build has failed because the unit tests aren't complete yet. I'll address this issue once the proposed implementation/approach is approved. Thanks!

emilyki:
May I ask for your feedback/comments on the approach proposed @efeg @kidkun ? Thanks!

kidkun:
Thanks for the well-organized and detailed design doc at issue #493 !
I think returning http-202 is a good idea to indicate the calculation is not done and inform client to poll later.

Note recently we have a refactor of servlet code and make in-progress response a top-level response(create a Class [ProgressResult](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ProgressResult.java)), I think instead of adding a new `writeInProgressResponse()` in interface(`CruiseControlResponse`) we can directly override method `writeSuccessResponse()` in `ProgressResult` to return `SC_ACCEPTED` in header.

emilyki:
Thanks for your suggestions @kidkun! I'll update the pull request with the new approach (as soon as my final exams are done, by the end of this week).

emilyki:
Hello @kidkun and @efeg, I've rebased the branch to the latest master and updated the implementation to override the `writeSuccessResponse` method of `ProgressResult` instead. May I ask for another review on this? Thanks!

Issue #875, opened 2019-07-31
Feature Request: cruise-control topic_configuration should accept topic regex as POST data


mgrubent:
# Background
`cruise-control`'s `topic_configuration` endpoint accepts a parameter, `topic`, whose value is a regular expression that specifies which topics to operate on.

When the URI passed to `cruise-control` is too long, `cruise-control` responds with
```
<h1>Bad Message 414</h1><pre>reason: URI Too Long</pre>
```
# Issue
For long topic regexes, it is not possible in a single session with `cruise-control`, to start a `topic_configuration` proposal execution, without somehow shrinking the topic regex.  
This is common in clusters with heterogenous replication factors among topics, where there is no short regular expression that can arbitrarily select any set of topics in the cluster.

For regexes that are constructed as a `|`-delimited list of topic names, this means either that:  
1. Clients must know the longest URI that `cruise-control` will accept ahead of time, and only construct URIs that are shorter than this, or
2. Clients must half their requests to `topic_configuration` each time they receive a `414` from `cruise-control`.

The effect of either is that a client is not able to start a single `cruise-control` `topic_configuration` proposal given a single too-long list of topics.  
This greatly inhibits the programmatic use of `cruise-control` as, rather than a single response, clients must serialize their requests and manage numerous sequential long-running tasks.

# Proposal
For the `topic_configuration` endpoint, there should be some way to select a very-large set of topics, in the course of a single `cruise-control` session.

There are a few different viable approaches for this:  
1. Allow for the topic regex to be sent in the data payload of the `topic_configuration` POST.  Any topic regex sent in this way would override any 'topic=value' parameter in the URI.  
2. Allow for some list or set of exact topic names to be sent in the data payload of the `topic_configuration` POST, if for some reason long regexes in the implementation above are disfavored.  `cruise-control` could then parse this into a `set` and do topic matching against this set.  Yes, regular expressions provide a superset of this functionality.
3. Implement a small 'filter' function on the server side of `topic_configuration`.  This would allow the user to pass logic like `filter=rf_leq2`, to select all topics whose replication factor is less than or equal to 2.  One major drawback of this approach is that if for some reason the replication factor of this set of topics needs to be increased yet again, it may be that those topics cannot be selected by the use of any replication-factor-based filter, for instance if all the topics in the cluster now have the same replication factor.  
4. Do nothing.  Require clients to implement recursion logic.  This is an option, but as noted above it places a high burden on clients seeking to perform replication factor operations programmatically.

[...]

PR #877, opened 2019-07-31
Specify topic to change replication factor in request body to topic_configuration endpoint.

kidkun:
Fix issue #875

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -1211,6 +1211,15 @@ public class KafkaCruiseControl {
     if (topicsToChangeByReplicationFactor.isEmpty()) {
       throw new IllegalStateException("All Topics matching given pattern already have target replication factor.");
     }
+    // Sanity check that no topic is set with more than one target replication factor.
+    Set<String> topicsToChange = new HashSet<>();
+    for (Set<String> topics : topicsToChangeByReplicationFactor.values()) {
+      for (String topic : topics) {
+        if (!topicsToChange.add(topic)) {
+          throw new IllegalStateException(String.format("Topic %s is requested with more than one target replication factor.", topic));
+        }
+      }
+    }
     return topicsToChangeByReplicationFactor;
   }
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -1086,8 +1086,8 @@ public class KafkaCruiseControl {
    * If partition's current replication factor is larger than target replication factor, remove one or more follower replicas
    * from the partition. Replicas are removed following the reverse order of position in partition's replica list.
    *
-   * @param topicPatternsByReplicationFactor The name patterns of topics to apply the change with the target replication factor.
-   *                                         If no topic in the cluster matches the pattern, an exception will be thrown.
+   * @param topicPatternByReplicationFactor The name patterns of topic to apply the change with the target replication factor.
+   *                                        If no topic in the cluster matches the patterns, an exception will be thrown.
    * @param goals The goals to be met during the new replica assignment. When empty all goals will be used.
    * @param skipTopicRackAwarenessCheck Whether ignore rack awareness property if number of rack in cluster is less
    *                                    than target replication factor.
@@ -1111,7 +1111,7 @@ public class KafkaCruiseControl {
    * @return The optimization result.
    * @throws KafkaCruiseControlException When any exception occurred during the topic configuration updating.
    */
-  public OptimizerResult updateTopicConfiguration(Map<Short, Set<Pattern>> topicPatternsByReplicationFactor,
+  public OptimizerResult updateTopicConfiguration(Map<Short, Pattern> topicPatternByReplicationFactor,
                                                   List<String> goals,
                                                   boolean skipTopicRackAwarenessCheck,
                                                   ModelCompletenessRequirements requirements,
@@ -1134,7 +1134,7 @@ public class KafkaCruiseControl {
     Cluster cluster = kafkaCluster();
     // Ensure there is no offline replica in the cluster.
     sanityCheckNoOfflineReplica(cluster);
-    Map<Short, Set<String>> topicsToChangeByReplicationFactor = topicsForReplicationFactorChange(topicPatternsByReplicationFactor, cluster);
+    Map<Short, Set<String>> topicsToChangeByReplicationFactor = topicsForReplicationFactorChange(topicPatternByReplicationFactor, cluster);
 
     // Generate cluster model and get proposal
     OptimizerResult result;
@@ -1184,27 +1184,22 @@ public class KafkaCruiseControl {
     return result;
   }
 
-  private static Map<Short, Set<String>> topicsForReplicationFactorChange(Map<Short, Set<Pattern>> topicPatternsByReplicationFactor,
+  private static Map<Short, Set<String>> topicsForReplicationFactorChange(Map<Short, Pattern> topicPatternByReplicationFactor,
                                                                           Cluster cluster) {
-    Map<Short, Set<String>> topicsToChangeByReplicationFactor = new HashMap<>(topicPatternsByReplicationFactor.size());
-    for (Map.Entry<Short, Set<Pattern>> entry : topicPatternsByReplicationFactor.entrySet()) {
+    Map<Short, Set<String>> topicsToChangeByReplicationFactor = new HashMap<>(topicPatternByReplicationFactor.size());
+    for (Map.Entry<Short, Pattern> entry : topicPatternByReplicationFactor.entrySet()) {
       short replicationFactor = entry.getKey();
-      for (Pattern topicPattern : entry.getValue()) {
-        Set<String> topics = cluster.topics().stream().filter(t -> topicPattern.matcher(t).matches()).collect(Collectors.toSet());
-        // Ensure there are topics matching the requested topic pattern.
-        if (topics.isEmpty()) {
-          throw new IllegalStateException("There is no topic in cluster matching pattern " + topicPattern);
-        }
-        Set<String> topicsToChange = topics.stream()
-                                           .filter(t -> cluster.partitionsForTopic(t).stream().anyMatch(p -> p.replicas().length != replicationFactor))
-                                           .collect(Collectors.toSet());
-        if (!topicsToChange.isEmpty()) {
-          if (topicsToChangeByReplicationFactor.containsKey(replicationFactor)) {
-            topicsToChangeByReplicationFactor.get(replicationFactor).addAll(topicsToChange);
-          } else  {
-            topicsToChangeByReplicationFactor.put(replicationFactor, topicsToChange);
-          }
-        }
+      Pattern topicPattern = entry.getValue();
+      Set<String> topics = cluster.topics().stream().filter(t -> topicPattern.matcher(t).matches()).collect(Collectors.toSet());
+      // Ensure there are topics matching the requested topic pattern.
+      if (topics.isEmpty()) {
+        throw new IllegalStateException("There is no topic in cluster matching pattern " + topicPattern);
+      }
+      Set<String> topicsToChange = topics.stream()
+                                         .filter(t -> cluster.partitionsForTopic(t).stream().anyMatch(p -> p.replicas().length != replicationFactor))
+                                         .collect(Collectors.toSet());
+      if (!topicsToChange.isEmpty()) {
+        topicsToChangeByReplicationFactor.put(replicationFactor, topicsToChange);
       }
     }
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -1063,7 +1063,7 @@ public class KafkaCruiseControl {
   }
 
   /**
-   * Update configuration of topics which match topic pattern. Currently only support changing topic's replication factor.
+   * Update configuration of topics which match topic patterns. Currently only support changing topic's replication factor.
    *
    * If partition's current replication factor is less than target replication factor, new replicas are added to the partition
    * in two steps.
@@ -1086,10 +1086,9 @@ public class KafkaCruiseControl {
    * If partition's current replication factor is larger than target replication factor, remove one or more follower replicas
    * from the partition. Replicas are removed following the reverse order of position in partition's replica list.
    *
-   * @param topicPattern The name pattern of topics to apply the change. If no topic in the cluster matches the pattern, an
-   *                     exception will be thrown.
+   * @param topicPatternsByReplicationFactor The name patterns of topics to apply the change with the target replication factor.
+   *                                         If no topic in the cluster matches the pattern, an exception will be thrown.
    * @param goals The goals to be met during the new replica assignment. When empty all goals will be used.
-   * @param replicationFactor The replication factor to set for the topics.
    * @param skipTopicRackAwarenessCheck Whether ignore rack awareness property if number of rack in cluster is less
    *                                    than target replication factor.
    * @param requirements The cluster model completeness requirements.
@@ -1112,9 +1111,8 @@ public class KafkaCruiseControl {
    * @return The optimization result.
    * @throws KafkaCruiseControlException When any exception occurred during the topic configuration updating.
    */
-  public OptimizerResult updateTopicConfiguration(Pattern topicPattern,
+  public OptimizerResult updateTopicConfiguration(Map<Short, Set<Pattern>> topicPatternsByReplicationFactor,
                                                   List<String> goals,
-                                                  short replicationFactor,
                                                   boolean skipTopicRackAwarenessCheck,
                                                   ModelCompletenessRequirements requirements,
                                                   OperationProgress operationProgress,
@@ -1136,7 +1134,7 @@ public class KafkaCruiseControl {
     Cluster cluster = kafkaCluster();
     // Ensure there is no offline replica in the cluster.
     sanityCheckNoOfflineReplica(cluster);
-    Set<String> topicsForReplicationFactorChange = topicsForReplicationFactorChange(topicPattern, cluster, replicationFactor);
+    Map<Short, Set<String>> topicsToChangeByReplicationFactor = topicsForReplicationFactorChange(topicPatternsByReplicationFactor, cluster);
 
     // Generate cluster model and get proposal
     OptimizerResult result;
@@ -1149,7 +1147,7 @@ public class KafkaCruiseControl {
                                                                                 : Collections.emptySet();
       Set<Integer> excludedBrokersForReplicaMove = excludeRecentlyRemovedBrokers ? executorState.recentlyRemovedBrokers()
                                                                                  : Collections.emptySet();
-      populateRackInfoForReplicationFactorChange(topicsForReplicationFactorChange, replicationFactor, cluster, excludedBrokersForReplicaMove,
+      populateRackInfoForReplicationFactorChange(topicsToChangeByReplicationFactor, cluster, excludedBrokersForReplicaMove,
                                                  skipTopicRackAwarenessCheck, brokersByRack, rackByBroker);
 
       ClusterModel clusterModel = _loadMonitor.clusterModel(-1,
@@ -1160,7 +1158,7 @@ public class KafkaCruiseControl {
       Map<TopicPartition, List<ReplicaPlacementInfo>> initReplicaDistribution = clusterModel.getReplicaDistribution();
 
       // First try to add and remove replicas to achieve the replication factor for topics of interest.
-      createOrDeleteReplicasInClusterModel(topicsForReplicationFactorChange, brokersByRack, rackByBroker, cluster, replicationFactor, clusterModel);
+      createOrDeleteReplicasInClusterModel(topicsToChangeByReplicationFactor, brokersByRack, rackByBroker, cluster, clusterModel);
 
       // Then further optimize the location of newly added replicas based on goals. Here we restrict the replica movement to
       // only considering newly added replicas, in order to minimize the total bytes to move.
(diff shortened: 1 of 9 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -1185,24 +1183,36 @@ public synchronized void userTriggeredStopExecution() {
...
+    Map<Short, Set<String>> topicsToChangeByReplicationFactor = new HashMap<>(topicPatternsByReplicationFactor.size());
+    for (Map.Entry<Short, Set<Pattern>> entry : topicPatternsByReplicationFactor.entrySet()) {
+      short replicationFactor = entry.getKey();
+      for (Pattern topicPattern : entry.getValue()) {
+        Set<String> topics = cluster.topics().stream().filter(t -> topicPattern.matcher(t).matches()).collect(Collectors.toSet());
+        // Ensure there are topics matching the requested topic pattern.
+        if (topics.isEmpty()) {
+          throw new IllegalStateException("There is no topic in cluster matching pattern " + topicPattern);
+        }
+        Set<String> topicsToChange = topics.stream()
```
mgrubent: Not sure where is the best place in the method for this suggestion, but something should validate whether there is an intersection between any of the topic sets built by the different provided patterns.

Otherwise there will be an insidious bug where two patterns are provided that both match `topicA`, and it is not deterministic what replication factor `topicA` will get.

kidkun: The current logic is that if request URL contains topic and replication factor, whatever specified in request body will be ignored. 

The idea is that the two approach to specify request should be mutual-exclusive. I document this in [Javadoc](https://github.com/linkedin/cruise-control/blob/b818bc16eed52544d7b584e5bf7caa6ef3619ad2/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java#L30), once this patch get merged, I will update wiki page about it.

mgrubent: Ah, I might be mis-parsing this method then.  
Does this PR allow for the body to specify more than one replication factor : pattern mapping, or only one?

If only one, then my comment is a non-issue.  
If more than one, then my concern is that there may be a topic set intersection, such that it's ambiguous which replication factor a colliding topic should have.

kidkun: Ah, I see. good point! I will add a sanity check that there will be no more  than one target replication factor for each topic.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
```diff
@@ -17,6 +20,14 @@
  *
  * <ul>
  *   <li>Note that "review_id" is mutually exclusive to the other parameters -- i.e. they cannot be used together.</li>
+ *   <li>If topics to change replication factor and target replication factor is not specified in URL, they can also be
```
efeg: 1. I am worried that this format makes the `topic_configuration` endpoint almost dedicated to replication factor change requests for topics. Is it possible to preserve the generic purpose of this endpoint by adding a relevant `key` to the provided configuration body? -- i.e. can users specify what topic configuration to change first in a `key`, then indicate the relevant configs under that `key`?

2. I think ideally each JSON value should have an associated key that describes what the value is -- i.e. in the current design, it is not clear what the following file indicates (but we know that it is a map of topic patterns by replication factor):
```
{
    3: E.*,
    2: K.*
}
```
Note that with such an explicit naming, users can also specify parameters that this endpoint requires as input, such as `goals`, under the relevant config change request type within the JSON payload.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
```diff
@@ -30,8 +41,7 @@
  * </pre>
  */
 public class TopicConfigurationParameters extends GoalBasedOptimizationParameters {
-  protected Pattern _topic;
-  protected short _replicationFactor;
+  protected Map<Short, Set<Pattern>> _topicPatternsByReplicationFactor;
```
efeg: I am worried that this change makes the `TopicConfigurationParameters` strongly tied to the replication factor change request, which is just an instance of topic configuration change request. Is it possible to abstract away the different type topic configuration requests here? -- i.e. `ReplicationFactorParameters` could be used to bundle changes corresponding to replication factor of topics.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
```diff
@@ -48,12 +58,18 @@ public TopicConfigurationParameters() {
...
     super.initParameters();
-    _topic = ParameterUtils.topic(_request);
-    if (_topic == null) {
-      throw new IllegalArgumentException("Topic to update configuration is not specified.");
+    Pattern topic = ParameterUtils.topic(_request);
+    Short replicationFactor = ParameterUtils.replicationFactor(_request);
+    if (topic != null && replicationFactor != null) {
+      _topicPatternsByReplicationFactor = Collections.singletonMap(replicationFactor, Collections.singleton(topic));
+    } else if (topic == null && replicationFactor == null) {
+      _topicPatternsByReplicationFactor = ParameterUtils.topicPatternsByReplicationFactor(_request);
```
efeg: `topicPatternsByReplicationFactor` -> `proposedConfigurations` / `proposedConfigs` (i.e. something that applies to all topic configurations)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
mgrubent: Would something like the following be sufficiently well-scoped?
```
{
    "replication_factor": [
        {"3": "E.*"},
        {"2": "K.*"}
    ]
}
```
Full JSON payload support for all potentially-passed parameters, even only for the `topic_configuration` endpoint, seems like it is a larger issue, even if it is a valid method of sending information to `cruise-control`'s API.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
kidkun: When I am refactoring the code, I realize that to keep subgroups of parameter inside endpoint's parameter(for `admin` and `topic_configuration` endpoint) become non-trivial so I think it is a good idea to separate the refactor into another patch, I created issue #886 to track it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
kidkun: @efeg  for 1, I feel it might be unnecessary for this patch. What we want to achieve is that if certain parameter is too long, we specify it in request body. So ideally it should be 1:1 mapping with whatever valid parameter for this endpoint. I think it is a good idea that we preserve what parameter is for what purpose(as described in issue #886 ), but explicitly have another layer in request body seems too much

for 2, I agree, I will add a field name `topic_pattern_by_replication_factor` to accurately describe what the value represent.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
kidkun: As we address issue #886 , proposed topic configurations will not be mixed in a signal map, we can address this generalization in that patch.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -111,6 +112,7 @@
   public static final String POPULATE_DISK_INFO_PARAM = "populate_disk_info";
   public static final String BROKER_ID_AND_LOGDIRS_PARAM = "brokerid_and_logdirs";
   public static final String REPLICATION_FACTOR_PARAM = "replication_factor";
+  public static final String TOPIC_BY_REPLICATION_FACTOR_PARAM = "topic_by_replication_factor";
```
mgrubent: I'm a little fuzzy on the desired API currently.

Is `topic_by_replication_factor` intended to signify a mapping from RF (positive integers) to topic regex?

Is the `topic_by_replication_factor` codepath expected to be accessible only from valid JSON submitted in the body of the request (this appears to be what the subsequent `topicPatternByReplicationFactorFromBody` method implements) , or would it potentially also be a new parameter that could be sent to `cruise-control` (I don't see this anywhere, nor do I think this would be a good idea, I just see `TOPIC_BY_REPLICATION_FACTOR_PARAM` next to other existing parameter Strings, and I thought I would ask)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
```diff
@@ -21,13 +19,17 @@
...
- *   specified in request body. The body format will be
+ *   specified in request body. The body format is expected to be valid JSON format. e.g.
+ *    <pre><code>
  *   {
- *     target_replicaiton_factor_1 : topic_regex_1
- *     target_replication_factor_2 : topic_regex_2
- *     ...
- *   }.
- *   If it is specified both in URL and body, what specified in URL will be picked up by Cruise Control.</li>
+ *     topic_by_replication_factor : {
```
mgrubent: Thanks for updating this :)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
mgrubent: After reading the updates to `TopicConfigurationParameters.java`, most of my questions are answered.

I do still have a question for whether `topic_by_replication_factor` is ever potentially a parameter in the URI, even though it's expected to be a key in a JSON body sent to `cruise-control`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
efeg: If we intend to postpone the work to support for updating general topic configurations to https://github.com/linkedin/cruise-control/issues/886, I think we should still strive to make sure that this patch does not introduce conflicting/breaking changes for the intended work.

> So ideally it should be 1:1 mapping with whatever valid parameter for this endpoint.

Achieving a 1-1 mapping may not be possible for several reasons. For example, consider the case where the user intends to change `config-1` of `selected-topics-1` and `config-2` of `selected-topics-2`, where `selected-topics-1` != `selected-topics-2`; hence, `topic` parameter wouldn't be common for these changes.

kidkun: agree,  since we want to support specifying multiple pair of replication factor and topic pattern, we are unable to do 1-1 mapping for `topic` and `replication_factor` parameter. This patch allows client to specify the pairs in body under `topic_by_replication_factor`. the format is 

>    {
      topic_by_replication_factor : {
          target_replication_factor_1 : topic_regex_1,
          target_replication_factor_2 : topic_regex_2,
          ...
      }
    }

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
kidkun: `topic_by_replication_factor` will be only used in request body, `topic_configuration` endpoint's parameter list does not change in this patch.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
```diff
@@ -30,8 +43,7 @@
  * </pre>
  */
 public class TopicConfigurationParameters extends GoalBasedOptimizationParameters {
-  protected Pattern _topic;
-  protected short _replicationFactor;
```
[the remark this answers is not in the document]
kidkun: This patch does not drop `topic` parameter,  actually it introduce no change to current support parameter of `topic_configuration` endpoint.
It just introduce a new field name ,`topic_by_replication_factor`, which user can specify in the request body.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
```diff
@@ -48,13 +60,9 @@ public TopicConfigurationParameters() {
...
   protected void initParameters() throws UnsupportedEncodingException {
     super.initParameters();
-    _topic = ParameterUtils.topic(_request);
-    if (_topic == null) {
-      throw new IllegalArgumentException("Topic to update configuration is not specified.");
-    }
-    _replicationFactor = ParameterUtils.replicationFactor(_request);
-    if (_replicationFactor < 1) {
-      throw new IllegalArgumentException("Target replication factor cannot be set to smaller than 1.");
+    _topicPatternByReplicationFactor = ParameterUtils.topicPatternByReplicationFactor(_request);
```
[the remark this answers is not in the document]
kidkun: This patch bring in new parameter, `topic_by_replication_factor` is just a new field user can specify in `JSON` body.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
mgrubent: Thanks Kun, all of my questions here have been addressed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -1204,17 +1205,11 @@ public OptimizerResult updateTopicConfiguration(Map<Short, Pattern> topicPattern
     }
 
     if (topicsToChangeByReplicationFactor.isEmpty()) {
-      throw new IllegalStateException("All Topics matching given pattern already have target replication factor.");
+      throw new IllegalStateException(String.format("All topics matching given pattern already have target replication factor. requested "
```
efeg: Nit: `requested` -> `Requested`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -406,4 +406,20 @@ public static void sanityCheckBrokersHavingOfflineReplicasOnBadDisks(List<String
...
+   * Check that only one target replication factor is set for each topic.
+   *
+   * @param topicsToChangeByReplicationFactor Topics to change replication factor by target replication factor.
+   */
+  public static void sanityCheckTargetReplicationFactorForTopic(Map<Short, Set<String>> topicsToChangeByReplicationFactor) {
+    Set<String> topicsToChange = new HashSet<>();
+    for (Set<String> topics : topicsToChangeByReplicationFactor.values()) {
+      for (String topic : topics) {
+        if (!topicsToChange.add(topic)) {
+          throw new IllegalStateException(String.format("Topic %s is requested with more than one target replication factor.", topic));
```
efeg: For ease of fixing the issue in one shot, is it possible to print all topic names, which specify multiple replication factors? -- i.e. the current message prints the first violation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -1186,24 +1185,35 @@ public OptimizerResult updateTopicConfiguration(Pattern topicPattern,
     return result;
   }
 
-  private static Set<String> topicsForReplicationFactorChange(Pattern topicPattern, Cluster cluster, short replicationFactor) {
-    Set<String> topics = cluster.topics().stream().filter(t -> topicPattern.matcher(t).matches()).collect(Collectors.toSet());
-    // Ensure there are topics matching the requested topic pattern.
-    if (topics.isEmpty()) {
-      throw new IllegalStateException("There is no topic in cluster matching pattern " + topicPattern);
+  private static Map<Short, Set<String>> topicsForReplicationFactorChange(Map<Short, Pattern> topicPatternByReplicationFactor,
```
efeg: Nit: Should we move this function to the util class?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -575,6 +577,52 @@ static Pattern topic(HttpServletRequest request) {
...
+      }
+      return topicPatternByReplicationFactorFromBody;
+    } else {
+      if (topic == null && replicationFactor != null) {
+        throw new UserRequestException("Topic is not specified in URL while target replication factor is specified.");
+      }
+      if ((topic != null && replicationFactor == null)) {
+        throw new UserRequestException("Topic's replication factor is not specified in URL while subject topic is specified.");
+      }
+      if (topic != null && replicationFactor != null) {
```
efeg: Redundant check for ` replicationFactor != null`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -575,6 +577,52 @@ static Pattern topic(HttpServletRequest request) {
     return parameterString == null ? null : Pattern.compile(request.getParameter(parameterString));
   }
 
+  @SuppressWarnings("unchecked")
+  static Map<Short, Pattern> topicPatternByReplicationFactorFromBody(HttpServletRequest request) {
```
efeg: Would it be possible to make this function `private` -- i.e. is it expected to be used by a function other than `topicPatternByReplicationFactor()`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -575,6 +577,52 @@ static Pattern topic(HttpServletRequest request) {
...
+    Map<Short, Pattern> topicPatternByReplicationFactor;
+    try {
+      Gson gson = new Gson();
+      Map<String, Object> json = gson.fromJson(request.getReader(), Map.class);
+      if (!json.containsKey(TOPIC_BY_REPLICATION_FACTOR)) {
+        return null;
+      }
+      topicPatternByReplicationFactor = new HashMap<>();
+      for (Map.Entry<String, String> entry : ((Map<String, String>) json.get(TOPIC_BY_REPLICATION_FACTOR)).entrySet()) {
+        Short replicationFactor = Short.parseShort(entry.getKey().trim());
```
efeg: Nit: Any reason why we are not using a `short` rather than `Short`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
efeg: I am concerned that if we avoid adding a top-layer `request-type` (e.g. `replication_factor`) in this patch, any followup patch (e.g. https://github.com/linkedin/cruise-control/issues/886) will need to introduce backwards incompatible changes in the payload structure.

Here's what I propose:

```
{
    replication_factor: {
        topic_by_replication_factor : {
            target_replication_factor_1 : topic_regex_1
            target_replication_factor_2 : topic_regex_2,
            ...
    }
}
```

kidkun: Had an offline discussion, we want to support specifying multiple different configuration changes for different topics via the same POST body. 
Thus having an extra layer of `request-type` helps to organize parameter for different sub-requests in one HTTP request.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -406,4 +407,59 @@ public static void sanityCheckBrokersHavingOfflineReplicasOnBadDisks(List<String
   public static boolean isKafkaAssignerMode(Collection<String> goals) {
     return goals.stream().anyMatch(KAFKA_ASSIGNER_GOALS::contains);
   }
+
+  /**
+   * Check that only one target replication factor is set for each topic.
+   *
+   * @param topicsToChangeByReplicationFactor Topics to change replication factor by target replication factor.
+   */
+  public static void sanityCheckTargetReplicationFactorForTopic(Map<Short, Set<String>> topicsToChangeByReplicationFactor) {
```
efeg: Nit: Should this function be `private`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -1221,83 +1203,87 @@ private static void populateRackInfoForReplicationFactorChange(Set<String> topic
...
-   * @param topics The potential topics to modify replication factor.
+   * @param topicsByReplicationFactor The topics to modify replication factor with target replication factor.
    * @param brokersByRack A map from rack to broker.
    * @param rackByBroker A map from broker to rack.
    * @param cluster The metadata of the cluster.
-   * @param replicationFactor The replication factor to set for the topics.
    * @param clusterModel The cluster model fo the cluster.
    */
-  public static void createOrDeleteReplicasInClusterModel(Set<String> topics,
+  public static void createOrDeleteReplicasInClusterModel(Map<Short, Set<String>> topicsByReplicationFactor,
```
efeg: Should we move this method to the relevant util class?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java
kidkun: @mgrubent FYI, the final POST body format have another layer to specify `request-type`. As Efe mentioned, since `admin` and `topic_configuration` are "multi-purpose" endpoints, another layer helps to clarify in body which parameter goes to which sub-request.

mgrubent: @kidkun Got it; I think with this schema you actually allow for all types of information which are currently passed through a combination of endpoint and parameter-value mappings, to be passed through the POST body instead.

I'm not sure how far you want to allow the use of the POST body, from an API perspective, but I agree that this additional `request-type` wrapper solves the narrow problem of allowing for a sufficiently-well-scoped passing of `target_replication_factor : topic_regex` mappings

kidkun: So far, `target_replication_factor : topic_regex` is the only thing client can specify in body and that is solely because the regex could be too long.
Currently I do not have plan to parse other parameter from body, but if things changes, like we can do more and more things to these two endpoints, we may consider to support specifying all parameters in the body. Then this hierarchy in body can be beneficial.

PR #880, opened 2019-08-02
Allow easier passing of data payload

mgrubent:
# Background
Currently, there is no support for `cruise-control-client` passing a data payload to `cruise-control`.  
Accomplishing this at present would require some deep subclassing of the existing `Responder` classes.

Conceivably (e.g. as an outcome of #875), `cruise-control` will accept data payloads in the future.

# Proposal
This PR seeks to allow for `cruise-control-client` to easily pass arbitrary data payloads to `cruise-control`, by adhering more-closely to how `requests` creates and sends HTTP requests.

In short:  
```python
# Get the endpoint that the parsed args specify
endpoint = get_endpoint(args=args, execution_context=e)

# Get the socket address for the cruise-control we're communicating with
cc_socket_address = args.socket_address

# Define some JSON payload to send to cruise-control
data = {'a': 1}

# Retrieve the response and display it
json_responder = CruiseControlResponder()
response = json_responder.retrieve_response_from_Endpoint(cc_socket_address, endpoint, json=data)
```

# Changes
This PR:  
1. Introduces a new `CruiseControlResponder` class, which is a thin subclass of `requests.Session`, with some convenience methods.
2. Refactors `cccli` to make use of the new `CruiseControlResponder` class and paradigm.
3. Deprecates all previous classes in `Responder.py`, as these are needlessly deep and overcomplicated ways to construct and create responses.
4. Deprecates the `Endpoint.compose_endpoint` method and the `generate_url_from_cc_socket_address` function for similar reasons.
5. Fixes a long-existing bug with the `BooleanParameters` where most of them were just always True 🤦‍♂️
6. Introduces a minor version bump, as this PR introduces new classes and deprecates old ones.

### DIFF
--- a/cruise-control-client/cruisecontrolclient/client/CCParameter/BooleanParameter.py
+++ b/cruise-control-client/cruisecontrolclient/client/CCParameter/BooleanParameter.py
@@ -2,6 +2,7 @@ from typing import Union
 
 from cruisecontrolclient.client.CCParameter.Parameter import AbstractParameter
 
+set_of_choices = {'true', 'false'}
 
 class AbstractBooleanParameter(AbstractParameter):
     def __init__(self, value: Union[bool, str]):
@@ -30,7 +31,7 @@ class ClearMetricsParameter(AbstractBooleanParameter):
     argparse_properties = {
         'args': ('--clearmetrics', '--clear-metrics'),
         # There is no sensible default here; allow for ternary state
-        'kwargs': dict(help=description, type=bool, metavar='BOOL')
+        'kwargs': dict(help=description, choices=set_of_choices, metavar='BOOL')
     }
 
 
@@ -53,7 +54,7 @@ class ExcludeFollowerDemotionParameter(AbstractBooleanParameter):
     argparse_properties = {
         'args': ('--exclude-follower-demotion',),
         # There is no sensible default here; allow for ternary state
-        'kwargs': dict(help=description, type=bool, metavar='BOOL')
+        'kwargs': dict(help=description, choices=set_of_choices, metavar='BOOL')
     }
 
 
@@ -64,7 +65,7 @@ class ExcludeRecentlyDemotedBrokersParameter(AbstractBooleanParameter):
     argparse_properties = {
         'args': ('--exclude-recently-demoted-brokers', '--exclude-recently-demoted', '--exclude-demoted'),
         # There is no sensible default here; allow for ternary state
-        'kwargs': dict(help=description, type=bool, metavar='BOOL')
+        'kwargs': dict(help=description, choices=set_of_choices, metavar='BOOL')
     }
 
 
@@ -75,7 +76,7 @@ class ExcludeRecentlyRemovedBrokersParameter(AbstractBooleanParameter):
     argparse_properties = {
         'args': ('--exclude-recently-removed-brokers', '--exclude-recently-removed', '--exclude-removed'),
         # There is no sensible default here; allow for ternary state
-        'kwargs': dict(help=description, type=bool, metavar='BOOL')
+        'kwargs': dict(help=description, choices=set_of_choices, metavar='BOOL')
     }
 
 
@@ -86,7 +87,7 @@ class IgnoreProposalCacheParameter(AbstractBooleanParameter):
     argparse_properties = {
         'args': ('--ignore-proposal-cache',),
         # There is no sensible default here; allow for ternary state
-        'kwargs': dict(help=description, type=bool, metavar='BOOL')
+        'kwargs': dict(help=description, choices=set_of_choices, metavar='BOOL')
     }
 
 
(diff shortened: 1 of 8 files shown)


cruise-control-client/cruisecontrolclient/client/CCParameter/BooleanParameter.py
```diff
@@ -185,5 +186,5 @@ class VerboseParameter(AbstractBooleanParameter):
     argparse_properties = {
         'args': ('--verbose',),
         # There is no sensible default here; allow for ternary state
-        'kwargs': dict(help=description, type=bool, metavar='BOOL')
+        'kwargs': dict(help=description, choices=set_of_choices, metavar='BOOL')
```
efeg: Are (1) `allow_capacity_estimation`, (2) `skip_rack_awareness_check`, and (3) `dryrun` intentionally inconsistent with the rest of the requests?

cruise-control-client/cruisecontrolclient/client/CCParameter/BooleanParameter.py
```diff
@@ -2,6 +2,7 @@
 
 from cruisecontrolclient.client.CCParameter.Parameter import AbstractParameter
 
+set_of_choices = {'true', 'false'}
```
efeg: Would it be possible to make set of choices `case-insensitive` (unless they already are)?

cruise-control-client/cruisecontrolclient/client/Endpoint.py
```diff
@@ -159,6 +158,27 @@ def compose_endpoint(self) -> str:
         if self.parameter_name_to_value:
             combined_parameter_to_value.update(self.parameter_name_to_value)
 
+        return combined_parameter_to_value
+
+    def compose_endpoint(self) -> str:
+        warnings.warn("This method is deprecated as of 1.0.0, as it needlessly recreates requests functionality. "
```
efeg: Nit (there is one more similar usage): Should the deprecation warning be placed after or before documentation?

cruise-control-client/cruisecontrolclient/client/Responder.py
```diff
@@ -16,21 +22,104 @@
 # To be able to define a multithreaded way of interacting with cruise-control
 from threading import Thread
 
+# To be able to deprecate classes
+import warnings
+
+
+class CruiseControlResponder(requests.Session):
+    """
+    This class is intented to lightly wrap requests' Session class,
```
efeg: Nit: `intented` -> `intended`

cruise-control-client/cruisecontrolclient/client/Responder.py
```diff
@@ -16,21 +22,104 @@
...
+        return response
+
+    def retrieve_response(self, method, url, **kwargs) -> requests.Response:
+        """
+        Returns a final requests.Response object from cruise-control
+        where Response.text is JSON-formatted.
+
+        :return: requests.Response
+        """
+        # cruise-control's JSON response has a 'progress' key in it so long
```
efeg: Just FYI for the upcoming feature: https://github.com/linkedin/cruise-control/pull/876

cruise-control-client/setup.py
```diff
@@ -11,7 +11,7 @@
 
 setuptools.setup(
     name='cruise-control-client',
-    version='0.1.5',
+    version='1.0.0',
```
efeg: Just out of curiosity, why is this a major version bump up?

cruise-control-client/cruisecontrolclient/client/CCParameter/BooleanParameter.py
mgrubent: (whoops, comment then resolve, not the other way around 😛)

Why [yes](https://stackoverflow.com/a/27616814/11280049), it is possible to have `argparse` lowercase the provided values for these switches, so that they will be compared against (and provided to `cruise-control` as) lowercase strings.

db11aacf9f2c826980086e47de59362f602d3429 should provide this functionality.

cruise-control-client/cruisecontrolclient/client/Endpoint.py
mgrubent: While PEP appears to be silent on the issues, I think the current placement actually breaks proper parsing of the docstring.

Accordingly, move the `DeprecationWarning` after the docstring, where applicable.
3a16e82c57936577d2d347235f2ffba06ffe10cf

cruise-control-client/cruisecontrolclient/client/Responder.py
mgrubent: Agreed; 22f8639

cruise-control-client/setup.py
mgrubent: Solely due to the deletion of the `get_responder` function in `cccli.py`, which still works after this PR, just makes use of deprecated classes.

Given that, it is probably better to add this function back in and include a deprecation warning, and relegate to minor version bump.
0a2c3a8

cruise-control-client/cruisecontrolclient/client/CCParameter/BooleanParameter.py
mgrubent: Yes this is intentional.

The reasoning here is that there is a sensible default value for each of these parameters (namely, `false`) which is explicitly passed by `cccli` to `cruise-control` whenever the appropriate endpoint is called.

I am aware that this effectively shadows existing `cruise-control` server-side defaults.

PR #730, opened 2019-05-29
Expose linger.ms and batch.size config of KafkaProducer used in CruiseControlMetricsReporter.

kidkun:
Fix issue #729

PR #882, opened 2019-08-02
Better clarify sampling/reporting frequency validations

stanislavkozlovski:
I see that the validation requires us to have a larger sampling rate than interval, yet the comments say the opposite

This was probably a typo in the original PR:
https://github.com/linkedin/cruise-control/pull/730/files#diff-23050e53166d8fdaaca2ce7299086068R1399

cc @efeg @kidkun


stanislavkozlovski:
Got it. How do you feel if we made it dead simple, like "Ensure we report more often than we sample". I think it's okay for comments to be informal like that

efeg:
> How do you feel if we made it dead simple, like "Ensure we report more often than we sample". I think it's okay for comments to be informal like that

Sure, that sounds better! It might be a little more explicit to specify who "`we`" refers to -- e.g. `Ensure that the metrics reporter reports more often that the sample fetcher samples.`

efeg:
@stanislavkozlovski Would you like to make the discussed changes?

stanislavkozlovski:
Yes, sorry for taking forever to respond

stanislavkozlovski:
Does this sound good?

Issue #884, opened 2019-08-03
Add Cruise Control support for heterogeneous CPU capacity.


efeg:
Kafka clusters may have heterogeneous brokers in terms of their CPU cores / processor counts. However, this heterogeneity breaks the existing implicit assumption of Cruise Control regarding the homogeneous processing capacity of brokers. Hence, in such clusters, the actual distribution of load in the cluster would be different from the one that is perceived by Cruise Control.

Cruise Control should handle brokers with different number of CPU cores.

PR #885, opened 2019-08-03
Enable sample process to take broker CPU cores into account.

efeg:
The part-1 of the work towards addressing the issue: https://github.com/linkedin/cruise-control/issues/884.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/MetricFetcherManager.java
```diff
@@ -79,7 +80,8 @@ public MetricFetcherManager(KafkaCruiseControlConfig config,
    * @param metricDef the metric definitions.
    * @param time        The time object.
    * @param dropwizardMetricRegistry The Metric Registry object.
-   * @param fetchers    A list of metric fetchers.
+   * @param brokerCapacityConfigResolver The resolver for retrieving broker capacities.
+   * @param fetcher Metric fetcher or {@code null} to create one using {@link KafkaCruiseControlConfig#METRIC_SAMPLER_CLASS_CONFIG}.
```
kidkun: Could we use `sampler` for this parameter? `fetcher` reminds me of `MetricFetcher`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
```diff
@@ -57,6 +65,21 @@ void addMetric(CruiseControlMetric metric) {
       brokerLoad.recordMetric(metric);
       return brokerLoad;
     });
+
+    // Compute cached number of cores by broker id if they have not been cached already.
+    Node node = cluster.nodeById(brokerId);
+    _cachedNumCoresByBroker.computeIfAbsent(brokerId, bid -> {
+      BrokerCapacityInfo capacity = _brokerCapacityConfigResolver.capacityForBroker(getRackHandleNull(node), node.host(), bid);
+      // No mapping shall be recorded if capacity is estimated, but estimation is not allowed.
+      return (!_allowCpuCapacityEstimation && capacity.isEstimated()) ? null : capacity.numCpuCores();
```
kidkun: I am thinking should we throw exception when we detect this condition since core information is a necessary thing to calculate cpu utilization. 

Anther thing to consider is should we try to cache core number in this method. For one thing, `AddMetric()` is call frequently; for another thing, the purpose of this method is to process reporter's metrics record. I think we can collect core information in `process()` method.

we can revisit this issue in the next patch.

efeg: > I am thinking should we throw exception when we detect this condition since core information is a necessary thing to calculate cpu utilization.

I believe this information is only relevant when building a `PartitionMetricSample`, which will not be created if (1) there are no partitions in the cluster, or (2) sampling mode is `MetricSampler.SamplingMode.BROKER_METRICS_ONLY`. Hence, this part will be handled when creating partition samples (i.e. also added a `TODO` item for the next patch).

-- addressed the second feedback.

PR #892, opened 2019-08-09
Make CLI compatible with Python 3.6

tomncooper:
Small tweak which switches from using re.Pattern as the expected type of re.compile (which is Python 3.7 only) to using typing.Pattern. 

This makes the CLI runnable using Python 3.6 which many LTS Linux distros are still running.


mgrubent:
Looks good to me; good catch 😄

mgrubent:
FYI for tracking, this has been merged into [`cruise-control-client` `0.2.1`](https://pypi.org/project/cruise-control-client/0.2.1/)

Issue #901, opened 2019-08-13
Add balancedness metric


efeg:
### Balancedness
A cumulative score that represents how well the load distribution in a healthy cluster satisfies the Cruise Control goals on a scale between `0` (worst) to `100.0` (best). The balancedness of a cluster is periodically evaluated with an interval configured by `anomaly.detection.interval.ms` (i.e. default: 1 hour), using the goals configured by `anomaly.detection.goals`. Violation of each such configured goal means a deduction of balancedness score – i.e. if all goals are satisfied, then the balancedness score is `100.0`. 

The balancedness weight of a goal is determined based on two factors:
* `goal priority`: A higher priority goal has a higher balancedness weight. This weight is configurable via `goal.balancedness.priority.weight`.
* `goal strictness`: A hard goal has a different (i.e. typically higher) balancedness weight compared to a soft goal. This weight is configurable via `goal.balancedness.strictness.weight`.

#### Expected Balancedness After an Execution
* An on-demand request (e.g. for a rebalance, add/remove/demote broker operation, PLE, fix-offline-replicas, or RF increase), the corresponding optimization result response will indicate the balancedness score of the cluster before and the expected balancedness after the execution.
* A self-healing execution, the corresponding self-healing result response will indicate the current and the expected balancedness score of the cluster after the execution.
* JMX metric `balancedness-score` will show the periodic evaluation of the balancedness.
* Anomaly detector substate shows the periodic evaluation of the balancedness.

PR #902, opened 2019-08-13
Add balancedness metric.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/901.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -524,11 +524,14 @@ public class KafkaCruiseControlUtils {
   public static Map<String, Double> balancednessCostByGoal(List<Goal> goals, double priorityWeight, double strictnessWeight) {
     if (goals.isEmpty()) {
       throw new IllegalArgumentException("At least one goal must be provided to get the balancedness cost.");
+    } else if (priorityWeight <= 0 || strictnessWeight <= 0) {
+      throw new IllegalArgumentException(String.format("Balancedness weights must be positive (priority:%f, strictness:%f).",
+                                                       priorityWeight, strictnessWeight));
     }
     Map<String, Double> balancednessCostByGoal = new HashMap<>(goals.size());
     // Step-1: Get weights.
     double weightSum = 0.0;
-    double previousGoalPriorityWeight = 1;
+    double previousGoalPriorityWeight = (1 / priorityWeight);
     for (int i = goals.size() - 1; i >= 0; i--) {
       Goal goal = goals.get(i);
       double currentGoalPriorityWeight = priorityWeight * previousGoalPriorityWeight;
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -528,11 +528,14 @@ public class KafkaCruiseControlUtils {
     Map<String, Double> balancednessCostByGoal = new HashMap<>(goals.size());
     // Step-1: Get weights.
     double weightSum = 0.0;
+    double previousGoalPriorityWeight = 1;
     for (int i = goals.size() - 1; i >= 0; i--) {
       Goal goal = goals.get(i);
-      double cost = Math.pow(priorityWeight, (goals.size() - 1 - i)) * (goal.isHardGoal() ? strictnessWeight : 1);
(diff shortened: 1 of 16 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
```diff
@@ -189,11 +180,13 @@ public void run() {
           LOG.warn("Skipping goal violation detection for {} because load completeness requirement is not met.", goal);
         }
       }
-      if (!goalViolations.violatedGoalsByFixability().isEmpty()) {
+      Map<Boolean, List<String>> violatedGoalsByFixability = goalViolations.violatedGoalsByFixability();
+      if (!violatedGoalsByFixability.isEmpty()) {
         _anomalies.add(goalViolations);
       }
+      refreshBalancednessScore(violatedGoalsByFixability);
```
kidkun: If there is dead broker in the cluster, we will not reaches here, looks like currently this balancedness score(in `state` endpoint and the metric) works only under scenario there is not dead broker?

efeg: Good catch -- it should be `0`!

#### Balancedness under Unhealthy Cluster State
A cluster is considered as unhealthy if it contains offline logDirs with replicas. Hence, depending on the replication factor of the corresponding partitions, minISR configuration of the topic, and the client's consistency requirements (e.g. acks=all), clients that produce to, or consume from the corresponding partitions with offline replicas are likely to experience undesired side-effects (e.g. higher latency, data loss). To indicate the severity of the issue, such clusters have "0" balancedness score. Note that similar approach does not apply to on-demand requests, because user-initiated requests  (1) allow manually marking healthy brokers as dead for forcing the removal of replicas from healthy brokers/disks, (2) evaluate the balancedness of each goal cumulatively (i.e. based on the incrementally optimized cluster state) rather than individually (i.e. based on the same initial cluster state) for optimization.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -528,11 +528,14 @@ public static void populateRackInfoForReplicationFactorChange(Map<Short, Set<Str
     Map<String, Double> balancednessCostByGoal = new HashMap<>(goals.size());
     // Step-1: Get weights.
     double weightSum = 0.0;
+    double previousGoalPriorityWeight = 1;
     for (int i = goals.size() - 1; i >= 0; i--) {
       Goal goal = goals.get(i);
-      double cost = Math.pow(priorityWeight, (goals.size() - 1 - i)) * (goal.isHardGoal() ? strictnessWeight : 1);
+      double currentGoalPriorityWeight = priorityWeight * previousGoalPriorityWeight;
```
kidkun: Looks with this change, the behavior changes.
for the last goal(first iteration), its weight should be 1.0(original it is `pow(priorityWeight, 0)`, but now it is `priorityWeight`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -781,15 +781,16 @@
...
 
   /**
    * <code>goal.balancedness.strictness.weight</code>
    */
   public static final String GOAL_BALANCEDNESS_STRICTNESS_WEIGHT_CONFIG = "goal.balancedness.strictness.weight";
-  private static final String GOAL_BALANCEDNESS_STRICTNESS_WEIGHT_DOC = "The impact of strictness on the relative "
-       + " balancedness score. For example, 1.5 means that a hard goal will have the 1.5x balancedness weight of a soft "
-       + "goal (assuming goal.balancedness.priority.weight is 1).";
+  private static final String GOAL_BALANCEDNESS_STRICTNESS_WEIGHT_DOC = "The impact of strictness (i.e. hard or soft goal)"
+       + " on the relative  balancedness score. For example, 1.5 means that a hard goal will have the 1.5x balancedness "
```
kidkun: nit: redundant white space here(double whitespace)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
efeg: Good catch!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
efeg: -- went over the code to fix similar issues globally.

PR #904, opened 2019-08-14
Add support for text responses

### LINKED ISSUE #873: cruise-control-client should not swallow error codes
# Background
`cruise-control` implements a REST API.

As such, there are all sorts of HTTP codes that it can return (for example, a [414](https://http.cat/414)).

# Issue
Unfortunately, `cruise-control-client`'s internal logic _always_ attempts to parse the response from `cruise-control` as JSON.

The implicit assumption here (that `cruise-control` will always return a text response, parseable as JSON) is incorrect.

One clear place where this is incorrect is if users of `cruise-control-client` tries to pass a too-long URI to `cruise-control`, to change the replication factor of lots of topics via the `topic_configuration` endpoint.

This results in the following ugly stack trace:  
```
Traceback (most recent call last):
  File "/export/apps/python/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/export/apps/python/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/bin/thing.pyz/__main__.py", line 3, in <module>
  File "/usr/bin/thing.pyz/_bootstrap/__init__.py", line 181, in bootstrap
  File "/usr/bin/thing.pyz/_bootstrap/interpreter.py", line 52, in execute_interpreter
  File "/usr/bin/thing.pyz/_bootstrap/interpreter.py", line 29, in execute_content
  File "/usr/bin/thing.pyz/_bootstrap/interpreter.py", line 12, in _exec_function
  File "./rf_change_executor.py", line 145, in <module>
    main()
  File "/place/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/place/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/place/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/place/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "./rf_change_executor.py", line 141, in main
    set_rf(fabric, tag, topic_regex, rf, moves=moves, execute=execute)
  File "./rf_change_executor.py", line 86, in set_rf
    response = json_responder.retrieve_response()
  File "/place/site-packages/cruisecontrolclient/client/Responder.py", line 125, in retrieve_response
    while 'progress' in response.json().keys():
  File "/place/site-packages/requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "/place/site-packages/simplejson/__init__.py", line 518, in loads
    return _default_decoder.decode(s)
  File "/place/site-packages/simplejson/decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "/place/site-packages/simplejson/decoder.py", line 400, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

The offending line is, of course, 
```
  File "/place/site-packages/cruisecontrolclient/client/Responder.py", line 125, in retrieve_response
    while 'progress' in response.json().keys():
```

# Proposal
The `Responder` class in `cruise-control-client` should not falsely assume that `response.json()` will succeed.

It should probably still bubble up the `Requests.Response` to the caller, since the caller would need to examine this `Response` for its `status_code`

### LINKED ISSUE #898: Support in-progress async response logic in cruise-control-client
# Background
With the merging of #876, `cruise-control` now has much better support for knowing when a received response is final.

`cruise-control-client` checks whether a received response is final by examining the body of the response.
# Proposal
`cruise-control-client` should make use of the functionality implemented in #876 to check whether a received response is final.

Optionally also, `cruise-control-client` should support `json=false` queries to `cruise-control`, since examining the `json` response for finality should not be required after #876.

mgrubent:
Mostly fixes #898 
Also fixes #873 

# Changes
* Minor version bump, since `cccli.py` no longer includes `json=true` parameter by default
* Deprecate every function in `Display.py`, since these functions duplicate the human-readable text functionality of `cruise-control` itself
* Add support in `Display.display_response` for displaying non-JSON responses
* Remove `CruiseControlResponder.send` method override, which caused a correctness error (specifically, that `CruiseControlResponder` would not maintain the session in subsequent requests)
* Add response-finality checks for whether response code is `202`, and which `Cruise-Control-Version` string is returned in the header, in addition to whether the response is JSON (and has a `progress` key) or text

### DIFF
--- a/cruise-control-client/cruisecontrolclient/client/Display.py
+++ b/cruise-control-client/cruisecontrolclient/client/Display.py
@@ -22,6 +22,9 @@ import json
 # To be able to indent the correct level for conveying JSON descent
 import textwrap
 
+# To be able to deprecate functions
+import warnings
+
 
 def get_key_to_display_function() -> Dict[str, Callable]:
     """
@@ -29,6 +32,13 @@ def get_key_to_display_function() -> Dict[str, Callable]:
 
     :return: a mapping of keys to expected functions
     """
+    warnings.warn("This function is deprecated as of 0.3.0, as cruise-control already "
+                  "provides human-readable text when supplied the parameter json=false. "
+                  "Please ensure that your Cruise-Control-Version is 2.0.61 or higher, "
+                  "and use json=false for retrieving human-readable text. "
+                  "This function may be removed entirely in future versions.",
+                  DeprecationWarning,
+                  stacklevel=2)
     return {
         'AnomalyDetectorState': display_anomaly_detector_state,
         'brokers': display_brokers,
@@ -51,6 +61,13 @@ def print_with_indent(s: str, level=0) -> None:
     :param level:
     :return:
     """
+    warnings.warn("This function is deprecated as of 0.3.0, as cruise-control already "
+                  "provides human-readable text when supplied the parameter json=false. "
+                  "Please ensure that your Cruise-Control-Version is 2.0.61 or higher, "
+                  "and use json=false for retrieving human-readable text. "
+                  "This function may be removed entirely in future versions.",
+                  DeprecationWarning,
+                  stacklevel=2)
     print_error(textwrap.indent(s, "\t" * level))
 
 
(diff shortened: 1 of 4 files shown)


mgrubent:
This PR looks like it would also fix #873

Issue #886, opened 2019-08-05
Better process request of general-purpose endpoints


kidkun:
Currently in Cruise Control, `admin` and `topic configuration` endpoints are designed as a general-purpose endpoint, which means request sent to these endpoints can be targeting different purposes. 

For example, an request sent to `admin` endpoint may want to 
1. change operation concurrency
2. drop recently removed/demoted brokers
3. enable/disable self-healing for certain anomaly
4. combination of above

The processing logic for these endpoints could be more modularized, which includes
1. group parameters for different purpose into different parameter group under the endpoint
2. based on the whether parameter for different parameter group is set in request or not, optionally execute handling logic and return response.

Issue #795, opened 2019-06-22
Wrap optimization options in user request to avoid tedious passing down to optimizer


kidkun:
For endpoints like `add_broker`/`remove_broker`/`rebalance`, there are more and more optimization options can be specified to maximize the flexibility of different operation requirement. At the same time, it make the code long and error-prone to pass these options from servlet all the way down to optimizer.
 It should be good to wrap them in servlet to centralize management of the options.

PR #907, opened 2019-08-16
Group parameters of admin and topic_configuration endpoint into different request types.

kidkun:
Fix issue #886 
This patches makes parameter handling logic for `admin` and `topic_configuration` endpoint more modular and easier to extend for new purpose.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/AdminParameters.java
```diff
@@ -27,14 +29,11 @@
...
-  protected Integer _concurrentInterBrokerPartitionMovements;
-  protected Integer _concurrentIntraBrokerPartitionMovements;
-  protected Integer _concurrentLeaderMovements;
   protected Integer _reviewId;
-  protected Set<Integer> _dropRecentlyRemovedBrokers;
-  protected Set<Integer> _dropRecentlyDemotedBrokers;
+  protected Map<String, ?> _configs;
+  protected Optional<DropRecentBrokersParameters> _dropBrokersParameters;
+  protected Optional<UpdateSelfHealingParameters> _updateSelfHealingParameters;
+  protected Optional<ChangeExecutionConcurrencyParameters> _changeExecutionConcurrencyParameters;
```
efeg: [There are multiple use cases of this pattern] As indicated by the `Java` language architect `Brian Goetz`, we should almost never use `Optional` as a field of something. He also says that:

```
Of course, people will do what they want.
But we did have a clear intention when adding this feature, and it was not to be a general purpose Maybe type, as much as many people would have liked us to do so.
Our intention was to provide a limited mechanism for library method return types where there needed to be a clear way to represent "no result", and using null for such was overwhelmingly likely to cause errors.
```
*Source: https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555

I believe we can be more conservative about the use of `Optional` -- i.e. avoid using it when it is not absolutely needed. In our case, is using `null` to check for no result cases possible?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/UpdateTopicConfigurationRunnable.java
```diff
@@ -9,74 +9,55 @@
...
-                                                                                   _replicationThrottle,
-                                                                                   _excludeRecentlyDemotedBrokers,
-                                                                                   _excludeRecentlyRemovedBrokers,
-                                                                                   _dryRun,
-                                                                                   _uuid),
-                                  _config);
+    if (_topicReplicationFactorChangeParameters != null) {
+      return new OptimizationResult(
+          _kafkaCruiseControl.updateTopicReplicationFactor(_topicReplicationFactorChangeParameters.topicPatternByReplicationFactor(),
+                                                           _topicReplicationFactorChangeParameters.goals(),
```
efeg: I am wondering whether we can restructure the hierarchy, such that it would be possible to compress `13` parameters in a single `_topicReplicationFactorChangeParameters` parameter passed to `updateTopicReplicationFactor` function?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -1109,6 +1110,16 @@ static boolean fetchCompletedTask(HttpServletRequest request) {
     return getBooleanParam(request, FETCH_COMPLETED_TASK_PARAM, false);
   }
 
+  /**
+   * Sanity check to ensure that at least one of the optional parameter is not specified.
```
efeg: `not specified` -> `specified`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/UpdateTopicConfigurationRunnable.java
```diff
@@ -9,74 +9,55 @@
...
+                                                           _topicReplicationFactorChangeParameters.skipHardGoalCheck(),
+                                                           _topicReplicationFactorChangeParameters.replicaMovementStrategy(),
+                                                           _topicReplicationFactorChangeParameters.replicationThrottle(),
+                                                           _topicReplicationFactorChangeParameters.excludeRecentlyDemotedBrokers(),
+                                                           _topicReplicationFactorChangeParameters.excludeRecentlyRemovedBrokers(),
+                                                           _topicReplicationFactorChangeParameters.dryRun(),
+                                                           _uuid),
+          _config);
+    }
+    // Never reaches here.
```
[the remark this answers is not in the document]
kidkun: Yes, we have the sanity check in [TopicConfigurationParameters.initParameters()](https://github.com/linkedin/cruise-control/blob/fe41eb90fa22d86b7bd5e3a810ebe192fee8c0ec/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/TopicConfigurationParameters.java#L62) so we do fail early and confident never reach here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/AdminParameters.java
kidkun: Sure, I think that holds. I did hesitate between using `Optional` or not, the reason I choose to use it is because I want to explicitly signal the fact that these `Parameters` object can be null.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/UpdateTopicConfigurationRunnable.java
kidkun: This is a general issue for all the endpoint, I have issue #795 in my to do list, I will address issue as a whole there.

Issue #908, opened 2019-08-16
Fail fast during a rebalance if allow_capacity_estimation is false and the cluster has estimation


efeg:
Currently the failure happens after the optimization of goals with the bad cluster model -- it should be before this process.

PR #914, opened 2019-08-22
Fail fast during a rebalance if allow_capacity_estimation is false and the cluster model has estimation.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/908.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
@@ -148,8 +148,9 @@ public class GoalOptimizer implements Runnable {
             clearCachedProposal();
 
             long start = System.nanoTime();
-            // Proposal precomputation runs with the default topics to exclude.
-            computeCachedProposal();
+            // Proposal precomputation runs with the default topics to exclude, and allows capacity estimation.
+            // TODO: Make allowing/disallowing capacity estimation during proposal precomputation configurable.
+            computeCachedProposal(true);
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -148,8 +148,9 @@ public void run() {
             clearCachedProposal();
 
             long start = System.nanoTime();
-            // Proposal precomputation runs with the default topics to exclude.
-            computeCachedProposal();
+            // Proposal precomputation runs with the default topics to exclude, and allows capacity estimation.
+            // TODO: Make allowing/disallowing capacity estimation during proposal precomputation configurable.
+            computeCachedProposal(true);
```
kidkun: Looks like this will be an issue if we turned on proposal precomputing(`num.proposal.precompute.threads` >0).
consider the scenario that we want a proposal with no capacity estimation and we everything else is default, so we will enter [this method](https://github.com/linkedin/cruise-control/blob/79fbbf77cbefd0ef1c0cfeaa0e47006de94698ec/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java#L264), and we will interrupt the `_proposalPrecomputingSchedulerThread` , but the result this thread finally get is still proposal with estimation?

efeg: That is a good catch -- thanks!

This indeed is an existing issue (not introduced by this patch) that we should fix.

Issue #689, opened 2019-05-06
Features not yet supported in Kafka 2.1+


nat-stripe:
In the README, there is an interesting line:
> Support for Apache Kafka 2.1+ is in progress.

Are some things known to not yet work or be untested with Kafka 2.1+? And are there particular areas where contributions would be appreciated?

aravindvs:
+1 - will be good to know what might not yet work on 2.1+?

prune998:
And what about Kafka 2.3 ?

stanislavkozlovski:
It would be great to understand how far in progress we are and whether contributors can help.

I want to bump it to 2.3 and eventually to 2.4 so I can contribute changes that make use of the new reassignment API in Kafka (KIP-455).

I actually have a PR that bumps it to 2.3 - https://github.com/linkedin/cruise-control/pull/924
Are there any known blockers that may prevent us from merging that PR?

austinnichols101:
I think this is addressed by #990 / #991

PR #924, opened 2019-08-31
Bump Kafka dependency to 2.3 and replace NetworkClient inside MetadataClient with the public AdminClient interface

stanislavkozlovski:
It is better to depend on public interfaces where possible to improve API compatibility with newer Kafka versions.
This patch refactors the MetadataClient, removing the use of the Kafka's internal NetworkClient class and the associated network.client.provider.class config.
It also changes the metadata.max.age.ms config from a Long to an Integer, as the AdminClient interface uses ints for its timeouts


stanislavkozlovski:
Hey everybody, as you might know, there is a KIP (KIP-455 https://cwiki.apache.org/confluence/display/KAFKA/KIP-455%3A+Create+an+Administrative+API+for+Replica+Reassignment) for improving reassignment in Kafka. I've been working on it and I believe it might land in 2.4. It allows for scheduling/viewing reassignments through a new Admin API with the ability to incrementally reassign partitions (no more run one reassignment and wait for it to complete) + the ability to properly revert an on-going reassignment.

I'd like to contribute changes to Cruise Control that make use of this new API.

To begin implementing that, we'd need to bump the Kafka dependency to the latest version (2.3). The changes from 2.0->2.3 broke the current Cruise Control MetadataClient implementation.
I thought I would replace that class to use an AdminClient rather than the non-public NetworkClient, as it is generally better to depend on well-tested public interfaces for better compatibility.

Let me know if you are okay with bumping the dependency and if so, whether I should target the `migrate_to_kafka_2_0` branch.
I believe the changes here will work with Kafka 2.0

stanislavkozlovski:
cc @efeg - what are your thoughts on this PR?

efeg:
@stanislavkozlovski thanks for the patch and sorry for the delay! -- will try to review the PR this week.

stanislavkozlovski:
Thanks for the response. Sounds good to me

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -4,123 +4,164 @@
...
-  public synchronized void close() {
-    _networkClient.close();
+  private Cluster doRefreshMetadata(int timeout) throws InterruptedException, ExecutionException, TimeoutException {
+    // We use both AdminClient timeout options and `future.get(timeout)`
+    // The former ensures the timeouts are enforced in the brokers
+    int remaining = timeout;
+    long start = _time.milliseconds();
+
+    Set<String> topicNames = _adminClient.listTopics(new ListTopicsOptions().timeoutMs(remaining))
+        .names().get(remaining, TimeUnit.MILLISECONDS);
```
efeg: Overall timeout can potentially be more than `remaining`, because 
1) `_adminClient.listTopics(new ListTopicsOptions().timeoutMs(remaining))` can take up to `remaining` time, and
2) `.names().get(remaining, TimeUnit.MILLISECONDS);` can take up to remaining timeout, as well.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -4,123 +4,164 @@
...
+    long end = _time.milliseconds();
+    long elapsedTimeMs = end - start;
+    remaining -= elapsedTimeMs;
+
+    DescribeClusterResult result = _adminClient.describeCluster(new DescribeClusterOptions().timeoutMs(remaining));
+    Collection<Node> nodes = result.nodes().get(remaining, TimeUnit.MILLISECONDS);
+    String clusterId = result.clusterId().get(remaining, TimeUnit.MILLISECONDS);
+    Node controller = result.controller().get(remaining, TimeUnit.MILLISECONDS);
```
efeg: (Similar to the issue above) Remaining timeout is not adjusted properly between get calls to future.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -4,123 +4,164 @@
...
+    elapsedTimeMs = _time.milliseconds() - end;
+    remaining -= elapsedTimeMs;
+
+    Map<String, TopicDescription> topicDescriptions = _adminClient.describeTopics(topicNames, new DescribeTopicsOptions().timeoutMs(remaining))
+        .all().get(remaining, TimeUnit.MILLISECONDS);
```
efeg: (Similar to the issue above) Overall timeout can potentially be more than `remaining`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -4,123 +4,164 @@
...
+    Set<String> internalTopics = new HashSet<>();
+
+    for (Map.Entry<String, TopicDescription> topicDescription: describeTopicResult.entrySet()) {
+      if (topicDescription.getValue().isInternal()) {
+        internalTopics.add(topicDescription.getKey());
+      }
+
+      for (TopicPartitionInfo topicPartitionInfo : topicDescription.getValue().partitions()) {
+        PartitionInfo partitionInfo = new PartitionInfo(topicDescription.getKey(),
+            topicPartitionInfo.partition(),
```
efeg: Nit: Bad indentation

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -4,123 +4,164 @@
...
+            topicPartitionInfo.leader(),
+            topicPartitionInfo.replicas().toArray(new Node[0]),
+            topicPartitionInfo.isr().toArray(new Node[0]));
+        partitionInfos.add(partitionInfo);
+      }
+    }
+    return new Cluster(clusterId, nodes, partitionInfos, Collections.emptySet(),
+        internalTopics, controller);
```
efeg: Nit: Bad indentation

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -4,123 +4,164 @@
...
+    List<PartitionInfo> partitionInfos = new LinkedList<>();
+    Set<String> internalTopics = new HashSet<>();
+
+    for (Map.Entry<String, TopicDescription> topicDescription: describeTopicResult.entrySet()) {
+      if (topicDescription.getValue().isInternal()) {
+        internalTopics.add(topicDescription.getKey());
+      }
+
+      for (TopicPartitionInfo topicPartitionInfo : topicDescription.getValue().partitions()) {
+        PartitionInfo partitionInfo = new PartitionInfo(topicDescription.getKey(),
```
efeg: How does this call capture offline replicas?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -4,123 +4,164 @@
...
-    return new ClusterAndGeneration(_metadata.fetch(), _metadataGeneration.get());
+    return new ClusterAndGeneration(_cluster, _metadataGeneration.get());
   }
 
-  /**
-   * Close the metadata client. Synchronized to avoid interrupting the network client during a poll.
-   */
-  public synchronized void close() {
-    _networkClient.close();
+  private Cluster doRefreshMetadata(int timeout) throws InterruptedException, ExecutionException, TimeoutException {
```
efeg: Looks like retrieving `Cluster` requires 3X more NW round trip with this patch compared to the current way of getting metadata -- is it possible to reduce this overhead?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/common/MetadataClientTest.java
```diff
@@ -0,0 +1,291 @@
+/*
+ * Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
```
efeg: Nit: `2017` -> `2019`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/common/MetadataClientTest.java
```diff
@@ -0,0 +1,291 @@
...
+    @Override
+    protected void assertElementsEqual(Object expected, Object actual) {
+      PartitionInfo expectedInfo = (PartitionInfo) expected;
+      PartitionInfo actualInfo = (PartitionInfo) actual;
+      Assert.assertEquals(expectedInfo.topic(), actualInfo.topic());
+      Assert.assertEquals(expectedInfo.partition(), actualInfo.partition());
+      Assert.assertEquals(expectedInfo.leader(), actualInfo.leader());
+      Assert.assertArrayEquals(expectedInfo.replicas(), actualInfo.replicas());
+      Assert.assertArrayEquals(expectedInfo.inSyncReplicas(), actualInfo.inSyncReplicas());
+      Assert.assertArrayEquals(expectedInfo.offlineReplicas(), actualInfo.offlineReplicas());
```
efeg: Can we add test cases to verify offline replicas, logdirs, brokers, and partitions in response?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
stanislavkozlovski: I can't think of a way to reduce the NW round trips unless we are okay with not getting the latest data each time. e.g the DescribeCluster call can be cached for less frequent calls.

Do you think 2 extra round trips are a blocker? I would imagine CC is deployed close to the Kafka cluster and round trips would be pretty fast.

I was wondering why we default to refreshing the metadata every 10 seconds, both within the LoadMonitor and the executor (`EXECUTION_PROGRESS_CHECK_INTERVAL_MS_CONFIG` default).
Is that not too often anyway?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
stanislavkozlovski: `topicPartitionInfo.replicas()` would return any replicas that are offline, unless you are talking about something else

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
stanislavkozlovski: I don't think that's the case because `_adminClient.listTopics(new ListTopicsOptions().timeoutMs(remaining))` simply enqueues the call in the background. The timeout there tells the background thread to give up after `remaining` (+ also pass the timeoutMs in the request, if applicable)

`.names().get(remaining, TimeUnit.MILLISECONDS);` is the blocking code which will not take more than `remaining`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
stanislavkozlovski: I think the same thing applies here - `_adminClient.describeCluster(new DescribeClusterOptions().timeoutMs(remaining))` simply instructs the background thread to give up after `remaining`.

The consequent three `get(remaining, TimeUnit.MILLISECONDS)` calls do not adjust the `remaining` because they rely on the fact that only the first get will take any time - the others should be instant because the data has been fetched

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
stanislavkozlovski: ditto response

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -4,123 +4,164 @@
...
+    // We use both AdminClient timeout options and `future.get(timeout)`
+    // The former ensures the timeouts are enforced in the brokers
+    int remaining = timeout;
+    long start = _time.milliseconds();
+
+    Set<String> topicNames = _adminClient.listTopics(new ListTopicsOptions().timeoutMs(remaining))
```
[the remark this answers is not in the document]
stanislavkozlovski: Great catch!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -4,123 +4,164 @@
...
-      while (!isMetadataUpdated && remaining > 0) {
-        _metadata.requestUpdate();
-        long start = _time.milliseconds();
-        _networkClient.poll(remaining, start);
-        remaining -= (_time.milliseconds() - start);
-        isMetadataUpdated = _metadata.version() > version;
+    if (_time.milliseconds() >= _lastSuccessfulUpdate + _metadataTTL) {
+      // Cruise Control always fetches metadata for all the topics.
+      boolean isMetadataUpdated = false;
+      Optional<Cluster> refreshedCluster = Optional.empty();
```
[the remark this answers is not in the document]
stanislavkozlovski: I added the check for redundancy. I went with `Optional` because I generally prefer it as it avoids NPEs. I don't think it's a big deal when the usage literally a couple of lines below - I'll change to null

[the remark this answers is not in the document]
stanislavkozlovski: In fact, I consolidated the if block into the try

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
stanislavkozlovski: We could consider extending Kafka to support a `describeTopics()` which describes all present topics, including internal ones

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
stanislavkozlovski: Good catch

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/common/MetadataClientTest.java
stanislavkozlovski: I don't see where we have logdirs in the Cluster object. I improved the test to check for all others

stanislavkozlovski:
Thanks for the review! I think I addressed most of the nit comments, the two important points re internal topics and offline replicas and iterated on the tests. Two open questions:

- What should the indentation be? I see that in other code you align variables with the first one passed to the function. e.g
```
func(a,
         b)
```

- Two extra round trips discussion

stanislavkozlovski:
I'll rebase this with master. @efeg do you think we may be able to iterate on this PR next week?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
gwenshap: I think it is better to use the public, supported API that guarantees compatibility.

If we think the public API isn't efficient enough and we need a single API call that will return the list of topics, their descriptions and the cluster description, we can open an improvement KIP to add this to Apache Kafka. 

I agree with Stan that this doesn't have significant impact on Cruise Control users or operations. Right now, the time consuming part is the generation of proposals. And as Stan said, we can reduce the impact farther by reducing the time between metadata refreshes.

stanislavkozlovski:
Rebased with `migrate_to_kafka_2_0`

TaLoN1x:
looking forward for release with this commit. There are lots of useful bug fixes in Kafka 2.3

stanislavkozlovski:
Rebased with `migrate_to_kafka_2_0` again. @efeg would it be possible to get an ETA on this review?

stanislavkozlovski:
@kidkun are we aiming at bumping master to 2.3?
By the time we merge this, we may have 2.4 available

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -1,126 +1,163 @@
...
+        internalTopics.add(topicDescription.getKey());
+      }
+
+      for (TopicPartitionInfo topicPartitionInfo : topicDescription.getValue().partitions()) {
+        PartitionInfo partitionInfo = new PartitionInfo(topicDescription.getKey(),
+            topicPartitionInfo.partition(),
+            topicPartitionInfo.leader(),
+            topicPartitionInfo.replicas().toArray(new Node[0]),
+            topicPartitionInfo.isr().toArray(new Node[0]),
+            topicPartitionInfo.replicas()
```
efeg: `ISR` indicates the subset of the replicas that are in sync, that is caught-up to the leader and ready to take over as leader if the leader should fail. Hence, if a replica is not in sync, it could be either (1) offline, or (2) online, but falling behind. So IIUC, this filter not necessarily yields the offline replicas. Am I missing something?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
efeg: I definitely support/prefer using public APIs when possible. That being said, I have concerns on
1. Consistency of the aggregated metadata,
2. Missing/erroneous `offline replicas` information, and
3. (Now less of a concern) Additional NW cost of retrieving `Cluster`.

I believe ideally, we should potentially consider
* using `PartitionInfo` rather than `TopicPartitionInfo` in `TopicDescription` to capture offline replicas, 
* sticky routing to pick nodes across different calls to admin client inside `doRefreshMetadata`, and 
* considering addition of a one-shot `Cluster` retrieval API in `AdminClient`.

efeg:
> looking forward for release with this commit. There are lots of useful bug fixes in Kafka 2.3

Hi @TaLoN1x 

I am not sure if this commit introduces bug fixes regarding `Kafka 2.3` -- if the intention is to use Cruise Control to manage Kafka 2.3 clusters, it should already be possible to do so today. Would you please clarify your use-case?

austinnichols101:
@efeg 

> if the intention is to use Cruise Control to manage Kafka 2.3 clusters, it should already be possible to do so today

Referring back to issue #689 - the README.md file states that 
`Support for Apache Kafka 2.1+ is in progress.`

Would be good if there was an "official" statement regarding 2.3 support (updated README.md).  Also note that 2.4 is schedule for release on 15-Nov.

We have been holding off upgrades waiting on CC 2.3 support...

efeg:
@austinnichols101 
PR to update Kafka dependency to `2.3` https://github.com/linkedin/cruise-control/pull/991 and the outdated `README.md`.

austinnichols101:
Thanks @efeg!

efeg:
@stanislavkozlovski  Happy new year Stan! :)

Would you be interested in a followup patch to address concerns in https://github.com/linkedin/cruise-control/pull/924#discussion_r340907797?

stanislavkozlovski:
Happy new year!
I'm sorry for forgetting about this. I will try to revisit this in the next two weeks

stanislavkozlovski:
Rebased with master

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -1,126 +1,163 @@
...
+    Collection<Node> nodes = result.nodes().get(remainingMs, TimeUnit.MILLISECONDS);
+    String clusterId = result.clusterId().get(remainingMs, TimeUnit.MILLISECONDS);
+    Node controller = result.controller().get(remainingMs, TimeUnit.MILLISECONDS);
+    elapsedTimeMs = _time.milliseconds() - endMs;
+    remainingMs -= elapsedTimeMs;
+
+    Map<String, TopicDescription> topicDescriptions = _adminClient.describeTopics(topicNames, new DescribeTopicsOptions().timeoutMs(remainingMs))
+        .all().get(remainingMs, TimeUnit.MILLISECONDS);
+
+    return cluster(clusterId, nodes, controller, topicDescriptions);
```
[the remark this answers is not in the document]
stanislavkozlovski: Although rare, that is true. We could have a check that errors out and retries in this case, WDYT?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
```diff
@@ -1,126 +1,163 @@
...
+    int remainingMs = timeoutMs;
+    long startMs = _time.milliseconds();
+
+    Set<String> topicNames = _adminClient.listTopics(new ListTopicsOptions().timeoutMs(remainingMs).listInternal(true))
+        .names().get(remainingMs, TimeUnit.MILLISECONDS);
+    long endMs = _time.milliseconds();
+    long elapsedTimeMs = endMs - startMs;
+    remainingMs -= elapsedTimeMs;
+
+    DescribeClusterResult result = _adminClient.describeCluster(new DescribeClusterOptions().timeoutMs(remainingMs));
```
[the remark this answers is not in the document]
stanislavkozlovski: Yes, our flow of `listTopics->describeCluster->describeTopics` means that each call could return different, stale information.

Thankfully, I think the calls are very tightly related to one another. I envision a small amount of possible scenarios:
* node in describeCluster but not in topic placements - likely no replicas have been placed in the node yet
* node NOT in describeCluster but in topic placements - this is obviously inconsistent and something is wrong. I proposed retrying here
* topic from listTopics is deleted and describeTopics returns an error for it - we should just catch this I think

stanislavkozlovski:
* consistency of the aggregated metadata - I covered the possible inconsistency issues here https://github.com/linkedin/cruise-control/pull/924/files#r369892274 - do these sound exhaustive to you?

* missing/erroneous offline replicas information - Good point that a replica may be offline or not caught up yet. The API does not expose this, unfortunately. I think that for the time being we may infer this by reconciling the describeTopics result with the describeCluster result. Any replica that is not present in ISR and not in the describeCluster result we will infer to be as an offlineReplica

We can do an additional KIP to add offline replicas to the API.

What are your thoughts?

amuraru:
@stanislavkozlovski I am testing latest released version of cruise-control with kafka 2.4.0 (scala 13 build) and all partition reassigments in kafka are stuck. No progress is made when e.g. a broker is removed via CC.
Is there a change in KIP-455 patches that may lead to this? (I saw you also contributed that). Thanks in advance

stanislavkozlovski:
Thanks for reporting this @amuraru. I am not aware of any such issue. The KIP-455 changes are fully backwards compatible and should work fine. I have very briefly tested Cruise Control with Kafka 2.4 and did not see any issues, albeit I was running a manual rebalance via the HTTP API

amuraru:
> The KIP-455 changes are fully backwards compatible and should work fine
Thanks @stanislavkozlovski - I managed to validate this and it's working fine. Initial test was impacted by some other env specific issues in my cluster.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataClient.java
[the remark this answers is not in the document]
efeg: > node in describeCluster but not in topic placements - likely no replicas have been placed in the node yet

I suspect that this might also indicate one of the following scenarios:
* Metadata staleness of the broker that responds to `listTopics` -- e.g. 
   1. The cluster has 2 brokers: `B1` and `B2`.
   2. `B1` has replicas, but `B2` has no replicas.
   3. A topic is created, which puts a replica on `B2`.
   4. `listTopics` request goes to `B1`, which returns topics excluding the new topic, because it hasn't heard about topic creation, yet.
* There was a replica on this node, but it was moved away from it after `describeCluster`, but before `describeTopics`.

> node NOT in describeCluster but in topic placements - this is obviously inconsistent and something is wrong. I proposed retrying here

As you correctly pointed out, this could be due to multiple reasons, including ongoing interactions of broker addition, topic creation, replica moves, and transient broker failures. In case this inconsistency occurs during an ongoing replica move, since this process relies on retrieval of metadata to make progress, I suspect that a retry logic may have adverse affects (e.g. prolonged inconsistencies in retrieved nodes and topic placements due to lack of guarantees on which broker receives the request) on the progress of execution.

> topic from listTopics is deleted and describeTopics returns an error for it - we should just catch this I think

I agree that this is also a possible outcome, and catching this (and retrying) sounds reasonable to me. Again though, in clusters with high topic churn (deletion/creation), this may have similar adverse affect on execution as described on the prev. item.

I guess my other concern is, when we put pieces of metadata together, we are always under the risk of generating an inconsistent world view, not necessarily shared by any broker at a given time.

amuraru:
Based on your comments @efeg @stanislavkozlovski I am wondering if Kafka should provide a new metadata method to return a consistent snapshot of ` listTopics->describeCluster->describeTopics` in single call

efeg:
Closing this PR as it is outdated -- please see https://github.com/linkedin/cruise-control/pull/924#pullrequestreview-354836760 for the discussion regarding proposed improvements in adminClient API, which will likely require a KIP.

As of July 2020, 
```
* The current master branch of Cruise Control is compatible with Apache Kafka 2.0, 2.1, 2.2, and 2.3 (i.e. Releases with 2.0.*)
* The kafka_0_11_and_1_0 branch of Cruise Control is compatible with Apache Kafka 0.11.0.0, 1.0, and 1.1 (i.e. Releases with 0.1.*)
* The migrate_to_kafka_2_4 (development) branch of Cruise Control is compatible with Apache Kafka 2.4 (i.e. Releases with 2.4.*)
* The migrate_to_kafka_2_5 (development) branch of Cruise Control is compatible with Apache Kafka 2.5 (i.e. Releases with 2.5.*)
```

PR #925, opened 2019-09-02
Standarize logic to filter and sort candidate replica of each goal.

kidkun:
Fix issue #802


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
```diff
@@ -185,16 +184,28 @@ static boolean legitMoveBetweenDisks(Replica replica,
    *
    * @param clusterModel The state of the cluster.
    * @param sourceReplica Source replica for intended swap operation.
-   * @param cb Candidate broker containing candidate replicas to swap with the source replica in the order of attempts to swap.
+   * @param candidateReplicas Candidate replicas(from the same candidate broker) to swap with the source replica in the order
```
efeg: `replicas(from ` -> missing space before parenthesis.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Set;
+import java.util.function.Function;
+
+
+/**
+ * A helper class to create new {@link SortedReplicas} for cluster/broker/disk.
+ * This class provides fluent API to enable user to customize {@link SortedReplicas} by adding new selection/priority/score
+ * functions one by one.
+ *
+ * One example is as follows( to register a SortedReplica to clusterModel which includes only follower replicas, prioritize offline
```
efeg: `follows( to` -> Missing space before parenthesis, and redundant space after it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+ * replicas and sort replicas based on replica'disk utilization).
```
efeg: `replica'disk utilization` -> the disk utilization of replicas?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
efeg: Why is the explanation of example is in the parentheses? Is it optional to read it to understand the example?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+    broker.trackSortedReplicas(sortName, new HashSet<>(_selectionFuncs), new ArrayList<>(_priorityFuncs), _scoreFunc);
+  }
+
+  /**
+   * Create {@link SortedReplicas} for each brokers in the given {@link ClusterModel} with added selection/priority/score functions.
+   *
+   * @param sortName The name of {@link SortedReplicas}.
+   * @param clusterModel The cluster model to create {@link SortedReplicas}.
+   */
+  public void trackSortedReplicasFor(String sortName, ClusterModel clusterModel) {
```
efeg: Ideally, all `trackSortedReplicasFor` functions should be using one generic interface (e.g. with the help of generic types). It is OK to leave this to a separate patch.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+  private List<Function<Replica, Integer>> _priorityFuncs;
+  private Function<Replica, Double> _scoreFunc;
+
+  public SortedReplicasHelper() {
+    _selectionFuncs = new HashSet<>();
+    _priorityFuncs = new ArrayList<>();
+  }
+
+  /**
+   * If condition hold, add a selection function to the {@link SortedReplicas} to be created.
```
efeg: Nit: `If condition hold` -> `If condition holds`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+   * @param selectionFunc The selection function to add.
+   * @return The helper object itself.
+   */
+  public SortedReplicasHelper addSelectionFunc(Function<Replica, Boolean> selectionFunc) {
+    return maybeAddSelectionFunc(selectionFunc, true);
+  }
+
+
+  /**
+   * If condition hold, add a priority function to the {@link SortedReplicas} to be created.
```
efeg: Nit: `If condition hold` -> `If condition holds`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * @param priorityFunc The priority function to add.
+   * @return The helper object itself.
+   */
+  public SortedReplicasHelper addPriorityFunc(Function<Replica, Integer> priorityFunc) {
+    return maybeAddPriorityFunc(priorityFunc, true);
+  }
+
+  /**
+   * If condition hold, add a score function to the {@link SortedReplicas} to be created.
```
efeg: Nit: `If condition hold` -> `If condition holds`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * @param scoreFunc The score function to add.
+   * @param condition Whether add the score function or not.
+   * @return The helper object itself.
+   */
+  public SortedReplicasHelper maybeAddScoreFunc(Function<Replica, Double> scoreFunc, boolean condition) {
```
efeg: (Also applies to other JavaDocs) `condition` -> can we use more descriptive names? Also this function seems redundant, would you clarify its expected use-case?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+   */
+  public SortedReplicasHelper addPriorityFunc(Function<Replica, Integer> priorityFunc) {
+    return maybeAddPriorityFunc(priorityFunc, true);
+  }
+
+  /**
+   * If condition hold, add a score function to the {@link SortedReplicas} to be created.
+   *
+   * @param scoreFunc The score function to add.
+   * @param condition Whether add the score function or not.
```
efeg: Does it `add` or `overwrite`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+  public SortedReplicasHelper maybeAddPriorityFunc(Function<Replica, Integer> priorityFunc, boolean condition) {
+    if (condition && !_priorityFuncs.contains(priorityFunc)) {
+      _priorityFuncs.add(priorityFunc);
+    }
+    return this;
+  }
+
+  /**
+   * Add a priority function to the {@link SortedReplicas} to be created.
+   * Note the order adding priority function matters. Each call of this method or {@link this#maybeAddPriorityFunc(Function, boolean)}
```
efeg: redundant use of `this` in `this#maybeAddPriorityFunc(Function, boolean)}`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+   * will add new priority function to the end of priority function list, which will be applied only when all the previous
```
efeg: (Also applies to other JavaDocs) JavaDoc of one method should not be documenting the behavior of another. We can provide a link in the form of `See XXX` to another function, but let's avoid describing the behavior of other functions. For example, here it is not clear that the other function will actually `not`  add a new priority function to the end of priority function list if the boolean parameter is `false`.

efeg: `will add new priority` -> `will add a new priority`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
```diff
@@ -0,0 +1,151 @@
...
+   * Note the order adding priority function matters. Each call of this method or {@link this#addPriorityFunc(Function)}
+   * will add new priority function to the end of priority function list, which will be applied only when all the previous
+   * priority functions cannot resolve two replica's priority.
+   *
+   * @param priorityFunc The priority function to add.
+   * @param condition Whether add the priority function or not.
+   * @return The helper object itself.
+   */
+  public SortedReplicasHelper maybeAddPriorityFunc(Function<Replica, Integer> priorityFunc, boolean condition) {
+    if (condition && !_priorityFuncs.contains(priorityFunc)) {
```
efeg: If the intention is to ensure that the `priorityFunc` is inserted just once, is there another more suitable data structure we can use (e.g. `SortedSet`)?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
kidkun: The main reason I want to have these `maybeAddxx` methods are to make the API more "confluent".
Consider [use case at CapacityGoal](https://github.com/linkedin/cruise-control/blob/d046302370e20c3bc780d09afabe2bf9342ede19/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java#L182), without this API, I need to write 
```   
SortedReplicaHelper helper = new SortedReplicasHelper();
if (onlyMoveImmigrantReplicas) {
    helper.addSelectionFunc(ReplicaSortFunctionFactory.selectImmigrants();
}                     
 helper.addSelectionFunc(ReplicaSortFunctionFactory.selectReplicasNotFromExcludedTopics(excludedTopics))
           .addPriorityFunc(ReplicaSortFunctionFactory.prioritizeOfflineReplicas())
           .addPriorityFunc(ReplicaSortFunctionFactory.prioritizeImmigrants())
           ...
```

kidkun: you are referring only to the score function,  you are right. it it not used, will remove it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicasHelper.java
kidkun: Changed to LinkedHashSet since here we want to preserve the insertion order.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -294,16 +299,16 @@ private boolean rebalanceByMovingLeaderReplicasIn(Broker broker,
     List<Broker> candidateBrokers = Collections.singletonList(broker);
```
efeg: Nit (existing issue): Bad indentation on `rebalanceByMovingLeaderReplicasIn()` function.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
```diff
@@ -294,16 +299,16 @@ private boolean rebalanceByMovingLeaderReplicasIn(Broker broker,
...
     Set<String> excludedTopics = optimizationOptions.excludedTopics();
     boolean onlyMoveImmigrantReplicas = optimizationOptions.onlyMoveImmigrantReplicas();
+    String replicaSortName = replicaSortName(this, false, true);
+    new SortedReplicasHelper().addSelectionFunc(ReplicaSortFunctionFactory.selectReplicasNotFromExcludedTopics(excludedTopics))
+                              .addSelectionFunc(ReplicaSortFunctionFactory.selectLeaders())
+                              .maybeAddSelectionFunc(ReplicaSortFunctionFactory.selectImmigrants(),
+                                                     !clusterModel.brokenBrokers().isEmpty() || onlyMoveImmigrantReplicas)
+                              .trackSortedReplicasFor(replicaSortName, clusterModel);
```
efeg: Looks like we are potentially re-filtering (i.e. re-selecting) replicas for the whole cluster every time `rebalanceByMovingLeaderReplicasIn` is called. This seems expensive and maybe unnecessary. I am wondering if we can improve the performance here a little -- e.g. via calling `trackSortedReplicasFor` for the whole cluster just once during the initialization as we do in other goals?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareGoal.java
```diff
@@ -206,14 +215,18 @@ protected void initGoalState(ClusterModel clusterModel, OptimizationOptions opti
...
+      // One pass is sufficient to satisfy or alert impossibility of this goal.
+      // Sanity check to confirm that the final distribution is rack aware.
+      ensureRackAware(clusterModel, excludedTopics);
+      // Sanity check: No self-healing eligible replica should remain at a dead broker/disk.
+      GoalUtils.ensureNoOfflineReplicas(clusterModel, name());
+      // Sanity check: No replica should be moved to a broker, which used to host any replica of the same partition on its broken disk.
+      GoalUtils.ensureReplicasMoveOffBrokersWithBadDisks(clusterModel, name());
+      finish();
+    } finally {
+      clusterModel.clearSortedReplicas();
```
efeg: Do we have a unit test to verify that the sortedReplicas are always cleared after each goal? (if not can we add one)?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ReplicaSortFunctionFactory.java
```diff
@@ -98,85 +88,73 @@ private ReplicaSortFunctionFactory() {
...
   }
 
   /**
-   * @return a selection function that only includes immigrant leader replicas.
+   * @return a selection function that only includes replicas whose metric value for certain resource is below limit.
    */
-  public static Function<Replica, Boolean> selectImmigrantLeaders() {
-    return SELECT_IMMIGRANT_LEADERS;
+  public static Function<Replica, Boolean> selectReplicasBelowLimit(Resource resource, Double limit) {
+    return r ->  r.load().expectedUtilizationFor(resource) < limit;
```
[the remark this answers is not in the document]
kidkun: unlike `selectFollowers()`, to create the function object here, we need to pass in information(`resource` and `limit` here) from the calling context. so we cannot have a singleton static function object in advance.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -535,63 +535,11 @@ public Broker broker(int brokerId) {
...
-
-  /**
-   * Ask the cluster model to keep track of the replicas sorted with the given priority function and score function.
-   *
-   * The sort will first use the priority function then the score function. The priority function allows the
-   * caller to prioritize a certain type of replicas, e.g immigrant replicas. The selection function determines
-   * which replicas to be included in the sorted replicas.
+   * The selection functions determine which replicas to be included in the sorted replicas. Then the sort will first
+   * use the priority functions then the score function. The priority functions allow the caller to prioritize a certain
+   * type of replicas, e.g immigrant replicas.
```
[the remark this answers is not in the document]
kidkun: That is covered `A replica is added to or removed from a broker`, in that case replica becomes `immigrant` replica of the destination broker.
I think the list here is completed, basically it maps to the 2 possible operation could happen to the cluster(replica movement, leadership change).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicas.java
```diff
@@ -36,106 +33,96 @@
...
  */
 public class SortedReplicas {
   private final Broker _broker;
   private final Disk _disk;
-  private final Map<Replica, ReplicaWrapper> _replicaWrapperMap;
-  private final NavigableSet<ReplicaWrapper> _sortedReplicas;
-  private final Function<Replica, Boolean> _selectionFunc;
-  private final Function<Replica, Integer> _priorityFunc;
+  private final SortedSet<Replica> _sortedReplicas;
+  private final Set<Function<Replica, Boolean>> _selectionFuncs;
```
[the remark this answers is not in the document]
kidkun: it would hard to sort to between function objects here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
```diff
@@ -289,18 +298,12 @@ protected void rebalanceForBroker(Broker broker,
...
       // Sort replicas by descending order of preference to relocate. Preference is based on resource cost.
       // Only leaders in the source broker are sorted.
       // Note that if the replica is offline, it cannot currently be a leader.
-      List<Replica> sortedLeadersInSourceBroker =
-          broker.trackedSortedReplicas(sortNameByLeader()).reverselySortedReplicas();
-      for (Replica leader : sortedLeadersInSourceBroker) {
-        if (shouldExclude(leader, excludedTopics)) {
-          continue;
-        }
+      for (Replica leader : broker.trackedSortedReplicas(replicaSortName(this, true, true)).sortedReplicas(true)) {
```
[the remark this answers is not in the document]
kidkun: good catch! Currently I am trying to clear sorted replicas in `updateGoalState` method, but in this case the exception is thrown even before reaching there.
I will do another scan over all goals.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderReplicaDistributionGoal.java
kidkun: the sorted replica is created lazily, in most case we just do that for one or two broker and in each iteration it is probably different broker. Therefore doing in `initialization` probably would not save much. The main reason I do not do it in initialization code is that in `rebalanceByMovingReplicasOut` we also create sorted replica for some brokers, there will be  `replicaSortName` conflict for different sorted replica set for the same broker.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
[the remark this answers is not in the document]
kidkun: I ends up in extracting `clearSortedReplicas`  out of `updateGoalState` and put it in a `final` statement of `goal.optimize()`. In this way we can guarantee any sorted replicas is cleared after goal optimize the cluster mode(no matter succeed/failed/exception thrown).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareGoal.java
kidkun: I changed the location calling this method, so that it is easy to reason that each goal will call it unconditionally.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskUsageDistributionGoal.java
```diff
@@ -412,32 +403,22 @@ private void rebalanceBySwappingLoadOut(Disk disk,
 
     while (!candidateDiskPQ.isEmpty()) {
       Disk candidateDisk = candidateDiskPQ.poll();
-      candidateDisk.trackSortedReplicas(name(),
-                                        ReplicaSortFunctionFactory.selectOnlineReplicas(),
-                                        ReplicaSortFunctionFactory.sortByMetricGroupValue(RESOURCE.name()));
-      for (Iterator<Replica> iterator = disk.trackedSortedReplicas(name()).reverselySortedReplicas().iterator();
+      for (Iterator<Replica> iterator = disk.trackedSortedReplicas(replicaSortName(this, true, false)).sortedReplicas(false).iterator();
```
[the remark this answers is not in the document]
kidkun: Every time the swap succeed, we will [break out of the `for` loop](https://github.com/linkedin/cruise-control/blob/4c1979da27e5ad55ac462a385915bfedfac9e851/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskUsageDistributionGoal.java#L419). Next time we enter the `for` loop, we will have a consistent iterator, so the exception will not be thrown.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ReplicaSortFunctionFactory.java
```diff
@@ -98,85 +88,73 @@ private ReplicaSortFunctionFactory() {
...
    */
-  public static Function<Replica, Boolean> selectLeaders() {
-    return SELECT_LEADERS;
+  public static Function<Replica, Boolean> selectReplicasBasedOnExcludedTopics(Set<String> excludedTopics) {
+    return r -> r.isOriginalOffline() || !excludedTopics.contains(r.topicPartition().topic());
   }
 
   /**
-   * @return a selection function that only includes online replicas.
+   * @return a selection function that only includes replicas whose metric value for certain resource is above limit.
```
efeg: Nit: Missing JavaDoc for params.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ReplicaSortFunctionFactory.java
```diff
@@ -98,85 +88,73 @@ private ReplicaSortFunctionFactory() {
...
    */
-  public static Function<Replica, Boolean> selectOnlineReplicas() {
-    return SELECT_ONLINE_REPLICAS;
+  public static Function<Replica, Boolean> selectReplicasAboveLimit(Resource resource, double limit) {
+    return r -> r.load().expectedUtilizationFor(resource) > limit;
   }
 
   /**
-   * @return a selection function that only includes immigrant leader replicas.
+   * @return a selection function that only includes replicas whose metric value for certain resource is below limit.
```
efeg: Nit: Missing JavaDoc for params.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ReplicaSortFunctionFactory.java
```diff
@@ -98,85 +88,73 @@ private ReplicaSortFunctionFactory() {
...
    */
-  public static Function<Replica, Boolean> selectImmigrants() {
-    return SELECT_IMMIGRANTS;
+  public static Function<Replica, Boolean> selectOnlineReplicas() {
+    return SELECT_ONLINE_REPLICAS;
   }
 
   /**
-   * @return a selection function that only includes leaders.
+   * @return a selection function that filters out replicas which are online and from topics which should be excluded.
```
efeg: Nit: Missing JavaDoc for param.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicas.java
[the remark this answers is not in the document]
efeg: Can't we achieve this via the insertion order -- e.g. using `LinkedHashSet`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskUsageDistributionGoal.java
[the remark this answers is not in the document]
efeg: Thanks for the clarification!
Can we add a unit test to ensure that any unintended future change to this behavior can be captured via tests rather than potentially during run time?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskUsageDistributionGoal.java
```diff
@@ -479,32 +455,22 @@ private void rebalanceBySwappingLoadIn(Disk disk,
 
     while (!candidateDiskPQ.isEmpty()) {
       Disk candidateDisk = candidateDiskPQ.poll();
-      candidateDisk.trackSortedReplicas(name(),
-                                        ReplicaSortFunctionFactory.selectOnlineReplicas(),
-                                        ReplicaSortFunctionFactory.sortByMetricGroupValue(RESOURCE.name()));
-      for (Iterator<Replica> iterator = disk.trackedSortedReplicas(name()).sortedReplicas().iterator();
+      for (Iterator<Replica> iterator = disk.trackedSortedReplicas(replicaSortName(this, false, false)).sortedReplicas(false).iterator();
```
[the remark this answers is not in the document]
efeg: Can we add a unit test to ensure that any unintended future change to this behavior can be captured via tests rather than potentially during run time?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareGoal.java
efeg: Thanks for the changes! Can we also add an `assert` / `validation` regarding whether goal states are cleared after execution?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicas.java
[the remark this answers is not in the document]
kidkun: Ah, good idea!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskUsageDistributionGoal.java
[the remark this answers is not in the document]
kidkun: Added the test.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -58,18 +56,23 @@
   private boolean _fixOfflineReplicasOnly;
   private double _balanceUpperThreshold;
   private double _balanceLowerThreshold;
+  private Comparator<Broker> _brokerComparator;
```
efeg: Nit: `private Comparator<Broker> _brokerComparator;` -> `private final Comparator<Broker> _brokerComparator;`

Issue #930, opened 2019-09-04
Epsilon value for disk resource is insufficient for broker with large number of replicas


kidkun:
We observed in our production environment that when we try to remove a broker hosting 11000+ replicas from the cluster,  sometimes the request will fail with exception

> Error processing POST request '/remove_broker' due to: 'com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException: java.lang.IllegalArgumentException: Inconsistent load distribution. Broker utilization for disk is different from the total replica utilization in the broker with id: xxx. Sum of the replica utilization: 0.000000, broker utilization:132.109833'

The cause for this issue is the same as issue #846 , i.e. there is precision lost in adding/subtracting `float` values. For broker hosting large number of replicas, this precision lost accumulate to a non-trivial value. Therefore we have defined epsilon value to account for the nuance. Currently for disk resource it is set to [100](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/Resource.java#L21). It turns out it is not large enough for broker with 11000+ replicas.

PR #931, opened 2019-09-04
Increse absolute epsilon value of disk resource.

kidkun:
Fix issue #930

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/Resource.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/Resource.java
@@ -18,7 +18,7 @@ public enum Resource {
   CPU("cpu", 0, true, true, 0.001),
   NW_IN("networkInbound", 1, true, false, 10),
   NW_OUT("networkOutbound", 2, true, false, 10),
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/Resource.java
```diff
@@ -18,7 +18,7 @@
   CPU("cpu", 0, true, true, 0.001),
   NW_IN("networkInbound", 1, true, false, 10),
   NW_OUT("networkOutbound", 2, true, false, 10),
-  DISK("disk", 3, false, true, 100);
+  DISK("disk", 3, false, true, 200);
```
efeg: It is a little suspicious that the https://github.com/linkedin/cruise-control/issues/930 indicates a `0` value for sum of replica utilization. I am curious if there is a special reason for this case?

kidkun: The sanity check is on the cluster model after all goals are applied. In this case(broker removal), all the replicas are moved off the broker so the `0` value is accurate, the inaccurate values is actually the broker utilization...

Issue #942, opened 2019-09-16
Cruise Control does not honor replica movement concurrency throttling in replication factor increasing operation


kidkun:
This misbehavior is due to a bug in the where Cruise Control determines whether an ongoing replica movement task is finished or not.

The current logic is to compare the assigned replica list for the subject topic partition against the task's target replica list. 
This works if we are moving replica, e.g. (1,2,3) -> (1,2,4), what happens is Kafka controller first add broker 4 into replica list, i.e. (1,2,3) -> (1,2,3,4). then when new replica in broker 4 catches up, the replica list is shrunk from (1,2,3,4) to (1,2,4) and Cruise Control thinks the task is done.

However, in increasing replication factor case, e.g. (1,2) -> (1,2,3), when Controller add 3 in to replica list and **before replica on broker 3 catches up**, Cruise Control will think this task is done. Over the time, Cruise Control keeps adding new replica movement tasks and performs a DOS attack to the Kafka cluster.

PR #943, opened 2019-09-16
Fix the bug that Cruise Control prematurely thinks some replica movement tasks are done.

kidkun:
Fix issue #942

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionProposal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionProposal.java
@@ -90,13 +90,15 @@ public class ExecutionProposal {
 
   /**
    * Check whether the successful completion of inter-broker replica movement from this proposal is reflected in the current
-   * ordered replicas in the given cluster.
+   * ordered replicas in the given cluster and all replicas are in-sync.
    *
    * @param currentOrderedReplicas Current ordered replica list from the cluster.
+   * @param currentInSyncReplicas Current in-sync replica list from the cluster.
    * @return True if successfully completed, false otherwise.
    */
-  public boolean isInterBrokerMovementCompleted(Node[] currentOrderedReplicas) {
-    return brokerOrderMatched(currentOrderedReplicas, _newReplicas);
+  public boolean isInterBrokerMovementCompleted(Node[] currentOrderedReplicas, Node [] currentInSyncReplicas) {
+    return brokerOrderMatched(currentOrderedReplicas, _newReplicas)
+           && currentInSyncReplicas.length == currentOrderedReplicas.length;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1028,7 +1028,8 @@ private boolean isInterBrokerReplicaActionDone(Cluster cluster, TopicPartition t
       Node[] currentOrderedReplicas = cluster.partition(tp).replicas();
       switch (task.state()) {
         case IN_PROGRESS:
-          return task.proposal().isInterBrokerMovementCompleted(currentOrderedReplicas);
+          return task.proposal().isInterBrokerMovementCompleted(currentOrderedReplicas)
```
efeg: `ExecutionProposal#isInterBrokerMovementCompleted` is called by other functions, as well. 

Does the same update not apply them? If it does, should we move `&& currentOrderedReplicas.length == cluster.partition(tp).inSyncReplicas().length;` in this function? If not, should we rename this function to something else to make it more clear?

kidkun: Good point! I would like to apply the logic to all the place so we better honor the concurrency throttling.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1017,7 +1017,7 @@ private boolean isTaskDone(Cluster cluster,
 
     /**
      * For a inter-broker replica movement action, the completion depends on the task state:
-     * IN_PROGRESS: when the current replica list is the same as the new replica list.
+     * IN_PROGRESS: when the current replica list is the same as the new replica list and all replicas are in-sync.
```
efeg: Nit: Same update for `ABORTING` is missing.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionProposal.java
```diff
@@ -106,10 +108,12 @@ public boolean isInterBrokerMovementCompleted(Node[] currentOrderedReplicas) {
    * In that case, we treat it as aborted as well.
    *
    * @param currentOrderedReplicas Current ordered replica list from the cluster.
+   * @param currentInSyncReplicas Current in-sync replica list from the cluster.
    * @return True if aborted, false otherwise.
    */
-  public boolean isInterBrokerMovementAborted(Node[] currentOrderedReplicas) {
-    return isInterBrokerMovementCompleted(currentOrderedReplicas) || brokerOrderMatched(currentOrderedReplicas, _oldReplicas);
+  public boolean isInterBrokerMovementAborted(Node[] currentOrderedReplicas, Node [] currentInSyncReplicas) {
```
efeg: Nit: [Also applies to `isInterBrokerMovementCompleted`] Should we pass a `PartitionInfo` rather than array of nodes?

PR #944, opened 2019-09-17
Fixing Snapshot coming into the version #

AdrielVelazquez:
```
    integer_semver = lambda x: [int(elem) for elem in x.split('.')]
ValueError: invalid literal for int() with base 10: '0-SNAPSHOT'
```

ValueError occurs because logic not stripping out non-numeric characters

### DIFF
--- a/cruise-control-client/cruisecontrolclient/client/Responder.py
+++ b/cruise-control-client/cruisecontrolclient/client/Responder.py
@@ -96,6 +96,7 @@ class CruiseControlResponder(requests.Session):
             else:
                 # Guess about whether this version of cruise-control supports 202
                 if "Cruise-Control-Version" in response.headers:
+                    # define a regex for extracting only leading digits and periods from the Cruise-Control-Version
                     non_decimal = re.compile(r'[^\d.]+')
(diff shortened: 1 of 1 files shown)


cruise-control-client/cruisecontrolclient/client/Responder.py
```diff
@@ -93,8 +96,9 @@ def json_or_text_guesser():
             else:
                 # Guess about whether this version of cruise-control supports 202
                 if "Cruise-Control-Version" in response.headers:
+                    non_decimal = re.compile(r'[^\d.]+')
```
mgrubent: This is a nit, but suggest a comment like  
```python
# define a regex for extracting only leading digits and periods from the Cruise-Control-Version
```

mgrubent: Otherwise, I agree this looks good

AdrielVelazquez: I'll add this. I'll use your exact comment.

mgrubent:
@AdrielVelazquez following the merge of #954, I've released `cruise-control-client` `1.0.1` [to PyPI](https://pypi.org/project/cruise-control-client/)

Issue #949, opened 2019-09-18
Demoting brokers should not cause continual Zookeeper disconnects


efeg:
-- Thanks `Krit Petty` for reporting the issue!

It appears CC keeps disconnecting and re-connecting to ZK and demote process never proceeds, this is observed when demoting a large number of brokers with many partitions at once -- e.g. `25` brokers with many replicas on each broker.
```
INFO [ZooKeeperClient] [ProposalExecutor-0] [kafka-cruise-control] [] [ZooKeeperClient] Waiting until connected.
INFO [ZooKeeperClient] [ProposalExecutor-0] [kafka-cruise-control] [] [ZooKeeperClient] Connected.
INFO [ClientCnxn] [kafka-cruise-control] [] Opening socket connection to server XXX. Will not attempt to authenticate using SASL (unknown error)
INFO [ClientCnxn] [kafka-cruise-control] [] Socket connection established, initiating session, client: XXX, server: XXX
INFO [ClientCnxn] [kafka-cruise-control] [] Session establishment complete on server XXX, sessionid = XXX, negotiated timeout = XXX
INFO [ClientCnxn] [kafka-cruise-control] [] Unable to read additional data from server sessionid XXX, likely server has closed socket, closing socket connection and attempting reconnect
```

bgrishinko:
No problem. there isn't a full workaround, but temporarily we can demote in smaller batches. This will not guarantee all hosts are removed of all leadership as a demote in the 2nd batch could reassign leadership back to the first batch.

PR #950, opened 2019-09-19
Ensure that executions do not cause continual Zookeeper disconnects.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/949.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java
```diff
@@ -87,21 +81,35 @@ public ExecutionTaskManager(int defaultInterBrokerPartitionMovementConcurrency,
...
 
   /**
    * Dynamically set the intra-broker partition movement concurrency.
+   * Ensure that the requested concurrency is smaller than the maximum number of allowed partition movements in cluster.
    *
    * @param requestedIntraBrokerPartitionMovementConcurrency The maximum number of concurrent intra-broker partition movements
    *                                                         (if null, use {@link #_defaultIntraBrokerPartitionMovementConcurrency}).
    */
   public synchronized void setRequestedIntraBrokerPartitionMovementConcurrency(Integer requestedIntraBrokerPartitionMovementConcurrency) {
+    if (requestedIntraBrokerPartitionMovementConcurrency != null
```
kidkun: we should also apply similar sanity check for requested leadership movement concurrency?

Issue #897, opened 2019-08-12
cruise-control-client is not API complete with cruise-control


mgrubent:
# Problem
Currently, `cruise-control-client` lags behind `cruise-control` in the parameters that are available to some of its endpoints.

All the endpoints are there, so a user with the `cruise-control` documentation would be able to use `cruise-control-client`'s `--add-parameter` flag to achieve future compatibility.
# Proposal
`cruise-control-client` should be updated to reflect all of the parameters that `cruise-control` provides for its endpoints

AnvarKhatik:
This Issue Is Still Open

rishabh8:
Hi Team,
Is this issue still open?
new to open source community
Please update 
Thanks

mhraza95:
Hi @mgrubent is this issue still open??

tn31boy:
Hi @mgrubent  , is this issue still open?

PR #955, opened 2019-09-20
Drop brokers from list

mgrubent:
This should make progress on #897 by adding `drop_recently_demoted_brokers` and `drop_recently_removed_brokers` to the `admin` Endpoint.


efeg:
@mgrubent did you have a chance to resolve the conflict?

Issue #956, opened 2019-09-27
The executor state information in `state` response is confusing


kidkun:
Currently if there are ongoing execution triggered by Cruise Control itself for self-healing purpose, the executor state returned in `state` endpoint looks like 

> ExecutorState: {state: INTER_BROKER_REPLICA_MOVEMENT_TASK_IN_PROGRESS, pending/in-progress/aborting/finished/total inter-broker partition movement 7117/0/0/8007/15124, completed/total bytes(MB): 1409073/3978370, maximum concurrent inter-broker partition movements per-broker: 1, **triggeredUserTaskId**: BROKER_FAILURE-bbe-9478-ac9bff60593b, recentlyRemovedBrokers: []}

The value of `triggeredUserTaskId` field is not a valid user task id which user can use to query in `user_tasks` endpoint.

PR #957, opened 2019-09-27
Rename triggeredUserTaskId field in executor state to triggeredTaskId.

kidkun:
fix issue #956


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorState.java
```diff
@@ -15,7 +15,7 @@
 import static com.linkedin.kafka.cruisecontrol.executor.ExecutionTaskTracker.ExecutionTasksSummary;
 
 public class ExecutorState {
-  private static final String TRIGGERED_USER_TASK_ID = "triggeredUserTaskId";
+  private static final String TRIGGERED_TASK_ID = "triggeredTaskId";
```
efeg: Should we refrain from changing the existing field name to avoid introducing a backwards incompatible change (e.g. this change will potentially break the `CCUI`)? Can we instead introduce a new field, such as `triggeredSelfHealingTaskId` to show task ids triggered by self-healing?

Issue #517, opened 2019-02-03
Feature Request: schematize JSON responses


mgrubent:
### Issue
When asking `cruise-control` for a JSON response, it is not clear what structure that response will have, except by examining the classes in `cruise-control/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response`.

This makes it difficult to programmatically know how to parse the received response, requiring a human to follow the `response` class code paths and their conditional parameter branches, to recreate a structure whose permutations the `cruise-control` servlet arguably already knows about.

Further, as `cruise-control` continues to mature and add features, the shape of the JSON response will grow, and a statically-defined response parser will miss the new fields.
### Proposal
`cruise-control` should define JSON schemas for the responses it can return from the different (endpoint, parameter) combinations.

`cruise-control` could then optionally pass the schema's identifier in the header of its response, as a low-overhead way of signaling the shape of the response.  
Returning a schema in this way could be opt-in, by including a parameter such as `json_schema_in_header=true` along with the `json=true` parameter.

Additionally, clients could use this proposed mapping of (endpoint, parameter) -> schema to know the expected schema ahead of time.
If `cruise-control` included the schema identifier in its header to the client, then the clients would affirmatively know whether `cruise-control` has updated its response schema for that (endpoint, parameter) tuple.  
Even if the schema identifier were not included, a client would be able to validate the received response against the expected schema and arrive at the same conclusion.

ian952:
I will be working on this issue!

efeg:
Closing this issue as it has been addressed in https://github.com/linkedin/cruise-control/pull/960.

PR #960, opened 2019-10-01
Include JSON schema in response header

ian952:
Addresses #517


efeg:
@ian952 Looks like there are some `FindBugs rule violations` -- would you update the PR to resolve these issues?

ian952:
@efeg Will do! There are also some small changes I want to make. Let me get back to you when the PR is ready.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/servlet/parameters/CruiseControlParameters.java
```diff
@@ -37,9 +37,9 @@
   boolean json();
 
   /**
-   * @return True if requested response should contain a JSON schema in the header, false otherwise.
+   * @return True if rejquested response should contain a JSON schema in the header, false otherwise.
```
efeg: `rejquested`?

ian952: Fixed!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ResponseUtils.java
```diff
@@ -57,11 +57,24 @@ static void writeResponseToOutputStream(HttpServletResponse response,
                                           boolean json,
                                           String responseMessage,
                                           KafkaCruiseControlConfig config)
+    throws IOException {
+    writeResponseToOutputStream(response, responseCode, json, null, responseMessage, config);
+  }
+
+  static void writeResponseToOutputStream(HttpServletResponse response,
```
kidkun: we should also check whether to populate schema information in header when returning [in-progress](https://github.com/linkedin/cruise-control/blob/d3619c650aaff1f3750a7d7b0f3ec5ab5f1c4741/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ProgressResult.java#L64)/[error](https://github.com/linkedin/cruise-control/blob/213fd8e348e62a7cadbfbc6b4dbe48e6700d9d12/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ResponseUtils.java#L96) response

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/AbstractParameters.java
```diff
@@ -34,6 +35,7 @@
   static {
     SortedSet<String> validParameterNames = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
     validParameterNames.add(JSON_PARAM);
+    validParameterNames.add(GET_RESPONSE_SCHEMA);
```
kidkun: we should also update the Java doc for all `Parameters`  class about adding this new parameter(e.g. [RebalanceParameters](https://github.com/linkedin/cruise-control/blob/db6281c2c458ded9dddad72606048bcc01f135ae/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/RebalanceParameters.java#L35))

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/AbstractCruiseControlResponse.java
```diff
@@ -49,4 +61,59 @@ public synchronized void discardIrrelevantResponse(CruiseControlParameters param
   public String cachedResponse() {
     return _cachedResponse;
   }
+
+  private String getJsonSchema() {
+    JsonElement response = new JsonParser().parse(_cachedResponse);
+    String schema = convertNodeToStringSchemaNode(response, null);
```
kidkun: we could directly return the `String` here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ResponseUtils.java
ian952: I moved the schema logic to `ResponseUtils` so the logic can be shared for these responses.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/AbstractParameters.java
ian952: Updated all existing java doc under the paramters/ folder. Let me know there are any I missed!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ResponseUtils.java
```diff
@@ -112,6 +121,60 @@ public static void writeErrorResponse(HttpServletResponse response,
...
+  private static String convertNodeToStringSchemaNode(
+          JsonElement node, String key) {
+    StringBuilder result = new StringBuilder();
+
+    if (key != null) {
+      result.append("\"" + key + "\": { \"type\": \"");
+    } else {
+      result.append("{ \"type\": \"");
+    }
+    if (node.isJsonArray()) {
```
kidkun: From the JSON schema's spec on [array](https://json-schema.org/understanding-json-schema/reference/array.html),looks like we should not iterate all the element, extract and print their schema info, instead we should just do it once?

Also if it is an empty array, can we just output `{ "type": "array" }` instead of `{ "type": "array", "items": []}`?

ian952: > From the JSON schema's spec on array, looks like we should not iterate all the elements, extract and print their schema info one by one, instead we should just do it once?

The reason why we iterate through every element is that each element can potentially have a different schema. I have modified the code to first put them into a set to avoid any duplicates. However, this will not be perfect as we could still run into cases where some elements in the array have optional fields and some don't, in which case, there will be two entries in items since we do not have enough context to tell they are the same items. This is the limitation of dynamically generating the schema as we do not have the complete schema information at generation time.

> Also if it is an empty array, can we just output { "type": "array" } instead of { "type": "array", "items": []}?
I have added this change. Empty arrays will just output `{ "type": "array" }` now.

ian952:
@kidkun I have updated the PR with suggested fixes. Let me know what you think regarding my comment on array items schema.

efeg:
@ian952 Are there any remaining feedback to address for this PR? (cc @kidkun)

ian952:
All comments have been addressed. Let me know if there are any other feedback.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ResponseUtils.java
kidkun: The issue with print all the element is that the output will be very huge in some case(consider we are print the partition load for 1m partitions in a cluster). 

For optional fields, whether it exists in element or not should purely depend on `verbose`/`super_verbose` parameter in the request, which means in one specific response, it should still be consistent between elements. 

If elements in array are not consistent, it should be a bug in code since it means to be an `array` instead of `tuple`. Therefore for header schema, one element's schema should be enough.

ian952: Sounds good. I have updated the PR to use the first element for non-empty arrays.

Issue #961, opened 2019-10-01
Make Cruise Control able to run operations when it is loading samples


kidkun:
Currently after Cruise Control boots up, there is period of time that it is unable to execute any operations(generated by either admin or itself). And this time could be several hours for large cluster. 
This actually harms CC's availability and sometimes user has urgent operations to run which does not need the whole load history(e.g. `preferred leader election`).

PR #962, opened 2019-10-02
Support executing operations which require no valid window when load monitor is loading.

kidkun:
Fix issue #961

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -33,6 +33,7 @@ import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
 import com.linkedin.kafka.cruisecontrol.monitor.LoadMonitor;
 import com.linkedin.kafka.cruisecontrol.monitor.MonitorUtils;
 import com.linkedin.kafka.cruisecontrol.monitor.metricdefinition.KafkaMetricDef;
+import com.linkedin.kafka.cruisecontrol.monitor.task.LoadMonitorTaskRunner;
 import com.linkedin.kafka.cruisecontrol.servlet.UserTaskManager;
 import com.linkedin.kafka.cruisecontrol.servlet.response.stats.BrokerStats;
 import java.io.InputStream;
@@ -98,17 +99,23 @@ public class KafkaCruiseControl {
     ModelParameters.init(config);
 
     // Instantiate the components.
-    _loadMonitor = new LoadMonitor(config, _time, dropwizardMetricRegistry, KafkaMetricDef.commonMetricDef());
-    _goalOptimizerExecutor =
-        Executors.newSingleThreadExecutor(new KafkaCruiseControlThreadFactory("GoalOptimizerExecutor", true, null));
+    _anomalyDetector = new AnomalyDetector(this, _time, dropwizardMetricRegistry);
     long demotionHistoryRetentionTimeMs = config.getLong(KafkaCruiseControlConfig.DEMOTION_HISTORY_RETENTION_TIME_MS_CONFIG);
     long removalHistoryRetentionTimeMs = config.getLong(KafkaCruiseControlConfig.REMOVAL_HISTORY_RETENTION_TIME_MS_CONFIG);
-    _anomalyDetector = new AnomalyDetector(_loadMonitor, this, _time, dropwizardMetricRegistry);
     _executor = new Executor(config, _time, dropwizardMetricRegistry, demotionHistoryRetentionTimeMs,
                              removalHistoryRetentionTimeMs, _anomalyDetector);
+    _loadMonitor = new LoadMonitor(config, _time, _executor, dropwizardMetricRegistry, KafkaMetricDef.commonMetricDef());
+    _goalOptimizerExecutor =
+        Executors.newSingleThreadExecutor(new KafkaCruiseControlThreadFactory("GoalOptimizerExecutor", true, null));
     _goalOptimizer = new GoalOptimizer(config, _loadMonitor, _time, dropwizardMetricRegistry, _executor);
   }
 
+  /**
+   * Get the load monitor.
+   */
+  public LoadMonitor loadMonitor() {
+    return _loadMonitor;
+  }
   /**
    * Refresh the cluster metadata and get the corresponding cluster and generation information.
    *
@@ -116,6 +123,13 @@ public class KafkaCruiseControl {
    */
   public MetadataClient.ClusterAndGeneration refreshClusterAndGeneration() {
     return _loadMonitor.refreshClusterAndGeneration();
+  }
+
+  /**
+   * Get the state of load monitor's task runner.
+   */
+  public LoadMonitorTaskRunner.LoadMonitorTaskRunnerState getLoadMonitorTaskRunnerState() {
+    return _loadMonitor.taskRunnerState();
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -7,6 +7,8 @@ package com.linkedin.kafka.cruisecontrol;
 import com.linkedin.kafka.cruisecontrol.analyzer.AnalyzerUtils;
 import com.linkedin.kafka.cruisecontrol.analyzer.goals.PreferredLeaderElectionGoal;
 import com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig;
+import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
+import com.linkedin.kafka.cruisecontrol.monitor.task.LoadMonitorTaskRunner;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Collection;
@@ -178,6 +180,25 @@ public class KafkaCruiseControlUtils {
                                                          + "parameter to ignore this sanity check.", hardGoals, goals,
                                                          SKIP_HARD_GOAL_CHECK_PARAM));
       }
+    }
+  }
+
+  /**
+   * Check to ensure that if requested cluster model completeness requires non-zero number of valid windows in cluster model,
+   * load monitor should have already finished loading all the samples.
+   * Note that even if only one valid window is required, we still need to wait load monitor finish loading all samples. This
+   * is because sample loading follow time order, i.e. the first window loaded is the oldest window and the load information
+   * in that window is probably stale.
+   *
+   * @param completenessRequirements Requested cluster model completeness requirement.
+   * @param loadMonitorTaskRunnerState Current state of load monitor's task runner.
+   */
+  public static void sanityCheckLoadMonitorReadiness(ModelCompletenessRequirements completenessRequirements,
+                                                     LoadMonitorTaskRunner.LoadMonitorTaskRunnerState loadMonitorTaskRunnerState) {
+    if (completenessRequirements.minRequiredNumWindows() > 0 &&
+        loadMonitorTaskRunnerState == LoadMonitorTaskRunner.LoadMonitorTaskRunnerState.LOADING) {
+      throw new IllegalStateException("Unable to generate proposal since load monitor is in "
+                                      + LoadMonitorTaskRunner.LoadMonitorTaskRunnerState.LOADING + " state.");
     }
   }
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
@@ -18,7 +18,6 @@ import com.linkedin.kafka.cruisecontrol.executor.ExecutionProposal;
 import com.linkedin.kafka.cruisecontrol.executor.ExecutorState;
 import com.linkedin.kafka.cruisecontrol.model.ClusterModel;
 import com.linkedin.kafka.cruisecontrol.model.ReplicaPlacementInfo;
-import com.linkedin.kafka.cruisecontrol.monitor.LoadMonitor;
 import com.linkedin.kafka.cruisecontrol.monitor.ModelGeneration;
 import java.util.Collections;
 import java.util.HashSet;
@@ -29,7 +28,6 @@ import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -48,9 +46,7 @@ import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.KAF
 public class GoalViolationDetector implements Runnable {
   private static final Logger LOG = LoggerFactory.getLogger(GoalViolationDetector.class);
   private final KafkaCruiseControl _kafkaCruiseControl;
-  private final LoadMonitor _loadMonitor;
   private final List<Goal> _detectionGoals;
-  private final Time _time;
   private final Queue<Anomaly> _anomalies;
   private ModelGeneration _lastCheckedModelGeneration;
   private final Pattern _excludedTopics;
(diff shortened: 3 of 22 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/SampleLoadingTask.java
```diff
@@ -33,8 +39,14 @@ public void run() {
                                                             _brokerMetricSampleAggregator));
       ModelParameters.updateModelCoefficient();
     } finally {
-      // The sample loading task is run before the load monitor starts.
-      _loadMonitorTaskRunner.setState(LoadMonitorTaskRunner.LoadMonitorTaskRunnerState.RUNNING);
+      // If Cruise Control has triggered operations before loading is finished, transfer monitor state to paused state;
```
efeg: Isn't there a race condition here? -- i.e. in general, if we would like to avoid such race conditions, shouldn't we update executor state and load monitor state atomically?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/SampleLoadingTask.java
```diff
@@ -33,8 +39,14 @@ public void run() {
...
+      // otherwise transfer to running state.
+      if (_executor.state().state() != NO_TASK_IN_PROGRESS) {
```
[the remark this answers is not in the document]
kidkun: Added  a sanity check in constructor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -98,7 +98,8 @@ public KafkaCruiseControl(KafkaCruiseControlConfig config, MetricRegistry dropwi
     ModelParameters.init(config);
 
     // Instantiate the components.
-    _loadMonitor = new LoadMonitor(config, _time, dropwizardMetricRegistry, KafkaMetricDef.commonMetricDef());
+    _executor = null;
+    _loadMonitor = new LoadMonitor(config, _time, _executor, dropwizardMetricRegistry, KafkaMetricDef.commonMetricDef());
```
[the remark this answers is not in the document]
kidkun: This is a hacky way to get around the lint check...
 There is a circular dependency between the constructor of `Executor`, `LoadMonitor` and `AnomalyDetector`...

`Executor` needs reference of `AnomalyDetector`
`AnomalyDetector` needs reference of `LoadMonitor`
`LoadMonitor` needs reference of `Executor`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -656,11 +657,12 @@ private void execute() {
       _executorState = ExecutorState.executionStarted(_uuid, _recentlyDemotedBrokers, _recentlyRemovedBrokers);
       OPERATION_LOG.info("Task [{}] execution starts.", _uuid);
       try {
-        // Pause the metric sampling to avoid the loss of accuracy during execution.
         while (true) {
           try {
-            // Ensure that the temporary states in the load monitor are explicitly handled -- e.g. SAMPLING.
-            _loadMonitor.pauseMetricSampling(String.format("Paused-By-Cruise-Control-Before-Starting-Execution (Date: %s)", currentUtcDate()));
+            if (_loadMonitor.taskRunnerState() != LoadMonitorTaskRunner.LoadMonitorTaskRunnerState.LOADING) {
```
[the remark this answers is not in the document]
kidkun: You are right! Moved this part of logic into the synchronized `pauseSampling()` .

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/SampleLoadingTask.java
kidkun: Wrap this part of logic with `synchronized (_loadMonitorTaskRunner)`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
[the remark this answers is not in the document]
efeg: IIUC, `ExecutorState` is the only state needed by `LoadMonitor` (i.e. not the whole `Executor`). I am curious if creating initial executor state out of the executor, ensuring that its state can be updated externally (i.e. no reassignment, but object modification) and passing it as an argument to the needed classes would eliminate this hack?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -33,6 +33,7 @@
 import com.linkedin.kafka.cruisecontrol.monitor.LoadMonitor;
 import com.linkedin.kafka.cruisecontrol.monitor.MonitorUtils;
 import com.linkedin.kafka.cruisecontrol.monitor.metricdefinition.KafkaMetricDef;
+import com.linkedin.kafka.cruisecontrol.monitor.task.LoadMonitorTaskRunner;
```
efeg: Can we update the title, commit message, and issue details of this PR to reflect the latest specific functionality (i.e. not all executions but executions with selected set of goals)?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -173,6 +175,25 @@ public static void sanityCheckGoals(List<String> goals, boolean skipHardGoalChec
...
+   *
+   * @param completenessRequirements Requested cluster model completeness requirement.
+   * @param loadMonitorTaskRunnerState Current state of load monitor's task runner.
+   */
+  public static void sanityCheckLoadMonitorReadiness(ModelCompletenessRequirements completenessRequirements,
+                                                     LoadMonitorTaskRunner.LoadMonitorTaskRunnerState loadMonitorTaskRunnerState) {
+    if (completenessRequirements.minRequiredNumWindows() > 0 &&
+        loadMonitorTaskRunnerState == LoadMonitorTaskRunner.LoadMonitorTaskRunnerState.LOADING) {
+      throw new IllegalStateException("Unable to generate proposal since load monitor is in "
+                                      + LoadMonitorTaskRunner.LoadMonitorTaskRunnerState.LOADING + "state.");
```
efeg: Nit: missing space before `state`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -780,6 +778,7 @@ private void execute() {
         }
         // Clear completed execution.
         clearCompletedExecution();
+        _loadMonitor.resumeMetricSampling(String.format("Resumed-By-Cruise-Control-After-Completed-Execution (Date: %s)", currentUtcDate()));
```
efeg: Nit: Should we move it inside `clearCompletedExecution`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -107,11 +108,13 @@
    *
    * @param config The load monitor configuration.
    * @param time   The time object.
+   * @param executor The proposal executor.
    * @param dropwizardMetricRegistry The sensor registry for cruise control
    * @param metricDef The metric definitions.
    */
   public LoadMonitor(KafkaCruiseControlConfig config,
                      Time time,
+                     Executor executor,
```
efeg: I suspect that we can limit ourselves to passing only the state (not the whole `executor`) -- please see the other comment above.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -73,13 +76,15 @@ public LoadMonitorTaskRunner(KafkaCruiseControlConfig config,
                                MetricDef metricDef,
                                Time time,
                                MetricRegistry dropwizardMetricRegistry,
-                               BrokerCapacityConfigResolver brokerCapacityConfigResolver) {
+                               BrokerCapacityConfigResolver brokerCapacityConfigResolver,
+                               Executor executor) {
```
efeg: Ditto on not passing executor, but its state.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -269,32 +278,49 @@ public void shutdown() {
...
     } else {
       _awaitingPauseSampling = false;
-      _reasonOfLatestPauseOrResume = reason;
+      setReasonOfLatestPauseOrResume(reason);
     }
   }
 
   /**
    * Resume the scheduled sampling tasks.
+   * Note if load monitor is still in loading state,, the method will be a noop.
```
efeg: `loading state,,` -> `loading state,`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -269,32 +278,49 @@ public void shutdown() {
   }
 
   /**
-   * Pause the scheduled sampling tasks..
+   * Pause the scheduled sampling tasks.
+   * Note if load monitor is still in loading state,, the method will be a noop.
```
efeg: `loading state,,` -> `loading state,`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
```diff
@@ -189,7 +197,8 @@ private void loadSamples() {
       _samplingScheduler.submit(new SampleLoadingTask(_sampleStore,
                                                       _partitionMetricSampleAggregator,
                                                       _brokerMetricSampleAggregator,
-                                                      this));
+                                                      this,
+                                                      _executor));
```
efeg: [`This can potentially be done on a separate issue`] Looks like this function (i.e. `loadSamples()`) is called only when `skipLoadingSamples` is `false` in `#start(boolean skipLoadingSamples)` -- I am curious if we can make this value configurable (currently `false` is hard-coded)?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/SampleLoadingTask.java
```diff
@@ -4,26 +4,35 @@
...
 
+import static com.linkedin.kafka.cruisecontrol.executor.ExecutorState.State.NO_TASK_IN_PROGRESS;
+
 
 public class SampleLoadingTask implements Runnable {
   private final SampleStore _sampleStore;
   private final KafkaPartitionMetricSampleAggregator _partitionMetricSampleAggregator;
   private final KafkaBrokerMetricSampleAggregator _brokerMetricSampleAggregator;
   private final LoadMonitorTaskRunner _loadMonitorTaskRunner;
+  private final Executor _executor;
```
efeg: Ditto on passing `ExecutorState` rather than `executor`.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorTest.java
```diff
@@ -146,6 +147,8 @@ public void testDelayedCheck() throws InterruptedException {
     Properties props = KafkaCruiseControlUnitTestUtils.getKafkaCruiseControlProperties();
     KafkaCruiseControlConfig kafkaCruiseControlConfig = new KafkaCruiseControlConfig(props);
     EasyMock.expect(mockKafkaCruiseControl.config()).andReturn(kafkaCruiseControlConfig).times(1, 2);
+    EasyMock.expect(mockKafkaCruiseControl.getLoadMonitorTaskRunnerState())
+            .andReturn(LoadMonitorTaskRunner.LoadMonitorTaskRunnerState.RUNNING).anyTimes();
```
efeg: Is this really expected to be called `anyTimes()`? Would it be possible to be more conservative? (applies to the other places).

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -198,6 +200,13 @@ public void testTimeoutLeaderActions() throws InterruptedException {
     EasyMock.expect(mockMetadataClient.refreshMetadata()).andReturn(clusterAndGeneration).anyTimes();
     EasyMock.expect(mockMetadataClient.cluster()).andReturn(clusterAndGeneration.cluster()).anyTimes();
     EasyMock.replay(mockMetadataClient);
+    LoadMonitor mockLoadMonitor = EasyMock.mock(LoadMonitor.class);
+    EasyMock.expect(mockLoadMonitor.taskRunnerState())
+            .andReturn(LoadMonitorTaskRunner.LoadMonitorTaskRunnerState.RUNNING)
+            .anyTimes();
```
efeg: Ditto on the use of `anyTimes()`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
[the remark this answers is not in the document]
kidkun: I ends in making `AnomalyDetector`  gets everything needed from `LoadMonitor` via `KafkaCruiseControl`, thus break the dependency loop.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/LoadMonitorTaskRunner.java
kidkun: Will do in another patch.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorTest.java
kidkun: changed to `times()`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -66,7 +67,7 @@
   private final LoadMonitor _loadMonitor;
   private final GoalOptimizer _goalOptimizer;
   private final ExecutorService _goalOptimizerExecutor;
-  private final Executor _executor;
+  private Executor _executor;
```
efeg: It is not clear why we drop `final` here. Can we avoid this change?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
```diff
@@ -148,19 +143,18 @@ public void run() {
...
               clusterModelSemaphore.close();
             }
-            clusterModelSemaphore = _loadMonitor.acquireForModelGeneration(new OperationProgress());
+            clusterModelSemaphore = _kafkaCruiseControl.acquireForModelGeneration(new OperationProgress());
             // Make cluster model null before generating a new cluster model so the current one can be GCed.
             clusterModel = null;
-            clusterModel = _loadMonitor.clusterModel(now,
-                                                     goal.clusterModelCompletenessRequirements(),
-                                                     new OperationProgress());
+            clusterModel = _kafkaCruiseControl.clusterModel(goal.clusterModelCompletenessRequirements(),
```
efeg: With this change, we do not need to pass `_time` to this class. Can we
1. drop _time from this class,
2. make the same change in other anomaly detectors?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/SampleLoadingTask.java
```diff
@@ -4,26 +4,35 @@
...
 
   SampleLoadingTask(SampleStore sampleStore,
                     KafkaPartitionMetricSampleAggregator partitionMetricSampleAggregator,
                     KafkaBrokerMetricSampleAggregator brokerMetricSampleAggregator,
-                    LoadMonitorTaskRunner loadMonitorTaskRunner) {
+                    LoadMonitorTaskRunner loadMonitorTaskRunner,
+                    Executor executor) {
```
efeg: Why don't we pass executorState rather than executor to this constructor?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
kidkun: This is part of the previous hacky way to get around of the cycler dependency. You are right, this is not needed anymore!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/task/SampleLoadingTask.java
kidkun: This is because currently every time we update the `_executorState` field inside `Executor`, we always doing by assign that field a new  `ExecutorState` object. 

This helps to make `ExecutorState` objects immutable and the good thing about that is the executor state returned in `state` endpoint is always consistent without extra locking.

Issue #963, opened 2019-10-03
Add support to dynamically change the metadata check frequency during executions


efeg:
The interval that the executor will check on the execution progress is configured by `execution.progress.check.interval.ms`. This interval helps rate-limiting the metadata requests from Cruise Control to Kafka.

Allow users to dynamically update this configuration during an ongoing execution. This helps being able alleviate the metadata pressure on Kafka.

PR #970, opened 2019-10-10
Support dynamically changing the metadata check frequency during executions.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/963.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -320,6 +320,16 @@ public class KafkaCruiseControl {
   }
 
   /**
+   * Dynamically set the interval between checking and updating (if needed) the progress of an initiated execution.
+   *
+   * @param requestedExecutionProgressCheckIntervalMs The interval between checking and updating the progress of an initiated
+   *                                                 execution (if null, use {@link Executor#MIN_EXECUTION_PROGRESS_CHECK_INTERVAL_MS}).
+   */
+  public void setRequestedExecutionProgressCheckIntervalMs(Long requestedExecutionProgressCheckIntervalMs) {
+    _executor.setRequestedExecutionProgressCheckIntervalMs(requestedExecutionProgressCheckIntervalMs);
+  }
+
+  /**
    * Dynamically set the inter-broker partition movement concurrency per broker.
    *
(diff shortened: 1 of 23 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -319,6 +319,16 @@ public void addRecentBrokersPermanently(Set<Integer> brokersToAdd, boolean isRem
     return isRemoved ? _executor.recentlyRemovedBrokers() : _executor.recentlyDemotedBrokers();
   }
 
+  /**
+   * Dynamically set the interval between checking and updating (if needed) the progress of an initiated execution.
+   *
+   * @param requestedExecutionProgressCheckIntervalMs The interval between checking and updating the progress of an initiated
+   *                                                 execution (if null, use {@link Executor#MIN_EXECUTION_PROGRESS_CHECK_INTERVAL_MS}).
```
kidkun: if requested is null, looks like we are using `_defaultExecutionProgressCheckIntervalMs` instead of `MIN_EXECUTION_PROGRESS_CHECK_INTERVAL_MS`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -649,6 +650,17 @@ public static Integer reviewId(HttpServletRequest request, boolean twoStepVerifi
...
 
+  /**
+   * Get the execution progress check interval in milliseconds. Default: {@code null}.
+   *
+   * @param request The Http request.
+   * @return Execution progress check interval in milliseconds.
+   */
+  static Long executionProgressCheckIntervalMs(HttpServletRequest request) {
+    String parameterString = caseSensitiveParameterName(request.getParameterMap(), EXECUTION_PROGRESS_CHECK_INTERVAL_MS_PARAM);
+    return parameterString == null ? null : Long.parseLong(request.getParameter(parameterString));
```
kidkun: maybe we can move the sanity check against `MIN_EXECUTION_PROGRESS_CHECK_INTERVAL_MS` here so if fails much faster for operations.

efeg: Copied the sanity check here for early termination -- kept the original one for other current (e.g. `admin` request), and potential future use-cases of this public method.

Issue #975, opened 2019-10-22
Use Cruise Control to detect slow broker


kidkun:
In our production environment, we observe Kafka broker could go wrong more than fail stop and has some offline disk(running on JBOD).  The broker could just have performance degradation and become a slow broker. And this could still be visible to clients, which result in increasing backlog in the producer buffer and making message producing call blocking.

Therefore, beside broker failure and disk failure, we need another anomaly type to detect and fix this. In our experience, the slow broker always due to disk performance degradation. For example, if one physical disk of the RAID10 goes down, the RAID disk does not go down as a whole but begin to perform at lower capacity.

There is a broker metric, `logFlushTime`, to indicate disk performance, but this metric alone is not enough, naively monitoring on it will cause a lot of false-positive alert(e.g. during traffic shift).  To accurate detect the anomaly, we should consider based on traffic metric and flush metic comprehensively.

And for the self-healing operation for detect slow broker, as an immediately remediation broker demotion is enough, it will remove the slow broker out of client's direct read/write patch in most case. If the broker consistently performs poor for an extended time, we may also consider remove the broker.

PR #976, opened 2019-10-22
Add support to detect and self heal slow broker.

kidkun:
Fix issue #975


kidkun:
Create a new PR based on the offline discussion #985 
The main change is to make `slow brokers` a sub type of `metric anomaly`. 
I do not want to override the commits here since we may revisit the approach in this PR(making `slow brokers` a top-level anomaly).

Issue #977, opened 2019-10-23
Force user to specify reason for requests which trigger operation


kidkun:
It is good to ask user to specify the reason in the request  when (s)he want to trigger an `add_broker`/`remove_broker`/`demote_broker`/`fix_offline_replicas`/`change_topic_replication_factor` operation.

These operation are alway heavy(take long time to run), if anything goes wrong in this process, checking the reason of original request helps user to determine what will the right thing to do next(i.e. re-trigger or abandon the request)

Also Cruise Control should return the reason when user checking state of `Executor` via `state` endpoint, which gives user a clear view of what Cruise Control is busy with.

PR #978, opened 2019-10-23
Add support to force user specifying reason for requests which trigger operation.

kidkun:
Fix issue #977

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
@@ -51,6 +51,7 @@ import org.apache.kafka.common.config.ConfigException;
 
 import static com.linkedin.kafka.cruisecontrol.config.CruiseControlParametersConfig.*;
 import static com.linkedin.kafka.cruisecontrol.config.CruiseControlRequestConfigs.*;
+import static com.linkedin.kafka.cruisecontrol.servlet.parameters.ParameterUtils.REASON_PARAM;
 import static org.apache.kafka.common.config.ConfigDef.Range.atLeast;
 import static org.apache.kafka.common.config.ConfigDef.Range.between;
 
@@ -935,6 +936,13 @@ public class KafkaCruiseControlConfig extends AbstractConfig {
       + "(verification) purgatory.";
 
   /**
+   * <code>request.reason.required</code>
+   */
+  public static final String REQUEST_REASON_REQUIRED_CONFIG = "request.reason.required";
+  private static final String REQUEST_REASON_REQUIRED_DOC = "Require specifying reason via " + REASON_PARAM + " parameter "
+      + "for non-dryrun POST request";
+
+  /**
    * <code>logdir.response.timeout.ms</code>
    */
   public static final String LOGDIR_RESPONSE_TIMEOUT_MS_CONFIG = "logdir.response.timeout.ms";
@@ -1119,6 +1127,11 @@ public class KafkaCruiseControlConfig extends AbstractConfig {
                 25,
                 atLeast(1),
                 ConfigDef.Importance.MEDIUM, TWO_STEP_PURGATORY_MAX_REQUESTS_DOC)
+        .define(REQUEST_REASON_REQUIRED_CONFIG,
+                ConfigDef.Type.BOOLEAN,
+                false,
+                ConfigDef.Importance.MEDIUM,
+                REQUEST_REASON_REQUIRED_DOC)
         .define(MAX_CACHED_COMPLETED_KAFKA_MONITOR_USER_TASKS_CONFIG,
                 ConfigDef.Type.INT,
                 null,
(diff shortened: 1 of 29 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -934,6 +935,13 @@
   private static final String TWO_STEP_PURGATORY_MAX_REQUESTS_DOC = "The maximum number of requests in two-step "
       + "(verification) purgatory.";
 
+  /**
+   * <code>request.reason.required</code>
+   */
+  public static final String REQUEST_REASON_REQUIRED_CONFIG = "request.reason.required";
+  private static final String REQUEST_REASON_REQUIRED_DOC = "Require specifying reason via " + REASON_PARAM + " parameter "
+      + "for non-dryrun POST request";
```
efeg: Is this config controls this behavior for _all_  or _selected_ POST requests -- i.e. should `stop_proposal_execution` get a `reason` parameter, as well? If it is only for selected endpoints, can we enumerate them?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailures.java
```diff
@@ -93,7 +93,8 @@ public void configure(Map<String, ?> configs) {
                                                            allowCapacityEstimation,
                                                            excludeRecentlyDemotedBrokers,
                                                            excludeRecentlyRemovedBrokers,
-                                                           _anomalyId)
+                                                           _anomalyId,
+                                                           String.format("Self healing for broker failure : %s", this))
```
efeg: Nit (applies to similar uses in other anomaly types for the sake of consistency): `broker failure` -> `%s` that uses `AnomalyType.BROKER_FAILURE`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailures.java
```diff
@@ -20,13 +19,14 @@
...
 import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.getSelfHealingGoalNames;
 import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.KAFKA_CRUISE_CONTROL_OBJECT_CONFIG;
+import static com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyType.BROKER_FAILURE;
 
 
 /**
  * The broker failures that have been detected.
  */
 public class BrokerFailures extends KafkaAnomaly implements CruiseControlConfigurable {
-  protected static final String ID_PREFIX = AnomalyType.BROKER_FAILURE.toString();
```
efeg: Now that we have a separation between the following:
```
  TRIGGERED_USER_TASK_ID = "triggeredUserTaskId";
  TRIGGERED_SELF_HEALING_TASK_ID = "triggeredSelfHealingTaskId";
  TRIGGERED_TASK_REASON = "triggeredTaskReason";
```

I am curious if we can drop `ID_PREFIX` overwrite from individual anomaly types (and handle the drop accordingly in the rest of the codebase) -> thus have an `anomalyId` that is a proper `uuid`?

kidkun: we have some places in our code depend on the assumption that self-healing task's uuid starts with anomaly type. 

One example is [determine whether mark `AnomalyDetector` or `UserTaskManager` when execution starts](https://github.com/linkedin/cruise-control/blob/d10aa73209ce0f542afca44ad7145e7b8fe8112d/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L699).
If we remove the prefix, we need to make decision based on `_reason` field, we can chose a reason template for self-healing tasks; but for user-triggered task, the `_reason` is provided by user, so theoretically user can accidentally provide a reason which matchs our pre-defined self-healing reason template.

efeg: Why not change the places that depends on `uuid` prefix? For the same reason, I don't think we should rely on parsing the `_reason` field.

For the specific example above, can't we just pass the information regarding `who` started the execution? Internally, we always know whether the execution was triggered by self healing or not; hence, it is possible to pass this information down to the executor.

kidkun: That's a good idea! 
I removed the prefix  in UUID and pass down the information of who triggers operation from `Runnable` to `Executor`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -692,16 +698,14 @@ public void run() {
...
       // retrieve the associated user task information.
-      if (AnomalyType.cachedValues().stream().anyMatch(type -> _uuid.startsWith(type.toString()))) {
-        userTaskInfo = null;
-      } else {
+      if (_isTriggeredByUserRequest) {
         userTaskInfo = _userTaskManager.markTaskExecutionBegan(_uuid);
       }
       _state = STARTING_EXECUTION;
-      _executorState = ExecutorState.executionStarted(_uuid, _reason, _recentlyDemotedBrokers, _recentlyRemovedBrokers);
+      _executorState = ExecutorState.executionStarted(_uuid, _reason, _recentlyDemotedBrokers, _recentlyRemovedBrokers, _isTriggeredByUserRequest);
```
efeg: Can we also log the `_reason` in the  `OPERATION_LOG` (please see the next line)?

Issue #982, opened 2019-10-25
Add support to uncleanly cancel hang replica/leadership movements


kidkun:
Currently Cruise Control does not guarantee that the `stop_proposal_execution` request always stop ongoing execution task. This is because without [KIP-455](https://cwiki.apache.org/confluence/display/KAFKA/KIP-455%3A+Create+an+Administrative+API+for+Replica+Reassignment)(which is going to be merged in Kafka release 2.4), there could be some ongoing replica/leadership movements hang at Kafka side.

For now, one way we can  interrupt hang replica/leadership movements is to 
(1) shut down Cruise Control
(2) delete the znode under `/admin`
(3) delete the znode `/controller` to triggers a controller re-election
(4) restart Cruise Control
(1) and (3) are needed to prevent Cruise Control or Controller to write back the task to znode once we manually delete it in (2)

It would be good if Cruise Control automate this process and also save user to reboot Cruise Control.

PR #983, opened 2019-10-25
Add support to uncleanly stop proposal execution.

kidkun:
For issue #982

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -87,6 +87,7 @@ public class Executor {
 
   // Some state for external service to query
   private final AtomicBoolean _stopRequested;
+  private final AtomicBoolean _forceStop;
   private final Time _time;
   private volatile boolean _hasOngoingExecution;
   private volatile ExecutorState _executorState;
@@ -180,6 +181,7 @@ public class Executor {
     _latestRemoveStartTimeMsByBrokerId = new ConcurrentHashMap<>();
     _executorState = ExecutorState.noTaskInProgress(recentlyDemotedBrokers(), recentlyRemovedBrokers());
     _stopRequested = new AtomicBoolean(false);
+    _forceStop = new AtomicBoolean(false);
     _hasOngoingExecution = false;
     _uuid = null;
     _executorNotifier = executorNotifier != null ? executorNotifier
@@ -519,6 +521,7 @@ public class Executor {
     _hasOngoingExecution = true;
     _anomalyDetector.maybeClearOngoingAnomalyDetectionTimeMs();
     _stopRequested.set(false);
+    _forceStop.set(false);
     _executionStoppedByUser.set(false);
     if (_isKafkaAssignerMode) {
       _numExecutionStartedInKafkaAssignerMode.incrementAndGet();
@@ -551,9 +554,11 @@ public class Executor {
 
   /**
    * Request the executor to stop any ongoing execution.
+   *
+   * @param forceExecutionStop Whether force execution to stop.
    */
-  public synchronized void userTriggeredStopExecution() {
-    if (stopExecution()) {
+  public synchronized void userTriggeredStopExecution(boolean forceExecutionStop) {
+    if (stopExecution(forceExecutionStop)) {
       LOG.info("User requested to stop the ongoing proposal execution.");
       _numExecutionStoppedByUser.incrementAndGet();
       _executionStoppedByUser.set(true);
@@ -563,10 +568,12 @@ public class Executor {
   /**
    * Request the executor to stop any ongoing execution.
    *
+   * @param forceExecutionStop Whether force execution to stop.
    * @return True if the flag to stop the execution is set after the call (i.e. was not set already), false otherwise.
    */
-  private synchronized boolean stopExecution() {
-    if (_stopRequested.compareAndSet(false, true)) {
+  private synchronized boolean stopExecution(boolean forceExecutionStop) {
+    if (_stopRequested.compareAndSet(false, true) &&
+        (!forceExecutionStop || _forceStop.compareAndSet(false, true))) {
       _numExecutionStopped.incrementAndGet();
       _executionTaskManager.setStopRequested();
       return true;
@@ -756,6 +763,16 @@ public class Executor {
           moveLeaderships();
           updateOngoingExecutionState();
         }
+
+        // 4. Delete tasks from Zookeeper if needed.
+        if (_state == STOPPING_EXECUTION && _forceStop.get()) {
+          // Delete znode of ongoing replica movement tasks.
+          _kafkaZkClient.deletePartitionReassignment();
+          // delete znode of ongoing leadership movement tasks.
+          _kafkaZkClient.deletePreferredReplicaElection();
+          // Delete controller znode to trigger a controller re-election.
+          _kafkaZkClient.deleteController();
+        }
       } catch (Throwable t) {
         LOG.error("Executor got exception during execution", t);
         _executionException = t;
@@ -801,6 +818,7 @@ public class Executor {
       _executorState = ExecutorState.noTaskInProgress(_recentlyDemotedBrokers, _recentlyRemovedBrokers);
       _hasOngoingExecution = false;
       _stopRequested.set(false);
+      _forceStop.set(false);
       _executionStoppedByUser.set(false);
     }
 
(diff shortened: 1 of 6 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -89,6 +89,7 @@
 
   // Some state for external service to query
   private final AtomicBoolean _stopRequested;
+  private final AtomicBoolean _forceStop;
```
efeg: Rather than having a separate `AtomicBoolean` here, do you think we can mimic the way Unix kill signals work -- e.g. we can assign a signal id using a single `AtomicInteger`? This also helps us make the kill logic more extensible.

Please see: https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -564,10 +569,12 @@ public synchronized void userTriggeredStopExecution() {
   /**
    * Request the executor to stop any ongoing execution.
    *
+   * @param forceExecutionStop Whether force execution to stop.
    * @return True if the flag to stop the execution is set after the call (i.e. was not set already), false otherwise.
    */
-  private synchronized boolean stopExecution() {
-    if (_stopRequested.compareAndSet(false, true)) {
+  private synchronized boolean stopExecution(boolean forceExecutionStop) {
```
efeg: If an execution is already being stopped either due to a previous user request that is not a force stop or due to a dead broker during an ongoing execution (i.e. triggered by CC) -- then it is possible that the execution will be stuck during stopping an execution.

If the user sends a stop execution request with a force option enabled, we should allow cleaning up the relevant znodes (i.e. we should allow unclean stop requests while there is an ongoing clean stop).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -757,6 +764,16 @@ private void execute() {
           moveLeaderships();
           updateOngoingExecutionState();
         }
+
+        // 4. Delete tasks from Zookeeper if needed.
+        if (_state == STOPPING_EXECUTION && _forceStop.get()) {
```
efeg: Nit: Can we move/extract this section to the relevant function?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -231,6 +231,58 @@ public void testTimeoutLeaderActions() throws InterruptedException {
...
+    Cluster cluster = new Cluster("id", Arrays.asList(node0, node1), Collections.singleton(partitionInfo),
+                                  Collections.emptySet(), Collections.emptySet());
+    MetadataClient.ClusterAndGeneration clusterAndGeneration = new MetadataClient.ClusterAndGeneration(cluster, 0);
+    EasyMock.expect(mockMetadataClient.refreshMetadata()).andReturn(clusterAndGeneration).anyTimes();
+    EasyMock.expect(mockMetadataClient.cluster()).andReturn(clusterAndGeneration.cluster()).anyTimes();
+    EasyMock.replay(mockMetadataClient);
+
+    Collection<ExecutionProposal> proposalsToExecute = Collections.singletonList(proposal);
+    Executor executor = new Executor(configs, time, new MetricRegistry(), mockMetadataClient, 86400000L,
+                                     43200000L, null, getMockUserTaskManager(RANDOM_UUID),
```
efeg: Nit: Can we avoid hardcoding constants? -- i.e. can move them to static variables?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -231,6 +231,58 @@ public void testTimeoutLeaderActions() throws InterruptedException {
     waitUntilExecutionFinishes(executor);
   }
 
+  @Test
+  public void testForceStopExecution() throws InterruptedException {
```
efeg: Running this test locally took more than 10seconds in my local test env. I noticed that it is using `Thread.sleep()`. Can we avoid using it in unit test?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
kidkun: good catch!

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
kidkun: All the tests in `ExecutorTest` are slow mainly due to the slowness of setting up a kafka cluster(and zookeeper).  

Here the `Thread.sleep(10);` is used in a while loop when we are waiting for executor to reach `INTER_BROKER_REPLICA_MOVEMENT_TASK_IN_PROGRESS` state and it is because we do not want to free burning CPU cycle waiting waiting Cruise Control executing. the time unit here is `millisecond` so it would not block unnecessarily long.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -552,9 +555,11 @@ private void sanityCheckOngoingReplicaMovement() {
 
   /**
    * Request the executor to stop any ongoing execution.
+   *
+   * @param forceExecutionStop Whether force execution to stop.
    */
-  public synchronized void userTriggeredStopExecution() {
-    if (stopExecution()) {
+  public synchronized void userTriggeredStopExecution(boolean forceExecutionStop) {
+    if (stopExecution(forceExecutionStop)) {
```
efeg: Can we update the log to indicate whether it was a force stop or not?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1060,9 +1086,9 @@ private int moveLeadershipInBatch() {
         if (!deadOrAbortingTasks.isEmpty()) {
           // TODO: re-enable this rollback action when KAFKA-6304 is available.
           // ExecutorAdminUtils.executeReplicaReassignmentTasks(_kafkaZkClient, deadOrAbortingTasks);
-          if (!_stopRequested.get()) {
+          if (_stopSignal.get() == NO_STOP_EXECUTION) {
             // If there is task aborted or dead, we stop the execution.
-            stopExecution();
+            stopExecution(false);
```
efeg: Nit: We are missing a log here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -996,20 +1009,29 @@ private int moveLeadershipInBatch() {
...
+        while (!_executionTaskManager.inExecutionTasks().isEmpty() && _stopSignal.get() == NO_STOP_EXECUTION) {
           waitForExecutionTaskToFinish();
         }
       }
       return numLeadershipToMove;
     }
 
+    private void deleteZnodesToStopExecution() {
+      // Delete znode of ongoing replica movement tasks.
+      _kafkaZkClient.deletePartitionReassignment();
```
efeg: Before deleting the znode, let's retrieve the stuck partitions and return them to the user in the response of force stop request (we should also log them). So that the user would know which topics are potentially inconsistent wrt RF of their partitions.

-- this also applies to `getControllerId()`.

kidkun: Returning these information in response will be hard.  
Since the real proposal execution happens asynchronously, i.e in the request handling logic, we just set the flag, The `ProposalExecutionRunnable` running in `ProposalExecutor` thread may pick it up [a little bit later](https://github.com/linkedin/cruise-control/blob/ba0c60ce531068f6c5f8192ff9a4ec1b70201d64/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L1044). 
Adding log is a good idea.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
efeg: I am curious if we can reduce the overall time by combining the cluster setup of multiple tests -- i.e. do we need to setup the cluster from scratch for each individual test case or can we do it once for selected test cases?

kidkun: There will be one issue if we pursue that way. 

Several tests in `ExecutorTest` will execute some partition movement/leadership movement tasks, If they share the same Kafka cluster under the hood, the tests will intervene each other.  It will be hard to implement/debug the tests because the cluster state at the beginning of test execution is changing.  And we will also observe different behavior between running specific test alone and running the suite.

Issue #649, opened 2019-04-09
cruise-control should provide an OpenAPI specification


mgrubent:
# Background
`cruise-control` currently uses servlets to define its available endpoints, parameters, and interaction mechanisms (e.g. header manipulation, POST/GET).
# Issue
While this server-side implementation is correct, its implementation does not allow for `cruise-control` client implementations except by human-inspection of the various endpoint and parameter classes.  

This in turn means that `cruise-control` client implementations are highly-manual, subject to implementation bugs, and do not update whenever the `cruise-control` server resources are extended.
# Proposal
`cruise-control` should adopt the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification) to fully define its REST interface.

By adopting this widely-used specification, `cruise-control`'s [server and client APIs would derive from a single source of truth](https://swagger.io/tools/swagger-codegen/), meaning that extensions to the REST interface could more-quickly propagate to both server- and client-side code.

Additionally, by adopting OpenAPI, `cruise-control` would make it significantly easier for users to [compile their own client APIs in a wide variety of popular languages](https://github.com/swagger-api/swagger-codegen#template-creator), including Java, Python, Go, PHP, JavaScript, Bash, Swift, etc.

efeg:
Closing the issue as it has been addressed in PRs linked to this ticket.

PR #984, opened 2019-10-29
Change optional fields in the JSON response of endpoints to nullable.

kidkun:
This patches is step one of issue #649 , i.e. to make Cruise Control compatible with `openAPI`.
With this change, we can specify the originally optional fields using [nullable](https://swagger.io/docs/specification/data-models/data-types/#null) keyword in `yaml` file describing the endpoints.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitorState.java
```diff
@@ -172,39 +172,41 @@ public static LoadMonitorState loading(int numValidSnapshotWindows,
...
+      loadMonitorState.put(NUM_FLAWED_PARTITIONS, _sampleExtrapolations.size());
+    } else {
+      loadMonitorState.put(TRAINED, null);
+      loadMonitorState.put(TRAINING_PCT, null);
+      loadMonitorState.put(NUM_MONITORED_WINDOWS, null);
+      loadMonitorState.put(MONITORED_WINDOWS, null);
+      loadMonitorState.put(NUM_VALID_PARTITIONS, null);
+      loadMonitorState.put(NUM_TOTAL_PARTITIONS, null);
+      loadMonitorState.put(MONITORING_COVERAGE_PCT, null);
+      loadMonitorState.put(NUM_FLAWED_PARTITIONS, null);
```
efeg: This `nullifying` pattern seems to involve so much repetition and a little error prone. I am curious whether it is possible to generalize this by creating some common wrapper for response keyset and then building util functions to nullify them?

kidkun: I did thought about that, my concern there is many case we ends in setting the value for the same field twice(first `null` then real value).  The good side is that the code will be clearer. 
Looks like it is worthy the cost.:)

kidkun:
@efeg , thanks for the feedback! It turns out optional is supported in open api spec(by default all fields are optional). We will use `required` keyword for the non-optional fields in our coming open api spec. 
Meanwhile, this patch is no longer needed.

PR #985, opened 2019-11-06
Add support to detect and self heal slow broker.

kidkun:
Fix issue #975


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -60,21 +60,25 @@
...
   private volatile long _numCheckedWithDelay;
   private final Object _shutdownLock;
 
   public AnomalyDetector(LoadMonitor loadMonitor,
                          KafkaCruiseControl kafkaCruiseControl,
                          Time time,
                          MetricRegistry dropwizardMetricRegistry) {
-    _anomalies = new LinkedBlockingDeque<>();
+    // For anomalies of different types, prioritize handling anomaly of higher priority;
+    // otherwise, handle anomaly in order of detected time.
```
efeg: I believe this _best-effort_ prioritization (because except for the broker failure detection, all other anomalies are detected periodically) should be avoided as 
1. the detection logic for each anomaly (should) already handle when to generate an anomaly in a best effort fashion -- e.g. please see (1) `GoalViolationDetector#shouldSkipGoalViolationDetection` (2) `DiskFailureDetector#shouldSkipDiskFailureDetection` and (3) `MetricAnomalyDetector#shouldSkipMetricAnomalyDetection`
2. it may cause starvation of certain anomaly types,
3. deprioritized anomalies will become irrelevant as their priority is dropped in the queue -- e.g. when would it be helpful to handle a `goal_violation` after handling a `broker_failure` that occurred after it? a broker failure self-healing will start a long-running execution and by the time the goal violation is picked from the anomaly queue, goal violation anomaly will be disposed due to the ongoing execution.
4. this prioritization makes the debugging harder.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -175,7 +179,7 @@ public void shutdown() {
     synchronized (_shutdownLock) {
       _shutdown = true;
     }
-    _anomalies.addFirst(SHUTDOWN_ANOMALY);
+    _anomalies.add(SHUTDOWN_ANOMALY);
```
efeg: This will not necessarily become the first item in the queue in case there are other broker failures -- am I missing something?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/DiskFailureDetector.java
```diff
@@ -26,44 +24,38 @@
...
 import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.shouldSkipAnomalyDetection;
+import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.ANOMALY_DETECTION_TIME_MS_CONFIG;
+import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.KAFKA_CRUISE_CONTROL_OBJECT_CONFIG;
 
 /**
  * This class detects disk failures.
  **/
 public class DiskFailureDetector implements Runnable {
   private static final Logger LOG = LoggerFactory.getLogger(DiskFailureDetector.class);
+  public static final String FAILED_DISKS_CONFIG = "failed.disks";
```
efeg: For consistent naming:
(1) `failed.disks` -> `failed.disks.object`?
(2) `FAILED_DISKS_CONFIG ` -> `FAILED_DISKS_OBJECT_CONFIG `?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
```diff
@@ -24,26 +22,14 @@
 public class AnomalyDetectorUtils {
   private static final Logger LOG = LoggerFactory.getLogger(AnomalyDetectorUtils.class);
   public static final String KAFKA_CRUISE_CONTROL_OBJECT_CONFIG = "kafka.cruise.control.object";
+  public static final String TIME_OBJECT_CONFIG = "time.object";
+  public static final String ANOMALY_DETECTION_TIME_MS_CONFIG = "anomaly.detection.time.ms";
```
efeg: Nit: `public static final String ANOMALY_DETECTION_TIME_MS_CONFIG = "anomaly.detection.time.ms";` -> `public static final String ANOMALY_DETECTION_TIME_MS_OBJECT_CONFIG = "anomaly.detection.time.ms.object";`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/KafkaAnomaly.java
```diff
@@ -4,17 +4,21 @@
...
 package com.linkedin.kafka.cruisecontrol.detector;
 
+import com.linkedin.cruisecontrol.common.CruiseControlConfigurable;
 import com.linkedin.cruisecontrol.detector.Anomaly;
+import com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyType;
 import com.linkedin.kafka.cruisecontrol.servlet.response.OptimizationResult;
+import java.util.Map;
 
-import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.getAnomalyType;
+import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.*;
```
efeg: Nit: Wildcard import.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/KafkaAnomaly.java
```diff
@@ -29,8 +33,35 @@ String optimizationResult(boolean isJson) {
...
+   * @return The type of anomaly.
+   */
+  public abstract AnomalyType anomalyType();
+
+  /**
+   * Get the detection time of anomaly.
+   *
+   * @return The detection time of anomaly.
+   */
+  public long detectionTimeMs() {
```
efeg: Can `detectionTimeMs` and `anomalyType` be moved to the `Anomaly` interface, as well? -- note that in this case, there should ideally be an interface for `AnomalyType`,as well. (please see `EndPoint` and `EndpointType` interfaces for similar use-cases).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/KafkaMetricAnomaly.java
```diff
@@ -121,4 +87,26 @@ String optimizationResult(boolean isJson) {
...
+    return METRIC_ANOMALY;
+  }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public void configure(Map<String, ?> configs) {
+    super.configure(configs);
+    _anomalyId = String.format("%s-%s", ID_PREFIX, UUID.randomUUID().toString().substring(ID_PREFIX.length() + 1));
+    _description = (String) configs.get(METRIC_ANOMALY_DESCRIPTION_CONFIG);
+    _fixable = (Boolean) configs.get(METRIC_ANOMALY_FIXABLE_CONFIG);
```
efeg: Nit: (1) `METRIC_ANOMALY_DESCRIPTION_CONFIG ` -> `METRIC_ANOMALY_DESCRIPTION_OBJECT_CONFIG `
(2) `METRIC_ANOMALY_FIXABLE_CONFIG ` -> `METRIC_ANOMALY_FIXABLE_OBJECT_CONFIG `

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/detector/metricanomaly/MetricAnomaly.java
```diff
@@ -6,27 +6,22 @@
 
 import com.linkedin.cruisecontrol.detector.Anomaly;
 import com.linkedin.cruisecontrol.model.Entity;
-import java.util.List;
+import java.util.Map;
```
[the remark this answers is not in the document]
kidkun: To do this, we need to add the dependency of `Kafka` to `cruise-control-core` submodule and jeopardize it as a generic library.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/detector/metricanomaly/MetricAnomalyFinderUtils.java
```diff
@@ -0,0 +1,37 @@
...
+   */
+  public static boolean isDataSufficient(int sampleCount,
+                                         double upperPercentile,
+                                         double lowerPercentile) {
+    if (upperPercentile >= 100.0 || upperPercentile <= 0.0 || lowerPercentile >= 100.0 || lowerPercentile <= 0.0) {
+      throw new IllegalArgumentException("Invalid percentile.");
+    }
+
+    int minNumValues = (int) Math.ceil(100 / (upperPercentile > 50.0 ? (100 - upperPercentile) : upperPercentile));
+    minNumValues = Math.max(minNumValues, (int) Math.ceil(100 / (lowerPercentile > 50.0 ? (100 - lowerPercentile) : lowerPercentile)));
```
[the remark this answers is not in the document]
kidkun: For example, the lower percentile is `10.0` and the upper percentile is `20.0`.  

In the existing logic, the minimal number of sample will be `Max(100.0 / 80.0, 100.0 / 90.0)=1`.  And one sample is not enough to to get an accurate `10.0` and `20.0` percentile.

After the change, the sample requirement become `Max(100.0 / 20.0, 100.0 / 10.0)`, which is 10.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
kidkun: The main reason I make this change is for scenario like the following.
1. A goal violation is detected at time 100
2. A slow broker is detected at time 101
3. AnomalyHandler picks up the next anomaly to work on

In this scenario, I want the slow broker to be picked up first because if CC first works on goal violation and trigger a long-time rebalance operation, it will not help slow broker issue and the existence of slow broker may cause backlog in the producer and cascade issue.
Therefore, it is better to first demote the slow broker(which will be fast), mark it as recently demoted broker, then do rebalance(thus no new leader will be moved to this slow broker).

For the concerns above, here are some of my thoughts.
1. This is preempting mechanism in anomaly detection, the priority queue gives us a similar preempting mechanism in anomaly handling. 

2. I feel this is ok, because for one thing, the high priority anomaly means more serious ones and deserves being addressed first; for another thing,  when CC is doing `remove_broker`(for broker failure or slow broker) or `fix_offline_replicas` (for disk failure), it actually runs customized rebalance, i.e. it still use the same set of goals thus helps to remedy goal violation(the starved type).

3. in that example, the order does not matter. but I think in the scenario I described above, the order change matters. In both scenario, changing to priority queue does not make the case worse.

4. I feel that is acceptable, as long as we can get the history of anomaly handling via [log](https://github.com/linkedin/cruise-control/blob/e73256a79d6ef9c3c1063452dd5e9eacd02e7d1c/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java#L419) and [operation log](https://github.com/linkedin/cruise-control/blob/db6281c2c458ded9dddad72606048bcc01f135ae/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L696).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -60,21 +60,25 @@
   private final DiskFailureDetector _diskFailureDetector;
   private final ScheduledExecutorService _detectorScheduler;
   private final long _anomalyDetectionIntervalMs;
-  private final LinkedBlockingDeque<Anomaly> _anomalies;
+  private final PriorityBlockingQueue<KafkaAnomaly> _anomalies;
```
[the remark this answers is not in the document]
kidkun: Please see my comments below.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
kidkun: Here is an implicit assumption.
`SHUTDOWN_ANOMALY` is a `BrokerFailures` with detection time to be 0, based on the current comparator(first check priority then detection time), it will be added to the head to the queue.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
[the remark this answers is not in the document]
kidkun: It is `detectionTimeMs` and `anomalyType`, I will move these API to `Anomaly` and revert the change here.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/detector/metricanomaly/PercentileMetricAnomalyFinderUtils.java
```diff
@@ -0,0 +1,48 @@
...
+    if (upperPercentile >= 100.0 || upperPercentile <= 0.0) {
+      throw new IllegalArgumentException(String.format("Provided upper percentile (%f) is invalid, it should be within (0.0, 100.0).",
+                                                       upperPercentile));
+    }
+    if (lowerPercentile >= 100.0 || lowerPercentile <= 0.0) {
+      throw new IllegalArgumentException(String.format("Provided lower percentile (%f) is invalid, it should be within (0.0, 100.0).",
+                                                       lowerPercentile));
+    }
+
+    if (lowerPercentile > lowerPercentile) {
```
efeg: Incorrect comparison between same values -- i.e. both are `lowerPercentile`.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/detector/metricanomaly/MetricAnomalyFinderUtils.java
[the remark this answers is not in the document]
efeg: Got it -- but does it ever make sense to set upper bound smaller than `50` (i.e. median) and the lower bound more than `50`? Maybe we should add sanity checks upon getting the config and avoid changing the way we compute `minNumValues`.

Also, can we add this explanation to the JavaDoc -- currently it is not very clear how this function works.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -58,19 +60,23 @@
...
   private final ScheduledExecutorService _detectorScheduler;
   private final long _anomalyDetectionIntervalMs;
-  private final LinkedBlockingDeque<Anomaly> _anomalies;
+  private final PriorityBlockingQueue<Anomaly> _anomalies;
   private volatile boolean _shutdown;
   private final AnomalyDetectorState _anomalyDetectorState;
   private final List<String> _selfHealingGoals;
   private final ExecutorService _anomalyLoggerExecutor;
-  private volatile Anomaly _anomalyInProgress;
+  private volatile KafkaAnomaly _anomalyInProgress;
```
efeg: Why are we using `KafkaAnomaly` rather than `Anomaly`? -- i.e. `KafkaAnomaly` is just one implementation of `Anomaly`.
Are there remaining changes needed in the interface to continue using the generic interface type?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
efeg: If slow broker handling is intended have a higher priority, then would scheduling its periodic check before goal violation checks help? -- i.e. in `AnomalyDetector#startDetection`, for `scheduleAtFixedRate` call can we set a _sufficiently_ earlier initial delay to set higher priority anomaly detectors?
As long as the `slow-broker-anomaly-generation-time` is less than the `difference between the other lower priority anomaly detector's start time` + `other-detector's-anomaly-generation` time, it will have higher priority. Note that we use the same frequency in (1) slow broker, (2) disk failure, and (3) goal violation detection; hence, the period of anomaly detection will stay fixed. Moreover, if the intention is to capture slow brokers with higher frequency, can't we run its detector more frequently, as well.

kidkun: Delay the detection start time only helps to some extent. 
The priority does not maintain after a self-healing operation because self-healing can take arbitrary time. It could be
 `slow broker detection skipped` -> `self-healing finished` -> `goal violation triggered`

Increase detection frequency sounds like a good idea. I will add configs for each anomaly type.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
kidkun: It is missed in the effort to revert `KafkaAnomaly` to `Anomaly`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -1744,11 +1775,28 @@
...
+                ConfigDef.Importance.LOW,
+                DISK_FAILURE_DETECTION_INTERVAL_MS_DOC)
         .define(ANOMALY_DETECTION_ALLOW_CAPACITY_ESTIMATION_CONFIG,
                 ConfigDef.Type.BOOLEAN,
                 DEFAULT_ANOMALY_DETECTION_ALLOW_CAPACITY_ESTIMATION_CONFIG,
                 ConfigDef.Importance.LOW,
                 ANOMALY_DETECTION_ALLOW_CAPACITY_ESTIMATION_DOC)
+        .define(BROKER_FAILURE_DETECTION_BACKOFF_MS_CONFIG,
+                ConfigDef.Type.LONG,
+                300000L,
```
efeg: Nit: Can we move the hardcoded value to a static variable?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -79,7 +82,17 @@ public AnomalyDetector(KafkaCruiseControl kafkaCruiseControl,
...
-    _anomalyDetectionIntervalMs = config.getLong(KafkaCruiseControlConfig.ANOMALY_DETECTION_INTERVAL_MS_CONFIG);
+    long anomalyDetectionIntervalMs = config.getLong(KafkaCruiseControlConfig.ANOMALY_DETECTION_INTERVAL_MS_CONFIG);
+    _goalViolationDetectionIntervalMs = config.getLong(KafkaCruiseControlConfig.GOAL_VIOLATION_DETECTION_INTERVAL_MS_CONFIG) != null ?
+                                        config.getLong(KafkaCruiseControlConfig.GOAL_VIOLATION_DETECTION_INTERVAL_MS_CONFIG) :
+                                        anomalyDetectionIntervalMs;
+    _metricAnomalyDetectionIntervalMs = config.getLong(KafkaCruiseControlConfig.METRIC_ANOMALY_DETECTION_INTERVAL_MS_CONFIG) != null ?
+                                        config.getLong(KafkaCruiseControlConfig.METRIC_ANOMALY_DETECTION_INTERVAL_MS_CONFIG) :
+                                        anomalyDetectionIntervalMs;
+    _diskFailureDetectionIntervalMs = config.getLong(KafkaCruiseControlConfig.DISK_FAILURE_DETECTION_INTERVAL_MS_CONFIG) != null ?
+                                      config.getLong(KafkaCruiseControlConfig.DISK_FAILURE_DETECTION_INTERVAL_MS_CONFIG) :
```
efeg: Nit: Can we avoid calling `getLong` twice in case the value is not `null` -- please see `SelfHealingNotifier#configure`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -1744,11 +1775,28 @@
                 300000L,
                 ConfigDef.Importance.LOW,
                 ANOMALY_DETECTION_INTERVAL_MS_DOC)
+        .define(GOAL_VIOLATION_DETECTION_INTERVAL_MS_CONFIG,
+                ConfigDef.Type.LONG,
+                ConfigDef.Importance.LOW,
+                GOAL_VIOLATION_DETECTION_INTERVAL_MS_DOC)
+        .define(METRIC_ANOMALY_DETECTION_INTERVAL_MS_CONFIG,
```
efeg: Should we add a sanity check to ensure that the value of this config is less frequent compared to the sampling? -- i.e. otherwise it is no-op.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -182,9 +198,9 @@ public void shutdown() {
     KafkaCruiseControlUtils.closeAdminClientWithTimeout(_adminClient);
 
     try {
-      _detectorScheduler.awaitTermination(_anomalyDetectionIntervalMs, TimeUnit.MILLISECONDS);
+      _detectorScheduler.awaitTermination(_goalViolationDetectionIntervalMs, TimeUnit.MILLISECONDS);
```
efeg: It is not clear why the `_detectorScheduler` termination time is associated with `_goalViolationDetectionIntervalMs`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -441,7 +457,7 @@ private void fixAnomalyInProgress(AnomalyType anomalyType) throws Exception {
             if (isReadyToFix) {
               LOG.info("Fixing anomaly {}", _anomalyInProgress);
               fixStarted = _anomalyInProgress.fix();
-              String optimizationResult = _anomalyInProgress.optimizationResult(false);
+              String optimizationResult = ((KafkaAnomaly)_anomalyInProgress).optimizationResult(false);
```
efeg: This casting seems a little flaky -- can we avoid this by making the anomaly interface (i.e. relevant return types, supported functions) even more generic?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
kidkun: There are a lot of hard-coded default values in config definition.
 I think we'd better change all of them in a separate patch.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
kidkun: moved `String optimizationResult(boolean isJson)` method to `Anomaly` interface.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
kidkun: Set to the max of all intervals.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
kidkun: add the sanity check.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -0,0 +1,246 @@
...
+  private static final double HISTORY_METRIC_MARGIN = 3.0;
+  // The percentile threshold used to compare last metric value again peers' latest value.
+  private static final double PEER_METRIC_PERCENTILE_THRESHOLD = 50.0;
+  // The margin used to compare last metric value again peers' latest value.
+  private static final double PEER_METRIC_MARGIN = 5.0;
+  // The score threshold to trigger a demotion for slow broker.
+  private static final int SLOW_BROKER_DEMOTION_SCORE = 5;
+  // The score threshold to trigger a removal for slow broker.
+  private static final int SLOW_BROKER_DECOMMISSION_SCORE = 50;
+  // The maximal ratio of slow brokers in the cluster to trigger self-healing operation.
```
efeg: Looks like this ratio Is not for the `cluster`, but only for the brokers that receive bytes during the sampling period. Is this intentional -- can we update the doc?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -0,0 +1,246 @@
...
+      Set<BrokerEntity> detectedSlowBrokers = detectSlowBrokers(metricsHistoryByBroker, currentMetricsByBroker);
+
+      Map<BrokerEntity, Long> brokersToDemote = new HashMap<>();
+      Map<BrokerEntity, Long> brokersToRemove = new HashMap<>();
+      for (BrokerEntity broker : detectedSlowBrokers) {
+        // Update slow broker detection time and slowness score.
+        Long currentTimeMs = _kafkaCruiseControl.timeMs();
+        _detectedSlowBrokers.putIfAbsent(broker, currentTimeMs);
+        Integer slownessScore = _brokerSlownessScore.compute(broker,
+                                                             (k, v) -> (v == null) ? 1 : Math.min(v + 1, SLOW_BROKER_DECOMMISSION_SCORE));
```
efeg: This slowness score is a little confusing. Can we split the content of `metricAnomalies` function to parseable smaller functions?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorTest.java
```diff
@@ -5,37 +5,46 @@
...
 import org.junit.Test;
 
+import static com.linkedin.kafka.cruisecontrol.common.DeterministicCluster.smallClusterModel;
 import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorState.NUM_SELF_HEALING_STARTED;
 import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.KAFKA_CRUISE_CONTROL_OBJECT_CONFIG;
+import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.ANOMALY_DETECTION_TIME_MS_OBJECT_CONFIG;
 import static com.linkedin.kafka.cruisecontrol.detector.BrokerFailureDetector.FAILED_BROKERS_OBJECT_CONFIG;
+import static com.linkedin.kafka.cruisecontrol.detector.MetricAnomalyDetector.METRIC_ANOMALY_BROKER_ENTITIES_OBJECT_CONFIG;
+import static com.linkedin.kafka.cruisecontrol.detector.MetricAnomalyDetector.METRIC_ANOMALY_FIXABLE_OBJECT_CONFIG;
+import static com.linkedin.kafka.cruisecontrol.detector.SlowBrokerFinder.*;
```
efeg: Nit: wildcard import

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorTest.java
```diff
@@ -5,37 +5,46 @@
 package com.linkedin.kafka.cruisecontrol.detector;
 
 import com.linkedin.cruisecontrol.detector.Anomaly;
+import com.linkedin.cruisecontrol.detector.AnomalyType;
```
efeg: Can we add a test case to verify that Anomaly Detector takes the priorities into account -- i.e. `PriorityBlockingQueue` for anomalies.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/KafkaMetricAnomalyFinderTest.java
```diff
@@ -60,12 +61,8 @@ public void testMetricAnomalies() {
     assertTrue("There should be exactly a single metric anomaly", anomalies.size() == 1);
 
     MetricAnomaly<BrokerEntity> anomaly = anomalies.iterator().next();
-    List<Long> expectedWindow = new ArrayList<>();
-    expectedWindow.add(21L);
-
-    assertEquals(55, anomaly.metricId().intValue());
-    assertEquals(_brokerEntity, anomaly.entity());
-    assertEquals(expectedWindow, anomaly.windows());
+    assertTrue(anomaly.entities().keySet().contains(_brokerEntity));
```
efeg: Nit: `.keySet().contains` -> `.containsKey`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -24,10 +25,11 @@
...
 import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetector.METRIC_REGISTRY_NAME;
-import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.getAnomalyType;
-import static com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyType.GOAL_VIOLATION;
-import static com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyType.METRIC_ANOMALY;
-import static com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyType.BROKER_FAILURE;
+import static com.linkedin.kafka.cruisecontrol.detector.notifier.KafkaAnomalyType.GOAL_VIOLATION;
+import static com.linkedin.kafka.cruisecontrol.detector.notifier.KafkaAnomalyType.METRIC_ANOMALY;
+import static com.linkedin.kafka.cruisecontrol.detector.notifier.KafkaAnomalyType.BROKER_FAILURE;
+import static com.linkedin.kafka.cruisecontrol.detector.notifier.KafkaAnomalyType.DISK_FAILURE;
+
```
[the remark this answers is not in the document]
kidkun: Agree. we should also extract the generic logic in `AnomalyDetector`, I will do it in another path, create #1016  to track this.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -0,0 +1,246 @@
...
+    _brokerSlownessScore = new HashMap<>();
+    _detectedSlowBrokers = new HashMap<>();
+    _percentile = new Percentile();
+  }
+
+  private Set<BrokerEntity> detectSlowBrokers(Map<BrokerEntity, ValuesAndExtrapolations> metricsHistoryByBroker,
+                                              Map<BrokerEntity, ValuesAndExtrapolations> currentMetricsByBroker) {
+    // Preprocess raw metrics to get the derived metrics for each broker.
+    Map<BrokerEntity, List<Double>> historyValueByBroker = new HashMap<>();
+    Map<BrokerEntity, Double> currentValueByBroker = new HashMap<>();
```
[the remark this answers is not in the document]
kidkun: I prefer not for this method, since the extracting logic in this method is pretty customized.  For thing, we have a customized filter to filter candidate broker( i.e. filter out broker current does not service traffic); For another thing, a major chunk in this method is about calculating the special derived metric.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -0,0 +1,279 @@
...
+
+/**
+ * This class will check whether there is broker performance degradation (i.e. slow broker) from collected broker metrics.
+ *
+ * Slow brokers are identified by calculating a derived broker metric
+ * {@code BROKER_LOG_FLUSH_TIME_MS_999TH / (ALL_TOPIC_BYTES_IN + ALL_TOPIC_REPLICATION_BYTES_IN) for each broker and then
+ * checking in two ways.
+ * <ul>
+ *   <li>Comparing the latest metric value against broker's own history. If the latest value is larger than
+ *       {@link #HISTORY_METRIC_MARGIN} * {@link #HISTORY_METRIC_PERCENTILE_THRESHOLD} of historical values, it is
```
efeg: Nit: Missing parenthesis causing ambiguity.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -0,0 +1,279 @@
...
+ *       considered to be abnormally high.</li>
+ *   <li>Comparing the latest metric value against the latest metric value of all active brokers in cluster (i.e. brokers
+ *       which serve non-zero traffic). If the value is larger than {@link #PEER_METRIC_MARGIN} * {@link #PEER_METRIC_PERCENTILE_THRESHOLD}
```
efeg: Nit: Missing parenthesis causing ambiguity.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -0,0 +1,279 @@
...
+ *       of all metric values, it is considered to be abnormally high.</li>
+ * </ul>
+ *
+ * If certain broker's metric value is abnormally high, the broker is marked as a slow broker suspect by the finder.
+ * Then if this suspect broker's derived metric anomaly persists for some time, it is confirmed to be a slow broker and
+ * the finder will report {@link SlowBrokers}} anomaly with broker demotion as self-healing proposal. If the metric
+ * anomaly still persists for an extended time, the finder will eventually report {@link SlowBrokers}} anomaly with broker
+ * removal as self-healing proposal.
```
efeg: Nit: JavaDoc comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -0,0 +1,279 @@
...
+ *
+ * The time to report slow broker for demotion and removal is controlled by an internal broker scoring system.
+ * The system keeps a "slowness score" for brokers which have metric anomaly detected recently. The scores are updated in
+ * each round of detection with following rules.
+ * <ul>
+ *   <li> For any broker not in the scoring system, once there is metric anomaly detected on it, the broker is added to the system
+ *        with the initial "slowness score" of one. </li>
+ *   <li> For any broker in the scoring system, if there is metric anomaly detected on it, its "slowness score" increases
+ *        by 1. Once the score exceeds {@link #SLOW_BROKER_DEMOTION_SCORE}, finder begin to report the broker as slow broker
```
efeg: Nit: `begin` -> `begins`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorTest.java
```diff
@@ -97,7 +109,7 @@ private static void shutdownDetector(ScheduledExecutorService mockDetectorSchedu
                                        ScheduledExecutorService executorService) throws InterruptedException {
     mockDetectorScheduler.shutdown();
     EasyMock.expectLastCall().andDelegateTo(executorService);
-    EasyMock.expect(mockDetectorScheduler.awaitTermination(MOCK_ANOMALY_DETECTION_INTERVAL_MS, TimeUnit.MILLISECONDS)).andDelegateTo(executorService);
+    EasyMock.expect(mockDetectorScheduler.awaitTermination(5000, TimeUnit.MILLISECONDS)).andDelegateTo(executorService);
```
efeg: Nit: Can we use a constant rather than hardcoding the value?

Issue #992, opened 2019-11-13
Eliminate superfluous broker failure notifications


efeg:
Users may receive an excessive number of notifications and log messages regarding the same broker failures and the corresponding self healing activities.

Reasons for this include (1) rescheduled broker failure anomalies due to hard goal violations during self-healing, (2) flaky `brokers>ids` znode triggering repetitive children change events, and (3) missing the anomalies queue cleanup in case of optimization failures during self healing.

-- We should eliminate superfluous broker failure notifications.

PR #993, opened 2019-11-13
Eliminate superfluous broker failure notifications.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/992.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
@@ -267,7 +267,7 @@ public class AnomalyDetector {
           LOG.warn("Unexpected state prevents anomaly detector from handling the anomaly {}.", _anomalyInProgress, ise);
           // An illegal state may indicate a transient process blocking self-healing (e.g. an ongoing execution not
           // started by Cruise Control).
-          postProcessAnomalyInProgress = true;
+          postProcessAnomalyInProgress = false;
         } catch (Throwable t) {
           LOG.error("Uncaught exception in anomaly handler.", t);
           postProcessAnomalyInProgress = true;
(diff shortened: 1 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -256,16 +256,25 @@ public void run() {
...
+          LOG.warn("Encountered optimization failure when trying to fix the anomaly {}.", _anomalyInProgress, ofe);
+          // If self-healing failed due to an optimization failure, that indicates a hard goal violation; hence there is
+          // no further processing anomaly detector can do without human intervention for the anomaly (i.e. other than
+          // what has already been done in the {@link #handlePostFixAnomaly(boolean, boolean, String)}).
+          postProcessAnomalyInProgress = false;
+        } catch (IllegalStateException ise) {
+          LOG.warn("Unexpected state prevents anomaly detector from handling the anomaly {}.", _anomalyInProgress, ise);
+          // An illegal state may indicate a transient process blocking self-healing (e.g. an ongoing execution not
+          // started by Cruise Control).
+          postProcessAnomalyInProgress = true;
```
kidkun: Looks like `IllegalStateException` is thrown when notifier reports un unsupported notification result [here](https://github.com/linkedin/cruise-control/blob/011b04aefec72cda142a0d067c9f943646915af1/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java#L311). In this case should  we set `postProcessAnomalyInProgress `  to false here.

I think for the scenario mentioned in comment(i.e. `an ongoing execution not started by Cruise Control`), we seems to already call `postProcessAnomalyInProgress` to handle [here](https://github.com/linkedin/cruise-control/blob/33a1fef9ee75ef4baff5552019a6fc8bda4b7db5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java#L294).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
```diff
@@ -418,34 +434,40 @@ private void fixAnomalyInProgress(AnomalyType anomalyType) throws Exception {
...
-                LOG.info("[{}] Self-healing started successfully.", anomalyId);
-              } else {
-                LOG.warn("[{}] Self-healing failed to start.", anomalyId);
-              }
             }
+          } catch (OptimizationFailureException ofe) {
+            _anomalyLoggerExecutor.submit(() -> logSelfHealingOperation(anomalyId, ofe, null));
+            throw ofe;
+          } finally {
+            handlePostFixAnomaly(isReadyToFix, startedSuccessfully, anomalyId);
```
kidkun: If we get an `OptimizationFailureException`(due to hard goal violation) when fixing a `BrokerFailure` here, this `handlePostFixAnomaly` will be called and we schedule a `detectBrokerFailures()` with 0ms delay. 

This  `detectBrokerFailures()` is called with `skipReportingIfNotUpdated` is set to false,  so a new `BrokerFailure` will be reported immediately.

Then even if `SelfHealingNotifier` will not send a new alert(which is good) but handler will try to fix and fail again. Looks like it forms a loop?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
efeg: Good catch! We should ideally have another `Exception` type for ongoing execution exceptions, because `IllegalStateException` may be thrown from other code paths, as well -- but for now this seems like a reasonable assumption.
-- Created the following issue to address the issue better: https://github.com/linkedin/cruise-control/issues/1000

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetector.java
efeg: Thanks -- addressed the issue!

Issue #994, opened 2019-11-13
Show current controller information in Kafka Cluster State


efeg:
This information is already available in the metadata. Monitoring the controller, _a SPOF in Kafka_, information enriches the operational toolbox of cluster operators -- e.g. to alleviate the load on the controller node, they would prefer not to place any replicas on the broker that also has the controller.

PR #995, opened 2019-11-13
Show controller information in Kafka Cluster State.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/994.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/KafkaClusterState.java
```diff
@@ -204,6 +207,12 @@ protected void populateKafkaBrokerState(Map<Integer, Integer> leaderCountByBroke
         leaderCountByBrokerId.put(nodeId, 0);
       }
     }
+    // Part-3: Gather controller information.
+    replicaCountByBrokerId.keySet().forEach(brokerId -> isControllerByBrokerId.put(brokerId, false));
```
kidkun: how about brokers without replicas? Should we also include them here?

efeg: `replicaCountByBrokerId` includes both brokers with and without replicas -- i.e. has `0` replica count. Am I missing something?

kidkun: Ah, I missed in part 2, we actually added to map.

PR #1010, opened 2019-11-20
add sasl.login.callback.handler.class to AdminClient configs if exists

adoron:
Respect the `sasl.login.callback.handler.class` config if specified


adoron:
@efeg BTW, why don't you just add all the configs in KafkaCruiseControlConfig like you do for the KafkaConsumer?

efeg:
> @efeg BTW, why don't you just add all the configs in KafkaCruiseControlConfig like you do for the KafkaConsumer?

Would you clarify what you mean by adding all the configs in KafkaCruiseControlConfig -- i.e. exposing them via `KafkaCruiseControlConfig`?

adoron:
> > @efeg BTW, why don't you just add all the configs in KafkaCruiseControlConfig like you do for the KafkaConsumer?
> 
> Would you clarify what you mean by adding all the configs in KafkaCruiseControlConfig -- i.e. exposing them via `KafkaCruiseControlConfig`?

I mean props.putAll(configs) like here: https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsReporterSampler.java#L235
instead of adding the ssl/sasl configs one by one

Issue #1014, opened 2019-11-22
Batching the task slow execution alert


kidkun:
PR #981 introduce a way to send out alert if some tasks execution is too slow (i.e. execution time is longer than a pre-defined threshold). But in that patch the alert is sent out per task, which could result in a flood of alerting emails for large Kafka clusters.

It is better to just send out alerts periodically and each time one alert contains all tasks detected

PR #1015, opened 2019-11-22
Make slow broker execution alerts to be sent out in batch.

kidkun:
Patch for #1014

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1191,13 +1191,7 @@ public class Executor {
     private void handleDeadOrAbortingTasks(Set<ExecutionTask> deadOrAbortingNonLeadershipTasks,
                                            List<ExecutionTask> slowTasksToReport) {
       if (!slowTasksToReport.isEmpty()) {
-        StringBuilder sb = new StringBuilder();
-        sb.append("Slow tasks are detected:\n");
-        for (ExecutionTask task: slowTasksToReport) {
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1107,6 +1114,8 @@ private void deleteZnodesToStopExecution() {
           LOG.debug("Tasks in execution: {}", _executionTaskManager.inExecutionTasks());
         }
         List<ExecutionTask> deadOrAbortingTasks = new ArrayList<>();
+        List<ExecutionTask> slowTasksToReport = new ArrayList<>();
+        boolean shouldReportSlowTasks = _time.milliseconds() - _lastSlowTaskReportingTimeMs > SLOW_TASK_ALERTING_BACKOFF_TIME_MS;
```
efeg: Which code path updates the value of `_lastSlowTaskReportingTimeMs`? 
-- i.e. It seems it will always be `-1`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1137,6 +1148,15 @@ private void deleteZnodesToStopExecution() {
             }
           }
         }
+        if (!slowTasksToReport.isEmpty()) {
```
efeg: Nit: This blob seems to perform a specific task -- should we move it to a function?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
kidkun: Good catch!

PR #1025, opened 2019-11-27
Add support to perform sanity check between openApi spec and Java response class.

kidkun:
This patch is the first patch of the effort to add OpenApi spec to Cruise Control(see  #649).

This patch create a new test,  [JsonResponseTest](https://github.com/kidkun/cruise-control/blob/39be630db802c1fd81748afa2185710dcd109360/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/response/JsonResponseTest.java#L40), which can read in OpenApi spec written in `yaml` format and check its JSON schema definition against what is defined in Java response class.

To achieve this, we need to add some metadata in Java response classes which the can help the test to deduce the what is the actual JSON response layout. 
This patch adds three Annotations for this purpose.
1. [JsonResponseClass](https://github.com/kidkun/cruise-control/blob/39be630db802c1fd81748afa2185710dcd109360/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/JsonResponseClass.java#L23) to mark Java response class.
2. [JsonResponseField](https://github.com/kidkun/cruise-control/blob/39be630db802c1fd81748afa2185710dcd109360/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/JsonResponseField.java#L19) to mark the field keys defined in Java response class.
3. [JsonResponseExternalFields](https://github.com/kidkun/cruise-control/blob/39be630db802c1fd81748afa2185710dcd109360/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/JsonResponseExternalFields.java#L20) to marks field keys which is defined in other class but needed for Java response class.

This patch adds annotations for following endpoints
 1. `REBALANCE`
 2. `BOOTSTRAP`
 3. `TRAIN`
 4. `FIX_OFFLINE_REPLICAS`
 5. `STOP_PROPOSAL_EXECUTION`
 6. `PAUSE_SAMPLING`
 7. `RESUME_SAMPLING`
 8. `KAFKA_CLUSTER_STATE`
 9. `REVIEW_BOARD`
 10. `ADMIN`
 11. `REVIEW`

For the rest endpoints, the annotation will be added in separate patches(together with the `yaml` files).


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
```diff
@@ -115,4 +129,38 @@ private void ensureMutable() {
...
   }
+
+  @JsonResponseClass
+  private static class StepProgress {
+    @JsonResponseField
+    static final String STEP = "step";
+    @JsonResponseField
+    static final String DESCRIPTION = "description";
+    @JsonResponseField
+    static final String TIME_IN_MS = "timeInMs";
```
efeg: This patch seems to introduce backwards incompatible name changes, which potentially breaks the existing clients / UIs that depend on CC.

Can we ensure that no such backwards incompatible changes are introduced?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/Statistic.java
```diff
@@ -4,16 +4,21 @@
...
 public enum Statistic {
+  @JsonResponseField
   AVG("AVG"),
+  @JsonResponseField
   MAX("MAX"),
+  @JsonResponseField
   MIN("MIN"),
-  ST_DEV("STD");
+  @JsonResponseField
+  ST_DEV("ST_DEV");
```
efeg: Nit: Backwards incompatible change -- is it a breaking change used in json responses?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
```diff
@@ -218,28 +225,9 @@ public String getJSONString() {
...
-      resourceMap.put(AnalyzerUtils.POTENTIAL_NW_OUT, potentialNwOutUtilizationStats().get(stat));
-      resourceMap.put(AnalyzerUtils.REPLICAS, replicaStats().get(stat));
-      resourceMap.put(AnalyzerUtils.LEADER_REPLICAS, leaderReplicaStats().get(stat));
-      resourceMap.put(AnalyzerUtils.TOPIC_REPLICAS, topicReplicaStats().get(stat));
-      allStatMap.put(stat.stat(), resourceMap);
-    }
-    statMap.put(AnalyzerUtils.METADATA, basicMap);
-    statMap.put(AnalyzerUtils.STATISTICS, allStatMap);
+    Map<String, Object> statMap = new HashMap<>(2);
+    statMap.put(METADATA, new ClusterModelStatsMetaData().getJsonStructure());
```
efeg: Let's avoid backward incompatible name changes.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/DiskStats.java
```diff
@@ -4,23 +4,30 @@
...
   private final int _numReplicas;
   // For dead disk, its utilization will be null.
   private final Double _utilization;
   private final double _capacity;
-  private static final String DISK_MB = "DiskMB";
-  private static final String DISK_PCT = "DiskPct";
-  private static final String NUM_LEADER_REPLICAS = "NumLeaderReplicas";
-  private static final String NUM_REPLICAS = "NumReplicas";
+  @JsonResponseField
+  private static final String DISK_MB = "diskMB";
```
efeg: backwards incompatible name change

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
kidkun: Sure, I will revert them back.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
kidkun: Here I actually refactor out the sub layer of JSON response into another Classes(`ClusterModelStatsMetaData` and `  ClusterModelStatsValue` here), the actual data layout does not change.

The reason I do this is because it is hard to annotate a class if it contains keys of multiple JSON layers, in that case we need to distinguish which layer the key should resides in.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/JsonResponseField.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface JsonResponseField {
+
+  /**
+   * Whether the field is required in JSON response or not.
+   * @return True if it is a required field; otherwise it is an optional field.
+   */
+  boolean required() default true;
```
efeg: Why is the property of being required to is limited to `json`? -- i.e. in a given request, is there a case, where a field is required for json, but not required for `plaintext` request?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/KafkaClusterState.java
```diff
@@ -114,376 +97,412 @@ protected int minInsyncReplicas(String topic) {
...
-  protected String getPlaintext(CruiseControlParameters parameters) {
-    StringBuilder sb = new StringBuilder();
-    try {
-      // Broker summary.
-      writeBrokerSummary(sb);
-      // Partition summary.
-      writePartitionSummary(sb, parameters);
-    } catch (InterruptedException | ExecutionException e) {
-      throw new RuntimeException("Failed to populate broker logDir state.", e);
+
```
efeg: Nit: redundant space.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/KafkaClusterState.java
```diff
@@ -114,376 +97,412 @@ protected int minInsyncReplicas(String topic) {
...
+  @JsonResponseClass
+  protected static class PartitionState {
```
efeg: Nit on style: I am curious about why we prefer using a static inner class as opposed to a separate class.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/response/ResponseTest.java
```diff
@@ -0,0 +1,216 @@
...
+   * to build JSON response, which are annotated with {@link JsonResponseClass}.
+   *
+   * @throws ClassNotFoundException If the definition of the class is not found.
+   */
+  @Before
+  public void scanForResponseClass() throws ClassNotFoundException {
+    _schemaToClass = new HashMap<>();
+    ScanResult scanResult = new ClassGraph().ignoreClassVisibility()
+                                            .enableAnnotationInfo()
+                                            .whitelistPackages("com.linkedin.kafka.cruisecontrol")
```
efeg: Nit: Let's avoid hardcoding.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/response/ResponseTest.java
```diff
@@ -0,0 +1,216 @@
...
+      }
+    }
+  }
+
+  /**
+   * For each endpoint, check all the possible responses defined in OpenAPI spec.
+   *
+   * @param operation The OpenAPI operation for certain endpoint.
+   */
+  private void checkOperation(Operation operation) {
```
efeg: Nit: Let's avoid hardcoding in this function.

cruise-control/src/yaml/responses/clusterModelStats.yaml
```diff
@@ -32,15 +32,15 @@ ClusterModelStatsValue:
     - AVG
     - MAX
     - MIN
-    - ST_DEV
+    - STD
```
efeg: Is this change intentional? Shouldn't we have a test/sanity check that fails due to mismatch between `ST_DEV` and `STD`?

cruise-control/src/yaml/responses/progressResult.yaml
```diff
@@ -15,14 +15,14 @@ StepProgress:
   required:
     - step
     - description
-    - timeInMs
+    - time-in-ms
```
efeg: I see that the current parameter name is `time-in-ms` -- were we missing the sanity check in the earlier version about name verification?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
```diff
@@ -20,14 +23,24 @@
...
+  protected static final String OPERATION = "operation";
+  @JsonResponseField
+  protected static final String OPERATION_PROGRESS = "operationProgress";
   private boolean _mutable = true;
   private List<OperationStep> _steps = new ArrayList<>();
   private List<Long> _startTimes = new ArrayList<>();
+  private String _operation;
+
+  public  OperationProgress() {
+    this("");
```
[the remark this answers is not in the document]
kidkun: This is to keep the old API and it is used in when we start an async job internally(e.g. from [self-healing ](https://github.com/linkedin/cruise-control/blob/3b0ea9cc47fc7c73c009ca9bb7903d5a1ffbfa3d/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java#L153)). Here we do not care the operation name since it is invisible from client.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationProgress.java
```diff
@@ -115,4 +129,38 @@ private void ensureMutable() {
       throw new IllegalStateException("Cannot change this operation progress because it is immutable.");
     }
   }
+
+  @JsonResponseClass
+  private static class StepProgress {
```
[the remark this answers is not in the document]
kidkun: discussed offline, change the inner class to a seperate class.

cruise-control/src/yaml/responses/clusterModelStats.yaml
kidkun: This is before the [latest commit](https://github.com/linkedin/cruise-control/pull/1025/commits/fbfa8eebc29003e064dd24da6f4d32198f84903f) it is `ST_DEV` in both Java class and `yaml` file, and this commit make change in both place at the same time.

cruise-control/src/yaml/responses/progressResult.yaml
kidkun: same as above, the [latest commit](https://github.com/linkedin/cruise-control/pull/1025/commits/fbfa8eebc29003e064dd24da6f4d32198f84903f) make the change in both places at the same time.

efeg: Was this due to a backwards incompatible change in a previous commit? Should we change it back to its original version (if this was unintentional)?

kidkun: In #1031, we checked in some `yaml` files which has the wrong key(e.g. `timeInMs`) defined, it is missed in review. 

In this patch, since we add the sanity check for response,  `gradle build` failed and I noticed these inconsistency. 

My initial fix in this PR is to change the field in `Java` class(i.e. real response) since I feel the key in `yaml` keys look better(standard camel format). But it is back-incompatible, so in latest commit of this PR, I change the field in `yaml` instead.

 So there is no backwards incompatible introduced by previous commits, it is bug in `yaml` files.

efeg: Got it -- thanks!

Issue #1027, opened 2019-11-27
Rebalance with PreferredLeaderElectionGoal should not change the the order of replica in the replica list.


kidkun:
Currently, when running a PLE, Cruise Control tries to make the first replica which is (1) not offline and (2) not on a demoted broker/disk to be the new leader by making it the first replica in the partition's replica list, which may not be the ideal way for several reasons. 

(1) Normally when user are running PLE, they want the leader replica to be the one initially assigned when the partition created, i.e. the first replica in partition's **current** replica list.
(2) Swapping position of replicas in replica list is considered as a replica reassignment (even through no data is moved), thus too many replica position swaps may unnecessarily prolong the time to run PLE.
(3) If there is offline replica in the partition, running unnecessary replica position swap may get stuck for undefined long time (to wait for the offline replica goes back to online and catch up)

PR #1028, opened 2019-11-27
Make PreferredLeaderElectionGoal to always pick the first replica in replica list as new leader.

kidkun:
For #1027

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/PreferredLeaderElectionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/PreferredLeaderElectionGoal.java
@@ -112,7 +112,7 @@ public class PreferredLeaderElectionGoal implements Goal {
         if (hasBrokerOrDiskToBeDemoted && !partitionsToMove.contains(p.topicPartition())) {
           continue;
         }
-        for (int i = 0; i < p.replicas().size() ; i++) {
+        for (int i = 0; i < p.replicas().size(); i++) {
           // If there is no broker or disk to be demoted, only try to transfer the leadership to the first replica of the partition.
           if (!hasBrokerOrDiskToBeDemoted && i > 0) {
             break;
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/PreferredLeaderElectionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/PreferredLeaderElectionGoal.java
@@ -112,7 +112,12 @@ public class PreferredLeaderElectionGoal implements Goal {
         if (hasBrokerOrDiskToBeDemoted && !partitionsToMove.contains(p.topicPartition())) {
           continue;
         }
-        for (Replica r : p.replicas()) {
+        for (int i = 0; i < p.replicas().size() ; i++) {
+          // If there is no broker or disk to be demoted, only try to transfer the leadership to the first replica of the partition.
+          if (!hasBrokerOrDiskToBeDemoted && i > 0) {
+            break;
+          }
+          Replica r = p.replicas().get(i);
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/PreferredLeaderElectionGoal.java
```diff
@@ -103,50 +103,46 @@ public boolean optimize(ClusterModel clusterModel, Set<Goal> optimizedGoals, Opt
...
+                                    p.topicPartition(), candidateBroker.id(), candidateReplica.disk().logDir()));
+            failedPartitions++;
+          } else {
+            clusterModel.relocateLeadership(candidateReplica.topicPartition(), p.leader().broker().id(), candidateBroker.id());
           }
         }
       }
     }
+    if (failedPartitions > 0) {
+      LOG.warn("Leadership transfer of {} partitions are skipped. Details:\n{}", failedPartitions, sb.toString());
```
efeg: Nit: Using `\n` makes it harder to parse and print `warn` logs -- can we avoid it?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/PreferredLeaderElectionGoal.java
```diff
@@ -103,50 +103,46 @@ public boolean optimize(ClusterModel clusterModel, Set<Goal> optimizedGoals, Opt
...
-                       + "replicas are demoted.", p.topicPartition(), leaderCandidate.id());
-            }
-            if (r.disk() != null && r.disk().state() == Disk.State.DEMOTED) {
-              LOG.warn("The leader of partition {} has to be on a demoted disk {} of broker {} because all the alive "
-                       + "replicas are demoted.", p.topicPartition(), r.disk().logDir(), leaderCandidate.id());
-            }
-            break;
+        Replica candidateReplica = p.replicas().get(0);
+        Broker candidateBroker = candidateReplica.broker();
+        if (!candidateReplica.isLeader()) {
```
efeg: If this goal is called due to a demote request, then (if possible to do so) the leadership should be removed from the current leader regardless of whether it is transferred to the most preferred replica or not.

Overall, rather than having all these changes, can't we just
1. Identify if this request is purely a PLE as opposed to a demote request, and
2. If it is a pure PLE request, skip the partition if its most preferred replica is offline?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/PreferredLeaderElectionGoal.java
```diff
@@ -103,50 +103,46 @@ public boolean optimize(ClusterModel clusterModel, Set<Goal> optimizedGoals, Opt
...
+          } else if (excludedBrokersForLeadership.contains(candidateBroker.id())) {
+            sb.append(String.format("Skipped leadership transfer of partition %s to broker %d because it is among brokers excluded"
+                                    + " for leadership %s.%n", p.topicPartition(), candidateBroker.id(), excludedBrokersForLeadership));
+            failedPartitions++;
+          } else if (clusterModel.demotedBrokers().contains(candidateBroker)) {
+            sb.append(String.format("Skipped leadership transfer of partition %s to broker %d because it is among recently "
+                                    + "demoted brokers %s.%n\" ", p.topicPartition(), candidateBroker.id(), clusterModel.demotedBrokers()));
+            failedPartitions++;
+          } else if (candidateReplica.disk() != null && candidateReplica.disk().state() == Disk.State.DEMOTED) {
+            sb.append(String.format("Skipped leadership transfer of partition %s to broker %d because the hosting disk %s is recently demoted.%n",
```
efeg: Nit: For the same reason discussed below, can we avoid adding new lines to each string? Also having the same string repeated for each such error case seems wasteful and hard to read. Can't we just combine each relevant information in a collection and then dump it to a log (with a single prefix for each class of errors) when we are out of the loop?

Issue #1026, opened 2019-11-27
Ensure that PreferredReplicaLeaderElectionPath is never updated when it znode exists


efeg:
If an ongoing leadership movement attempts to update `PreferredReplicaLeaderElectionPath`, then the execution fails with `kafka.common.AdminCommandFailedException: Admin command failed`.

Ensure that regardless of whether `Executor#isLeadershipMovementDone` confirms the completed leadership movement, a new leadership movement batch is not initiated unless `PreferredReplicaLeaderElectionPath` does not exist.

PR #1029, opened 2019-11-28
Ensure that PreferredReplicaLeaderElectionPath is never updated when the zNode exists.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/1026.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -537,7 +537,7 @@ public class Executor {
                               Long replicationThrottle,
                               boolean isTriggeredByUserRequest) {
     _executionStoppedByUser.set(false);
-    sanityCheckOngoingReplicaMovement();
+    sanityCheckOngoingMovement();
     _hasOngoingExecution = true;
     _anomalyDetector.maybeClearOngoingAnomalyDetectionTimeMs();
     _stopSignal.set(NO_STOP_EXECUTION);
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1062,7 +1075,16 @@ private int moveLeadershipInBatch() {
         // Mark leadership movements in progress.
         _executionTaskManager.markTasksInProgress(leadershipMovementTasks);
 
-        // Run preferred leader election.
+        // Run preferred leader election unless there is already an ongoing leadership movement.
```
kidkun: nit: can we make this chunk of code before line 1075. It grantees that we do not do anything until PLE zNode is cleared.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1137,23 +1165,31 @@ private void deleteZnodesToStopExecution() {
...
-            stopExecution(false);
-          }
-        }
+        handleDeadOrAbortingTasks(deadOrAbortingNonLeadershipTasks);
         updateOngoingExecutionState();
       } while (!_executionTaskManager.inExecutionTasks().isEmpty() && finishedTasks.isEmpty());
-      LOG.info("Completed tasks: {}", finishedTasks);
+
+      LOG.info("Completed tasks: {}.{}{}{}", finishedTasks,
+               forceStoppedTaskIds.isEmpty() ? "" : String.format("[Force-stopped: %s]", forceStoppedTaskIds),
```
kidkun: nit: can we also add a new line in `String.format` to the log more readable?

PR #1031, opened 2019-12-03
Initialize openapi integration

Will-Lo:
- Defines the yaml schemas for `PartitionLoad` and `Rebalance` endpoints
- Defines schema structure for endpoints moving forward


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterJsonStructureTest.java
```diff
@@ -0,0 +1,141 @@
...
+    _endpointToClass.put("/kafkacruisecontrol/partition_load", PartitionLoadParameters.class);
+  }
+
+  /**
+   * Load the OpenAPI files for endpoints and compares them against the source code
+   */
+  @Test
+  public void loadOpenApiSpec() throws Exception {
+    OpenAPIV3Parser openApiParser = new OpenAPIV3Parser();
+    String baseFileName = "../schemas/base.yaml";
```
kidkun: nit: can we get rid of relative path here, we could use something like
`System.getProperty("user.dir") + "/src/yaml/base.yaml";` and put it into a static variable.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterJsonStructureTest.java
```diff
@@ -0,0 +1,141 @@
...
+  public Schema parseSchema() {
+    Schema parsedSchema = new Schema();
+    Paths paths = _openAPI.getPaths();
+    for (Map.Entry<String, PathItem> endpoint : paths.entrySet()) {
+      parsedSchema.addEndpoint(endpoint.getKey(), parseEndpoint(endpoint.getValue()));
+    }
+    return parsedSchema;
+  }
+
+  static class Schema {
```
kidkun: Is there any reason that we do not directly use a `Map<String, Set<String>>` here( and define a specific inner static class `Schema`)?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterJsonStructureTest.java
```diff
@@ -0,0 +1,141 @@
...
+    Set<String> parameterSet = new TreeSet<>();
+    if (pathItem.getGet() != null) {
+      parameterList = pathItem.getGet().getParameters();
+    } else if (pathItem.getPost() != null) {
+      parameterList = pathItem.getPost().getParameters();
+    } else {
+      throw new IllegalArgumentException("Schema Parser does not support HTTP methods other than GET/POST");
+    }
+
+    for (Parameter parameter : parameterList) {
```
kidkun: I feel we should not deduplicate here. If in spec, the same parameter are defined twice or (name conflict between two parameters), it is a bug and the test is expected to fail.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterJsonStructureTest.java
```diff
@@ -0,0 +1,141 @@
...
+    OpenAPIV3Parser openApiParser = new OpenAPIV3Parser();
+    String baseFileName = "../schemas/base.yaml";
+    ParseOptions options = new ParseOptions();
+    options.setResolve(true);
+    options.setFlatten(true);
+
+    SwaggerParseResult parseResult = openApiParser.readLocation(baseFileName, null, options);
+    _openAPI = parseResult.getOpenAPI();
+    Schema schema = parseSchema();
+    for (Map.Entry<String, Set<String>> endpoint: schema.getEndpointsAndParameters()) {
```
kidkun: Can we add a to do item here to remind us to check the number of endpoint defined in spec is the same as entries in `_endpointToClass`

kidkun:
also can we move the `yaml` files under `cruise-control/cruise-control/src/yaml`?

schemas/responses/progressResult.yaml
```diff
@@ -0,0 +1,8 @@
+ProgressResult:
```
kidkun: nit: I feel we can merge the `response` with `endpoint`  so that `endpoint.yaml` contains more complete information of each endpoint.  for `progressResult` we can move it to `component` folder since it is shared by multiple endpoints( also it be merged with `stepProgress.yaml`.

schemas/endpoints/rebalance.yaml
```diff
@@ -0,0 +1,169 @@
...
+        in: query
+        description: Whether to balance load between brokers or between disks within broker
+        schema:
+          type: boolean
+          default: false
+      - name: review_id
+        in: query
+        description: Review id for 2FA
+        schema:
+          type: string
```
kidkun: [review_id](https://github.com/linkedin/cruise-control/blob/b31598a5b57f0f9d8123a35e4b4a0173d42416fc/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/RebalanceParameters.java#L74) shoud be an int32 here

schemas/endpoints/partitionLoad.yaml
```diff
@@ -0,0 +1,103 @@
...
+        schema:
+          type: boolean
+    responses:
+      '200':
+        description: |-
+          Successful partition load state response
+        content:
+          application/json:
+            schema:
+              $ref: '../responses/partitionLoadState.yaml#/PartitionLoadState'
```
kidkun: missing plaintext response here

schemas/endpoints/rebalance.yaml
```diff
@@ -0,0 +1,169 @@
...
+      '202':
+        description: |-
+          Rebalancing in progress
+        content:
+          application/json:
+            schema:
+              $ref: '../responses/progressResult.yaml#/ProgressResult'
+          text/plain:
+            schema:
+              type: string
```
kidkun: we are missing[ error response](https://github.com/linkedin/cruise-control/blob/d3619c650aaff1f3750a7d7b0f3ec5ab5f1c4741/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java#L129) here, I think we can add a default response for all error response since they share the same schema.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterJsonStructureTest.java
```diff
@@ -0,0 +1,141 @@
...
+  private Map<String, Class> _endpointToClass;
+  OpenAPI _openAPI;
+  /**
+   * Specify endpoints to be tested
+   */
+  @Before
+  public void setupParameterClasses() {
+    _endpointToClass = new HashMap<>();
+    _endpointToClass.put("/kafkacruisecontrol/rebalance", RebalanceParameters.class);
+    _endpointToClass.put("/kafkacruisecontrol/partition_load", PartitionLoadParameters.class);
```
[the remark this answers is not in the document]
ian952: It seems that I need additional config overrides for the configure call to succeed. I tried the code you suggested but ran into the following error:
```
java.lang.IllegalArgumentException: HttpServletRequest configuration is missing from the request.

	at com.linkedin.kafka.cruisecontrol.servlet.parameters.AbstractParameters.configure(AbstractParameters.java:95)
	at com.linkedin.kafka.cruisecontrol.servlet.parameters.BootstrapParameters.configure(BootstrapParameters.java:76)
	at com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig.getConfiguredInstance(KafkaCruiseControlConfig.java:1838)
	at com.linkedin.kafka.cruisecontrol.servlet.parameters.ParameterJsonStructureTest.setupParameterClasses(ParameterJsonStructureTest.java:45)
```
Do you know how I can create the overrides for the test?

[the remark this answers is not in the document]
kidkun: Ah, I forget `getConfiguredInstance()` will call `object.configure()` after create the object, which will fail here since we are missing required config(`HttpServletRequest` for `CruiseControlParameters` here). 

Looks like the original way  (i.e. `(CruiseControlParameters (_endpointToClass.get(endpoint.getKey()).newInstance())`  at line 54) is the right way to create object.

schemas/responses/optimzationResult.yaml
```diff
@@ -0,0 +1,24 @@
+OptimizationResult:
```
kidkun: There is a typo in this file's name.  `optimzationResult` -> `optimizationResult`

schemas/endpoints/rebalance.yaml
```diff
@@ -0,0 +1,169 @@
...
+        description: Upper bound on the bandwidth used to move replicas
+        schema:
+          type: integer
+          format: int32
+      - name: execution_progress_check_interval_ms
+        in: query
+        description: Execution progress check interval in milliseconds
+        schema:
+          type: integer
+          format: int32
```
kidkun: `int32` -> `int64`

schemas/endpoints/partitionLoad.yaml
```diff
@@ -0,0 +1,103 @@
+PartitionLoadEndpoint:
+  get:
+    summary: Get the resource load for each partition
```
kidkun: to be consistent across all endpoint, can we add a `operationId` here(like in `rebalance.yaml`)?

schemas/components/stepProgress.yaml
```diff
@@ -0,0 +1,18 @@
+StepProgress:
```
kidkun: The boundary between `component` and `response` are kind blur. Maybe we can merge the two directories into one(since all the `yaml`s in these directories are for response). Also if one `component` is specific to another, like here `StepProgress` to `progressResult`, can we put them in one file(to reduce the number of files), just like we put `ClusterModelStatsMetaData` etc. in `ClusterModelStats.yaml`

schemas/responses/progressResult.yaml
```diff
@@ -0,0 +1,8 @@
...
+  type: object
```
kidkun: missing `required` field.

schemas/components/statistic.yaml
```diff
@@ -0,0 +1,12 @@
+Statistic:
+  type: object
+  properties:
+    potentialNwOut:
+      type: number
+      format: double
+    replicas:
+      type: number
```
kidkun: (also apply to the following 2 fields)
type: integer
format: int32

schemas/components/stepProgress.yaml
Will-Lo: I'll put it all in `responses`, thanks!

schemas/endpoints/rebalance.yaml
```diff
@@ -0,0 +1,169 @@
...
+          default: false
+      - name: exclude_recently_demoted_brokers
+        in: query
+        description: Whether to allow leader replicas to be moved to recently demoted broker
+        schema:
+          type: boolean
+          default: false
+      - name: exclude_recently_removed_brokers
+        in: query
+        description: Whether to allow leader replicas to be moved to recently removed broker
```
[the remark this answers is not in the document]
Will-Lo: Will this be also true for `exclude_recently_demoted_brokers`?

schemas/base.yaml
```diff
@@ -0,0 +1,10 @@
+openapi: 3.0.0
+info:
+  title: Cruise Control API
+  description: Cruise-control is the first of its kind to fully automate the dynamic workload rebalance and self-healing of a kafka cluster. It provides great value to Kafka users by simplifying the operation of Kafka clusters.
+  version: 19.10.0
```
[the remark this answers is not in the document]
Will-Lo: Ah we just chose it based on the date, 2019 oct, but I guess we can just make it 1.0.0

cruise-control/src/yaml/endpoints/partitionLoad.yaml
```diff
@@ -0,0 +1,150 @@
...
+        description: |-
+          Partition load state fetch in progress
+        content:
+          application/json:
+            schema:
+              $ref: '../responses/progressResult.yaml#/ProgressResult'
+          text/plain:
+            schema:
+              type: string
+      '400':
```
kidkun: (Also apply to `rebalance.yaml`)the error response for `400`,`403` and `500` share the same format, can we just define once using the [default response syntax](https://swagger.io/docs/specification/describing-responses/)?

cruise-control/src/yaml/endpoints/rebalance.yaml
```diff
@@ -0,0 +1,200 @@
...
+          default: VALID_WINDOWS
+      - name: verbose
+        in: query
+        description: Return detailed state information
+        schema:
+          type: boolean
+          default: false
+      - name: exclude_recently_demoted_brokers
+        in: query
+        description: Whether to allow replicas to be moved to recently demoted broker
```
kidkun: `replicas` -> `leader replicas`, broker demotion is to transfer all the leadership out of broker, so only leader replica matters here.

cruise-control/src/yaml/endpoints/partitionLoad.yaml
Will-Lo: Ah didn't realize that was a thing, that's much better

Issue #1034, opened 2019-12-06
Add support to modify ongoing executions


efeg:
This provides flexibility so that current executions can be paused and reissued with modifications -- e.g. adding additional brokers to an ongoing `remove_broker` request.

Current workflow is:
1. user sends a request to stop execution,
2. user waits for the ongoing execution to stop.
3. user reissues a new request with modification.

The proposed modification should also ensure that it keeps track of all modifications to ongoing tasks (i.e. for auditability).

PR #1036, opened 2019-12-07
Add support to modify ongoing executions

efeg:
Addresses https://github.com/linkedin/cruise-control/issues/1034.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
@@ -8,6 +8,8 @@ import com.linkedin.kafka.cruisecontrol.KafkaCruiseControl;
 import com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils;
 import com.linkedin.kafka.cruisecontrol.analyzer.kafkaassigner.KafkaAssignerDiskUsageDistributionGoal;
 import com.linkedin.kafka.cruisecontrol.analyzer.kafkaassigner.KafkaAssignerEvenRackAwareGoal;
+import com.linkedin.kafka.cruisecontrol.async.progress.OperationProgress;
+import com.linkedin.kafka.cruisecontrol.async.progress.WaitingForOngoingExecutionToStop;
 import com.linkedin.kafka.cruisecontrol.executor.strategy.ReplicaMovementStrategy;
 import com.linkedin.kafka.cruisecontrol.model.Broker;
 import com.linkedin.kafka.cruisecontrol.model.ClusterModel;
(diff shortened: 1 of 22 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
```diff
@@ -203,6 +209,28 @@ public static void sanityCheckBrokersHavingOfflineReplicasOnBadDisks(List<String
...
+   *
+   * @param kafkaCruiseControl The Kafka Cruise Control instance.
+   */
+  public static void maybeStopOngoingExecutionToModifyAndWait(KafkaCruiseControl kafkaCruiseControl) {
+    if (kafkaCruiseControl.hasOngoingExecution() && kafkaCruiseControl.modifyOngoingExecution(true)) {
+      LOG.info("Gracefully stopping the ongoing execution... Use {} endpoint with {}=true to force-stop it now.",
+               STOP_PROPOSAL_EXECUTION, FORCE_STOP_PARAM);
+      while (kafkaCruiseControl.executionState() != NO_TASK_IN_PROGRESS) {
+        // Stop ongoing execution, and wait for the execution to stop.
+        try {
```
kidkun: The wait here could be a very long time(or even forever since we do not force stop here), maybe we should assign a [`OperationStep`](https://github.com/linkedin/cruise-control/blob/472173237d96dbe7e47bbd16672ee45472706921/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/async/progress/OperationStep.java) to it, so user know that Cruise Control is current wait for ongoing task to finish.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
```diff
@@ -203,6 +209,28 @@ public static void sanityCheckBrokersHavingOfflineReplicasOnBadDisks(List<String
     }
   }
 
+  /**
+   * Gracefully stop the ongoing execution (if any) and wait until the execution stops.
+   *
+   * @param kafkaCruiseControl The Kafka Cruise Control instance.
+   */
+  public static void maybeStopOngoingExecutionToModifyAndWait(KafkaCruiseControl kafkaCruiseControl) {
+    if (kafkaCruiseControl.hasOngoingExecution() && kafkaCruiseControl.modifyOngoingExecution(true)) {
```
kidkun: Looks like there is race condition here. 
If there are two requests try to modify the same ongoing execution.
The first one reaches this method first and wait on ongoing tasks to stop, the second come and `if` will false and it proceeds, then in the end one of the two request may fail with error saying something like "there is ongoing execution"?

efeg: Indeed in this case one of the executions will fail with `there is ongoing execution` exception; hence, it will guarantee safety. However, it will be wasteful due to creating a new cluster model, and that can be avoided.

Updated the logic and added JavaDoc to explain acceptable / potential scenarios.

PR #1035, opened 2019-12-07
Add reason parameter in python client for multiple POST endpoints.

kidkun:
For #1007

PR #1038, opened 2019-12-09
[ccclient-1.0.3] Implement reasons for requests

### LINKED ISSUE #1007: Add support to cruise-control-client to let use specify reason for requests
As a follow up of #977 , we need to expose the newly-added parameter `reason` out of the python client.

mgrubent:
Fixes #1007 .

This also cleans up the previous implementation of reason for a review, which would display help text as if it were a reason for a pause / resume of sampling.


cruise-control-client/cruisecontrolclient/client/CCParameter/ReasonParameter.py
```diff
@@ -0,0 +1,51 @@
+from cruisecontrolclient.client.CCParameter import AbstractParameter
+
+
+class AbstractReasonParameter(AbstractParameter):
+    def validate_value(self):
+        """
+        The validation strategy here is to avoid a Bobby Tables situation
```
efeg: `Nit`: Missing link https://xkcd.com/327/ :))

cruise-control-client/setup.py
```diff
@@ -11,7 +11,7 @@
 
 setuptools.setup(
     name='cruise-control-client',
-    version='1.0.2',
+    version='1.0.3',
```
efeg: Nit: Is it possible to add a documentation on when to bump up client version?

cruise-control-client/cruisecontrolclient/client/CCParameter/ReasonParameter.py
mgrubent: Agreed; 196aec9

cruise-control-client/setup.py
mgrubent: > Thanks for the patch -- @mgrubent I see some naming conflicts between this patch and #1035. Should this patch be prioritized over the other one?

Let's merge this one first then I'll work with @kidkun to merge in #1035

mgrubent:
> Fixes #1007 .
> 
> This also cleans up the previous implementation of reason for a review, which would display help text as if it were a reason for a pause / resume of sampling.

Let's merge this one first then I'll work with @kidkun to merge in #1035

mgrubent:
> Thanks for the patch -- @mgrubent I see some naming conflicts between this patch and #1035. Should this patch be prioritized over the other one?

Let's merge this one first then I'll work with @kidkun to merge in #1035

PR #1058, opened 2020-01-03
Add OpenAPI yaml spec for Load endpoint

annaleeli:
Add OpenAPI Load endpoint for only cluster load as partition load is accessed via a different path and capacity is not yet merged.

### DIFF
--- a/cruise-control/src/yaml/endpoints/load.yaml
+++ b/cruise-control/src/yaml/endpoints/load.yaml
@@ -1,0 +1,64 @@
+LoadEndpoint:
+  get:
+    operationId: load
+    summary: Get the cluster load once Cruise Control Load Monitor is running.
+    parameters:
+      - name: start
+        in: query
+        description: Start time of the cluster load.
+        schema:
+          type: long
+          format: int64
+          # Default is time of earliest valid window.
+      - name: end
+        in: query
+        description: End time of the cluster load.
+        schema:
+          type: long
+          format: int64
+          # Default is current system time.
+      - name: time
+        in: query
+        description: End time of the clutser load.
+        schema:
+          type: long
+          format: int64
+        # Default is current system time, mutually exclusive from end parameter.
+      - name: allow_capacity_estimation
+        in: query
+        description: Whether to allow capacity estimation when cruise-control is unable to obtain all per-broker capacity information.
+        schema:
+          type: boolean
+          default: true
+      - name: json
+        in: query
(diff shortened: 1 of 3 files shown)


cruise-control/src/yaml/endpoints/load.yaml
```diff
@@ -0,0 +1,64 @@
...
+  get:
+    operationId: load
+    summary: Get the cluster load once Cruise Control Load Monitor is running.
+    parameters:
+      - name: start
+        in: query
+        description: Start time of the cluster load.
+        schema:
+          type: long
+          format: int64
```
kidkun: can we add a `minimum` and `default` field for this parameter? (see [partitionLoad.yaml](https://github.com/linkedin/cruise-control/blob/b15fbffd04b8bd4fa561e13a3bc247f02fc794dc/cruise-control/src/yaml/endpoints/partitionLoad.yaml))

cruise-control/src/yaml/endpoints/load.yaml
```diff
@@ -0,0 +1,64 @@
...
+          # Default is time of earliest valid window.
+      - name: end
+        in: query
+        description: End time of the cluster load.
+        schema:
+          type: long
+          format: int64
```
kidkun: can we add a `minimum` field for this parameter? (see [partitionLoad.yaml](https://github.com/linkedin/cruise-control/blob/b15fbffd04b8bd4fa561e13a3bc247f02fc794dc/cruise-control/src/yaml/endpoints/partitionLoad.yaml))

cruise-control/src/yaml/endpoints/load.yaml
```diff
@@ -0,0 +1,64 @@
...
+          # Default is current system time.
+      - name: time
+        in: query
+        description: End time of the clutser load.
+        schema:
+          type: long
+          format: int64
```
kidkun: can we add a `minimum` and `default` field for this parameter?

annaleeli: I added `minimum` for this parameter as per the comment. However, `default` for this field is the current time stamp at time of execution (which is difficult to express as a constant value). I omitted adding the field in the yaml.

Let me know if you have a strong preference for `default` here and I will investigate a way to add it in.

annaleeli:
@kidkun Thank you for the review! I addressed the comments you have left on the PR.

Issue #986, opened 2019-11-07
Dead brokers should not block executions of due to being unable retrieve their capacity


efeg:
Ensure that dead brokers do not block execution in case `allow_capacity_estimation` is `false` and the capacity of a dead broker(s) cannot be retrieved.

PR #1066, opened 2020-01-11
Tolerate broker capacity resolver unable to retrieve capacity information of dead broker.

kidkun:
Addresses the issue #986


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -523,10 +523,11 @@ public ClusterModel clusterModel(long from,
       }
 
       // Populate snapshots for the cluster model.
+      Set<Integer> deadBrokers = MonitorUtils.deadBrokersWithReplicas(cluster);
```
efeg: `MonitorUtils#deadBrokersWithReplicas(Cluster)` excludes dead brokers with 0 replicas. Shouldn't we include all dead brokers?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java
```diff
@@ -436,24 +437,33 @@ static void setBadBrokerState(ClusterModel clusterModel, Cluster cluster) {
...
-        // is dead. We need this to get the host resource capacity.
-        BrokerCapacityInfo brokerCapacity =
-            brokerCapacityConfigResolver.capacityForBroker(rack, replica.host(), replica.id());
+        BrokerCapacityInfo brokerCapacity;
+        try {
+          brokerCapacity = brokerCapacityConfigResolver.capacityForBroker(rack, replica.host(), replica.id());
+        } catch (Exception e) {
+          //Capacity resolver may not be able to return the capacity information of dead brokers.
+          if (deadBrokers.contains(replica.id())) {
+            brokerCapacity = new BrokerCapacityInfo(Collections.emptyMap());
```
efeg: Passing an empty map to `BrokerCapacityInfo` means that the broker has no resources -- I suspect that this will cause issues in the interpreting logic, which expects a value for each resource type (e.g. populating partition load on such brokers -- also see `Broker#_brokerCapacity`).

1. Can we add a sanity check in `BrokerCapacityInfo` constructors to ensure that any capacity sent to this broker has exactly all resource types in the keyset of map?
2. Can we add a unit test to verify our expected behavior and a cluster model can be generated properly under the tested scenario?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java
```diff
@@ -436,24 +437,33 @@ static void setBadBrokerState(ClusterModel clusterModel, Cluster cluster) {
...
         Node replica = partitionInfo.replicas()[index];
         String rack = getRackHandleNull(replica);
-        // Note that we assume the capacity resolver can still return the broker capacity even if the broker
-        // is dead. We need this to get the host resource capacity.
-        BrokerCapacityInfo brokerCapacity =
-            brokerCapacityConfigResolver.capacityForBroker(rack, replica.host(), replica.id());
+        BrokerCapacityInfo brokerCapacity;
+        try {
+          brokerCapacity = brokerCapacityConfigResolver.capacityForBroker(rack, replica.host(), replica.id());
+        } catch (Exception e) {
```
efeg: Instead of catching all exceptions, should we catch a `TimeoutException` and in the logic that implements `capacityForBroker`, ensure that a timeout exception is thrown in the desired scenario?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -473,7 +474,7 @@ public ClusterModel clusterModel(long from,
                                    ModelCompletenessRequirements requirements,
                                    boolean populateReplicaPlacementInfo,
                                    OperationProgress operationProgress)
-      throws NotEnoughValidWindowsException {
+      throws NotEnoughValidWindowsException, TimeoutException {
```
efeg: If we intend to propagate `TimeoutException` all the way to higher layers, can we update the JavaDoc as we did for `NotEnoughValidWindowsException` for consistency? Alternatively, would it be cleaner to return `null` if the cluster model creation has failed due to not being able to resolve the cluster capacity, and handle the null cluster capacity response?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java
```diff
@@ -442,18 +450,26 @@ static void populatePartitionLoad(Cluster cluster,
...
         String rack = getRackHandleNull(replica);
-        // Note that we assume the capacity resolver can still return the broker capacity even if the broker
-        // is dead. We need this to get the host resource capacity.
-        BrokerCapacityInfo brokerCapacity =
-            brokerCapacityConfigResolver.capacityForBroker(rack, replica.host(), replica.id());
+        BrokerCapacityInfo brokerCapacity;
+        try {
+          brokerCapacity = brokerCapacityConfigResolver.capacityForBroker(rack, replica.host(), replica.id());
+        } catch (TimeoutException tme) {
+          //Capacity resolver may not be able to return the capacity information of dead brokers.
```
efeg: Nit: Missing space before `Capacity`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java
```diff
@@ -442,18 +450,26 @@ static void populatePartitionLoad(Cluster cluster,
...
+          if (!aliveBrokers.contains(replica.id())) {
+            brokerCapacity = new BrokerCapacityInfo(EMPTY_BROKER_CAPACITY);
```
efeg: Should we (bulk) log this empty broker capacity assignment?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java
```diff
@@ -442,18 +450,26 @@ static void populatePartitionLoad(Cluster cluster,
                                     TopicPartition tp,
                                     ValuesAndExtrapolations valuesAndExtrapolations,
                                     Map<TopicPartition, Map<Integer, String>> replicaPlacementInfo,
-                                    BrokerCapacityConfigResolver brokerCapacityConfigResolver) {
+                                    BrokerCapacityConfigResolver brokerCapacityConfigResolver) throws TimeoutException {
     PartitionInfo partitionInfo = cluster.partition(tp);
+    Set<Integer> aliveBrokers = cluster.nodes().stream().mapToInt(Node::id).boxed().collect(Collectors.toSet());
```
efeg: Should we move this inside `if (partitionInfo != null)`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
```diff
@@ -76,9 +77,14 @@ private void updateCachedNumCoresByBroker(Cluster cluster) {
...
         }
-        BrokerCapacityInfo capacity = _brokerCapacityConfigResolver.capacityForBroker(getRackHandleNull(node), node.host(), bid);
-        // No mapping shall be recorded if capacity is estimated, but estimation is not allowed.
-        return (!_allowCpuCapacityEstimation && capacity.isEstimated()) ? null : capacity.numCpuCores();
+        try {
+          BrokerCapacityInfo capacity = _brokerCapacityConfigResolver.capacityForBroker(getRackHandleNull(node), node.host(), bid);
+          // No mapping shall be recorded if capacity is estimated, but estimation is not allowed.
+          return (!_allowCpuCapacityEstimation && capacity.isEstimated()) ? null : capacity.numCpuCores();
+        } catch (TimeoutException tme) {
+          LOG.warn("Unable to get core number of broker {}.", node.id());
```
efeg: Nit: `core number` -> `number of CPU cores`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
```diff
@@ -76,9 +77,14 @@ private void updateCachedNumCoresByBroker(Cluster cluster) {
...
+          return null;
```
efeg: Can we update the JavaDoc to indicate cases that a null response is expected? Can we add a unit test to verify the return of expected `null` response?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
kidkun: I prefer the first approach. Because if the cluster model is unable to generate, there is nothing CC should do other than notify the user about the error.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java
```diff
@@ -442,18 +451,28 @@ static void populatePartitionLoad(Cluster cluster,
...
                                     ValuesAndExtrapolations valuesAndExtrapolations,
                                     Map<TopicPartition, Map<Integer, String>> replicaPlacementInfo,
-                                    BrokerCapacityConfigResolver brokerCapacityConfigResolver) {
+                                    BrokerCapacityConfigResolver brokerCapacityConfigResolver) throws TimeoutException {
     PartitionInfo partitionInfo = cluster.partition(tp);
     // If partition info does not exist, the topic may have been deleted.
     if (partitionInfo != null) {
+      Set<Integer> aliveBrokers = cluster.nodes().stream().mapToInt(Node::id).boxed().collect(Collectors.toSet());
       boolean needToAdjustCpuUsage = true;
+      Set<Integer> deadBrokers = new HashSet<>();
```
efeg: Can we rename this to `deadBrokersWithUnknownCapacity`? -- i.e. it is possible that a broker can be dead, but its capacity can still be resolved depending on whether the capacity resolving service can provide the capacity or not.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java
```diff
@@ -479,6 +498,9 @@ static void populatePartitionLoad(Cluster cluster,
                                     valuesAndExtrapolations.windows());
         needToAdjustCpuUsage = false;
       }
+      if (!deadBrokers.isEmpty()) {
+        LOG.info("Assign empty capacity to dead brokers {} in cluster model.", deadBrokers);
```
efeg: Can we update the message to make it clear that these are not necessarily all the dead brokers, but only the dead brokers, whose capacity cannot be obtained from the capacity resolver?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/CruiseControlMetricsProcessor.java
```diff
@@ -76,9 +80,15 @@ private void updateCachedNumCoresByBroker(Cluster cluster) {
...
-        BrokerCapacityInfo capacity = _brokerCapacityConfigResolver.capacityForBroker(getRackHandleNull(node), node.host(), bid);
-        // No mapping shall be recorded if capacity is estimated, but estimation is not allowed.
-        return (!_allowCpuCapacityEstimation && capacity.isEstimated()) ? null : capacity.numCpuCores();
+        try {
+          BrokerCapacityInfo capacity =
+              _brokerCapacityConfigResolver.capacityForBroker(getRackHandleNull(node), node.host(), bid, BROKER_CAPACITY_FETCH_TIMEOUT_MS);
+          // No mapping shall be recorded if capacity is estimated, but estimation is not allowed.
+          return (!_allowCpuCapacityEstimation && capacity.isEstimated()) ? null : capacity.numCpuCores();
+        } catch (TimeoutException tme) {
+          LOG.warn("Unable to get number of CPU cores for broker {}.", node.id());
```
efeg: Should we add `, tme` to the log?

Issue #1081, opened 2020-01-29
CruiseControlReport should check and update metric topic's configs during bootstrap


kidkun:
Currently reporter is able to automatically create metric if it does not exist. 
If the topic already exists, reporter does not check the topic config(partition count, replication factor) is consistent with the its own config.  Therefore it the topic is misconfiged in the first place, we are unlikely to discover and leads to extraordinarily some large replicas in the cluster and make the cluster hard to balance.

PR #1082, opened 2020-01-30
 Make CruiseControlReporter check metrics topic's partition count during bootstrap

kidkun:
Addresses the issue #1081 .

### DIFF
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -191,14 +191,14 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
       DescribeConfigsResult describeConfigsResult = _adminClient.describeConfigs(Collections.singleton(topicResource));
       Config topicConfig = describeConfigsResult.values().get(topicResource).get();
       Set<AlterConfigOp> configsToBeSet = new HashSet<>(2);
-      if(topicConfig.get(LogConfig.RetentionMsProp()) == null ||
-         !topicConfig.get(LogConfig.RetentionMsProp()).value().equals(_newTopic.configs().get(LogConfig.RetentionMsProp()))) {
+      if (topicConfig.get(LogConfig.RetentionMsProp()) == null ||
+          !topicConfig.get(LogConfig.RetentionMsProp()).value().equals(_newTopic.configs().get(LogConfig.RetentionMsProp()))) {
         configsToBeSet.add(new AlterConfigOp(new ConfigEntry(LogConfig.RetentionMsProp(),
                                                              _newTopic.configs().get(LogConfig.RetentionMsProp())),
                                              AlterConfigOp.OpType.SET));
       }
-      if(topicConfig.get(LogConfig.CleanupPolicyProp()) == null ||
-         !topicConfig.get(LogConfig.CleanupPolicyProp()).value().equals(_newTopic.configs().get(LogConfig.CleanupPolicyProp()))) {
+      if (topicConfig.get(LogConfig.CleanupPolicyProp()) == null ||
+          topicConfig.get(LogConfig.CleanupPolicyProp()).value().equals(_newTopic.configs().get(LogConfig.CleanupPolicyProp()))) {
         configsToBeSet.add(new AlterConfigOp(new ConfigEntry(LogConfig.CleanupPolicyProp(),
                                                              _newTopic.configs().get(LogConfig.CleanupPolicyProp())),
                                              AlterConfigOp.OpType.SET));
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -14,16 +14,25 @@ import com.yammer.metrics.Metrics;
 import com.yammer.metrics.core.Metric;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import kafka.log.LogConfig;
 import kafka.server.KafkaConfig;
 import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.admin.AlterConfigOp;
+import org.apache.kafka.clients.admin.AlterConfigsResult;
+import org.apache.kafka.clients.admin.Config;
+import org.apache.kafka.clients.admin.ConfigEntry;
 import org.apache.kafka.clients.admin.CreateTopicsResult;
+import org.apache.kafka.clients.admin.DescribeConfigsResult;
 import org.apache.kafka.clients.admin.NewPartitions;
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.admin.TopicDescription;
@@ -33,6 +42,7 @@ import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.config.ConfigResource;
 import org.apache.kafka.common.errors.InterruptException;
 import org.apache.kafka.common.metrics.KafkaMetric;
 import org.apache.kafka.common.metrics.MetricsReporter;
@@ -140,7 +150,6 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
         LOG.warn("Cruise Control metrics topic auto creation was disabled", e);
       }
     }
-
   }
 
   protected NewTopic createNewTopicFromReporterConfig(CruiseControlMetricsReporterConfig reporterConfig)
@@ -151,12 +160,18 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
         reporterConfig.getInt(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_NUM_PARTITIONS_CONFIG);
     Short cruiseControlMetricsTopicReplicaFactor =
         reporterConfig.getShort(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR_CONFIG);
-
     if (cruiseControlMetricsTopicReplicaFactor <= 0 || cruiseControlMetricsTopicNumPartition <= 0) {
       throw new CruiseControlMetricsReporterException("The topic configuration must explicitly set the replication factor and the num partitions");
     }
+    NewTopic newTopic = new NewTopic(cruiseControlMetricsTopic, cruiseControlMetricsTopicNumPartition, cruiseControlMetricsTopicReplicaFactor);
 
-    return new NewTopic(cruiseControlMetricsTopic, cruiseControlMetricsTopicNumPartition, cruiseControlMetricsTopicReplicaFactor);
+    Map<String, String> config = new HashMap<>(2);
+    config.put(LogConfig.RetentionMsProp(),
+               Long.toString(reporterConfig.getLong(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_RETENTION_MS_CONFIG)));
+    config.put(LogConfig.CleanupPolicyProp(),
+               reporterConfig.getString(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_CLEAN_UP_POLICY_CONFIG));
+    newTopic.configs(config);
+    return newTopic;
   }
 
   protected void createCruiseControlMetricsTopic() {
@@ -171,6 +186,33 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
 
   protected void maybeUpdateCruiseControlMetricsTopic() {
     try {
+      // Retrieve topic config to check and update.
+      ConfigResource topicResource = new ConfigResource(ConfigResource.Type.TOPIC, _cruiseControlMetricsTopic);
+      DescribeConfigsResult describeConfigsResult = _adminClient.describeConfigs(Collections.singleton(topicResource));
+      Config topicConfig = describeConfigsResult.values().get(topicResource).get();
+      Set<AlterConfigOp> configsToBeSet = new HashSet<>(2);
+      if(topicConfig.get(LogConfig.RetentionMsProp()) == null ||
+         !topicConfig.get(LogConfig.RetentionMsProp()).value().equals(_newTopic.configs().get(LogConfig.RetentionMsProp()))) {
+        configsToBeSet.add(new AlterConfigOp(new ConfigEntry(LogConfig.RetentionMsProp(),
+                                                             _newTopic.configs().get(LogConfig.RetentionMsProp())),
+                                             AlterConfigOp.OpType.SET));
+      }
+      if(topicConfig.get(LogConfig.CleanupPolicyProp()) == null ||
+         !topicConfig.get(LogConfig.CleanupPolicyProp()).value().equals(_newTopic.configs().get(LogConfig.CleanupPolicyProp()))) {
+        configsToBeSet.add(new AlterConfigOp(new ConfigEntry(LogConfig.CleanupPolicyProp(),
+                                                             _newTopic.configs().get(LogConfig.CleanupPolicyProp())),
+                                             AlterConfigOp.OpType.SET));
+      }
+      if (!configsToBeSet.isEmpty()) {
+        AlterConfigsResult alterConfigsResult = _adminClient.incrementalAlterConfigs(Collections.singletonMap(topicResource, configsToBeSet));
+        alterConfigsResult.values().get(topicResource).get();
+      }
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.warn("Unable to update config of Cruise Cruise Control metrics topic {}", _cruiseControlMetricsTopic, e);
+    }
+
+    try {
+      // Retrieve topic partition count to check and update.
       TopicDescription topicDescription =
           _adminClient.describeTopics(Collections.singletonList(_cruiseControlMetricsTopic)).values()
                       .get(_cruiseControlMetricsTopic).get();
@@ -179,8 +221,8 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
                                                                NewPartitions.increaseTo(_newTopic.numPartitions())));
       }
     } catch (InterruptedException | ExecutionException e) {
-      LOG.warn(String.format("Unable to increase Cruise Cruise Control metrics topic %s's partition number to %d",
-                             _cruiseControlMetricsTopic, _newTopic.replicationFactor()), e);
+      LOG.warn("Unable to increase Cruise Cruise Control metrics topic {} partition number to {}",
+               _cruiseControlMetricsTopic, _newTopic.replicationFactor(), e);
       }
     }
 
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -30,7 +30,6 @@ import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AlterConfigOp;
 import org.apache.kafka.clients.admin.AlterConfigsResult;
 import org.apache.kafka.clients.admin.Config;
-import org.apache.kafka.clients.admin.ConfigEntry;
 import org.apache.kafka.clients.admin.CreateTopicsResult;
 import org.apache.kafka.clients.admin.DescribeConfigsResult;
 import org.apache.kafka.clients.admin.NewPartitions;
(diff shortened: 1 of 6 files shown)


cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -162,20 +165,48 @@ private void createCruiseControlMetricsTopic() {
       createTopicsResult.values().get(_newTopic.name()).get();
       LOG.info("Cruise Control metrics topic created: {}", _cruiseControlMetricsTopic);
     } catch (InterruptedException | ExecutionException e) {
-      if (!(e.getCause() instanceof TopicExistsException)) {
-        LOG.error("Unable to create Cruise Control topic", e);
-      }
+      LOG.warn("Unable to create Cruise Control metrics topic " + _cruiseControlMetricsTopic, e);
```
efeg: Nit: Should we use lazy logging to avoid string concat? -- i.e.
`LOG.warn("Unable to create Cruise Control metrics topic {}.", _cruiseControlMetricsTopic, e);`

efeg: We have some common logic regarding (1) creating non-existing topic and (2) modifying the selected configs of existing topics (i.e. here and in sample store topic creation).
Can we make the relevant logic (1) shared and (2) pluggable between the two? The first part will ensure a common code path for similar execution. The second part will enable users to add their custom logic that they may want to automate during topic creation (e.g. generating ACLs).

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -162,20 +165,48 @@ private void createCruiseControlMetricsTopic() {
...
+        }
+        if (topicExists) {
+          maybeUpdateCruiseControlMetricsTopic();
+        } else {
+          createCruiseControlMetricsTopic();
+        }
+      } catch (InterruptedException | ExecutionException e) {
+        LOG.warn("Skip checking Cruise Control metrics topic " + _newTopic.name(), e);
+      } finally {
+        CruiseControlMetricsUtils.closeAdminClientWithTimeout(_adminClient, 5000);
```
efeg: Nit: Can we avoid hard coding `5000`?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
kidkun: I was thinking about this.  
One concern prevent me doing that is where should I put the common code. Because `cruise-control` and `cruise-control-metrics-reporter` are two separate module and we do not want one to depends on the other. And it also seems inappropriate to put this code in `cruise-control-core` module since it only contains non-kafka-specific code.

efeg: Sure, I agree with your concerns regarding sharing the logic.
Are there any issues with making each logic pluggable within their corresponding modules?

kidkun: While I was taking a deeper look, I just realize that `CruiseControlMetricsReporter` and `KafkaSampleStore` themselves are `Configurable` class, then to make the topic management logic pluggable, we can simply change corresponding methods to `protected`.

For example, if someone want to add some more check for metrics topic, (s)he can directly extends from `CruiseControlMetricsReporter` and override `createNewTopicFromReporterConfig()`, `createCruiseControlMetricsTopic()` and `maybeUpdateCruiseControlMetricsTopic()` methods.

and then change the config `metric.reporters` to the extended class name.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -156,26 +159,54 @@ private NewTopic createNewTopicFromReporterConfig(CruiseControlMetricsReporterCo
...
+      LOG.info("Cruise Control metrics topic {} is created.", _cruiseControlMetricsTopic);
     } catch (InterruptedException | ExecutionException e) {
-      if (!(e.getCause() instanceof TopicExistsException)) {
-        LOG.error("Unable to create Cruise Control topic", e);
-      }
+      LOG.warn("Unable to create Cruise Control metrics topic {}.", _cruiseControlMetricsTopic, e);
     }
   }
 
+  protected void maybeUpdateCruiseControlMetricsTopic() {
```
efeg: (Applies to topic creation code path) Missing modification regarding cleanup policy and retention (please see `KafkaSampleStore#ensureTopicCreated`).

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -137,45 +150,105 @@ public void configure(Map<String, ?> configs) {
...
+      Config topicConfig = describeConfigsResult.values().get(topicResource).get();
+      Set<AlterConfigOp> configsToBeSet = new HashSet<>(2);
+      if (topicConfig.get(LogConfig.RetentionMsProp()) == null ||
+          !topicConfig.get(LogConfig.RetentionMsProp()).value().equals(_newTopic.configs().get(LogConfig.RetentionMsProp()))) {
+        configsToBeSet.add(new AlterConfigOp(new ConfigEntry(LogConfig.RetentionMsProp(),
+                                                             _newTopic.configs().get(LogConfig.RetentionMsProp())),
+                                             AlterConfigOp.OpType.SET));
+      }
+      if (topicConfig.get(LogConfig.CleanupPolicyProp()) == null ||
+          topicConfig.get(LogConfig.CleanupPolicyProp()).value().equals(_newTopic.configs().get(LogConfig.CleanupPolicyProp()))) {
```
efeg: Should it be `equals(...)` or `!equals(...)`?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -137,45 +150,105 @@ public void configure(Map<String, ?> configs) {
...
   }
 
+  protected void maybeUpdateCruiseControlMetricsTopic() {
+    try {
+      // Retrieve topic config to check and update.
+      ConfigResource topicResource = new ConfigResource(ConfigResource.Type.TOPIC, _cruiseControlMetricsTopic);
+      DescribeConfigsResult describeConfigsResult = _adminClient.describeConfigs(Collections.singleton(topicResource));
+      Config topicConfig = describeConfigsResult.values().get(topicResource).get();
+      Set<AlterConfigOp> configsToBeSet = new HashSet<>(2);
+      if (topicConfig.get(LogConfig.RetentionMsProp()) == null ||
```
efeg: This and the following if statements seem to be good candidates to move to a util function, which adds a config to the given set if it satisfies certain criteria.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -137,45 +150,105 @@ public void configure(Map<String, ?> configs) {
...
+      }
+      if (!configsToBeSet.isEmpty()) {
+        AlterConfigsResult alterConfigsResult = _adminClient.incrementalAlterConfigs(Collections.singletonMap(topicResource, configsToBeSet));
+        alterConfigsResult.values().get(topicResource).get();
+      }
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.warn("Unable to update config of Cruise Cruise Control metrics topic {}", _cruiseControlMetricsTopic, e);
+    }
+
+    try {
```
efeg: This and the preceding `try-catch` blocks seem to perform independent operations -- would it possible to move them to their own functions?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterConfig.java
```diff
@@ -48,6 +52,8 @@
   private static final Integer DEFAULT_CRUISE_CONTROL_METRICS_TOPIC_NUM_PARTITIONS = -1;
   private static final boolean DEFAULT_CRUISE_CONTROL_METRICS_TOPIC_AUTO_CREATE = false;
   private static final Short DEFAULT_CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR = -1;
+  private static final String DEFAULT_CRUISE_CONTROL_METRICS_TOPIC_CLEAN_UP_POLICY = "delete";
```
efeg: I don't think this should be configurable -- i.e. we do not support any other configuration, let's avoid exposing it.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterConfig.java
```diff
@@ -31,6 +31,8 @@
   private static final String CRUISE_CONTROL_METRICS_TOPIC_NUM_PARTITIONS_DOC = "The number of partitions of Cruise Control metrics topic";
   public static final String CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR_CONFIG = "cruise.control.metrics.topic.replication.factor";
   private static final String CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR_DOC = "The replication factor of Cruise Control metrics topic";
+  public static final String CRUISE_CONTROL_METRICS_TOPIC_RETENTION_MS_CONFIG = "cruise.control.metrics.topic.retention.ms";
+  private static final String CRUISE_CONTROL_METRICS_TOPIC_RETENTION_MS_DOC = "The retention time of Cruise Control metrics topic";
```
efeg: Nit-1: The description is missing the unit of the retention time.
Nit-2: Missing dot at the end.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -137,45 +151,116 @@ public void configure(Map<String, ?> configs) {
...
+        alterConfigsResult.values().get(topicResource).get();
+      }
     } catch (InterruptedException | ExecutionException e) {
-      if (!(e.getCause() instanceof TopicExistsException)) {
-        LOG.error("Unable to create Cruise Control topic", e);
+      LOG.warn("Unable to update config of Cruise Cruise Control metrics topic {}", _cruiseControlMetricsTopic, e);
+    }
+  }
+
+  protected void maybeUpdateConfig(Set<AlterConfigOp> configsToAlter,
```
efeg: Can we (1) make it `static`? ... -> `protected static void maybeUpdateConfig` and (2) move it to `CruiseControlMetricsUtils` with a JavaDoc?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -137,45 +151,116 @@ public void configure(Map<String, ?> configs) {
...
       final CreateTopicsResult createTopicsResult = _adminClient.createTopics(Collections.singletonList(_newTopic));
       createTopicsResult.values().get(_newTopic.name()).get();
-      LOG.info("Cruise Control metrics topic created: {}", _cruiseControlMetricsTopic);
+      LOG.info("Cruise Control metrics topic {} is created.", _cruiseControlMetricsTopic);
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.warn("Unable to create Cruise Control metrics topic {}.", _cruiseControlMetricsTopic, e);
+    }
+  }
+
+  protected void maybeUpdateCruiseControlMetricsTopic() {
```
efeg: Can we add a unit test to verify that this function works as expected?

kidkun: I find writing a unit test for this method is pretty hard given it is a broker dependency, I would like to verify this in deployment.

efeg: Can we use `CCKafkaClientsIntegrationTestHarness` to handle broker dependency?

kidkun: Added test.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterConfig.java
```diff
@@ -31,6 +31,8 @@
   private static final String CRUISE_CONTROL_METRICS_TOPIC_NUM_PARTITIONS_DOC = "The number of partitions of Cruise Control metrics topic";
   public static final String CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR_CONFIG = "cruise.control.metrics.topic.replication.factor";
   private static final String CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR_DOC = "The replication factor of Cruise Control metrics topic";
+  public static final String CRUISE_CONTROL_METRICS_TOPIC_RETENTION_MS_CONFIG = "cruise.control.metrics.topic.retention.ms";
+  private static final String CRUISE_CONTROL_METRICS_TOPIC_RETENTION_MS_DOC = "The retention time in milliseconds of Cruise Control metrics topic.";
```
efeg: Nit: Would it be possible to make this and the other `DEFAULT_*` `static final` variables public? This allows users to access these defaults in other pluggable implementations of metrics reporter in case of need.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -137,45 +153,107 @@ public void configure(Map<String, ?> configs) {
...
   @Override
   public void run() {
     LOG.info("Starting Cruise Control metrics reporter with reporting interval of {} ms.", _reportingIntervalMs);
-
     if (_newTopic != null && _adminClient != null) {
-      LOG.info("Try to create Cruise Control metrics topic if not exist");
-      createCruiseControlMetricsTopic();
-      CruiseControlMetricsUtils.closeAdminClientWithTimeout(_adminClient, 5000);
+      try {
+        Collection<TopicListing> topicListings = _adminClient.listTopics().listings().get();
```
efeg: This call to `listTopics()` introduces an additional network roundtrip in case the topic does not exist.
Wouldn't it be more efficient to attempt to create the topic and handle `TopicExistsException`? If the topic exists, then we can update it.

Maybe more importantly, if the topic is created after the topic list is retrieved here, the call to create topic will fail with `TopicExistsException` and the code won't be able to handle it.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -137,45 +153,107 @@ public void configure(Map<String, ?> configs) {
         LOG.warn("Cruise Control metrics topic auto creation was disabled", e);
       }
     }
-
   }
 
-  private NewTopic createNewTopicFromReporterConfig(CruiseControlMetricsReporterConfig reporterConfig)
-  throws CruiseControlMetricsReporterException {
+  protected NewTopic createNewTopicFromReporterConfig(CruiseControlMetricsReporterConfig reporterConfig)
```
efeg: Is it possible to make this static and move it to `CruiseControlMetricsUtils` with a JavaDoc?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -137,45 +153,107 @@ public void configure(Map<String, ?> configs) {
...
+            break;
+          }
+        }
+        if (topicExists) {
+          maybeUpdateCruiseControlMetricsTopic();
+        } else {
+          createCruiseControlMetricsTopic();
+        }
+      } catch (InterruptedException | ExecutionException e) {
+        LOG.warn("Skip checking Cruise Control metrics topic " + _newTopic.name(), e);
```
efeg: Nit: Can we use `{}` rather than string concat?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -137,45 +153,107 @@ public void configure(Map<String, ?> configs) {
...
+    maybeUpdateTopicConfig();
+    maybeIncreaseTopicPartitionCount();
+  }
+
+  protected void maybeUpdateTopicConfig() {
+    try {
+      // Retrieve topic config to check and update.
+      ConfigResource topicResource = new ConfigResource(ConfigResource.Type.TOPIC, _cruiseControlMetricsTopic);
+      DescribeConfigsResult describeConfigsResult = _adminClient.describeConfigs(Collections.singleton(topicResource));
+      Config topicConfig = describeConfigsResult.values().get(topicResource).get();
```
efeg: For `describeConfigsResult` and `alterConfigsResult`, we seem to be blocking indefinitely.
Can we add timeouts to relevant get calls to avoid potential deadlock?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsUtils.java
```diff
@@ -106,4 +110,20 @@ private static void setStringConfigIfExists(CruiseControlMetricsReporterConfig c
...
+  /**
+   * Create a config altering operation if config's current value does not equal to target value.
+   * @param configsToAlter Set of config altering operations to be applied.
+   * @param configName Name of config to check.
+   * @param targetConfigValue Target value of the config.
+   * @param currentConfig Current value of the config.
+   */
+  public static void maybeUpdateConfig(Set<AlterConfigOp> configsToAlter,
+                                       String configName,
+                                       String targetConfigValue,
```
efeg: Would it be possible to combine `configName` and `targetConfigValue` in a `Map`? This allows getting response from this util function in one call.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
kidkun: I feel the content of this method is specific to metric topic, I would like to keep it in this class.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsUtils.java
```diff
@@ -18,6 +19,7 @@
 public class CruiseControlMetricsUtils {
 
   public static final long ADMIN_CLIENT_CLOSE_TIMEOUT_MS = 10000;
+  public static final long CLIENT_REQUEST_TIMEOUT_MS = 10000;
```
efeg: Nit: `10000` -> `10000L`

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -130,52 +145,98 @@ public void configure(Map<String, ?> configs) {
...
+    return newTopic;
+  }
+
+  protected void createCruiseControlMetricsTopic() throws ExecutionException {
+    try {
+      CreateTopicsResult createTopicsResult = _adminClient.createTopics(Collections.singletonList(_metricsTopic));
+      createTopicsResult.values().get(_metricsTopic.name()).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      LOG.info("Cruise Control metrics topic {} is created.", _cruiseControlMetricsTopic);
+    } catch (InterruptedException | TimeoutException e) {
+      LOG.warn("Unable to create Cruise Control metrics topic {}.", _cruiseControlMetricsTopic, e);
```
efeg: Do we need to verify that the exception is a `TopicExistsException` and not due to another reason?

kidkun: In the test I found that `TopicExistsException` will not be directly thrown out here, it will be wrapped by an `ExecutionException`. That's why we do not explicitly handle `TopicExistsException`.

efeg: Isn't the `ExecutionException` thrown by the call to `get` for the future?

`TopicExistsException` seems to be stored in `CreateTopicsResponse`  (i.e. please see `CreateTopicsRequest#getErrorResponse`), which is handled in `KafkaAdminClient#createTopics` -> `handleResponse`. This completes the future exceptionally.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -130,52 +146,104 @@ public void configure(Map<String, ?> configs) {
...
+  }
+
+  protected void createCruiseControlMetricsTopic() throws TopicExistsException {
+    try {
+      CreateTopicsResult createTopicsResult = _adminClient.createTopics(Collections.singletonList(_metricsTopic));
+      createTopicsResult.values().get(_metricsTopic.name()).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      LOG.info("Cruise Control metrics topic {} is created.", _cruiseControlMetricsTopic);
+    } catch (ExecutionException e) {
+      if (e.getCause() instanceof TopicExistsException) {
+        throw (TopicExistsException) e.getCause();
```
efeg: Not sure why we need a cast here.

Issue #1099, opened 2020-02-14
Add a sensor to indicate whether the goal violation detector has unfixable goals.


efeg:
**Background**: An unfixable goal is a violated hard goal that CC cannot satisfy -- typically due to lack of physical resources in the cluster. For example, rack awareness goal can be violated if there are partitions with a replication factor > the number of racks in the cluster.

**Proposal**: Add a sensor that indicates whether there are unfixable goals in the cluster. This metric indicates positive response if the latest goal violation contains unfixable goals. It indicates negative response if either the latest goal violation contains no unfixable goals, or if any execution was started after the latest goal violation.

PR #1100, opened 2020-02-14
Add a sensor to indicate whether the goal violation detector has unfixable goals.

efeg:
Addresses the issue #1099.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -555,6 +555,7 @@ public class Executor {
     sanityCheckOngoingMovement();
     _hasOngoingExecution = true;
     _anomalyDetector.maybeClearOngoingAnomalyDetectionTimeMs();
+    _anomalyDetector.resetHasUnfixableGoals();
     _stopSignal.set(NO_STOP_EXECUTION);
     _executionStoppedByUser.set(false);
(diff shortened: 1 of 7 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -555,6 +555,7 @@ private void startExecution(LoadMonitor loadMonitor,
     sanityCheckOngoingMovement();
     _hasOngoingExecution = true;
     _anomalyDetector.maybeClearOngoingAnomalyDetectionTimeMs();
+    _anomalyDetector.resetHasUnfixableGoals();
```
kidkun: Why we need to reset `hasUnfixableGoals` here? Shouldn't refreshing this value in anomaly detector's anomaly handling logic enough?

efeg: An ongoing execution would potentially take hours to days to complete. During an ongoing execution, goal violation detector skips detection; hence, would fail to refresh this state.

Resetting the state during an ongoing execution avoids keeping this signal for non-actionable case. If an execution was manually started, the cluster operator already knows if there are unfixable goals, because s/he either fixes them or skips them explicitly (on purpose).

kidkun: I see, thanks for the explanation.

Issue #1104, opened 2020-02-17
Upgrading Kafka dependency to 2.4


viktorsomogyi:
Upgrading to Kafka 2.4 would open up the possibility to do reassignment through a Java API with 2.4+ brokers, however some parts of the source changed so we need to adapt those parts of the codebase.
I collected the list of source incompatible changes:
* 121308cc7 - (2019-05-29) KAFKA-8286; Generalized Leader Election Admin RPC (KIP-460) (#6686): the return type of getAllTopicsInCluster changed from Seq to Set
* I0Itec.zkclient has been removed from Kafka so there are compile failures in `BrokerFailureDetector`: probably we can add a direct dependency on I0Itec it or rewrite that part of the codebase to depend on KafkaZkClient
* 1df01d258 - (2019-10-24) KAFKA-9089; Reassignment should be resilient to unexpected errors (#7562): AdminZkClient.addPartitions' assignment list parameter type changed.

PR #1105, opened 2020-02-17
Upgrade Kafka dependency to 2.4

viktorsomogyi:
Addresses the issue #1104 

* Added I0Itec zkclient dependency as it has been removed from Kafka
* `KafkaZkClient.getAllTopicsInCluster`'s return type changed from Seq to Set
* `AdminZkClient.addPartition`'s `existingAssignment`'s parameter type changed

### DIFF
--- a/cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
+++ b/cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
@@ -131,8 +131,4 @@ object ExecutorUtils {
   def getAllLiveBrokerIdsInCluster(kafkaZkClient: KafkaZkClient): java.util.List[java.lang.Integer] = {
     seqAsJavaList(kafkaZkClient.getAllBrokersInCluster.map(_.id : java.lang.Integer))
   }
-
-  def getAllTopicsInCluster(kafkaZkClient: KafkaZkClient): java.util.List[String] = {
-    seqAsJavaList(kafkaZkClient.getAllTopicsInCluster)
-  }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -221,6 +221,7 @@ project(':cruise-control') {
     compile 'org.eclipse.jetty:jetty-server:9.4.26.v20200117'
     compile 'io.dropwizard.metrics:metrics-core:3.2.6'
     compile 'com.nimbusds:nimbus-jose-jwt:8.5'
+    compile 'com.101tec:zkclient:0.11'
     compile group: 'io.swagger.parser.v3', name: 'swagger-parser', version: '2.0.18'
     compile group: 'io.github.classgraph', name: 'classgraph', version: '4.8.65'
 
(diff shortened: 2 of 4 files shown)


cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -131,8 +131,4 @@ object ExecutorUtils {
   def getAllLiveBrokerIdsInCluster(kafkaZkClient: KafkaZkClient): java.util.List[java.lang.Integer] = {
     seqAsJavaList(kafkaZkClient.getAllBrokersInCluster.map(_.id : java.lang.Integer))
   }
-
-  def getAllTopicsInCluster(kafkaZkClient: KafkaZkClient): java.util.List[String] = {
```
viktorsomogyi: This method isn't called anywhere in the project so I removed it.

viktorsomogyi:

cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
efeg: Thanks for removing the dead code.

viktorsomogyi:
@efeg ok, made the change with `close`. 

You mentioned that it's a requirement to compile CC with different Kafka versions, but between 2.3 and 2.4 `AdminZkClient.addPartitions` changed its parameter types so after making that change which is made here in this PR, it makes CC uncompilable with 2.3. Do you think we should find a workaround for this? I'm afraid there might not be much more option than duplicating some of the code of `AdminZkClient` and implement `addPartitions`.

efeg:
> @efeg ok, made the change with `close`.
> 
> You mentioned that it's a requirement to compile CC with different Kafka versions, but between 2.3 and 2.4 `AdminZkClient.addPartitions` changed its parameter types so after making that change which is made here in this PR, it makes CC uncompilable with 2.3. Do you think we should find a workaround for this? I'm afraid there might not be much more option than duplicating some of the code of `AdminZkClient` and implement `addPartitions`.

@viktorsomogyi Backwards incompatible changes in this PR seem to be limited to `KafkaSampleStore`, which is a pluggable component. Presumably, such implementation changes in pluggable components should be acceptable as those, who compile CC with different Kafka versions (e.g. `pre-Kafka 2.4`), can avoid being affected using their own pluggable component with the old API. Note that none of these changes should affect the compatibility of CC with the version of managed Kafka cluster.

akatona84:
@efeg  Could you check this PR again, please?

viktorsomogyi:
Thanks @akatona84!

@efeg I checked the release notes of the 2.4 release and I don't think there is anything else that would break this kind of backward compatibility that you mentioned. Would you please check this PR again once you have some time for this?

viktorsomogyi:
Thanks @efeg, I rebased it and also changed the Kafka version to 2.4.1 as the first bugfix release came out since. Checked its release notes, there are a couple of bug fixes but I think they don't break anything related to backward compatibility.

Issue #1109, opened 2020-02-19
Ensure that self healing goal violations do not report empty goal violations in ExecutorState.


efeg:
Checking the executor substate during an ongoing self healing for goal violations reports the following `triggeredTaskReason`:

```
ExecutorState: {..., triggeredTaskReason: Self healing for GOAL_VIOLATION: {Unfixable goal violations: {}, Fixable goal violations: {}, Exclude brokers recently (removed: true demoted: true)}, ...}
```
Note that both `Unfixable goal violations` and `Fixable goal violations` are empty. CC should ensure that the goals that are being fixed due to self healing are reported properly.

PR #1110, opened 2020-02-19
Ensure that self healing goal violations do not report empty goal violations in ExecutorState.

efeg:
Addresses the issue #1109.


kidkun:
@efeg  Can you explain why originally both the goal lists are empty?

efeg:
@kidkun The `reason` string in `GoalViolations#configure(Map<String, ?>)`, which is expected to show fixable and unfixable goals, is passed to `RebalanceRunnable` w/o being populated with those goals (i.e. note that `_violatedGoalsByFixability` is initialized with an empty hash map). The fix opts for lazy evaluation of anomaly `reason`, which ensures that the anomalies can be updated to represent their full state upon evaluation -- e.g. please note that `GoalViolations#addViolation` modifies the state of `GoalViolations`.

PR #1113, opened 2020-02-24
Extract ZooKeeper version into gradle.properties

akatona84:


akatona84:
> Thanks for the patch @akatona84!
> -- To better understanding the reasoning, would you clarify why we specifically extract the ZK version.

Would be easier to build with custom kafka and zookeeper. saw the commit for kafka, thought we can do the same for zk. I know kafka commit is somewhat easing the version change of it, while ZK only has one entry in build.gradle...

Issue #1118, opened 2020-02-26
Slow broker detector throws ConcurrentModificationException


kidkun:
Full stack trace from log

> 2020/02/26 12:15:34.574 INFO [SlowBrokerFinder] [AnomalyDetector-3] [kafka-cruise-control] [] Slow broker detection started.
> 2020/02/26 12:15:34.578 WARN [SlowBrokerFinder] [AnomalyDetector-3] [kafka-cruise-control] [] Slow broker detector encountered exception: 
> java.util.ConcurrentModificationException: null
>         at java.util.HashMap$HashIterator.nextNode(HashMap.java:1442) ~[?:1.8.0_172]
>         at java.util.HashMap$EntryIterator.next(HashMap.java:1476) ~[?:1.8.0_172]
>         at java.util.HashMap$EntryIterator.next(HashMap.java:1474) ~[?:1.8.0_172]
>         at com.linkedin.kafka.cruisecontrol.detector.SlowBrokerFinder.updateBrokerSlownessScore(SlowBrokerFinder.java:238) ~[cruise-control-2.0.89.jar:?]
>         at com.linkedin.kafka.cruisecontrol.detector.SlowBrokerFinder.metricAnomalies(SlowBrokerFinder.java:219) [cruise-control-2.0.89.jar:?]
>         at com.linkedin.kafka.cruisecontrol.detector.MetricAnomalyDetector.run(MetricAnomalyDetector.java:83) [cruise-control-2.0.89.jar:?]
>         at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_172]
>         at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [?:1.8.0_172]
>         at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [?:1.8.0_172]
>         at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [?:1.8.0_172]
>         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_172]
>         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_172]
>         at java.lang.Thread.run(Thread.java:748) [?:1.8.0_172]
[...]

PR #1119, opened 2020-02-26
Fix ConcurrentModificationException thrown by SlowBrokerFinder.

kidkun:
Addresses the issue #1118 .


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -235,17 +235,21 @@ private void updateBrokerSlownessScore(Set<BrokerEntity> detectedMetricAnomalies
...
+    Set<BrokerEntity> brokersRecovered = new HashSet<>();
     for (Map.Entry<BrokerEntity, Integer> entry : _brokerSlownessScore.entrySet()) {
       BrokerEntity broker = entry.getKey();
       if (!detectedMetricAnomalies.contains(broker)) {
         Integer score = _brokerSlownessScore.get(broker);
-        // If the score reaches zero, remove its suspicion.
         if (score != null && --score == 0) {
-          _brokerSlownessScore.remove(broker);
-          _detectedSlowBrokers.remove(broker);
+          brokersRecovered.add(broker);
```
efeg: If this loop only uses the key of the `_brokerSlownessScore` map, why does it iterate over its entries rather than its keys?

PR #1146, opened 2020-03-12
[ccclient-1.0.6] Add stop_ongoing_execution parameter

### LINKED ISSUE #1141: cruise-control-client missing stop_ongoing_execution boolean parameter
This parameter should be added to `cruise-control-client`

mgrubent:
Fixes #1141


mgrubent:
Should probably follow #1145

cruise-control-client/setup.py
```diff
@@ -11,7 +11,7 @@
 
 setuptools.setup(
     name='cruise-control-client',
-    version='1.0.4',
+    version='1.0.6',
```
efeg: Were the `*.egg` files added intentionally to this patch? -- If not should we update `.gitignore` to prevent their addition by mistake?

mgrubent: The `*.egg` files were added unintentionally; I will add `*.egg` to `gitignore` and remove these `*.egg` files

PR #1154, opened 2020-03-24
Change webserver ssl keystore and key password to password type

viktorsomogyi:
Changing them to `Password` will hide the actual values at logging.


viktorsomogyi:

Issue #1160, opened 2020-04-01
SlowBrokerFinder throws NullPointerException


kidkun:
Currently `SlowBrokerFinder` holds an assumption that all brokers in the cluster should have a metric history, which may not be valid in some case.

 For example, a broker just joined the cluster, Cruise Control does not have enough time to collect and generate a metric history for it.

PR #1161, opened 2020-04-01
Fix NullPointerException thrown by SlowBrokerFinder.

kidkun:
Addresses the issue #1160 .


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -200,15 +200,19 @@ private void collectLogFlushTimeMetric(BrokerEntity broker,
...
+      double[] historicalLogFlushTime = aggregatedMetricValues.valuesFor(BROKER_LOG_FLUSH_TIME_MS_999TH_ID).doubleArray();
+      List<Double> historicalValue = new ArrayList<>(historicalLogFlushTime.length);
+      for (int i = 0; i < historicalLogFlushTime.length; i++) {
+        if (historicalLogFlushTime[i] > 5.0) {
+          historicalValue.add(historicalLogFlushTime[i]);
+        }
       }
+      historicalLogFlushTimeMetricValues.put(broker, historicalValue);
+    } else {
+      LOG.info("Metric history for broker {} is missing.", broker.brokerId());
```
efeg, line 214: (applies to the other log below) Is it possible to add reason regarding why it might be missing? Would a cold start for CC also cause this? Should this be an `info` or debug level log? -- i.e. would it be important for the user to see this in the log or is it internal to the slow broker finder design?

Issue #1155, opened 2020-03-27
stop_ongoing_execution parameter is broken


kidkun:
If user first run "remove_broker?broker_id=1&dryrun=false" and then run "remove_broker?broker_id=1,2&dryrun=false&stop_ongoing_execution=true" , Cruise Control should cancel the first execution and start the second execution. However, the user will receive a error response for the second request like

> Error processing POST request '/remove_broker' due to: 'java.lang.IllegalStateException: Cannot execute new proposals while there are ongoing partition reassignments.'.

PR #1162, opened 2020-04-01
Fix the bug that stop_ongoing_execution parameter does not work.

kidkun:
Addresses the issue #1155 .

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -217,14 +217,19 @@ public class KafkaCruiseControl {
     if (dryRun) {
       return;
     }
-    if (_executor.hasOngoingExecution() && !stopOngoingExecution) {
-      throw new IllegalStateException(String.format("Cannot start a new execution while there is an ongoing execution. "
-                                                    + "Please use %s=true to stop ongoing execution and start a new one.",
-                                                    STOP_ONGOING_EXECUTION_PARAM));
-    } else if (_executor.hasOngoingPartitionReassignments()) {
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -217,14 +217,19 @@ public void sanityCheckDryRun(boolean dryRun, boolean stopOngoingExecution) {
...
+      if (!stopOngoingExecution) {
+        throw new IllegalStateException(String.format("Cannot start a new execution while there is an ongoing execution. "
+                                                      + "Please use %s=true to stop ongoing execution and start a new one.",
+                                                      STOP_ONGOING_EXECUTION_PARAM));
+      }
+    } else {
+      if (_executor.hasOngoingPartitionReassignments()) {
+        throw new IllegalStateException("Cannot execute new proposals while there are ongoing partition reassignments.");
+      }
+      if (_executor.hasOngoingLeaderElection()) {
```
efeg, line 230: Nit: Is the change from `else if (_executor.hasOngoingLeaderElection())` to `if (_executor.hasOngoingLeaderElection())` needed?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -217,14 +217,19 @@ public void sanityCheckDryRun(boolean dryRun, boolean stopOngoingExecution) {
...
+        throw new IllegalStateException("Cannot execute new proposals while there are ongoing leadership reassignments.");
```
efeg, line 231: In these two exception messages, can we indicate that these ongoing executions are expected to have been started by an agent that is external to this Cruise Control instance?

PR #1168, opened 2020-04-06
Use ReassignPartitionsCommand to reassign partitions

### LINKED ISSUE #1167: Rebalance stuck, partitions rebalance not progressing in Kafka
I am seeing an issue with the partition reassignments performed by CC: basically any rebalance that triggers more than one partition reassignment batch (tasks?) in kafka never completes.
It gets stuck in the second batch. with cc waiting for reassignment to complete and kafka controller doing nothing.
The `reassign_partition` zk node is being updated by CC but Kafka Controller is not triggering the rebalance.
The only workaround is to delete /controller in kafka

Working with kafka 2.4.1 server side and a build of cc that has the 2.4.1 client libs (master)

Looking at 
https://github.com/linkedin/cruise-control/blob/b386141146dfe4cc013b8233c19434e09b50027c/cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala#L89 it seems that CC is not using `ReassignPartitionsCommand` but instead updating the zk node directly.
This apparently is not enough in newer version of Kafka.

amuraru:
Fixes #1167

### DIFF
--- a/cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
+++ b/cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
@@ -7,8 +7,9 @@ package com.linkedin.kafka.cruisecontrol.executor
 import java.util
 import java.util.Properties
 
-import kafka.admin.PreferredReplicaLeaderElectionCommand
-import kafka.zk.{AdminZkClient, KafkaZkClient, ZkVersion}
+import kafka.admin.{PreferredReplicaLeaderElectionCommand, ReassignPartitionsCommand}
+import kafka.common.AdminCommandFailedException
+import kafka.zk.{AdminZkClient, KafkaZkClient}
 import org.apache.kafka.common.TopicPartition
 import org.slf4j.{Logger, LoggerFactory}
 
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/scala/com/linkedin/kafka/cruisecontrol/executor/ExecutorUtils.scala
```diff
@@ -86,9 +87,15 @@ object ExecutorUtils {
           newReplicaAssignment += (tp -> replicasToWrite)
       })
 
-      // We do not use the ReassignPartitionsCommand here because we want to have incremental partition movement.
```
amuraru, line 89: Was there a strong reason why previously the RPC was not used?

efeg:
@amuraru Thanks for the patch!

Cruise Control has not yet been certified for managing `Kafka 2.4.0+` clusters. This patch may introduce some disruptions for managing `pre-Kafka 2.4.0` clusters (see https://github.com/linkedin/cruise-control/issues/1167#issuecomment-620314334). Ideally, we should look into adopting the public API to be introduced by [KIP-455](https://cwiki.apache.org/confluence/display/KAFKA/KIP-455%3A+Create+an+Administrative+API+for+Replica+Reassignment) for replica moves.

amuraru:
Thanks @efeg for explaining the full context - agree that KIP-455 would be the right lomg term solution. Wondering if this would be an acceptable short term fix though? For pre 2.4 clusters the ReassignmemtPartComm is also available right?

efeg:
Issues https://github.com/linkedin/cruise-control/issues/1208 and https://github.com/linkedin/cruise-control/issues/1167 are intended to be addressed `migrate_to_kafka_2_4` branch starting with the version https://github.com/linkedin/cruise-control/releases/tag/2.4.1

@amuraru Closing this issue -- please let us know if you see any issues on `Kafka 2.4+` using `2.4.*` releases of CC.

PR #1182, opened 2020-04-28
Fix typos and invalid types in OAS3 yaml

viktorsomogyi:
Fixes a couple of typos in the OAS3 API yaml descriptor files, such as typos, type errors (set and map definitions) so that the yaml will be compilable with the openapi-generator tool too.

### DIFF
--- a/cruise-control/src/yaml/endpoints/demoteBroker.yaml
+++ b/cruise-control/src/yaml/endpoints/demoteBroker.yaml
@@ -7,9 +7,10 @@ DemoteBrokerEndpoint:
         in: query
         description: List of target broker ids.
         schema:
-          type: set
+          type: array
           items:
-            type: int32
+            type: integer
+            format: int32
         required: true
       - name: dryrun
         in: query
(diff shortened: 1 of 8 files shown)


viktorsomogyi:
@efeg would you please have a look at it? I fixed a couple of typo and type related errors so that a json can be compiled from this (we generate docs based on the json).
Most of these are that I changed `set` types to `array` as there is no `set` type according to the OAS3 data model (https://swagger.io/docs/specification/data-models/dictionaries/). Also changed the `map` definition a bit as integer keys aren't possible with OAS3.

viktorsomogyi:
Another issue I found is in many places it uses the following construct which is invalid:
```
DiskMB:
      oneOf:
        - type: string
          enum:
            - "DEAD"
        - type: number
          format: double
```
oneOf, anyOf, allOf can be only used for schemas and not for primitive types. It won't create an error even when you try to generate a server stub but it will result in a compilation error in the generated code as it would expect classes such as `OneOfstringdouble` that doesn't exist. I guess this is fine now as we only use this for documentation purposes but later on we have to change the API to comply with OAS (or request this feature but that would take more time).
For reference I used this to generate a test server code just to see if everything is valid:
```
openapi-generator generate -i cruise-control/src/yaml/base.yaml -g jaxrs-jersey -o /tmp/test/
```

viktorsomogyi:
Thanks, that makes sense.

Issue #1183, opened 2020-04-28
Replica Movement Strategy Fails when determining order of tasks


matthinograb:
Received an error while using the `PrioritizeSmallReplicaMovementStrategy`. Not sure if this related to https://github.com/linkedin/cruise-control/issues/1127.

CC Version: `2.0.96`

Request:
```
POST /kafkacruisecontrol/rebalance?json=true&dryrun=true&use_ready_default_goals=true&concurrent_partition_movements_per_broker=1&concurrent_leader_movements=2
````

Error Code:
```
[2020-04-28 10:50:25,950] ERROR Error processing POST request '/rebalance' due to: 'java.lang.IllegalStateException: Replica movement strategy PrioritizeSmallReplicaMovementStrategy failed to determine order of ta
sks.'. (com.linkedin.kafka.cruisecontrol.servlet.KafkaCruiseControlServlet)
java.util.concurrent.ExecutionException: java.lang.IllegalStateException: Replica movement strategy PrioritizeSmallReplicaMovementStrategy failed to determine order of tasks.
        at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
        at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2022)
        at com.linkedin.kafka.cruisecontrol.servlet.handler.async.AbstractAsyncRequest.getResponse(AbstractAsyncRequest.java:57)
...
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.doProduce(EatWhatYouKill.java:313)
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.tryProduce(EatWhatYouKill.java:171)
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.run(EatWhatYouKill.java:129)
        at org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:388)
        at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:806)
        at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:938)
        at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.IllegalStateException: Replica movement strategy PrioritizeSmallReplicaMovementStrategy failed to determine order of tasks.
[...]

Issue #1126, opened 2020-03-02
DryRun flag is not correctly used on creating composite replica movement strategy 


daniyar-kulakhmetov-liftoff:
The [getReplicaMovementStrategy](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java#L618) method of `ParameterUtils` class checks the dryrun mode incorrectly.

Instead of creating a composite strategy if it is not a dryrun mode, it does vice versa:
```
    if (!getDryRun(request)) {
      return null;
    }
```

Should be:
```
    if (getDryRun(request)) {
      return null;
    }
```

That's for both branches: master and kafka_0_11_and_1_0.

efeg:
Closing this issue as it has been resolved on the linked PR.

PR #1184, opened 2020-04-29
Fix 1183 chained replica movement strategies

matthinograb:
Addresses the issue #1183, #1126 

- Change test case to re-produce the IllegalStateException in #1183.
- Copy commits made from https://github.com/liftoffio/cruise-control/pull/1 and bring into master branch
- Force chaining with BaseReplicaMovementStrategy when strategy changes dynamically

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
@@ -196,7 +196,10 @@ public class ExecutionTaskPlanner {
       _interPartMoveTaskByBrokerId = _defaultReplicaMovementTaskStrategy.applyStrategy(_remainingInterBrokerReplicaMovements,
                                                                                        cluster);
     } else {
-      _interPartMoveTaskByBrokerId = replicaMovementStrategy.applyStrategy(_remainingInterBrokerReplicaMovements, cluster);
+      // Chain the generated composite strategy with BaseReplicaMovementStrategy in the end to ensure the returned
+      // strategy can always determine the order of two tasks.
+      _interPartMoveTaskByBrokerId = replicaMovementStrategy.chain(new BaseReplicaMovementStrategy())
+              .applyStrategy(_remainingInterBrokerReplicaMovements, cluster);
     }
   }
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
@@ -4,7 +4,6 @@
 
 package com.linkedin.kafka.cruisecontrol.executor;
 
-import com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils;
 import com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig;
 import com.linkedin.kafka.cruisecontrol.executor.strategy.BaseReplicaMovementStrategy;
 import com.linkedin.kafka.cruisecontrol.executor.strategy.ReplicaMovementStrategy;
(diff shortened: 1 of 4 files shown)


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -62,7 +62,7 @@
       new ExecutionProposal(new TopicPartition(TOPIC1, 3), 0, _r3, Arrays.asList(_r3, _r2), Arrays.asList(_r2, _r3));
 
   private final ExecutionProposal _partitionMovement1 =
-      new ExecutionProposal(new TopicPartition(TOPIC2, 0), 4, _r0, Arrays.asList(_r0, _r2), Arrays.asList(_r2, _r1));
+      new ExecutionProposal(new TopicPartition(TOPIC2, 0), 1, _r0, Arrays.asList(_r0, _r2), Arrays.asList(_r2, _r1));
```
matthinograb, line 65: Note: We didn't need to change the PrioritizeLargeReplicaRebalance strategy tests after making this change, because of: https://github.com/linkedin/cruise-control/issues/1183#issuecomment-621460812.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
```diff
@@ -315,6 +317,13 @@ private void maybeAddLeaderChangeTasks(Collection<ExecutionProposal> proposals,
                                                                 Set<TopicPartition> inProgressPartitions) {
     LOG.trace("Getting inter-broker replica movement tasks for brokers with concurrency {}", readyBrokers);
     List<ExecutionTask> executableReplicaMovements = new ArrayList<>();
+
+    // Kafka doesn't allow new replica movements if there is an ongoing rebalance
+    if (!inProgressPartitions.isEmpty()) {
```
amuraru, line 322: Would this alliviate the issue described in https://github.com/linkedin/cruise-control/issues/1167#issuecomment-620314334 ? @efeg can you please comment?

kidkun: I do not think this check should be added.

The way CC submit tasks is in batch. we select a set of tasks based on the concurrency requirement and submit them in batch, later we check the finish status and select more tasks to current ongoing tasks(as long as we do not violate concurrency requirement) and submit in second batch.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
kidkun: Is there any specific reason we want to change the test case here? 
I just test locally, with the bug fix, If we do not change the test, it can still pass.

matthinograb: The test cases were missing the "equal partition size" case, which caused the original bug. Adding this test case proves that the fix is working as expected, and it should help to prevent future regressions (if the chaining breaks in the future).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
matthinograb: From my understanding, Kafka cannot start a new rebalance while a current one is in progress. So we don't need to create a second batch, if the first batch isn't completed yet.

But I'll remove this line, since it isn't related to the referenced issue -- and isn't important for the bugfix.

kidkun: Kafka can accept new task while there is ongoing task and that CC use that to ensure the task are always executed in desired concurrency.

efeg: @matthinograb
Just wanted to chime in to give a little more context on @kidkun 's explanation. 
Starting a new rebalance while a current one is in progress requires the workaround patch described in https://github.com/linkedin/cruise-control/issues/1167#issuecomment-620314334 -- i.e. it is indeed not supported in Apache Kafka out of the box.

Hope it helps!

Issue #1185, opened 2020-05-04
Suppress false positive alert from SlowBrokerFinder


kidkun:
Currently slow broker find detect slow broker by checking the log flush time and per-byte log flush time metrics against each broker's history and its peers. There are still some false positive alerts found in production. It is better to further filter the broker by checking broker's log flush time against a pre-defined threshold(the maximal value we know for sure it is a not slow broker).

PR #1186, opened 2020-05-04
Further filter detected slow broker against pre-defined flush time threshold.

kidkun:
Addresses the issue #1185 .


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -174,6 +184,7 @@ public SlowBrokerFinder() {
     }
 
     Set<BrokerEntity> detectMetricAnomalies = getMetricAnomalies(historicalLogFlushTimeMetricValues, currentLogFlushTimeMetricValues);
+    detectMetricAnomalies.retainAll(detectMetricAnomaliesFromThreshold(currentLogFlushTimeMetricValues, _logFlushTimeThresholdMs));
```
efeg, line 187: Can we avoid passing private variable `_logFlushTimeThresholdMs` to non-static function `detectMetricAnomaliesFromThreshold` and update the name of the function to indicate its custom role regarding log flush time?

Actually it is a little confusing that we have another function in this class called `getMetricAnomalies`, which is also used for detecting metric anomalies -- can we merge/rename them to make the logic more clear?

efeg, line 187: Nit: `detectMetricAnomalies` -> `detectedMetricAnomalies`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -286,6 +297,16 @@ private void detectMetricAnomaliesFromPeers(Map<BrokerEntity, Double> currentVal
     }
   }
 
+  private Set<BrokerEntity> detectMetricAnomaliesFromThreshold(Map<BrokerEntity, Double> currentValue, double threshold) {
+    Set<BrokerEntity> detectedMetricAnomalies = new HashSet<>();
+    for (Map.Entry<BrokerEntity, Double> entry : currentValue.entrySet()) {
```
efeg, line 302: The loop uses only the key, but iterates over entities. Can we either iterate over `keySet()` or update the logic to use `getValue()`, as well?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -416,6 +437,21 @@ public void configure(Map<String, ?> configs) {
       }
     }
 
+    String logFlushTimeThresholdMs = (String) originalConfig.get(SLOW_BROKER_LOG_FLUSH_TIME_THRESHOLD_MS_CONFIG);
```
efeg, line 440: This function (i.e. configure) is full of repeated code fragments. Can we use util functions to eliminate code repetition and be more concise?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinderTest.java
```diff
@@ -161,6 +161,36 @@ public void testNoFalsePositiveDetection() {
     assertTrue(anomalies.isEmpty());
   }
 
+  /**
+   * Test slow broker finder does not report false positive anomaly when the broker's absolute log flush time is small.
+   */
+  @Test
+  public void testNoFalsePositiveDetection2() {
```
efeg, line 168: Nit: Enumerating functions does not give much information about the difference between `testNoFalsePositiveDetection()` and `testNoFalsePositiveDetection2()`. Can we rename them with a suffix that describes what each function are?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
kidkun: I rename the method to make it more specific.
I was thinking about merge this part of logic into `getMetricAnomalies()`. The issue is the logic in `getMetricAnomalies()` is generic and this extra check is specific to log flush time metrics.

Issue #1193, opened 2020-05-15
Dependency on joda-time


urbandan:
CC depends on joda-time through swagger-parser and swagger-compat-spec-parser.
joda-time is unnecessary, and there is no need for using the old swagger-parser.
Footprint can be decreased through switching to swagger-parser-v3.

PR #1194, opened 2020-05-15
Use swagger-parser-v3 to eliminate compat dependencies

urbandan:
Addresses the issue #1193
Upgrading to swagger-parser-v3 removes a good chunk of dependencies, and decreases the footprint of the project.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -222,7 +222,7 @@ project(':cruise-control') {
     compile 'io.dropwizard.metrics:metrics-core:3.2.6'
     compile 'com.nimbusds:nimbus-jose-jwt:8.5'
     compile 'com.101tec:zkclient:0.11'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -222,7 +222,7 @@ project(':cruise-control') {
     compile 'io.dropwizard.metrics:metrics-core:3.2.6'
     compile 'com.nimbusds:nimbus-jose-jwt:8.5'
     compile 'com.101tec:zkclient:0.11'
-    compile group: 'io.swagger.parser.v3', name: 'swagger-parser', version: '2.0.18'
+    compile group: 'io.swagger.parser.v3', name: 'swagger-parser-v3', version: '2.0.18'
```
viktorsomogyi, line 225: I think we should bump this to 2.0.19 too, so we get the latest.

urbandan: I agree, bumped the version.

PR #1198, opened 2020-05-20
Environment variable resolution in configs

### LINKED ISSUE #1197: Resolve environment variables in config on startup
This is particularly useful if passwords can't be hardcoded in the config
files but instead passed to Cruise Control via an environment variable.
For instance if there is a MY_PASSWORD=test123 then defining the following
config in cruisecontrol.properties should be resolved on startup to the
actual password.

Config file on the disk: config.name=${env:MY_ENV_VAR}
Resolved runtime config: config.name=test123

viktorsomogyi:
This commit enables parsing config values from environment variables.
This is particularly useful if passwords can't be hardcoded in the config
files but instead passed to Cruise Control via an environment variable.
For instance if there is a MY_PASSWORD=test123 then defining the following
config in cruisecontrol.properties will be resolved on startup to the
actual password.

Config file on the disk: config.name=${env:MY_ENV_VAR}
Resolved runtime config: config.name=test123

This fixes #1197

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -20,6 +20,7 @@ import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.SaslConfigs;
 import org.apache.kafka.common.config.SslConfigs;
@@ -69,6 +70,8 @@ public class KafkaCruiseControlUtils {
   public static final String OPERATION_LOGGER = "operationLogger";
   // This will make MetaData.update() trigger a real metadata fetch.
   public static final int REQUEST_VERSION_UPDATE = -1;
+  public static final String ENV_CONFIG_PROVIDER_NAME = "env";
+  public static final String ENV_CONFIG_PROVIDER_CLASS_CONFIG = ".env.class";
 
   private KafkaCruiseControlUtils() {
 
@@ -549,7 +552,8 @@ public class KafkaCruiseControlUtils {
   }
 
   /**
-   * Reads the configuration file, parses and validates the configs.
+   * Reads the configuration file, parses and validates the configs. Enables the configs to be passed
+   * in from environment variables.
    * @param propertiesFile is the file containing the Cruise Control configuration.
    * @return a parsed {@link KafkaCruiseControlConfig}
    * @throws IOException if the configuration file can't be read.
@@ -557,8 +561,8 @@ public class KafkaCruiseControlUtils {
   public static KafkaCruiseControlConfig readConfig(String propertiesFile) throws IOException {
     Properties props = new Properties();
     try (InputStream propStream = new FileInputStream(propertiesFile)) {
-      props.put("config.providers", "env");
-      props.put("config.providers.env.class", EnvConfigProvider.class.getName());
+      props.put(AbstractConfig.CONFIG_PROVIDERS_CONFIG, ENV_CONFIG_PROVIDER_NAME);
+      props.put(AbstractConfig.CONFIG_PROVIDERS_CONFIG + ENV_CONFIG_PROVIDER_CLASS_CONFIG, EnvConfigProvider.class.getName());
       props.load(propStream);
     }
     return new KafkaCruiseControlConfig(props);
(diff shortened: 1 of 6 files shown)


viktorsomogyi:

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -541,4 +547,20 @@ public static void sanityCheckNonExistingGoal(List<String> goals, Map<String, Go
...
+   * Reads the configuration file, parses and validates the configs.
+   * @param propertiesFile is the file containing the Cruise Control configuration.
+   * @return a parsed {@link KafkaCruiseControlConfig}
+   * @throws IOException if the configuration file can't be read.
+   */
+  public static KafkaCruiseControlConfig readConfig(String propertiesFile) throws IOException {
+    Properties props = new Properties();
+    try (InputStream propStream = new FileInputStream(propertiesFile)) {
+      props.put("config.providers", "env");
+      props.put("config.providers.env.class", EnvConfigProvider.class.getName());
```
efeg, line 561: These hardcoded config names and the way the user is expected to name their environment variables to pass them to CC is not very clear. Would it help if we (1) refer to the following names rather than hardcoding these configs:
1. `AbstractConfig#CONFIG_PROVIDERS_CONFIG`, and
2. `AbstractConfig#CONFIG_PROVIDERS_PARAM`

and (2) provide a brief section in `Configurations.md` regarding how to name env variables to pass them to CC?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -541,4 +547,20 @@ public static void sanityCheckNonExistingGoal(List<String> goals, Map<String, Go
 
     return balancednessCostByGoal;
   }
+
+  /**
+   * Reads the configuration file, parses and validates the configs.
```
efeg, line 552: Should we also mention the properties that enable configs to be provided out of the configuration file? 
Is it intentional to override these hardcoded properties if they were also defined in the configuration file?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/EnvConfigProvider.java
```diff
@@ -0,0 +1,59 @@
...
+  public ConfigData get(String path, Set<String> keys) {
+    assertNoPath(path);
+    Map<String, String> filtered = new HashMap<>(getenv());
+    filtered.keySet().retainAll(keys);
+    return new ConfigData(filtered);
+  }
+
+  @Override
+  public void close() {
+    // nop
```
efeg, line 36: Should we `clear()` `_preConfiguredEnvironmentVariables`?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/EnvConfigProviderTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class EnvConfigProviderTest {
+
+  public static final String TEST_PASSWORD = "testPassword123";
+  public static final String NOT_SUBSTITUTED_CONFIG = "${env:SSL_KEY_PASSWORD}";
+
+  @Test
+  public void testEnvConfigProvider() throws IOException {
+    KafkaCruiseControlConfig configs = KafkaCruiseControlUtils.readConfig(
+        Objects.requireNonNull(this.getClass().getClassLoader().getResource("envConfigProviderTest.properties")).getPath());
```
efeg, line 25: Nit: Is it possible to move the hardcoded value to a static variable?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
viktorsomogyi: `AbstractConfig#CONFIG_PROVIDERS_PARAM` seems to be private but I'll resolve the other points.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/EnvConfigProvider.java
viktorsomogyi: Good point, will do so.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
viktorsomogyi: 1st: acked, will do
2nd: makes sense, our requirements were quite simple but I can agree that people may want to do something else so it isn't optimal to override their configs.

viktorsomogyi: 1st: acked, will do
2nd: well, I treat them as defaults here so if people want to do something else than what CC does by default, then they should be free to. What can I improve here?

Issue #1213, opened 2020-05-27
Update the Balancedness Score under Unhealthy Cluster State


efeg:
A cluster is considered as unhealthy if it contains offline logDirs with replicas. Hence, depending on the replication factor of the corresponding partitions, minISR configuration of the topic, and the client's consistency requirements (e.g. acks=all), clients that produce to, or consume from the corresponding partitions with offline replicas are likely to experience undesired side-effects (e.g. higher latency, data loss). To indicate the severity of the issue, such clusters have "0" balancedness score (exposed as a JMX metric).

While unlikely in practice, it is possible to have an extreme case that all configured goals are violated, which also yields a balancedness score of 0. To distinguish this edge case from the case with unhealthy cluster state, update the balancedness score under unhealthy cluster state to a value out of `[0, 100.0]` range -- e.g. `-1`.

PR #1214, opened 2020-05-27
Update the Balancedness Score under Unhealthy Cluster State

efeg:
Addresses the issue #1213.


mgrubent:
This is going to be a backwards-incompatible change; what version bump would this PR entail?

PR #1217, opened 2020-05-28
Compile using java 11

amuraru:
use java-11 to compile:
- migrated from findbugs to spotbugs (findbugs won't work with java 11). This required separation of spotBugs gradle task in a separate run in CircleCI to avoid out of memory issues
- fixed checkstyle config files discovery
- fixed on prio 1 bug `VO_VOLATILE_INCREMENT` reported in AnomalyDetector.java
- upgraded org.eclipse.jetty.security and fixed tests to provide a non-null Credential

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,16 +4,21 @@ jobs:
 
   build:
     environment:
-      _JAVA_OPTIONS: "-Xms512m -Xmx512m"
+      _JAVA_OPTIONS: "-Xms512m -Xmx1g"
     working_directory: ~/workspace
     docker:
-      - image: circleci/openjdk:8-jdk
+      - image: circleci/openjdk:11-jdk
     steps:
       - checkout
       - restore_cache:
           key: dependency-cache-{{ checksum "build.gradle" }}
       - run:
-          command: ./gradlew --no-daemon -PmaxParallelForks=1 clean javadoc build
+          command: ./gradlew --no-daemon clean javadoc
+      - run:
+          # run static analysis tasks standalone to avoid OOME in CircleCI
+          command: ./gradlew --max-workers=1 --no-daemon analyze
+      - run:
+          command: ./gradlew --no-daemon -PmaxParallelForks=1 build
       - save_cache:
           key: dependency-cache-{{ checksum "build.gradle" }}
           paths:
@@ -32,7 +37,7 @@ jobs:
   publish:
     working_directory: ~/workspace
     docker:
-      - image: circleci/openjdk:8-jdk
+      - image: circleci/openjdk:11-jdk
     steps:
       - checkout
       - restore_cache:
--- a/build.gradle
+++ b/build.gradle
@@ -8,6 +8,7 @@ plugins {
   id "idea"
   id "jacoco" // Java Code Coverage plugin
   id "com.github.ben-manes.versions" version "0.28.0"
+  id "com.github.spotbugs" version "4.2.4" apply false
 }
 
 group = 'com.linkedin.cruisecontrol'
@@ -76,7 +77,7 @@ subprojects {
 
   apply plugin: 'java'
   apply plugin: 'checkstyle'
-  apply plugin: 'findbugs'
+  apply plugin: "com.github.spotbugs"
   apply plugin: 'jacoco'
 
   task sourcesJar(type: Jar, dependsOn: classes) {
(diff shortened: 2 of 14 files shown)


.circleci/config.yml
```diff
@@ -5,9 +5,10 @@ jobs:
   build:
     environment:
       _JAVA_OPTIONS: "-Xms512m -Xmx512m"
+      GRADLE_OPTS: "-Dorg.gradle.daemon=false -Dorg.gradle.workers.max=1"
```
efeg, line 8: I suspect that with jdk 11, we may need something like https://stackoverflow.com/a/56939344 -- i.e. rather than relying on Xms and Xmx

build.gradle
```diff
@@ -76,7 +77,7 @@ subprojects {
 
   apply plugin: 'java'
   apply plugin: 'checkstyle'
-  apply plugin: 'findbugs'
+  apply plugin: "com.github.spotbugs"
```
efeg, line 80: Should we extract the migration from findbugs to spotbugs on a separate PR? This would let us avoid `finished with non-zero exit value 137` error from blocking us.

amuraru: circleci failures seems to be due to OutOfMemory exceptions when running build. Most probably due to increased spotbugs memory needs.

Are you suggesting migrating first to spotbugs and keep java8 compiler? I think we'd hit the same issue

amuraru: Splitting spotbugs in separate runs seems to workaround the OOM issue in circleci.
@efeg I am thinking of creating a "verify" gradle task to aggregate these two tasks. Would this be a decent workaround?

amuraru:
@efeg the build is passing now - let me know if this looks good to you. Thanks

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/MetricSamplerPartitionAssignor.java
```diff
@@ -17,13 +17,14 @@
 public interface MetricSamplerPartitionAssignor extends CruiseControlConfigurable {
 
   /**
-   * @deprecated Please use {@link #assignPartitions(Cluster)}.
+   * Please use {@link #assignPartitions(Cluster)}.
```
efeg, line 20: Nit: Should we keep `@deprecated` tag?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/security/trustedproxy/TrustedProxyAuthorizationService.java
```diff
@@ -20,12 +21,18 @@
...
+      return false;
+    }
+  };
   private final UserStore _adminUserStore;
   private final Pattern _trustedProxyIpPattern;
 
   TrustedProxyAuthorizationService(List<String> userNames, String trustedProxyIpPattern) {
     _adminUserStore = new UserStore();
-    userNames.forEach(u -> _adminUserStore.addUser(u, null, new String[] { DefaultRoleSecurityProvider.ADMIN }));
+    userNames.forEach(u -> _adminUserStore.addUser(u, NO_CREDENTIAL, new String[] { DefaultRoleSecurityProvider.ADMIN }));
```
efeg, line 35: I am curious why this change is needed -- doesn't it introduce a behavior change?
 -- Also, in case the change is needed, can we move it to a publicly accessible place and let the relevant unit tests (e.g. `JwtAuthenticatorTest`, `JwtLoginServiceTest`, `TrustedProxyLoginServiceTest`, and `SpnegoUserStoreAuthorizationServiceTest`) access this -- i.e. instead of defining their own static vars in unit tests?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/MetricSamplerPartitionAssignor.java
amuraru: this was to solve a compilation warning - iiuc the `@Deprecated` annotation needs to precede the method declaration and not embedded in the comment - so moved it here: https://github.com/linkedin/cruise-control/pull/1217/files#diff-71850ea35e26c703f1a1253519e0c41cR27

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/security/trustedproxy/TrustedProxyAuthorizationService.java
amuraru: right. the change is needed as `addUser` is now requiring non-null credentials. A NPE would throw if null is passed (regardless if it's being used or not).

amuraru: agree on reusing it - what would be a good place to define it ?

efeg: Thanks for the clarification! 
> agree on reusing it - what would be a good place to define it ?

Would a utility class work?
```
/*
 * Copyright 2020 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
 */

package com.linkedin.kafka.cruisecontrol.servlet.security;

import org.eclipse.jetty.util.security.Credential;


public final class SecurityUtils {

  private SecurityUtils() {
  }

  public static final Credential NO_CREDENTIAL = new Credential() {
    @Override
    public boolean check(Object credentials) {
      return false;
    }
  };
}
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/MetricSamplerPartitionAssignor.java
amuraru: Cool - makes sense. Let me add it back to javadoc comment

amuraru:
@efeg implemented your suggestions - lmk

Issue #1221, opened 2020-05-30
Requests to update replication factor to a value larger than the number of brokers causes deadlock


efeg:
Whenever a user attempts to manually change the replication factor of topics to a value larger than the number of brokers in the cluster and with `skip_rack_awareness_check=true`, the relevant thread takes the cluster lock and does not release it. This causes a deadlock, which blocks all user requests and self-healing attempts that require generation of a cluster model.

Make CC resilient against bad user input or against such misconfigurations.

PR #1223, opened 2020-06-02
Ensure that requests to update replication factor cannot cause a deadlock.

efeg:
Addresses the issue https://github.com/linkedin/cruise-control/issues/1221.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
@@ -64,8 +64,8 @@ public class RunnableUtils {
 
   /**
    * Populate cluster rack information for topics to change replication factor. In the process this method also conducts a sanity
-   * check to ensure that there are enough racks in the cluster to allocate new replicas to racks which do not host replica
-   * of the same partition.
+   * check to ensure that there are enough brokers and racks in the cluster to allocate new replicas to racks which do not host
+   * replica of the same partition.
    *
    * @param topicsByReplicationFactor Topics to change replication factor by target replication factor.
    * @param cluster Current cluster state.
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
```diff
@@ -93,7 +93,11 @@ public static void populateRackInfoForReplicationFactorChange(Map<Short, Set<Str
     }
 
     topicsByReplicationFactor.forEach((replicationFactor, topics) -> {
-      if (replicationFactor > brokersByRack.size()) {
+      if (replicationFactor > rackByBroker.size()) {
+        throw new RuntimeException(String.format("Unable to change replication factor (RF) of topics %s to %d since there are only %d "
```
smccauliff, line 97: I'm not sure Set<String>.toString() will always convert all the topics toString, it's behavior is not stable between different versions of Java.

efeg: I can't recall observing issues related to `toString()` call on Sets using `openjdk-jdk8`, `openjdk-jdk11`, or `zulu 8`. That being said, our repo has been using ootb set implementations , such as `HashSet`, `TreeSet`, and `LinkedHashSet` -- and openJDK implementation for their `toString()` implementation in `AbstractCollection.java` hasn't changed from 8 to 11.

Have you observed issues in specific implementations / versions of Java or specific implementation of collections?

efeg:
> So the fix here is really this check?
> 
> ```java
> if (replicationFactor > rackByBroker.size()) {
> ```

Correct.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
smccauliff: Never mind.  It's some trauma I experienced in some previous version of Java.  This program never truncates; even for very large numbers and different Set implementations.
```java
import java.util.*;

class tostring {
   public static void main(String[] argv) throws Exception {
        Set<Integer> s = new TreeSet<>();
        int count = Integer.parseInt(argv[0]);
        for (int i =0; i < count; i++) {
           s.add(i);
        }
        
        System.out.println(s);
   }
}
```

Issue #1230, opened 2020-06-06
Balancedness is reported as 100 when there are offline brokers


efeg:
This is a bug introduced in a recent PR (see [deleted line](https://github.com/linkedin/cruise-control/commit/ac2140f5a583d1cda24be6e715996293106bb286#diff-592d1d65991d45cc83e40b7d30f93eafL111)). As a result of it, the balancedness is reported as `100` when it actually should be `-1`.

PR #1231, opened 2020-06-06
Fix reported balancedness when there are offline brokers.

efeg:
Addresses the issue #1230.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectionStatus.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectionStatus.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2020 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.cruisecontrol.detector;
+
+import java.util.Arrays;
+import java.util.Collections;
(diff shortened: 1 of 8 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectionStatus.java
```diff
@@ -0,0 +1,37 @@
...
+  READY, SKIP_HAS_OFFLINE_REPLICAS, SKIP_LOAD_MONITOR_NOT_READY, SKIP_EXECUTOR_NOT_READY, SKIP_MODEL_GENERATION_NOT_CHANGED,
+  SKIP_HAS_DEAD_BROKERS;
+
+  private static final List<AnomalyDetectionStatus> CACHED_VALUES = Collections.unmodifiableList(Arrays.asList(values()));
+
+  /**
+   * Use this instead of values() because values() creates a new array each time.
+   * @return enumerated values in the same order as values()
+   */
+  public static List<AnomalyDetectionStatus> cachedValues() {
```
hzxa21, line 34: Minor: is this method being used anywhere?

efeg: For the top-level enums, we follow the practice of caching the immutable result of `values()` inside the enum itself in case it needs to be used multiple times.

PR #1241, opened 2020-06-17
Add support to retrieve capacity only via load endpoint.

### LINKED ISSUE #1191: Add capacity_only parameter to load endpoint
When user only want to know capacity information, there is no need to calculate cluster model which is very expensive.

efeg:
This PR resolves #1191.

-- 
Example plaintext response, where values are defined for entries in `<...>`:
```
           HOST         BROKER      RACK         DISK_CAP(MB)            DISK(MB)/_(%)_            CORE_NUM         CPU(%)          NW_IN_CAP(KB/s)       LEADER_NW_IN(KB/s)     FOLLOWER_NW_IN(KB/s)         NW_OUT_CAP(KB/s)        NW_OUT(KB/s)       PNW_OUT(KB/s)    LEADERS/REPLICAS
<hostname-here>,          <id>,     <id>,        <diskCap-mb>,              0.000/00.00,            <cores>,         0.000,              <capacity>,                   0.000,                   0.000,              <capacity>,              0.000,              0.000,             0/0
<hostname-here>,          <id>,     <id>,        <diskCap-mb>,              0.000/00.00,            <cores>,         0.000,              <capacity>,                   0.000,                   0.000,              <capacity>,              0.000,              0.000,             0/0
...
<hostname-here>,          <id>,     <id>,        <diskCap-mb>,              0.000/00.00,            <cores>,         0.000,              <capacity>,                   0.000,                   0.000,              <capacity>,              0.000,              0.000,             0/0
```

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
@@ -13,7 +13,10 @@ import com.linkedin.kafka.cruisecontrol.config.BrokerCapacityInfo;
 import com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig;
 import com.linkedin.kafka.cruisecontrol.monitor.ModelGeneration;
 import com.linkedin.kafka.cruisecontrol.servlet.response.stats.BrokerStats;
+import java.io.IOException;
+import java.io.OutputStream;
 import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -28,6 +31,7 @@ import java.util.TreeMap;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import org.apache.commons.math3.stat.descriptive.moment.Variance;
 import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.PartitionInfo;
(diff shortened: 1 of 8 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -1242,23 +1238,6 @@ public BrokerStats brokerStats(KafkaCruiseControlConfig config) {
...
-
-    double[] variance = new double[resources.length];
-
-    for (int resourceIndex = 0; resourceIndex < resources.length; resourceIndex++) {
-      Variance varianceCalculator = new Variance();
-      variance[resourceIndex] = varianceCalculator.evaluate(utilization[resourceIndex]);
-    }
-    return variance;
-  }
-
```
hzxa21, line 1261: Why do we remove this method? I did a code search and I guess it is not related to this patch but because it is not used anywhere, right?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -1299,20 +1278,6 @@ public BrokerStats brokerStats(KafkaCruiseControlConfig config) {
...
-   */
-  public void writeTo(OutputStream out) throws IOException {
-    String cluster = String.format("<Cluster maxPartitionReplicationFactor=\"%d\">%n", _maxReplicationFactor);
-    out.write(cluster.getBytes(StandardCharsets.UTF_8));
-    for (Rack rack : _racksById.values()) {
-      rack.writeTo(out);
-    }
-    out.write("</Cluster>".getBytes(StandardCharsets.UTF_8));
-  }
-
```
hzxa21, line 1315: Why do we remove this method? I did a code search and I guess it is not related to this patch but because it is not used anywhere, right?

I noticed that `writeTo` method exists in `Replica`, `Broker`, `Load`, `Rack`, `Disk` and `Host`. Should we also keep it in `ClusterModel`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
efeg: This refactoring is irrelevant to this patch. To keep this PR self-contained, I will revert this change, and create a new refactoring PR for the meta-changes.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
efeg: Ditto as above.

PR #1289, opened 2020-07-29
Automate replica reassignment concurrency adjustment based on broker metrics

### LINKED ISSUE #1288: Automate replica reassignment concurrency adjustment based on broker metrics
Setting up the inter-broker replica reassignment concurrency for executions is currently a human-driven process. A high concurrency would help with the faster completion of executions, but it might also overwhelm the cluster, causing visible impact to clients with respect to latency and availability. Whereas, a low concurrency is safer with its minimal user impact, but it might lead to excessively long maintenance operation times sometimes taking days to complete. Hence, manually finding the optimal value is often challenging, which requires multiple attempts to get it right.

Cruise Control should automate the dynamic adjustment of replica reassignment concurrency based on a feedback loop that uses selected broker metrics. PR #1255 enabled collecting broker metrics during ongoing executions. an automated concurrency adjuster should use the relevant broker metric to decide whether to increase, decrease, or maintain the current pace of replica reassignment.

efeg:
This PR resolves #1288.
-- Selected default values for `CONCURRENCY_ADJUSTER_LIMIT_*` are based on observations from a production cluster with ongoing maintenance operation.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -288,6 +295,50 @@ public void run() {
...
+    private void refreshConcurrency() {
+      if (canRefreshConcurrency()) {
+        Integer recommendedConcurrency = ExecutionUtils.recommendedConcurrency(_loadMonitor.currentBrokerMetricValues(),
+                                                                               _executionTaskManager.interBrokerPartitionMovementConcurrency(),
+                                                                               _maxPartitionMovementsPerBroker);
+        if (recommendedConcurrency != null) {
+          setRequestedInterBrokerPartitionMovementConcurrency(recommendedConcurrency);
+        }
+      }
+    }
```
hzxa21, line 330: I don't see a lock or other synchronization mechanisms in these functions. But since they are accessing the some shared members, will there by any race condition?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -30,10 +42,119 @@
...
+   */
+  static void init(KafkaCruiseControlConfig config) {
+    CONCURRENCY_ADJUSTER_LIMIT_BY_METRIC_NAME.put(BROKER_LOG_FLUSH_TIME_MS_999TH.name(),
+                                                  config.getDouble(ExecutorConfig.CONCURRENCY_ADJUSTER_LIMIT_LOG_FLUSH_TIME_MS_CONFIG));
+    CONCURRENCY_ADJUSTER_LIMIT_BY_METRIC_NAME.put(BROKER_FOLLOWER_FETCH_LOCAL_TIME_MS_999TH.name(),
+                                                  config.getDouble(ExecutorConfig.CONCURRENCY_ADJUSTER_LIMIT_FOLLOWER_FETCH_LOCAL_TIME_MS_CONFIG));
+    CONCURRENCY_ADJUSTER_LIMIT_BY_METRIC_NAME.put(BROKER_PRODUCE_LOCAL_TIME_MS_999TH.name(),
+                                                  config.getDouble(ExecutorConfig.CONCURRENCY_ADJUSTER_LIMIT_PRODUCE_LOCAL_TIME_MS_CONFIG));
+    CONCURRENCY_ADJUSTER_LIMIT_BY_METRIC_NAME.put(BROKER_CONSUMER_FETCH_LOCAL_TIME_MS_999TH.name(),
+                                                  config.getDouble(ExecutorConfig.CONCURRENCY_ADJUSTER_LIMIT_CONSUMER_FETCH_LOCAL_TIME_MS_CONFIG));
```
hzxa21, line 67: I noticed that the metrics used by the adjuster are all broker-side metrics related to consumption and production. It may not reflect problems related to delayed leadership changes. For example, there can be a long queue of LeaderAndIsrRequest or UpdateMetadataRequest causing some partitions to become leaderless while the fetch/produce local time metrics still look good because there is less I/O if partitions are leaderless.

Have you considered also looking at the controller metrics as well? The ones I am thinking about are:
- Controller to per broker request queue time: https://github.com/linkedin/kafka/blob/2.3-li-1/core/src/main/scala/kafka/controller/ControllerChannelManager.scala#L184
- Controller to per broker request queue size: https://github.com/linkedin/kafka/blob/2.3-li-1/core/src/main/scala/kafka/controller/ControllerChannelManager.scala#L192

This patch is already a big step forward and I don't think we should address this comment in this patch but it is something worth thinking.

efeg: Thanks for the feedback -- We can definitely look into enriching the set of tracked broker metrics! 

Overall, I feel that bursty requests might lead to an increase in the queue size, but the growing size would be acceptable as long as the per-request time spent on the queue does not increase beyond "some acceptable" limit. That being said, as a safety net, we can track `RequestQueueSize` with a reasonable limit (_updated the code to add this_).
Controller to per broker request queue time metric is interesting, but it breaks CC's metric reporting/sampling assumptions on 1-1 mapping between metrics and brokers -- i.e. all other broker metrics are reported per broker, however, in this case controller reports as many metrics as the number of brokers in the cluster.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: There is indeed a case that may potentially override the initial request concurrency by `ConcurrencyAdjuster` thread. Updated the code to prevent this -- thanks!

Note that the check on `_executorState.state()` is just an optimization to minimize updates to inter-broker partition movement concurrency. It has no affect to set it in any other execution state. Hence, no strict synchronization is needed.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
hzxa21: Some scenarios may not be reflected in `RequestQueueSize` on the broker side because controller is doing blocking send and receive for the controller request, i.e. it waits for the broker response before sending out the next request in queue. 

One example can be:
- Controller receives a lot of operations (e.g. partition reassignment, preferred leader election) at the same time, which results in a large number of requests enqueued in the controller broker request queue. Let's say there are 2001 requests needed to be sent out per broker.
- The broker is processing the controller request at a normal speed. There is no excessive load on the broker so broker side metrics look fine, including `RequestQueueSize` and `Request LocalTime` metrics. Let's say the average latency for processing each request is 50ms.
- Due to the natural of the blocking send and receive for controller request, the 2001th request in the controller broker queue can take 2000 * 50ms = 100s to reach the broker since the request is generated. If the controller broker request queue is imbalance across different brokers, it can easily lead to prolong leadership transition time for some partitions.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -288,6 +295,59 @@ public void run() {
...
+     * Initialize the inter-broker partition reassignment concurrency adjustment with the load monitor and the initially
+     * requested inter-broker partition reassignment concurrency.
+     *
+     * @param loadMonitor Load monitor.
+     * @param requestedInterBrokerPartitionMovementConcurrency The maximum number of concurrent inter-broker partition movements
+     *                                                         per broker(if null, use num.concurrent.partition.movements.per.broker).
+     */
+    public synchronized void initAdjustment(LoadMonitor loadMonitor, Integer requestedInterBrokerPartitionMovementConcurrency) {
+      _loadMonitor = loadMonitor;
+      setRequestedInterBrokerPartitionMovementConcurrency(requestedInterBrokerPartitionMovementConcurrency);
```
hzxa21, line 322: `setRequestedInterBrokerPartitionMovementConcurrency` is a public method in the Executor class. Can it be called elsewhere? My concern is that we only synchronize the methods in ConcurrencyAdjuster but they are calling methods in the outer class. If the outer class method is invoked without going through ConcurrencyAdjuster, there still can be a race condition.

efeg: The public method `Executor#setRequestedInterBrokerPartitionMovementConcurrency` can be called to handle `Admin` endpoint requests, which just calls the `synchronized` `ExecutionTaskManager#setRequestedInterBrokerPartitionMovementConcurrency` method to update the concurrency. Why is this a concern -- or is there some other race condition that I am missing?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
efeg: The controller indeed has a blocking send on a separate thread for each broker -- i.e. https://github.com/linkedin/kafka/blob/2.3-li-1/core/src/main/scala/kafka/controller/ControllerChannelManager.scala#L274. 

In arguably the typical case with more data plane requests (e.g. produce/consume), a backlogged `requestQueue` on the broker side would cause the relevant per-broker controller thread to be blocked  (i.e. controller cannot send out more requests to that broker) for a longer period of time on the broker response w/o the `KIP-291` support.

I believe the scenario you exemplified can happen if the controller receives a burst of control plane requests (i.e. `LeaderAndIsr`, `StopReplica`, or `UpdateMetadata`). This is more likely to happen during a 
1. Demote operation, 
2. PLE, 
3. Replication factor change, or 
4. A leadership transfer phase of the other goal-based CC executions, which are configured with a high "leadership" movement concurrency (i.e. leadership movement concurrency also applies to replica reorder concurrency during the demote operation -- this generates `UpdateMetadata` requests) and/or low "execution progress check" interval. 

Except for the replication factor changes, which should now be addressed with the latest commit, changes introduced in this PR were already _not_ relevant to these operations. Also note that Concurrency Adjuster comes with a `concurrency.adjuster.max.partition.movements.per.broker` config, which puts a bound on the extent of concurrency increase. Hence, even if this becomes an issue, there are means to workaround it.

However, once we want to add support to cover auto-concurrency adjustment for the listed cases above, controller metrics will be more relevant. Hence, I created https://github.com/linkedin/cruise-control/issues/1292 to address those cases.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
hzxa21: Oh I see. I think the missing part is I don't know `ExecutionTaskManager#setRequestedInterBrokerPartitionMovementConcurrency` is already synchronized. Thanks for the explanation. I don't have a concern then.

PR #1302, opened 2020-08-05
Support stop ongoing executions with rollback

### LINKED ISSUE #1195: Support stop ongoing executions with rollback
With Kafka 2.4, [KIP-455](https://cwiki.apache.org/confluence/display/KAFKA/KIP-455%3A+Create+an+Administrative+API+for+Replica+Reassignment) added support to rollback ongoing replica reassignments. The `stop_proposal_execution` endpoint of Cruise Control (CC) should have a rollback mode to support this feature. This will enable CC to cleanly and instantaneously cancel ongoing replica movements and restore the replica set to the one before the reassignment.

efeg:
This PR resolves #1195.
* Changes in this PR are relevant to `migrate_to_kafka_2_4` and `migrate_to_kafka_2_5` branches.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1335,19 +1335,26 @@ public class Executor {
           LOG.debug("Tasks in execution: {}", inExecutionTasks());
         }
         List<ExecutionTask> deadInterBrokerReplicaTasks = new ArrayList<>();
-        List<ExecutionTask> stoppedInterBrokerReplicaTasks = new ArrayList<>();
+        List<ExecutionTask> gracefullyStoppedInterBrokerReplicaTasks = new ArrayList<>();
         List<ExecutionTask> slowTasksToReport = new ArrayList<>();
         boolean shouldReportSlowTasks = _time.milliseconds() - _lastSlowTaskReportingTimeMs > SLOW_TASK_ALERTING_BACKOFF_TIME_MS;
         for (ExecutionTask task : inExecutionTasks()) {
           TopicPartition tp = task.proposal().topicPartition();
-          if (_stopSignal.get() != NO_STOP_EXECUTION) {
-            LOG.debug("Task {} is marked as dead to stop the execution.", task);
+          if (_stopSignal.get() == FORCE_STOP_EXECUTION
+              || (_stopSignal.get() == STOP_EXECUTION && task.type() == INTER_BROKER_REPLICA_ACTION)) {
+            // If the execution is (1) force-stopped or (2) gracefully-stopped during an ongoing inter-broker replica reassignment,
+            // then the executor will not just silently wait for the completion of the current in-progress tasks. For the former
+            // case, the executor will mark all in progress tasks as dead, cleanup the relevant reassignment zNodes, and then bounce
+            // the controller (see ExecutionUtils#deleteZNodesToStopExecution). For the latter case, the executor will mark in
+            // progress tasks as dead, and rollback the ongoing reassignment of gracefully stopped inter-broker replica reassignment.
+            LOG.debug("Task {} is marked as dead to {}.", task,
+                      _stopSignal.get() == FORCE_STOP_EXECUTION ? "force-stop the execution with a controller bounce"
+                                                                : "gracefully-stop the execution with a rollback");
             finishedTasks.add(task);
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1333,16 +1334,21 @@ private int moveLeadershipInBatch() throws InterruptedException, ExecutionExcept
...
-        List<ExecutionTask> deadOrAbortingInterBrokerReplicaActions = new ArrayList<>();
+        List<ExecutionTask> deadInterBrokerReplicaTasks = new ArrayList<>();
+        List<ExecutionTask> stoppedInterBrokerReplicaTasks = new ArrayList<>();
         List<ExecutionTask> slowTasksToReport = new ArrayList<>();
         boolean shouldReportSlowTasks = _time.milliseconds() - _lastSlowTaskReportingTimeMs > SLOW_TASK_ALERTING_BACKOFF_TIME_MS;
         for (ExecutionTask task : inExecutionTasks()) {
           TopicPartition tp = task.proposal().topicPartition();
-          if (_stopSignal.get() == FORCE_STOP_EXECUTION) {
-            LOG.debug("Task {} is marked as dead to force execution to stop.", task);
+          if (_stopSignal.get() != NO_STOP_EXECUTION) {
```
hzxa21, line 1343: Prior to this patch, we act on `STOP_EXECUTION`  in other conditions but we handle that in this condition now. Is it intentional?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1358,42 +1364,76 @@ private int moveLeadershipInBatch() throws InterruptedException, ExecutionExcept
...
+                                                   List<ExecutionTask> stoppedInterBrokerReplicaTasks)
         throws InterruptedException, ExecutionException, TimeoutException {
-      if (!deadOrAbortingInterBrokerReplicaActions.isEmpty()) {
-        ExecutionUtils.submitReplicaReassignmentTasks(_adminClient, deadOrAbortingInterBrokerReplicaActions);
+      List<ExecutionTask> tasksToCancel = new ArrayList<>(deadInterBrokerReplicaTasks);
+      tasksToCancel.addAll(stoppedInterBrokerReplicaTasks);
+
+      if (!tasksToCancel.isEmpty()) {
+        // Cancel/rollback reassignment of dead inter-broker replica tasks.
+        ExecutionUtils.submitReplicaReassignmentTasks(_adminClient, tasksToCancel);
```
hzxa21, line 1411: I am confused about the naming at the very beginning since the comment says "Cancel/rollback ... tasks" but we call `submitReplicaReassignmentTasks` here. When I look at the [implementation ](https://github.com/linkedin/cruise-control/blob/6b7157f8b4b5f77394d4577c3f3c68d623edfe45/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java#L235) I realized that the task will be cancelled if it is in a DEAD state. Are the tasks in `deadInterBrokerReplicaTasks` and `stoppedInterBrokerReplicaTasks` all guaranteed to be in DEAD state?

Maybe add more documentations here to make it clearer?

efeg: I added a sanity check along with `expected to be ...` phrase in the existing `param` in the JavaDoc.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: Prior to this PR, only `FORCE_STOP_EXECUTION` (i.e. a force-stop execution request) had a special handling in this loop. Whereas, `STOP_EXECUTION` did not have any handling -- i.e. all tasks in execution were treated like any other in-progress tasks, and were being tracked until the completion (In pre Kafka 2.4.* versions, the role of `STOP_EXECUTION` was just avoid submitting a new batch of execution and wait for the completion of the ongoing batch).

However, the current logic in the initial version of this patch was incomplete. More specifically, we should handle the rollback for gracefully stopped inter-broker replica reassignments. And also we should avoid running a rollback for force-stopped executions -- i.e. they are already handled without rollback.
I updated the handling logic along with comments to explain the reasoning. Thanks for pointing this out!

PR #1304, opened 2020-08-05
Log uncaught exception in Thread.setDefaultUncaughtExceptionHandler

### LINKED ISSUE #1303: Uncaught exceptions are printed to stderr
Uncaught exceptions are handled by the default uncaught exception handler and printed to stderr, which complicates debugging and investigation. Uncaught exceptions should be logged instead.

urbandan:
This PR resolves #1303.


urbandan:
> I am curious if you can share an example case, where an uncaught exception was printed to `stderr`?

@efeg I've encountered a bug in a custom metrics sampler, which was throwing NPE in some cases - those exceptions were printed to stderr. Clearly, this should be fixed in the sampler, but printing to the log will help a lot in these cases.

PR #1308, opened 2020-08-08
Eliminate inefficiencies in execution process.

### LINKED ISSUE #1307: Eliminate inefficiencies in execution process for CC versions supporting Kafka 2.4+
Cruise Control (CC) adopted the Admin client-based replica reassignment API (see [KIP-455](https://cwiki.apache.org/confluence/display/KAFKA/KIP-455%3A+Create+an+Administrative+API+for+Replica+Reassignment)) for its versions that support `Kafka 2.4+` (e.g. CC `2.4.*` and `2.5.*` under [releases](https://github.com/linkedin/cruise-control/releases)). However, the current execution (in particular, inter-broker replica reassignment) logic has several inefficiencies, including:

* Redundant retrieval of ongoing reassignments in `ExecutionUtils#submitReplicaReassignmentTasks(AdminClient, List<ExecutionTask>)`: This can be dropped to save an admin client call for every new inter-broker replica reassignment batch.
* Lack of processing `AlterPartitionReassignmentsResult` response, which is received upon submitting a new inter-broker replica reassignment batch causes the executor to fail to ensure that the corresponding request has been accepted (e.g. replica reassignment started being processed) on Kafka-side. As a result, the executor can haphazardly submit a redundant re-execution request for the same batch of inter-broker replica reassignment tasks, thinking that they are not in progress. While this does not impact the correctness, it causes extra load on Kafka cluster.
* Use of a common `waitForExecutionTaskToFinish` for all task types causes redundant Admin client calls. For example, every execution of a leadership reassignment retrieves a list of ongoing inter-broker reassignments to compare them with the in progress inter-broker reassignments started by CC. However, the executor would never move to the leadership reassignment phase w/o completing all in progress inter-broker reassignments. As a general rule, no two task types (i.e. `inter-broker`, `intra-broker`, or `leadership`) can execute at the same time.
* With `Kafka 2.4+`, the new graceful stop logic guarantees that there are no in execution tasks to wait after a cancellation/rollback request is processed. Hence, inter-broker replica move logic (i.e. `Executor#interBrokerMoveReplicas`) can be simplified correspondingly.

efeg:
This PR resolves #1307.
-- The correctness of execution has also been verified on a `Kafka 2.4` cluster with and w/o `stop_proposal_execution`.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1016,6 +1016,7 @@ public class Executor {
         }
 
         // 3. Transfer leadership if possible.
+        boolean forceStoppedLeadershipMoves = false;
         if (_executorState.state() == INTRA_BROKER_REPLICA_MOVEMENT_TASK_IN_PROGRESS) {
           _executorState = ExecutorState.operationInProgress(LEADER_MOVEMENT_TASK_IN_PROGRESS,
                                                              _executionTaskManager.getExecutionTasksSummary(
@@ -1028,13 +1029,13 @@ public class Executor {
                                                              _recentlyDemotedBrokers,
                                                              _recentlyRemovedBrokers,
                                                              _isTriggeredByUserRequest);
-          moveLeaderships();
+          forceStoppedLeadershipMoves = moveLeaderships();
           updateOngoingExecutionState();
         }
 
-        // 4. Delete tasks from Zookeeper if needed.
-        if (_executorState.state() == STOPPING_EXECUTION && _stopSignal.get() == FORCE_STOP_EXECUTION) {
-          ExecutionUtils.deleteZNodesToStopExecution(_kafkaZkClient);
+        // 4. Delete leadership tasks from Zookeeper if needed.
+        if (_executorState.state() == STOPPING_EXECUTION && _stopSignal.get() == FORCE_STOP_EXECUTION && forceStoppedLeadershipMoves) {
+          ExecutionUtils.deleteZNodesToForceStopLeadershipMoves(_kafkaZkClient);
         }
       } catch (Throwable t) {
         LOG.error("Executor got exception during execution", t);
@@ -1260,7 +1261,12 @@ public class Executor {
       }
     }
 
-    private void moveLeaderships() {
+    /**
+     * Executes leadership movement tasks.
+     *
+     * @return {@code true} if the leadership movements have been force stopped, {@code false} otherwise.
+     */
+    private boolean moveLeaderships() {
       int numTotalLeadershipMovements = _executionTaskManager.numRemainingLeadershipMovements();
       LOG.info("Starting {} leadership movements.", numTotalLeadershipMovements);
       int numFinishedLeadershipMovements = 0;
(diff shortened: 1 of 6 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1416,6 +1416,120 @@ private int moveLeadershipInBatch() throws InterruptedException, ExecutionExcept
...
+        maybeReexecuteLeadershipTasks();
+        Cluster cluster = getClusterForExecutionProgressCheck();
+
+        List<ExecutionTask> slowTasksToReport = new ArrayList<>();
+        boolean shouldReportSlowTasks = _time.milliseconds() - _lastSlowTaskReportingTimeMs > SLOW_TASK_ALERTING_BACKOFF_TIME_MS;
+        for (ExecutionTask task : inExecutionTasks()) {
+          TopicPartition tp = task.proposal().topicPartition();
+          if (_stopSignal.get() == FORCE_STOP_EXECUTION) {
+            // If the execution is force-stopped, the executor will mark all in progress tasks as dead, cleanup the
+            // relevant reassignment zNodes, and then bounce the controller (see ExecutionUtils#deleteZNodesToStopExecution).
```
hzxa21, line 1453: Is this comment correct? I thought CC no longer deletes the znode and moves controller in order to stop the execution.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1416,6 +1416,120 @@ private int moveLeadershipInBatch() throws InterruptedException, ExecutionExcept
...
+            _executionTaskManager.inExecutionTasks(Collections.singleton(INTRA_BROKER_REPLICA_ACTION)),
+            _adminClient, _config);
+
+        List<ExecutionTask> slowTasksToReport = new ArrayList<>();
+        boolean shouldReportSlowTasks = _time.milliseconds() - _lastSlowTaskReportingTimeMs > SLOW_TASK_ALERTING_BACKOFF_TIME_MS;
+        for (ExecutionTask task : inExecutionTasks()) {
+          TopicPartition tp = task.proposal().topicPartition();
+          if (_stopSignal.get() == FORCE_STOP_EXECUTION) {
+            // If the execution is force-stopped, the executor will mark all in progress tasks as dead, cleanup the
+            // relevant reassignment zNodes, and then bounce the controller (see ExecutionUtils#deleteZNodesToStopExecution).
```
hzxa21, line 1504: Is this comment correct? I thought CC no longer deletes the znode and moves controller in order to stop the execution.

efeg: Please see the response to the other comment above.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: Yes, I believe it is correct.
Leadership reassignments in Kafka use `PreferredReplicaElection` znode; hence, potentially stuck leadership movements may require a force-stop. Though in practice, a stuck / time-consuming leadership movement is not common.

Note that `KIP-455` was about a new API for `inter-broker` replica reassignment tasks. Thus, its impact is limited to changes in starting, stopping, and listing them.
I cleaned up the `force-stop` logic to ensure that it only applies to leadership movements along with a deprecation message for replica reassignments  -- i.e. with `KIP-455` inter-broker replica reassignments no longer need force stop and intra-broker replica reassignments were already not using zNodes.

One side-effect of this change would be the loss of ability to bounce controller upon force-stopping inter/intra broker replica reassignments, which would have helped in case the controller itself is stuck for some reason. We can address that in a future PR by moving intentional controller bounce functionality to `admin` endpoint of cruise control.

PR #1311, opened 2020-08-11
Add support for Kafka 2.6 brokers

### LINKED ISSUE #1309: Current metrics reporter does not work with Kafka 2.6.0
The current `CruiseControlMetricsReporter` does not work on Kafka 2.6.0 clusters. The issue seems to be linked to recent changes in Kafka that changed where the `MetricsRegistry` is obtained from. The default registry now needs to be pulled from the `kafka.metrics.KafkaYammerMetrics` class instead of the `com.yammer.metrics.Metrics` class.

I have a working fix but I was wondering if I should open a PR against the `migrate_to_kafka_2_5` branch or wether it would be better for a committer to create a `migrate_to_kafka_2_6` branch that I can open it against?

tomncooper:
This PR resolves #1309 and allows the Cruise Control Metrics reporter to work with Kafka 2.6 brokers.

This PR also refactors the `ExecutionTask.State` enum into its own separate class. This is needed as switching to the Kafka 2.6 libraries causes the gradle scala compiler to throw cyclical reference errors for this class.

### DIFF
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -10,8 +10,6 @@ import com.linkedin.kafka.cruisecontrol.metricsreporter.metric.MetricsUtils;
 import com.linkedin.kafka.cruisecontrol.metricsreporter.metric.MetricSerde;
 import com.linkedin.kafka.cruisecontrol.metricsreporter.metric.TopicMetric;
 import com.linkedin.kafka.cruisecontrol.metricsreporter.metric.YammerMetricProcessor;
-import com.yammer.metrics.Metrics;
-import com.yammer.metrics.core.Metric;
 
 import java.io.IOException;
 import java.time.Duration;
@@ -26,11 +24,9 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-
-import com.yammer.metrics.core.MetricsRegistry;
 import kafka.log.LogConfig;
-import kafka.metrics.KafkaYammerMetrics;
 import kafka.server.KafkaConfig;
+import kafka.metrics.KafkaYammerMetrics;
 import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AlterConfigOp;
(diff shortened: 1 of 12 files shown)


cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -332,10 +335,30 @@ public void onCompletion(RecordMetadata recordMetadata, Exception e) {
     });
   }
 
+  /**
+   * Starting with Kafka 2.6.0, a new class KafkaYammerMetrics provides the default Metrics Registry. The old default
+   * registry does not work with 2.6+. Therfore if the new class exists, we use it and if it doesn't exist we use the
+   * old one. More details can be found here: https://github.com/apache/kafka/blob/2.6.0/core/src/main/java/kafka/metrics/KafkaYammerMetrics.java
+   *
+   * Once CC supports only 2.6.0 and newer, we can clean this up and use only KafkaYammerMetrics all the time.
```
efeg, line 343: I initially assumed there were significant backwards incompatible changes from 2.5 to 2.6, but if this function would be sufficient to make `migrate_to_kafka_2_5` compatible with `Kafka 2.6`, can we rebase this patch on `migrate_to_kafka_2_5` branch to eliminate the overhead of managing one more branch?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTask.java
```diff
@@ -11,12 +11,10 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
```
efeg, line 14: Nit: Redundant space.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskState.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
```
efeg, line 2: Nit: `2017` -> `2020`

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -332,10 +335,30 @@ public void onCompletion(RecordMetadata recordMetadata, Exception e) {
...
+   *
+   * @return  MetricsRegistry with Kafka metrics
+   */
+  private MetricsRegistry metricsRegistry()   {
+    try {
+      Class.forName("kafka.metrics.KafkaYammerMetrics");
```
amuraru, line 349: This method seems to be called in each cycle of metrics reporting. Can we call the metricsRegistry() method once, cache the regustry and reuse it each time is required?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
tomncooper: I will run some tests to make sure this metrics reporter works ok with 2.5.

tomncooper: From my testing it seems this function works as expected and allows this code to be run on Kafka 2.5 and 2.6 brokers. So I will rebase this against the `migrate_to_kafka_2_5` branch and you can delete the 2.6 branch.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
tomncooper: Good point, I added a field and instantiated it in the `init()` method.

PR #1316, opened 2020-08-17
Added option to configure metrics topic minISR

### LINKED ISSUE #1314: The metrics topic min.insync.replicas is not configurable
Currently the `min.insync.replicas` setting for the metrics topic is not configurable. 

This becomes an issue when the cluster's default `min.insync.replicas` is higher than the metrics topic's number of replicas as the reporter uses `acks=all` (this is a [known issue](https://cwiki.apache.org/confluence/display/KAFKA/KIP-579%3A+new+exception+on+min.insync.replicas+%3E+replication.factor) in Kafka). This is probably not an issue if setting up a Kafka Cluster and CC server manually, as you would make sure that the metrics topic had replicas >= to the cluster`s minISR. 

However, in Strimzi we set up the metric topic for the user and unless they set the metric topic replicas explicitly we set them to 1. So if a user cluster minISR > 1 we get [errors](https://github.com/strimzi/strimzi-kafka-operator/issues/3389).  If we can set the  the topic's minISR we have the flexibility to avoid these errors in various ways.

tomncooper:
This PR resolves #1314 and allows the minimum insync replicas setting of the metrics topic to be configured. 

This is needed to prevent a cluster's default minISR overriding the metric topic which may have a lower replication factor and therefore preventing the `ack=all` of the metrics from ever being satisfied.


tomncooper:
Not sure why the CI has failed? That test passes locally? I don't have permission to rerun the build.

PR #1332, opened 2020-09-26
Support handling planned maintenance events submitted via a topic

### LINKED ISSUE #1248: Add support to handle planned maintenance events submitted via a generic store (e.g a topic)
Maintenance events are admin operations such as adding/removing/demoting brokers, rebalancing the cluster, running a PLE (preferred leader election), or clearing recently removed/demoted brokers. These operations could be triggered via a human, based on a priori knowledge about an upcoming maintenance event, or via Cruise Control self-healing as a result of an anomaly detection.

However, in case a system has knowledge about the upcoming planned maintenance event (e.g. a switch change, upcoming freeze or reboot of VMs on cloud), this information can be conveyed to Cruise Control to automatically start the maintenance event w/o human intervention. These planned maintenance events can be passed through a generic maintenance event store, and an implementation of this event store could use a Kafka topic  -- see [Sample Store](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/SampleStore.java) and [Kafka Sample Store](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/KafkaSampleStore.java) for a relevant use of Kafka topics for fault tolerance of broker and partition load history.

efeg:
This PR resolves #1248.

Known issues:
* If a detected maintenance event cannot be handled due to `AnomalyDetectionStatus` being not `READY`, it will be dropped. The support for stronger handling semantics will be added on later PRs.
* Idempotency support (handling duplicate maintenance events) will be added on later PRs.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -341,6 +341,13 @@ public class KafkaCruiseControlUtils {
    * transient network failures (e.g. Error sending fetch request XXX to node XXX: org.apache.kafka.common.errors.DisconnectException.)
    * that prevents the consumer from getting offset from some brokers for as long as reconnect.backoff.ms.
    *
+   * Note the following assumptions for the given parameters:
+   * <ul>
+   *   <li>if a partition has never been written to, its end offset is expected to be {@code 0}.</li>
+   *   <li>if the earliest offset, whose timestamp is greater than or equal to the given timestamp in the corresponding
+   *   partition does not exist, {@code null} is expected for the offset for the time of the partition.</li>
+   * </ul>
+   *
    * @param endOffsets End offsets retrieved by consumer.
    * @param offsetsForTimes Offsets for times retrieved by consumer.
    */
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -10,26 +10,48 @@ import com.linkedin.kafka.cruisecontrol.analyzer.goals.PreferredLeaderElectionGo
 import com.linkedin.kafka.cruisecontrol.config.EnvConfigProvider;
 import com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig;
 import com.linkedin.kafka.cruisecontrol.config.constants.AnalyzerConfig;
+import com.linkedin.kafka.cruisecontrol.exception.SamplingException;
 import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
 import com.linkedin.kafka.cruisecontrol.monitor.task.LoadMonitorTaskRunner;
+import java.util.Collections;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import kafka.zk.KafkaZkClient;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AdminClientConfig;
+import org.apache.kafka.clients.admin.AlterConfigOp;
+import org.apache.kafka.clients.admin.AlterConfigsResult;
+import org.apache.kafka.clients.admin.Config;
+import org.apache.kafka.clients.admin.ConfigEntry;
+import org.apache.kafka.clients.admin.CreatePartitionsResult;
+import org.apache.kafka.clients.admin.CreateTopicsResult;
+import org.apache.kafka.clients.admin.DescribeConfigsResult;
 import org.apache.kafka.clients.admin.DescribeLogDirsResult;
+import org.apache.kafka.clients.admin.NewPartitions;
+import org.apache.kafka.clients.admin.NewTopic;
+import org.apache.kafka.clients.admin.TopicDescription;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.OffsetAndTimestamp;
 import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.config.ConfigResource;
 import org.apache.kafka.common.config.SaslConfigs;
 import org.apache.kafka.common.config.SslConfigs;
+import org.apache.kafka.common.errors.ReassignmentInProgressException;
+import org.apache.kafka.common.errors.TopicExistsException;
 import org.apache.kafka.common.message.MetadataResponseData;
 import org.apache.kafka.common.record.RecordBatch;
 import org.apache.kafka.common.requests.AbstractResponse;
 import org.apache.kafka.common.requests.MetadataResponse;
 import org.apache.kafka.common.security.auth.SecurityProtocol;
 import org.apache.kafka.common.utils.SystemTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import scala.Option;
 
 import java.io.FileInputStream;
@@ -49,12 +71,15 @@ import java.util.TimeZone;
 import java.util.stream.Collectors;
 
 import static com.linkedin.kafka.cruisecontrol.servlet.parameters.ParameterUtils.SKIP_HARD_GOAL_CHECK_PARAM;
+import static kafka.log.LogConfig.CleanupPolicyProp;
+import static kafka.log.LogConfig.RetentionMsProp;
 
 
 /**
  * Util class for convenience.
  */
 public class KafkaCruiseControlUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(KafkaCruiseControlUtils.class);
   public static final double MAX_BALANCEDNESS_SCORE = 100.0;
   public static final int ZK_SESSION_TIMEOUT = 120000;
   public static final int ZK_CONNECTION_TIMEOUT = 120000;
@@ -72,6 +97,8 @@ public class KafkaCruiseControlUtils {
   public static final int REQUEST_VERSION_UPDATE = -1;
   public static final String ENV_CONFIG_PROVIDER_NAME = "env";
   public static final String ENV_CONFIG_PROVIDER_CLASS_CONFIG = ".env.class";
+  public static final long CLIENT_REQUEST_TIMEOUT_MS = 30000L;
+  public static final String DEFAULT_CLEANUP_POLICY = "delete";
 
   private KafkaCruiseControlUtils() {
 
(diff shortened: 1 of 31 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AddBrokerPlan.java
```diff
@@ -0,0 +1,85 @@
...
+  public byte planVersion() {
+    return PLAN_VERSION;
+  }
+
+  public Set<Integer> brokers() {
+    return _brokers;
+  }
+
+  @Override
+  public ByteBuffer toBuffer(int headerSize) {
```
smccauliff, line 39: Do you want to add a magic number and/or CRC to detect corruption?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
```diff
@@ -28,10 +38,42 @@
...
+      bootstrapServers = bootstrapServers.substring(1, bootstrapServers.length() - 1);
+    }
+
+    // Create consumer
+    long randomToken = RANDOM.nextLong();
+    Properties consumerProps = new Properties();
+    consumerProps.putAll(configs);
+    consumerProps.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+    consumerProps.setProperty(ConsumerConfig.GROUP_ID_CONFIG, groupIdPrefix + randomToken);
+    consumerProps.setProperty(ConsumerConfig.CLIENT_ID_CONFIG, groupIdPrefix + "-consumer-" + randomToken);
```
smccauliff, line 67: Randomly generating the group id makes it impossible to track offset commits using lagmontor.  Randomly generating the client id breaks quotas based on client ids.  Also each time there is a new client id a new set of metrics are generated on both the broker and the client itself.  Since the brokers are probably longer lived than the clients this results is accumulated garbage.  Open source kafka does periodically clean it out but for some reason we've opted to keep them around in memory forever.  Recommend never having random numbers be part of the group or client ids.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
```diff
@@ -28,10 +38,42 @@
...
+   * Create a Kafka consumer for retrieving reported Maintenance plans.
+   * The consumer uses {@link String} for keys and {@link MaintenancePlan} for values.
+   *
+   * @param configs The configurations for Cruise Control.
+   * @return A new Kafka consumer
+   */
+  public static Consumer<String, MaintenancePlan> createMaintenanceEventConsumer(Map<String, ?> configs, String groupIdPrefix) {
+    // Get bootstrap servers
+    String bootstrapServers = configs.get(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG).toString();
+    // Trim the brackets in List's String representation.
```
smccauliff, line 56: The output of this method is not guaranteed to be stable across Java versions.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventType.java
```diff
@@ -23,7 +23,38 @@
...
+  // Do not change the order of enums. Append new ones to the end.
+  ADD_BROKER((byte) 0),
+  REMOVE_BROKER((byte) 1),
+  FIX_OFFLINE_REPLICAS((byte) 2),
+  REBALANCE((byte) 3),
+  DEMOTE_BROKER((byte) 4),
+  TOPIC_REPLICATION_FACTOR((byte) 5);
+
+  // This id helps with serialization and deserialization of event types
+  private final byte _id;
```
smccauliff, line 35: I've always thought this was strange as this can be implemented as
```Java
  byte id() {
    return (byte) ordinal();
  }
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenancePlan.java
```diff
@@ -0,0 +1,75 @@
...
+   */
+  public abstract byte planVersion();
+
+  /**
+   * Serialize the maintenance plan to a byte buffer with the header size reserved.
+   *
+   * @param headerSize the header size to reserve.
+   * @return A {@link ByteBuffer} with header size reserved at the beginning.
+   */
+  public ByteBuffer toBuffer(int headerSize) {
```
smccauliff, line 59: Seems like this protocol is going to have many different types of messages, but low QPS.  Seems like a good use case for some kind of serialization library.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/SamplingUtils.java
```diff
@@ -346,124 +315,70 @@ private static boolean skipBuildingBrokerMetricSample(BrokerLoad brokerLoad, int
...
-  private static void maybeUpdateConfig(Set<AlterConfigOp> configsToAlter, Map<String, String> desiredConfig, Config currentConfig) {
-    for (Map.Entry<String, String> entry : desiredConfig.entrySet()) {
-      String configName = entry.getKey();
-      String targetConfigValue = entry.getValue();
-      ConfigEntry currentConfigEntry = currentConfig.get(configName);
-      if (currentConfigEntry == null || !currentConfigEntry.value().equals(targetConfigValue)) {
-        configsToAlter.add(new AlterConfigOp(new ConfigEntry(configName, targetConfigValue), AlterConfigOp.OpType.SET));
-      }
+    // Get group id
+    long randomToken = RANDOM.nextLong();
```
smccauliff, line 331: See comment about group id

docs/wiki/User Guide/Configurations.md
```diff
@@ -245,6 +246,14 @@ We are still trying to improve cruise control. And following are some configurat
...
 | skip.sample.store.topic.rack.awareness.check          | Boolean | N         | false         | The config to skip rack awareness sanity check for sample store topics                                                                                                                                  |
 
+### MaintenanceEventTopicReader configurations
+| Name                                          | Type    | Required? | Default Value           | Description                                                       |
+|-----------------------------------------------|---------|-----------|-------------------------|-------------------------------------------------------------------|
+| maintenance.event.topic                       | String  | N         | __MaintenanceEvent      | The name of the Kafka topic to consume maintenance events from    |
+| maintenance.event.topic.replication.factor    | Short   | N         | min(2, #alive-brokers)  | The replication factor of the maintenance event topic             |
+| maintenance.event.topic.partition.count       | Integer | N         | 32                      | The partition count of the maintenance event topic                |
```
smccauliff, line 254: This can probably be a much smaller topic.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
smccauliff: ArrayList.toString() has a specification, but that's not true for all implementations.

smccauliff:
I've looked at about 1/2 of the changes.  I will take another look at the remaining changes tomorrow.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -168,6 +195,194 @@ public static String getRequiredConfig(Map<String, ?> configs, String configName
...
   }
 
+  /**
+   * Creates the given topic if it does not exist.
+   *
+   * @param adminClient The adminClient to send createTopics request.
+   * @param topicToBeCreated A wrapper around the topic to be created.
+   * @return {@code false} if the topic to be created already exists, {@code true} otherwise.
+   */
+  public static boolean createTopic(AdminClient adminClient, NewTopic topicToBeCreated) {
```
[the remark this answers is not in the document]
efeg: In Cruise Control, components that create topics are already pluggable. Hence, it abstracts away the topic creation and update logic on the relevant caller end (e.g. `SampleStore`, `MaintenanceEventReader`).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -168,6 +195,194 @@ public static String getRequiredConfig(Map<String, ?> configs, String configName
...
+   *
+   * @param endOffsets End offsets retrieved by consumer.
+   * @param offsetsForTimes Offsets for times retrieved by consumer.
+   */
+  public static void sanityCheckOffsetFetch(Map<TopicPartition, Long> endOffsets,
+                                            Map<TopicPartition, OffsetAndTimestamp> offsetsForTimes)
+      throws SamplingException {
+    Set<TopicPartition> failedToFetchOffsets = new HashSet<>();
+    for (Map.Entry<TopicPartition, OffsetAndTimestamp> entry : offsetsForTimes.entrySet()) {
+      if (entry.getValue() == null && endOffsets.get(entry.getKey()) == null) {
```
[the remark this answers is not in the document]
efeg: * If a partition has never been written to, its end offset is expected to be`0`. (see `KafkaConsumer#endOffsets(Collection)`).
* If earliest offset whose timestamp is greater than or equal to the given timestamp in the corresponding partition does not exist, `null` is expected to be returned for the partition. (see `KafkaConsumer#offsetsForTimes(Map)`).

Consequently, this sanity check will succeed. -- updated the JavaDoc to indicate these parameter assumptions.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AddBrokerPlan.java
efeg: I believe we already have a `magic number` for each plan (please see `MaintenancePlan#planVersion()`) -- am I missing something?
-- Added `crc` to verify the integrity -- thanks for the suggestion!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
efeg: TL;DR Dropping group ids, but keeping random client ids.
_--_
> Randomly generating the group id makes it impossible to track offset commits using lagmontor

This consumer uses neither the group management functionality by using `subscribe(topic)` nor the Kafka-based offset management strategy. Hence, the `group.id` config is irrelevant. (Added a clarification to JavaDoc and dropped this setting).

> Randomly generating the client id ...

I see your point, but making this config deterministic is tricky:

If multiple CC instances point to the same cluster, which is a valid usage today (i.e. one active and multiple passive CC instances, for which all self-healing is disabled and `ReadOnlySampleStore` is used), then this will result in duplicate sensors for different clients. I will also share example internal issues caused by using the same client id for different consumers.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
efeg: I believe we discussed a similar issue under a different PR: https://github.com/linkedin/cruise-control/pull/1223#discussion_r433685111

It is true that the `List` interface itself does not enforce a specific `toString()` response format, but I don't suppose we expect to use / support collections that are not based on the standard https://docs.oracle.com/javase/7/docs/api/java/util/AbstractCollection.html#toString(), whose openJDK implementation for `toString()` hasn't changed from 8 to 11?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventType.java
efeg: This is indeed cleaner -- thanks.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenancePlan.java
efeg: Can you comment a little more on the advantages of using a serialization library as opposed to the current approach?
I feel that we may not get the same level of efficiency and may need to require users to have an extra dependency on their end -- and potentially on our end depending on the selected library.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/SamplingUtils.java
efeg: Addressed this similar to https://github.com/linkedin/cruise-control/pull/1332#discussion_r496353768.

docs/wiki/User Guide/Configurations.md
efeg: Changed this to `8`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
smccauliff: There's usually some i00X number used to identify different instances, this can be part of the client id.  Other people can have other solutions to this problem.  Host name?

So CC never commits offsets into Kafka?  The offset management and the consumer group coordination can be used independently from each other.

smccauliff: The previous code review was based on emitting the output in log messages.  I guess it's fine, but it seems strange because it's already a List it's possible to do the following in single line.
```java
  list.stream().collect(Collectors.joining( "," ) )
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
smccauliff: The previous code review was based on emitting the output in log messages.  I guess it's fine, but it seems strange because it's already a List it's possible to do the following in single line.
```java
  list.stream().collect(Collectors.joining( "," ) )
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
smccauliff: The i00X string can be added to different cruise control instances.  For those outside of LinkedIn their deployment system may have something like this or they can add hostname or something else.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenancePlan.java
smccauliff: You probably won't gain any efficiency.  Maybe protobuf since the integers are variable length encoded.  You gain maintainability.   Since this is low qps there's no need for efficiency here.  Gson looks like it is easy to use and has minimal dependencies. The real win is not when you write the initial code it's when you need to change the code: add additional messages, add additional fields, need to reason about versioning, need to debug messages by actually reading their contents.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReader.java
```diff
@@ -0,0 +1,356 @@
...
+            LOG.warn("Cannot parse record, please update your Cruise Control version.");
+            continue;
+          }
+
+          long planGenerationTimeMs = record.value().timeMs();
+          if (planGenerationTimeMs + _maintenancePlanExpirationMs < eventReadPeriodEndMs) {
+            LOG.warn("Discarding the expired plan {}. (Expired: {} Evaluated: {}).", record.value(),
+                     planGenerationTimeMs + _maintenancePlanExpirationMs, eventReadPeriodEndMs);
+          } else if (planGenerationTimeMs >= eventReadPeriodEndMs) {
+            TopicPartition tp = new TopicPartition(record.topic(), record.partition());
```
smccauliff, line 192: Not all the producers will have the same clock and so time is not monotonically increasing.  Does this still work if that is the case?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/RebalancePlan.java
```diff
@@ -0,0 +1,47 @@
...
+    verifyCrc(headerSize, buffer);
+    byte version = buffer.get();
+    if (version > PLAN_VERSION) {
+      throw new UnknownVersionException("Cannot deserialize the plan for version " + version + ". Current version: " + PLAN_VERSION);
+    }
+
+    long timeMs = buffer.getLong();
+    int brokerId = buffer.getInt();
+
+    return new RebalancePlan(timeMs, brokerId);
```
smccauliff, line 40: What does it mean to rebalance a particular broker?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/TopicReplicationFactorPlan.java
```diff
@@ -0,0 +1,111 @@
...
+
+    long timeMs = buffer.getLong();
+    int brokerId = buffer.getInt();
+    byte numRFUpdateEntries = buffer.get();
+    Map<Short, String> topicRegexWithRFUpdate = new HashMap<>(numRFUpdateEntries);
+
+    for (int i = 0; i < numRFUpdateEntries; i++) {
+      short replicationFactor = buffer.getShort();
+      int regexLength = buffer.getInt();
+      String regex = new String(buffer.array(), buffer.arrayOffset() + buffer.position(), regexLength, StandardCharsets.UTF_8);
```
smccauliff, line 98: Do you want to validate the regex here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/SamplingUtils.java
```diff
@@ -346,124 +313,70 @@ private static boolean skipBuildingBrokerMetricSample(BrokerLoad brokerLoad, int
...
-      if (currentConfigEntry == null || !currentConfigEntry.value().equals(targetConfigValue)) {
-        configsToAlter.add(new AlterConfigOp(new ConfigEntry(configName, targetConfigValue), AlterConfigOp.OpType.SET));
-      }
+  private static String bootstrapServers(Map<String, ?> configs) {
+    String bootstrapServers = configs.get(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG).toString();
+    // Trim the brackets in List's String representation.
+    if (bootstrapServers.length() > 2) {
+      bootstrapServers = bootstrapServers.substring(1, bootstrapServers.length() - 1);
     }
+    return bootstrapServers;
```
smccauliff, line 354: Is this duplicated elsewhere?

smccauliff (commented):
I just need to read the code for the unit tests.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
efeg: > There's usually some i00X number used to identify different instances, this can be part of the client id. Other people can have other solutions to this problem. Host name?

> The i00X string can be added to different cruise control instances. For those outside of LinkedIn their deployment system may have something like this or they can add hostname or something else.

Even if a local (i.e. dev) deployment has a `i00X` string, I suspect that this would not suffice to distinguish clients for multiple local deployments pointing to _randomly configured_ (i.e. different) destination clusters (e.g. for testing).
If possible, I would prefer a solution that is applicable to both internal and open source. I believe there are at least 3 pieces that should go into the client id: (1) source host, (2) destination cluster, and (3) whether CC is the active instance or a passive one (and if passive, which specific passive instance).

Presumably, this change is not directly relevant to this PR other than continue using the existing approach. I created an issue about this discussion and propose to address this on a later patch: https://github.com/linkedin/cruise-control/issues/1337

> So CC never commits offsets into Kafka? The offset management and the consumer group coordination can be used independently from each other.

That is correct.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
efeg: I propose the following, which simplifies the statement w/o using the stream API.

```
    String bootstrapServers = String.join(",", (List<String>) configs.get(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG));
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReader.java
```diff
@@ -0,0 +1,356 @@
...
+    Map<TopicPartition, OffsetAndTimestamp> offsetsForTimes = _consumer.offsetsForTimes(timestampToSeek);
+    sanityCheckOffsetFetch(endOffsets, offsetsForTimes);
+
+    // If offsets for times are provided for a partition, seek to the returned offset. Otherwise, i.e. for partitions
+    // without a record timestamp greater than or equal to the target timestamp, seek to the end offset.
+    for (Map.Entry<TopicPartition, OffsetAndTimestamp> entry : offsetsForTimes.entrySet()) {
+      TopicPartition tp = entry.getKey();
+      OffsetAndTimestamp offsetAndTimestamp = entry.getValue();
+      _consumer.seek(tp, offsetAndTimestamp != null ? offsetAndTimestamp.offset() : endOffsets.get(tp));
+    }
```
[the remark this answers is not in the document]
efeg: I assume by `this way` you mean the  _manual_ management of offsets?

The first iteration of CC used subscribe-based group management for its consumers -- i.e. both for sample store consumers and the metrics reporter sampler. However, we observed frequent consumer rebalances typically under unhealthy cluster state. This overhead has adversely affected CC, which is supposed to be robust under unhealthy cluster state; hence, we manage offsets ourselves.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReader.java
```diff
@@ -0,0 +1,356 @@
...
+ * </ul>
+ */
+public class MaintenanceEventTopicReader implements MaintenanceEventReader {
+  private static final Logger LOG = LoggerFactory.getLogger(MaintenanceEventTopicReader.class);
+  protected String _maintenanceEventTopic;
+  protected Consumer<String, MaintenancePlan> _consumer;
+  protected Set<TopicPartition> _currentPartitionAssignment;
+  protected volatile boolean _shutdown = false;
+  protected long _lastEventReadPeriodEndTimeMs;
+  protected KafkaCruiseControl _kafkaCruiseControl;
```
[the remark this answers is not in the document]
efeg: This class does not just retrieve the time and config from the `KafkaCruiseControl` object (i.e. as you mentioned, referencing them via their corresponding objects/interfaces would be trivial), but it also passes this config to `MaintenanceEvent` upon its creation for its API to be used by the underlying `OperationRunnable` objects. Going forward, we can further polish the existing code to abstract away the relevant logic that is used by them from `KafkaCruiseControl`; however, I feel that this is orthogonal to the goal of this PR.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReader.java
efeg: As (arguably) typically assumed across the participants of Kafka ecosystem, we also assume clocks are _loosely synchronized_. For example, we know that secure communication between brokers and clients can fail due to certificate expiration issues caused by clock skew. Similarly, if a producer has a clock that is significantly behind the others, then its plan can expire "faster" than it expects.
Please also see the comment above, which discusses that the delay can be introduced by Kafka producers:
```
  // A maintenance event has a certain validity period after which it expires and becomes invalid. A delay in handling
  // could be introduced by the Kafka producer that generates the plan, the consumer that retrieves the plan, or the
  // network that connects these clients to the Kafka cluster. Maintenance event topic reader discards the expired events.
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/RebalancePlan.java
efeg: This is to audit provenance of the maintenance plan.
Excerpt from JavaDoc of `MaintenancePlan#brokerId()`: `The id of the broker that reported this maintenance event.`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/TopicReplicationFactorPlan.java
efeg: The validity of the expression's syntax is verified in `MaintenanceEvent#topicPatternByReplicationFactor(Map<String, ?>)` -- i.e. upon creation of a `MaintenanceEvent` from `TopicReplicationFactorPlan`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/SamplingUtils.java
efeg: Moved the bootstrapServers extraction to a util function.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReader.java
smccauliff: I saw that comment.  My point is that this introduces additional latency and that one broker with a bad clock can block all maintenance events from being processed.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/RebalancePlan.java
smccauliff: Got it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/TopicReplicationFactorPlan.java
smccauliff: That's assuming your code is going to be generating these events which it may not be.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenancePlan.java
efeg: Makes sense -- updated the code to use Gson for de/serialization of maintenance plans.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReader.java
efeg: > one broker with a bad clock can block all maintenance events from being processed.

I am confused. Unless that broker is the _only_ one that produces maintenance events to the topic, why would all maintenance events be blocked?

efeg: > one broker with a bad clock can block all maintenance events from being processed.

I am a little confused. Unless that broker is either (1) the only one in the system that generates maintenance events or (2) generates maintenance events every `MaintenanceEventDetector#READ_EVENTS_TIMEOUT`, it will not block the other maintenance events. Am I missing something?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/TopicReplicationFactorPlan.java
efeg: Yes, because the semantic interpretation of this String, called `regex`, is relevant to `MaintenanceEvent`.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReaderTest.java
```diff
@@ -0,0 +1,213 @@
...
+  private void retrieveLatestMetadata() throws InterruptedException, ExecutionException {
+    TopicDescription description0;
+    TopicDescription description1;
+    Config topicConfig0;
+    Config topicConfig1;
+    ConfigResource topicResource = new ConfigResource(ConfigResource.Type.TOPIC, TEST_TOPIC);
+    AdminClient adminClient0 = KafkaCruiseControlUtils.createAdminClient(Collections.singletonMap(
+        AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, broker(0).plaintextAddr()));
+    AdminClient adminClient1 = KafkaCruiseControlUtils.createAdminClient(Collections.singletonMap(
+        AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, broker(1).plaintextAddr()));
```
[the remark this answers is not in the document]
efeg: If my understanding is correct, they will prefer a node with an existing connection, which will be the one specified as the bootstrap server.

In practice, not using this trick definitely leads to reproducible failure with sufficient repetition of this test case.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/TopicReplicationFactorPlan.java
```diff
@@ -0,0 +1,111 @@
...
+package com.linkedin.kafka.cruisecontrol.detector;
+
+import com.linkedin.kafka.cruisecontrol.metricsreporter.exception.UnknownVersionException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+
+
+public class TopicReplicationFactorPlan extends MaintenancePlan {
```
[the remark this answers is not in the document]
efeg: Added JavaDoc to each plan to clarify its purpose.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReader.java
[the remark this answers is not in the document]
smccauliff: Assign can be used with offset commit without needing to do consumer group coordination.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/TopicReplicationFactorPlan.java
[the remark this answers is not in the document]
smccauliff: This was the only one which was confusing to me, but thanks for updating everything.

smccauliff:
Did you decided to use GSON  and then changed your mind and consolidate the binary serialization code instead?

efeg:
>Did you decided to use GSON and then changed your mind and consolidate the binary serialization code instead?

@smccauliff Please see [MaintenancePlanSerde.java](https://github.com/linkedin/cruise-control/pull/1332/files#diff-96eadac9cc3f69e9a0d00bdf0cefb0d3), which is the GSON-based de/serializer.

-- i.e. updates after the commit, titled [Use Gson for de/serialization of maintenance plans](https://github.com/linkedin/cruise-control/pull/1332/commits/5b394034cd2064f253e70cfc74b4f699cffd6a68), are using GSON for serialization and deserialization. Is it possible that maybe you looked at an earlier commit?

PR #1352, opened 2020-10-16
Check entries value in partition load query to disallow negative values (#1351)

### LINKED ISSUE #1351: Query partition resource utilization should not allow "entries" parameter with negative values
Users can query partition resource utilization with the `entries` parameter. For example:

`GET http://<Cruise Control host>:<port>/kafkacruisecontrol/partition_load?entries=100`

If the `entries` parameter has negative values, for example `partition_load?entries=-123`, it would be treated as if it's 0. We should make it more strict by disallowing negative values and returning 4XX error code to indicate a bad request.

Lincong:
This PR resolves #1351

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
@@ -83,7 +83,7 @@ public class PartitionLoadParameters extends AbstractParameters {
     try {
       _resource = Resource.valueOf(resourceString.toUpperCase());
     } catch (IllegalArgumentException iae) {
-      throw new UserRequestException(String.format("Invalid resource type %s. The resource type must be one of the"
+      throw new UserRequestException(String.format("Invalid resource type %s. The resource type must be one of the "
                                                    + "following: CPU, DISK, NW_IN, NW_OUT", resourceString));
     }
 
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
```diff
@@ -97,10 +97,14 @@ protected void initParameters() throws UnsupportedEncodingException {
     _endMs = ParameterUtils.endMs(_request);
     _partitionLowerBoundary = ParameterUtils.partitionBoundary(_request, false);
     _partitionUpperBoundary = ParameterUtils.partitionBoundary(_request, true);
-    _entries = ParameterUtils.entries(_request);
     _minValidPartitionRatio = ParameterUtils.minValidPartitionRatio(_request);
     _allowCapacityEstimation = ParameterUtils.allowCapacityEstimation(_request);
     _brokerIds = ParameterUtils.brokerIds(_request, true);
+    _entries = ParameterUtils.entries(_request);
```
efeg, line 103: Looks like `ParameterUtils #entries(HttpServletRequest)` is used by other callers. Would it ever be acceptable to get a negative value for entries? If not, should we move the sanity check to `ParameterUtils #entries(HttpServletRequest)`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
```diff
@@ -97,10 +97,14 @@ protected void initParameters() throws UnsupportedEncodingException {
...
+    if (_entries < 0) {
+      throw new UserRequestException("Invalid entries value. The entries value must be"
+          + " greater than or equal to 0. Got negative value: " + _entries);
```
efeg, line 106: Nit: Can we make this error message more concise? -- e.g. something like
```
throw new UserRequestException("The requested entries must be positive (Requested: " + _entries + ").");
```

efeg (commented):
@Lincong thanks for the patch!
Left some comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
Lincong: That's a good idea. All `entries` should have positive values. I will move the check to `ParameterUtils`. Thanks

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/PartitionLoadParameters.java
Lincong: Good idea!

efeg (approved):
LGTM!

PR #1357, opened 2020-10-16
Add check to validate that time range start time is smaller than end time

### LINKED ISSUE #1355: Incorrect exception when query partition load with invalid start/end time range
Users can query partition load with a time range. For example:

`GET http://<Cruise Control host>:<port>/kafkacruisecontrol/partition_load?start=1602565200000&end=1602561600000`

In the above example, `start_time` >  `end_time`. However, the request gets the below response:

`com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException: java.util.NoSuchElementException`

The error message should clearly state that `"Invalid time range specified: the start time should not be greater than the end time."`

Lincong:
This PR resolves #1355.

Notes for code reviewer:

This PR adds the below methods `ParameterUtils.endMsOrDefault` and `ParameterUtils.startMsOrDefault`. The reason is that it should be up to the caller to decide the default values for start time and end time. Depending on which `AbstractParameters` class parses the request, they might have different default values for start time and/or end time.

The original `startMs` and `endMs` returns null if no start/end time is set in the request.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
@@ -394,13 +394,7 @@ public class ParameterUtils {
   }
 
   static Long replicationThrottle(HttpServletRequest request, KafkaCruiseControlConfig config) {
-    String parameterString = caseSensitiveParameterName(request.getParameterMap(), REPLICATION_THROTTLE_PARAM);
-    Long value;
-    if (parameterString == null) {
-      value = config.getLong(ExecutorConfig.DEFAULT_REPLICATION_THROTTLE_CONFIG);
-    } else {
-      value = Long.parseLong(request.getParameter(parameterString));
-    }
+    Long value = getLongParam(request, REPLICATION_THROTTLE_PARAM, config.getLong(ExecutorConfig.DEFAULT_REPLICATION_THROTTLE_CONFIG));
     if (value != null && value < 0) {
       throw new UserRequestException(String.format("Requested rebalance throttle must be non-negative (Requested: %s).", value));
     }
@@ -826,8 +820,7 @@ public class ParameterUtils {
    * @return Execution progress check interval in milliseconds.
    */
   static Long executionProgressCheckIntervalMs(HttpServletRequest request) {
-    String parameterString = caseSensitiveParameterName(request.getParameterMap(), EXECUTION_PROGRESS_CHECK_INTERVAL_MS_PARAM);
-    return parameterString == null ? null : Long.parseLong(request.getParameter(parameterString));
+    return getLongParam(request, EXECUTION_PROGRESS_CHECK_INTERVAL_MS_PARAM, null);
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
@@ -248,6 +248,23 @@ public class ParameterUtils {
   }
 
   /**
+   * Get the long parameter parameter.
+   *
+   * @param request HTTP request received by Cruise Control.
+   * @param parameter Parameter to parse from the request.
+   * @param defaultIfMissing Default value to set if the request does not contain the parameter.
+   * @return The specified value for the parameter, or defaultIfMissing if the parameter is missing.
+   */
+  public static Long getLongParam(HttpServletRequest request, String parameter, @Nullable Long defaultIfMissing) {
+    String parameterString = caseSensitiveParameterName(request.getParameterMap(), parameter);
+
+    if (parameterString == null) {
+      return defaultIfMissing;
+    }
+    return Long.parseLong(request.getParameter(parameterString));
+  }
+
+  /**
    * Get the {@link List} parameter.
    *
    * @param request HTTP request received by Cruise Control.
@@ -409,24 +426,12 @@ public class ParameterUtils {
     return timeString.toUpperCase().equals("NOW") ? System.currentTimeMillis() : Long.parseLong(timeString);
   }
 
-  @Nullable
-  static Long startMs(HttpServletRequest request) {
-    return parseParamAsLong(request, START_MS_PARAM);
+  static Long startMsOrDefault(HttpServletRequest request, @Nullable Long defaultIfMissing) {
+    return getLongParam(request, START_MS_PARAM, defaultIfMissing);
   }
 
-  @Nullable
-  static Long endMs(HttpServletRequest request) {
-    return parseParamAsLong(request, END_MS_PARAM);
-  }
-
-  static long startMsOrDefault(HttpServletRequest request, long defaultStartMs) {
-    Long startMs = parseParamAsLong(request, START_MS_PARAM);
-    return startMs == null ? defaultStartMs : startMs;
-  }
-
-  static long endMsOrDefault(HttpServletRequest request, long defaultEndMs) {
-    Long endMs = parseParamAsLong(request, END_MS_PARAM);
-    return endMs == null ? defaultEndMs : endMs;
+  static Long endMsOrDefault(HttpServletRequest request, @Nullable Long defaultIfMissing) {
+    return getLongParam(request, END_MS_PARAM, defaultIfMissing);
   }
 
   static void validateTimeRange(long startMs, long endMs) {
@@ -1072,12 +1077,6 @@ public class ParameterUtils {
 
   static boolean fetchCompletedTask(HttpServletRequest request) {
     return getBooleanParam(request, FETCH_COMPLETED_TASK_PARAM, false);
-  }
-
-  @Nullable
-  private static Long parseParamAsLong(HttpServletRequest request, String paramName) {
-    String parameterString = caseSensitiveParameterName(request.getParameterMap(), paramName);
-    return parameterString == null ? null : Long.parseLong(request.getParameter(parameterString));
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
@@ -257,11 +257,7 @@ public class ParameterUtils {
    */
   public static Long getLongParam(HttpServletRequest request, String parameter, @Nullable Long defaultIfMissing) {
     String parameterString = caseSensitiveParameterName(request.getParameterMap(), parameter);
-
-    if (parameterString == null) {
-      return defaultIfMissing;
-    }
-    return Long.parseLong(request.getParameter(parameterString));
+    return parameterString == null ? defaultIfMissing : Long.valueOf(request.getParameter(parameterString));
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
@@ -24,6 +24,7 @@ import com.linkedin.kafka.cruisecontrol.servlet.UserTaskManager;
 import com.linkedin.kafka.cruisecontrol.servlet.purgatory.ReviewStatus;
 import com.linkedin.kafka.cruisecontrol.servlet.response.CruiseControlState;
 
+import javax.annotation.Nullable;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -408,16 +409,31 @@ static Long time(HttpServletRequest request) {
...
   }
 
-  static long endMs(HttpServletRequest request) {
-    String parameterString = caseSensitiveParameterName(request.getParameterMap(), END_MS_PARAM);
-    return parameterString == null ? System.currentTimeMillis()
-                                   : Long.parseLong(request.getParameter(parameterString));
+  @Nullable
+  static Long endMs(HttpServletRequest request) {
+    return parseParamAsLong(request, END_MS_PARAM);
+  }
```
efeg, line 420: Having separate functions for `startMs` and `startMsOrDefault` (same issue for `end`) seems a little redundant (e.g. see code repetition for `parseParamAsLong(request, START_MS_PARAM)` between these functions). 

I wonder if we can make it more concise by dropping `endMs` and `startMs` functions. Their callers can call `startMsOrDefault` and `endMsOrDefault` with `null` as the default parameter.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -408,16 +409,31 @@ static Long time(HttpServletRequest request) {
...
+
+  static long startMsOrDefault(HttpServletRequest request, long defaultStartMs) {
+    Long startMs = parseParamAsLong(request, START_MS_PARAM);
+    return startMs == null ? defaultStartMs : startMs;
+  }
+
+  static long endMsOrDefault(HttpServletRequest request, long defaultEndMs) {
```
efeg, line 427: 1. For consistency of the parameter name for missing default values `defaultEndMs` -> `defaultIfMissing`. 
2. Also, can we make the return type and the type of the default parameter `Long`? Clearly, if the response is `null`, it would only be possible via a `null` default parameter passed to this function.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
```
efeg, line 2: `2017` -> `2020` (i.e. date of creation)

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -0,0 +1,68 @@
...
+    HttpServletRequest mockRequest = EasyMock.mock(HttpServletRequest.class);
+    String startTimeString = "12345";
+    String endTimeString = "23456";
+    Map<String, String[]> paramMap = new HashMap<>();
+    paramMap.put(ParameterUtils.START_MS_PARAM, new String[]{ParameterUtils.START_MS_PARAM});
+    paramMap.put(ParameterUtils.END_MS_PARAM, new String[]{ParameterUtils.END_MS_PARAM});
+
+    EasyMock.expect(mockRequest.getParameterMap()).andReturn(paramMap).anyTimes();
+    EasyMock.expect(mockRequest.getParameter(ParameterUtils.START_MS_PARAM)).andReturn(startTimeString).anyTimes();
+    EasyMock.expect(mockRequest.getParameter(ParameterUtils.END_MS_PARAM)).andReturn(endTimeString).anyTimes();
```
efeg, line 32: Is it acceptable for these mocked responses to be called `anyTimes`? Is it possible to be more specific?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -0,0 +1,68 @@
...
+    EasyMock.replay(mockRequest);
+
+    Long startMs = ParameterUtils.startMs(mockRequest);
+    Long endMs = ParameterUtils.endMs(mockRequest);
+    Assert.assertNotNull(startMs);
+    Assert.assertNotNull(endMs);
+    Assert.assertEquals((Long) Long.parseLong(startTimeString), startMs);
+    Assert.assertEquals((Long) Long.parseLong(endTimeString), endMs);
```
efeg, line 40: Why don't we use the `.longValue()` of start and endMs?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -0,0 +1,68 @@
...
+public class ParameterUtilsTest {
+
+  private static final Long DEFAULT_START_TIME_MS = -1L;
+  private static final Long DEFAULT_END_TIME_MS = -2L;
+
+  @Test
+  public void testParseTimeRangeSet() {
+    HttpServletRequest mockRequest = EasyMock.mock(HttpServletRequest.class);
+    String startTimeString = "12345";
+    String endTimeString = "23456";
```
efeg, line 25: Can we avoid hardcoding test values and move them to static variables?

efeg (commented):
@Lincong thanks for the patch -- left some comments

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
Lincong: OK, that makes sense

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
Lincong: I changed it to:

`Assert.assertEquals(Long.parseLong(START_TIME_STRING), startMs.longValue());`

Is the reason why you suggested to use `longValue()` to avoid casting?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
Lincong: Sure

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
Lincong: Sure. I made it specific

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -1058,6 +1074,12 @@ static boolean fetchCompletedTask(HttpServletRequest request) {
     return getBooleanParam(request, FETCH_COMPLETED_TASK_PARAM, false);
   }
 
+  @Nullable
+  private static Long parseParamAsLong(HttpServletRequest request, String paramName) {
+    String parameterString = caseSensitiveParameterName(request.getParameterMap(), paramName);
+    return parameterString == null ? null : Long.parseLong(request.getParameter(parameterString));
+  }
+
```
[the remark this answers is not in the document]
Lincong: Cool. I added the below method:

```
  public static Long getLongParam(HttpServletRequest request, String parameter, @Nullable Long defaultIfMissing) {
    String parameterString = caseSensitiveParameterName(request.getParameterMap(), parameter);

    if (parameterString == null) {
      return defaultIfMissing;
    }
    return Long.parseLong(request.getParameter(parameterString));
  }
```

Note that the below statement does not work since `defaultIfMissing` is nullable and the implicit unboxing process might get NPE

`return parameterString == null ? defaultIfMissing : Long.parseLong...`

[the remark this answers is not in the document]
Lincong: More specifically, when we use `condition ? a : b`, it tries to find the common denominator between a type and b type. In the below case, the common type denominator is primitive `long`. That means it will try to unbox `defaultIfMissing` which could be null in this case.

`parameterString == null ? defaultIfMissing : Long.parseLong(...)`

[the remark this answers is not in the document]
efeg: @Lincong Thanks for the explanation.
I guess what you are looking for is the following:
```
return parameterString == null ? defaultIfMissing : Long.valueOf(request.getParameter(parameterString));
```

This has the benefit of 
1. not doing the extra boxing to return Long value of parsed long, and
2. being more concise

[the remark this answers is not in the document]
Lincong: Right. This one works since `Long.valueOf(...)` return `Long` instead of `long`. So `defaultIfMissing` does not get unboxed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -246,6 +247,19 @@ public static boolean getBooleanParam(HttpServletRequest request, String paramet
...
 
+  /**
+   * Get the long parameter parameter.
+   *
+   * @param request HTTP request received by Cruise Control.
+   * @param parameter Parameter to parse from the request.
+   * @param defaultIfMissing Default value to set if the request does not contain the parameter.
+   * @return The specified value for the parameter, or defaultIfMissing if the parameter is missing.
+   */
+  public static Long getLongParam(HttpServletRequest request, String parameter, @Nullable Long defaultIfMissing) {
```
efeg, line 258: Should other `Long` parameters in this class also use of this function for parsing -- e.g. `EXECUTION_PROGRESS_CHECK_INTERVAL_MS_PARAM`?

efeg (commented):
@Lincong Thanks for the updates -- left one more comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
Lincong: Yes, I applied this `getLongParam` method to `executionProgressCheckIntervalMs` and `replicationThrottle` and added unit tests

efeg:
@Lincong the rebase seems to have created unintentional commits -- please let me know when the PR is ready for review.

Lincong:
> @Lincong the rebase seems to have created unintentional commits -- please let me know when the PR is ready for review.

I just fixed the unintentional commits issue and this PR is ready for review again. Let me know if you see any git commits related issue here. Thanks

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+  @Test
+  public void testParseReplicationThrottleWithNoDefault() {
+    HttpServletRequest mockRequest = EasyMock.mock(HttpServletRequest.class);
+    KafkaCruiseControlConfig controlConfig = EasyMock.mock(KafkaCruiseControlConfig.class);
+
+    Map<String, String[]> paramMap = new HashMap<>();
+    paramMap.put(ParameterUtils.REPLICATION_THROTTLE_PARAM, new String[]{ParameterUtils.REPLICATION_THROTTLE_PARAM});
+
+    EasyMock.expect(mockRequest.getParameterMap()).andReturn(paramMap).times(1);
```
efeg, line 82: (Applies to other similar uses in this test):
Nit: `times(1)` -> `once()` -- i.e. calling one time is special :)

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  @Test
+  public void testParseExecutionProgressCheckIntervalMsWithValue() {
+    HttpServletRequest mockRequest = EasyMock.mock(HttpServletRequest.class);
+
+    Map<String, String[]> paramMap = new HashMap<>();
+    paramMap.put(
+        ParameterUtils.EXECUTION_PROGRESS_CHECK_INTERVAL_MS_PARAM,
+        new String[]{ParameterUtils.EXECUTION_PROGRESS_CHECK_INTERVAL_MS_PARAM});
```
efeg, line 124: (One more similar use above):
Nit: This is a map with a single element -- should we use a `singletonMap`? -- i.e.
```
    Map<String, String[]> paramMap = Collections.singletonMap(ParameterUtils.EXECUTION_PROGRESS_CHECK_INTERVAL_MS_PARAM,
                                                              new String[]{ParameterUtils.EXECUTION_PROGRESS_CHECK_INTERVAL_MS_PARAM});
```

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+import org.easymock.EasyMock;
+import org.junit.Assert;
+import org.junit.Test;
+
+
+public class ParameterUtilsTest {
```
efeg, line 18: Looks the mocks are not verified in tests -- i.e. there is no actual verification of whether mock functions are called or how many times they are called.

Can we call `EasyMock.verify(...)` for each test after mocks are replayed and actual calls are made?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -0,0 +1,135 @@
...
+  public void testParseReplicationThrottleWithDefault() {
+    HttpServletRequest mockRequest = EasyMock.mock(HttpServletRequest.class);
+    KafkaCruiseControlConfig controlConfig = EasyMock.mock(KafkaCruiseControlConfig.class);
+    // No parameter string value in the parameter map
+    EasyMock.expect(mockRequest.getParameterMap()).andReturn(Collections.emptyMap()).times(1);
+    EasyMock.expect(controlConfig.getLong(ExecutorConfig.DEFAULT_REPLICATION_THROTTLE_CONFIG))
+            .andReturn(Long.valueOf(DEFAULT_REPLICATION_THROTTLE_STRING));
+
+    EasyMock.replay(mockRequest);
+    EasyMock.replay(controlConfig);
```
efeg, line 103: (There are other similar uses in this test):
This function takes `Object...` as input; hence, we can pass all mocks in a single call rather than sequential calls.

efeg (commented):
@Lincong Thanks for the update. Left some comments about tests.

Lincong:
@efeg I just fixed those issues. Thanks!

efeg (approved):
@Lincong thanks for the updates. LGTM.

PR #1358, opened 2020-10-16
Check whether a cluster is using JBOD when populate_disk_info is true

### LINKED ISSUE #1353: Error message should state that "populate_disk_info" should be set to true only if JBOD is used in Kafka cluster
Users can query Kafka cluster load along with `populate_disk_info`. For example:

`GET http://<Cruise Control host>:<port>/kafkacruisecontrol/load?populate_disk_info=true`

If the Kafka cluster is not using JBOD. It returns:

`com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException: java.lang.IllegalStateException: Missing disk information for disk...`

The error message should be more specific and make it clear that `"populate_disk_info" should be set to true only if JBOD is used in Kafka cluster`

Lincong:
This PR resolves #1353 

Notes for reviewers:

The implementation is based on the assumption that if a `Broker` has multiple `Disk`s, it must be using JBOD. Please correct me if this assumption is incorrect. Thanks!

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
@@ -202,8 +202,11 @@ public class Broker implements Serializable, Comparable<Broker> {
     return _state != State.DEAD;
   }
 
+  /**
+   * @return True if the broker is using JBOD, false otherwise
+   */
   public boolean isUsingJBOD() {
-    return _diskByLogdir.size() > 1;
+    return !_diskByLogdir.isEmpty();
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
@@ -202,6 +202,10 @@ public class Broker implements Serializable, Comparable<Broker> {
     return _state != State.DEAD;
   }
 
+  public boolean isUsingJBOD() {
+    return _diskByLogdir.size() > 1;
+  }
+
   /**
    * @return True if the broker is a new broker, false otherwise.
    */
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
@@ -9,7 +9,6 @@ import com.linkedin.kafka.cruisecontrol.async.progress.OperationProgress;
 import com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException;
 import com.linkedin.kafka.cruisecontrol.model.Broker;
 import com.linkedin.kafka.cruisecontrol.model.ClusterModel;
-import com.linkedin.kafka.cruisecontrol.monitor.LoadMonitor;
 import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
 import com.linkedin.kafka.cruisecontrol.servlet.UserRequestException;
 import com.linkedin.kafka.cruisecontrol.servlet.parameters.ClusterLoadParameters;
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
@@ -12,6 +12,7 @@ import com.linkedin.kafka.cruisecontrol.model.ClusterModel;
 import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
 import com.linkedin.kafka.cruisecontrol.servlet.UserRequestException;
 import com.linkedin.kafka.cruisecontrol.servlet.parameters.ClusterLoadParameters;
+import com.linkedin.kafka.cruisecontrol.servlet.parameters.ParameterUtils;
 import com.linkedin.kafka.cruisecontrol.servlet.parameters.PartitionLoadParameters;
 import com.linkedin.kafka.cruisecontrol.servlet.response.stats.BrokerStats;
 
@@ -65,9 +66,8 @@ public class LoadRunnable extends OperationRunnable {
       if (cachedBrokerStats != null) {
         return cachedBrokerStats;
       }
-    }
-    if (_populateDiskInfo && !isClusterUsingJBOD()) {
-      throw new UserRequestException("populate_disk_info = true when Kafka cluster is NOT using JBOD");
+    } else if (isClusterUsingJBOD()) {
+      throw new UserRequestException(String.format("Cannot set %s=true for non-JBOD Kafka clusters.", ParameterUtils.POPULATE_DISK_INFO_PARAM));
     }
 
     if (_start != DEFAULT_START_TIME_FOR_CLUSTER_MODEL) {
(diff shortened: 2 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
```diff
@@ -202,6 +202,10 @@ public boolean isAlive() {
     return _state != State.DEAD;
   }
 
+  public boolean isUsingJBOD() {
+    return _diskByLogdir.size() > 1;
```
efeg, line 206: 1. Can we add a JavaDoc?
2. Nit: `_diskByLogdir.size() > 1` -> `!_diskByLogdir.isEmpty()`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
```diff
@@ -64,13 +66,29 @@ protected BrokerStats getResult() throws Exception {
         return cachedBrokerStats;
       }
     }
+    if (_populateDiskInfo && !isClusterUsingJBOD()) {
```
efeg, line 69: Can we avoid checking `_populateDiskInfo` here again -- i.e. can't we do the following?
```
} else if (!isClusterUsingJBOD()) {
   ...
}
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
```diff
@@ -64,13 +66,29 @@ protected BrokerStats getResult() throws Exception {
...
+    // If and only if all brokers in the cluster are using JBOD, the cluster is considered to be using JBOD
+    for (Broker broker : clusterModel.brokers()) {
+      if (!broker.isUsingJBOD()) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+
```
efeg, line 91: Nit: extra space.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
```diff
@@ -64,13 +66,29 @@ protected BrokerStats getResult() throws Exception {
         return cachedBrokerStats;
       }
     }
+    if (_populateDiskInfo && !isClusterUsingJBOD()) {
+      throw new UserRequestException("populate_disk_info = true when Kafka cluster is NOT using JBOD");
```
efeg, line 70: Can we refrain from (1) using hardcoded parameter names and (2) emphasis via capitalization -- i.e. maybe we can use something like:
```
throw new UserRequestException(String.format("Cannot set %s=true for non-JBOD Kafka clusters.", POPULATE_DISK_INFO_PARAM));
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
```diff
@@ -64,13 +66,29 @@ protected BrokerStats getResult() throws Exception {
...
 
+  private boolean isClusterUsingJBOD() throws Exception {
+    ClusterModel clusterModel = _kafkaCruiseControl.loadMonitor().clusterCapacity();
+    // If and only if all brokers in the cluster are using JBOD, the cluster is considered to be using JBOD
+    for (Broker broker : clusterModel.brokers()) {
+      if (!broker.isUsingJBOD()) {
+        return false;
+      }
+    }
+    return true;
```
efeg, line 88: Nit: This seems like a good candidate to replace with stream API for conciseness:
```
return clusterModel.brokers().stream().allMatch(Broker::isUsingJBOD);
```

efeg (commented):
@Lincong Thanks for the patch!
Left some comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
Lincong: Good idea!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
Lincong: Fixed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
Lincong: Good idea

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
Lincong: If we use `!_diskByLogdir.isEmpty()`, does that mean if `_diskByLogdir` is not empty, the broker must be using JBOD?

I thought `_diskByLogdir.size() == 1` means that the broker is using non-JBOD and `_diskByLogdir .size() > 1`  means that the broker is using JBOD.

efeg: > If we use `!_diskByLogdir.isEmpty()`, does that mean that the broker is using JBOD if and only if `_diskByLogdir` is NOT empty?

Correct -- please see how we `_diskByLogdir` is initialized in constructor.

Note that if a broker has 1 disk (i.e. non-JBOD), then the cluster model does not add another layer of abstraction -- i.e. keeps track of replicas per broker as opposed to replicas per disk in the broker, which would always be the same disk/logDir in a non-JBOD broker.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
```diff
@@ -7,9 +7,12 @@
 import com.linkedin.kafka.cruisecontrol.KafkaCruiseControl;
 import com.linkedin.kafka.cruisecontrol.async.progress.OperationProgress;
 import com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException;
+import com.linkedin.kafka.cruisecontrol.model.Broker;
 import com.linkedin.kafka.cruisecontrol.model.ClusterModel;
 import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
+import com.linkedin.kafka.cruisecontrol.servlet.UserRequestException;
 import com.linkedin.kafka.cruisecontrol.servlet.parameters.ClusterLoadParameters;
+import com.linkedin.kafka.cruisecontrol.servlet.parameters.ParameterUtils;
```
efeg, line 15: Nit: Maybe we can avoid having the boilerplate `ParameterUtils.XXX` if we static import just the `POPULATE_DISK_INFO_PARAM`.
```
import static com.linkedin.kafka.cruisecontrol.servlet.parameters.ParameterUtils.POPULATE_DISK_INFO_PARAM;
```

efeg (commented):
Thanks for the update -- added just one more comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/LoadRunnable.java
Lincong: OK. Fixed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
Lincong: Got it. Thanks

efeg (approved):
Lgtm!

PR #1359, opened 2020-10-20
Catch and handle exception thrown while linear regression model state gets computed

### LINKED ISSUE #1356: Get error when query Cruise Control state with "super_verbose" parameter set true
Users can query Cruise Control state and request verbose information in the response. For example:

`GET http://<Cruise Control host>:<port>/kafkacruisecontrol/state?super_verbose=true`

It returns an error `Error processing GET request '/state' due to: 'matrix is singular'`

Lincong:
This PR resolves #1356.

The current way of handling the `SingularMatrixException` is very non-aggressive. Another more aggressive way to handle it could be throwing 5XX internal server error since it's technically the logic on the CC server cannot figure out the linear regression model.


efeg (approved):
@Lincong thanks for the patch -- LGTM!

PR #1360, opened 2020-10-26
Handle Wrapped AdminClient Timeouts as Timeouts

### LINKED ISSUE #1331: processAlterPartitionReassignmentsResult and TimeoutException
We started having an issue in one of our clusters this week, and I think I've narrowed it down, but I'm not sure what the expected behaviour is. We're running Cruise Control 2.4.20.

We were seeing the cluster start to self-heal, but then it would get an exception and stop. The executor would be in "No Task In Progress", and it would complain about Rack Awareness until the moves it itself had created were finished. Then it would do it all over again.

The issue seems to be that the admin API request has a timeout (hard-coded) of 120 seconds, and the future has a hard-coded timeout of 10 seconds.

But that's ok!
The smaller timeout isn't the problem. The problem is that when there are a lot of partitions moving, all of the admin API calls are started together, which starts the clocks for those timeouts, but the futures appear to be read serially.

So the first 11 get the first timeout (the one on the Future) after 10 seconds of waiting, which logs a warning and moves on, but the 12th one has a timeout waiting for it when it gets there, leading to a timeout wrapped in an exception.

The first one (not disruptive):
```
Failed to process AlterPartitionReassignmentsResult of _core-sieve-_core-shard56-data-dedup-changelog-25. java.util.concurrent.TimeoutException
  at org.apache.kafka.common.internals.KafkaFutureImpl$SingleWaiter.await(KafkaFutureImpl.java:108)
  at org.apache.kafka.common.internals.KafkaFutureImpl.get(KafkaFutureImpl.java:272)
  at com.linkedin.kafka.cruisecontrol.executor.ExecutionUtils.processAlterPartitionReassignmentsResult(ExecutionUtils.java:340)
  at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.waitForInterBrokerReplicaTasksToFinish(Executor.java:1359)
  at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.interBrokerMoveReplicas(Executor.java:1173)
  at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.execute(Executor.java:994)
  at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.run(Executor.java:920)
  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
  at java.util.concurrent.FutureTask.run(FutureTask.java:266)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
  at java.lang.Thread.run(Thread.java:748)
```

The second one (breaks everything):
```
Executor got exception during execution java.lang.IllegalStateException: core-sale_unit_costs-0 encountered an unknown execution exception.
  at com.linkedin.kafka.cruisecontrol.executor.ExecutionUtils.processAlterPartitionReassignmentsResult(ExecutionUtils.java:355)
  at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.waitForInterBrokerReplicaTasksToFinish(Executor.java:1359)
  at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.interBrokerMoveReplicas(Executor.java:1173)
  at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.execute(Executor.java:994)
  at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.run(Executor.java:920)
  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
  at java.util.concurrent.FutureTask.run(FutureTask.java:266)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
  at java.lang.Thread.run(Thread.java:748)
Caused by: java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.TimeoutException:
Aborted due to timeout.
  at org.apache.kafka.common.internals.KafkaFutureImpl.wrapAndThrow(KafkaFutureImpl.java:45)
  at org.apache.kafka.common.internals.KafkaFutureImpl.access$000(KafkaFutureImpl.java:32)
  at org.apache.kafka.common.internals.KafkaFutureImpl$SingleWaiter.await(KafkaFutureImpl.java:104)
  at org.apache.kafka.common.internals.KafkaFutureImpl.get(KafkaFutureImpl.java:272)
  at com.linkedin.kafka.cruisecontrol.executor.ExecutionUtils.processAlterPartitionReassignmentsResult(ExecutionUtils.java:340)
  ... 9 more
Caused by: org.apache.kafka.common.errors.TimeoutException: Aborted due to timeout.
```

So I have a few questions. Should we even be expecting so many admin API timeouts? Is this the expected source of back-pressure, or is something else breaking to even have this many timeouts in the first place? I've done some hacking to get things to run. We have a 78 node cluster, so if we turn the number of concurrent moves to 1, and hack the admin API timeout to be 900 seconds (15 minutes!) then it will work, because it only needs 780 seconds to read all the futures and timeout there. That obviously doesn't scale, though, right? If that _is_ the plan, then we should make that admin API client timeout configurable.

I feel like the real fix, assuming the answer to question 1 is "yes, we expect there to be many admin API timeouts", is to have the _inner_ (API) timeout be thrown to the outer (future) context _as_ a Timeout, rather than wrapped, so the two are now the same case. They get logged as a warning and we move on. Does that seem like the right solution? Is it troublesome to have all these futures hanging around with already thrown timeout exceptions waiting in them?

I don't know enough about how this system is expected to function to know how it's supposed to look.

psycotica0-shopify:
Currently we handle the timeout that comes from getting the result of a Future by logging a warning and then doing nothing, which will allow it to try again later.

But, since we get the results of the Futures in serial, but we kick off the admin request essentially in parallel, if there are enough of them then some of them can timeout before we get around to even checking the Future.
This means we'll see the admin client timeout immediately, rather than seeing the Future timeout.

Currently this is handled as exceptional and aborts the whole operation, which isn't helpful.

This change just copy and pastes the Future timeout logic into a new branch that deals with the underlying timeout in the same way.

This PR resolves #1331


psycotica0-shopify:
I wanted to have tests for this simple change, but I ended up going to a deep rabbit-hole. It seems like that file isn't tested currently, and also creating KafkaFutures appears to be prohibited from outside the client package, which made it hard to test.

If it's not too much trouble, I could use some help with that part.

andrewjamesbrown:
@Lincong wondering if you've any time to look at this next week?  🙏

Lincong:
> @Lincong wondering if you've any time to look at this next week? 🙏

Hi @andrewjamesbrown  @psycotica0-shopify , thank you for creating this PR and sorry for my delayed response!
To test this change, have you considered using mocks? It seems that you need to mock the behavior of the `AlterPartitionReassignmentsResult` object and pass the mock to the method (`processAlterPartitionReassignmentsResult`).

`KafkaFuture` can be mocked as well and then you can specify that the `AlterPartitionReassignmentsResult` returns a map of  `KafkaFuture` mocks when a specific method is invoked.

Cruise Control uses `EasyMock` and I think the above approach can be done using `EasyMock`. Let me know whether you have explored this option or if any difficulty that I underestimated. Thanks!

Lincong:
Hi @psycotica0-shopify, let me know if you need anything (ideas, comments, help, etc) from me. Thanks

Lincong:
Hi @psycotica0-shopify @andrewjamesbrown I just created [a PR](https://github.com/Shopify/cruise-control/pull/1) in the Shopify CC repo. I hope you can take a look. Once that PR is merged to the Shopify CC branch, the change will show up in this PR and we can merge it to the LinkedIn repo.

Let me know if you have any questions or feedback! Thanks!

bpux:
hi, I have faced the same issue in our cruise control(2.5 branch). applied this change and the executor won't quit anymore. wondering what's better solution of i

Lincong (approved):
This PR initially needed some unit tests. I added some unit tests ([commit](https://github.com/linkedin/cruise-control/pull/1360/commits/e7b54b0affeb05a6fd02e9badebfa5966a29cb34)) and the overall change has been reviewed merged to the Shopify CC repo. So, I will merge it to LinkedIn repo.

PR #1369, opened 2020-11-02
Enable Cruise Control to collect metrics from low traffic clusters by default

efeg:
Cruise Control (CC) can fail to collect sufficient metrics from low traffic Kafka clusters for an extended period of time, which would make CC unable to handle model-based (e.g. get cluster `load`) or goal-based (`rebalance` the cluster using the default goals) operations – e.g.
```
Error processing GET request '/partition_load' due to: 'com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException: com.linkedin.cruisecontrol.exception.NotEnoughValidWindowsException: There are only 0 valid windows when aggregating in range [-1, 1603951915056] for aggregation options (minValidEntityRatio=0.98, minValidEntityGroupRatio=0.00, minValidWindows=1, numEntitiesToInclude=55498, granularity=ENTITY)'.
```

We have already exposed the relevant config to fix similar observed issues in https://github.com/linkedin/cruise-control/pull/730. However, the default config value has not been updated, which requires users to manually change the config. This patch updates the default value of the relevant config to a value that enable collection of metrics from low traffic clusters by default.

### DIFF
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterConfig.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterConfig.java
@@ -67,7 +67,7 @@ public class CruiseControlMetricsReporterConfig extends AbstractConfig {
   public static final Short DEFAULT_CRUISE_CONTROL_MIN_INSYNC_REPLICAS = -1;
   public static final long DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_INTERVAL_MS = 60000;
   public static final String PRODUCER_ID = "CruiseControlMetricsReporter";
-  public static final int DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_LINGER_MS = 30 * 1000;
+  public static final int DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_LINGER_MS = 5000;
   public static final int DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_MAX_BLOCK_MS = 60 * 1000;
   public static final int DEFAULT_CRUISE_CONTROL_METRICS_BATCH_SIZE = 800 * 1000;
   public static final boolean DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_KUBERNETES_MODE = false;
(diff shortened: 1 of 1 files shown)


cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterConfig.java
```diff
@@ -67,7 +67,7 @@
   public static final Short DEFAULT_CRUISE_CONTROL_MIN_INSYNC_REPLICAS = -1;
   public static final long DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_INTERVAL_MS = 60000;
   public static final String PRODUCER_ID = "CruiseControlMetricsReporter";
-  public static final int DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_LINGER_MS = 30 * 1000;
+  public static final int DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_LINGER_MS = 5000;
```
wyuka, line 70: We might want to go with the convention and write it as 5 * 1000.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterConfig.java
```diff
@@ -67,7 +64,7 @@
   public static final Short DEFAULT_CRUISE_CONTROL_MIN_INSYNC_REPLICAS = -1;
   public static final long DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_INTERVAL_MS = 60000;
   public static final String PRODUCER_ID = "CruiseControlMetricsReporter";
-  public static final int DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_LINGER_MS = 30 * 1000;
+  public static final int DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_LINGER_MS = 5 * 1000;
```
Lincong, line 67: I highly recommend using `TimeUnit` to set default values for all time-related parameters. For example `DEFAULT_CRUISE_CONTROL_METRICS_REPORTER_LINGER_MS  = TimeUnit.SECONDS.toMillis(5);`

Lincong (changes requested):

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterConfig.java
Lincong: I will close this issue and I will create another PR to fix this issue.

Lincong (approved):

PR #1380, opened 2020-11-06
[WIP] Avoid self-healing of goal violation under low load of CPU, disk, and networkoutbound

### LINKED ISSUE #1147: DISK, CPU and NW_OUT low utilization thresholds are not used
I am setting `{cpu,disk,network.inbound,network.outbound}.low.utilization.threshold` parameters to avoid self-healing of Goal violations under low load but I noticed that `cpu,disk,network.outbound` are actually never used in the code.

The only reference is to `lowUtilizationThreshold` is in `LeaderBytesInDistributionGoal` to check  `network.inbound` threshold.

https://github.com/linkedin/cruise-control/blob/da27f36297d240978c1067da303e67d812d5f922/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/LeaderBytesInDistributionGoal.java#L252

Am I missing something? Thanks

Lincong:


Lincong:
Since this WIP, I will move this PR away from upstream. Once it's completed, another PR will be created

PR #1381, opened 2020-11-06
Fix a bug that might cause invalid throttle replica list to be used

### LINKED ISSUE #1322: The follower.replication.throttled.replicas written can be messed up if previous value was '*'
In the `setThrottledReplicas` function (https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java#L182) we suspect that the joining of the old list and new list with commas can lead to a situation where, if the old value is a single '*' which is allowed, then the result is something like `follower.replication.throttled.replicas: "*,11:1,11:2....."` which is not a valid value.

Normally Kafka's helper would validate this, but due a bug (https://issues.apache.org/jira/browse/KAFKA-10460) it didn't catch this case.

Lincong:
This PR resolves #1322.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
@@ -154,7 +154,7 @@ class ReplicationThrottleHelper {
     if (_throttleRate == null) {
       throw new IllegalStateException("Throttle rate cannot be null");
     }
-    if ((!configKey.equals(LEADER_THROTTLED_RATE)) && (!configKey.equals(FOLLOWER_THROTTLED_RATE))) {
+    if (!configKey.equals(LEADER_THROTTLED_RATE) && !configKey.equals(FOLLOWER_THROTTLED_RATE)) {
       throw new IllegalArgumentException(
           String.format("Config key must be either %s or %s", LEADER_THROTTLED_RATE, FOLLOWER_THROTTLED_RATE));
     }
@@ -179,7 +179,7 @@ class ReplicationThrottleHelper {
   }
 
   private void setThrottledReplicas(String topic, Set<String> replicas, String configKey) {
-    if ((!configKey.equals(LEADER_THROTTLED_REPLICAS)) && (!configKey.equals(FOLLOWER_THROTTLED_REPLICAS))) {
+    if (!configKey.equals(LEADER_THROTTLED_REPLICAS) && !configKey.equals(FOLLOWER_THROTTLED_REPLICAS)) {
       throw new IllegalArgumentException(
           String.format("Config key must be either %s or %s", LEADER_THROTTLED_REPLICAS, FOLLOWER_THROTTLED_REPLICAS));
     }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
@@ -151,8 +151,14 @@ class ReplicationThrottleHelper {
   }
 
   private void setThrottledRateIfUnset(int brokerId, String configKey) {
-    assert (_throttleRate != null);
-    assert (configKey.equals(LEADER_THROTTLED_RATE) || configKey.equals(FOLLOWER_THROTTLED_RATE));
+    if (_throttleRate == null) {
+      throw new IllegalStateException("Throttle rate cannot be null");
+    }
+    if ((!configKey.equals(LEADER_THROTTLED_RATE)) && (!configKey.equals(FOLLOWER_THROTTLED_RATE))) {
+      throw new IllegalArgumentException(
+          String.format("Config key must be either %s or %s", LEADER_THROTTLED_RATE, FOLLOWER_THROTTLED_RATE));
+    }
+
     Properties config = _kafkaZkClient.getEntityConfigs(ConfigType.Broker(), String.valueOf(brokerId));
     Object oldThrottleRate = config.setProperty(configKey, String.valueOf(_throttleRate));
     if (oldThrottleRate == null) {
@@ -173,15 +179,28 @@ class ReplicationThrottleHelper {
   }
 
   private void setThrottledReplicas(String topic, Set<String> replicas, String configKey) {
-    assert (configKey.equals(LEADER_THROTTLED_REPLICAS) || configKey.equals(FOLLOWER_THROTTLED_REPLICAS));
-    Properties config = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
-    // Merge new throttled replicas with existing configuration values.
-    Set<String> newThrottledReplicas = new TreeSet<>(replicas);
-    String oldThrottledReplicas = config.getProperty(configKey);
-    if (oldThrottledReplicas != null) {
-      newThrottledReplicas.addAll(Arrays.asList(oldThrottledReplicas.split(",")));
+    if ((!configKey.equals(LEADER_THROTTLED_REPLICAS)) && (!configKey.equals(FOLLOWER_THROTTLED_REPLICAS))) {
+      throw new IllegalArgumentException(
+          String.format("Config key must be either %s or %s", LEADER_THROTTLED_REPLICAS, FOLLOWER_THROTTLED_REPLICAS));
     }
-    config.setProperty(configKey, String.join(",", newThrottledReplicas));
+
+    Properties config = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
+    String oldThrottledReplicas = config.getProperty(configKey);
+    if (oldThrottledReplicas != null && oldThrottledReplicas.trim().equals("*")) {
+      // The existing setup throttles all replica. So, nothing needs to be changed
+      return;
+    }
+
+    if (replicas.size() == 1 && replicas.iterator().next().trim().equals("*")) {
+      config.setProperty(configKey, "*");
+    } else {
+      // Merge new throttled replicas with existing configuration values.
+      Set<String> newThrottledReplicas = new TreeSet<>(replicas);
+      if (oldThrottledReplicas != null) {
+        newThrottledReplicas.addAll(Arrays.asList(oldThrottledReplicas.split(",")));
+      }
+      config.setProperty(configKey, String.join(",", newThrottledReplicas));
+    }
     ExecutorUtils.changeTopicConfig(_adminZkClient, topic, config);
   }
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
@@ -156,13 +156,13 @@ class ReplicationThrottleHelper {
     }
     String replicaThrottleRateConfigKey = throttleLeaderReplicaRate ? LEADER_THROTTLED_RATE : FOLLOWER_THROTTLED_RATE;
     Properties config = _kafkaZkClient.getEntityConfigs(ConfigType.Broker(), String.valueOf(brokerId));
-    Object oldThrottleRate = config.setProperty(replicaThrottleRateConfigKey, String.valueOf(_throttleRate));
-    if (oldThrottleRate == null) {
+    Object currThrottleRate = config.setProperty(replicaThrottleRateConfigKey, String.valueOf(_throttleRate));
+    if (currThrottleRate == null) {
       LOG.debug("Setting {} to {} bytes/second for broker {}", replicaThrottleRateConfigKey, _throttleRate, brokerId);
       ExecutorUtils.changeBrokerConfig(_adminZkClient, brokerId, config);
     } else {
       LOG.debug("Not setting {} for broker {} because pre-existing throttle of {} was already set",
-          replicaThrottleRateConfigKey, brokerId, oldThrottleRate);
+          replicaThrottleRateConfigKey, brokerId, currThrottleRate);
     }
   }
 
@@ -178,16 +178,16 @@ class ReplicationThrottleHelper {
     String replicaThrottleConfigKey = throttleLeaderReplica ? LEADER_THROTTLED_REPLICAS : FOLLOWER_THROTTLED_REPLICAS;
 
     Properties config = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
-    String oldThrottledReplicas = config.getProperty(replicaThrottleConfigKey);
-    if (oldThrottledReplicas != null && oldThrottledReplicas.trim().equals(WILD_CARD_ASTERISK)) {
+    String currThrottledReplicas = config.getProperty(replicaThrottleConfigKey);
+    if (currThrottledReplicas != null && currThrottledReplicas.trim().equals(WILD_CARD_ASTERISK)) {
       // The existing setup throttles all replica. So, nothing needs to be changed.
       return;
     }
 
     // Merge new throttled replicas with existing configuration values.
     Set<String> newThrottledReplicas = new TreeSet<>(replicas);
-    if (oldThrottledReplicas != null) {
-      newThrottledReplicas.addAll(Arrays.asList(oldThrottledReplicas.split(",")));
+    if (currThrottledReplicas != null) {
+      newThrottledReplicas.addAll(Arrays.asList(currThrottledReplicas.split(",")));
     }
     config.setProperty(replicaThrottleConfigKey, String.join(",", newThrottledReplicas));
     ExecutorUtils.changeTopicConfig(_adminZkClient, topic, config);
@@ -210,15 +210,15 @@ class ReplicationThrottleHelper {
    * @return true if the given configuration properties are modified and false otherwise
    */
   private boolean removeLeaderThrottledReplicasFromTopic(Properties config, String topic, Set<String> replicas) {
-    String oldLeaderThrottledReplicas = config.getProperty(LEADER_THROTTLED_REPLICAS);
-    if (oldLeaderThrottledReplicas != null) {
-      if (oldLeaderThrottledReplicas.equals(WILD_CARD_ASTERISK)) {
+    String currLeaderThrottledReplicas = config.getProperty(LEADER_THROTTLED_REPLICAS);
+    if (currLeaderThrottledReplicas != null) {
+      if (currLeaderThrottledReplicas.equals(WILD_CARD_ASTERISK)) {
         LOG.debug("Existing config throttles all leader replicas. So, not remove any leader replica throttle");
         return false;
       }
 
       replicas.forEach(r -> LOG.debug("Removing leader throttles for topic {} on replica {}", topic, r));
-      String newLeaderThrottledReplicas = removeReplicasFromConfig(oldLeaderThrottledReplicas, replicas);
+      String newLeaderThrottledReplicas = removeReplicasFromConfig(currLeaderThrottledReplicas, replicas);
       if (newLeaderThrottledReplicas.isEmpty()) {
         config.remove(LEADER_THROTTLED_REPLICAS);
       } else {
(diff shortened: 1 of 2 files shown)


MatthewBerk:
Looks good 👍

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -151,8 +151,14 @@ private void setFollowerThrottledRateIfUnset(int brokerId) {
   }
 
   private void setThrottledRateIfUnset(int brokerId, String configKey) {
-    assert (_throttleRate != null);
-    assert (configKey.equals(LEADER_THROTTLED_RATE) || configKey.equals(FOLLOWER_THROTTLED_RATE));
+    if (_throttleRate == null) {
+      throw new IllegalStateException("Throttle rate cannot be null");
+    }
+    if ((!configKey.equals(LEADER_THROTTLED_RATE)) && (!configKey.equals(FOLLOWER_THROTTLED_RATE))) {
```
efeg, line 157: 1. Superfluous parentheses around (1) `!configKey.equals(LEADER_THROTTLED_RATE)` and (2) `!configKey.equals(FOLLOWER_THROTTLED_RATE)`. Can we drop them for better readability?

* Please see https://docs.oracle.com/javase/tutorial/java/nutsandbolts/operators.html

2. Nit: This is a private function of the class -- i.e. no public / protected / package private access. Hence, I feel that this sanity check is redundant all together, but the decision to keep or remove it is up to you.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -173,15 +179,28 @@ private void setFollowerThrottledReplicas(String topic, Set<String> replicas) {
   }
 
   private void setThrottledReplicas(String topic, Set<String> replicas, String configKey) {
-    assert (configKey.equals(LEADER_THROTTLED_REPLICAS) || configKey.equals(FOLLOWER_THROTTLED_REPLICAS));
+    if ((!configKey.equals(LEADER_THROTTLED_REPLICAS)) && (!configKey.equals(FOLLOWER_THROTTLED_REPLICAS))) {
```
efeg, line 182: Superfluous parentheses - please see the comment above.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -173,15 +179,28 @@ private void setFollowerThrottledReplicas(String topic, Set<String> replicas) {
...
-    if (oldThrottledReplicas != null) {
-      newThrottledReplicas.addAll(Arrays.asList(oldThrottledReplicas.split(",")));
+    if (oldThrottledReplicas != null && oldThrottledReplicas.trim().equals("*")) {
+      // The existing setup throttles all replica. So, nothing needs to be changed
+      return;
+    }
+
+    if (replicas.size() == 1 && replicas.iterator().next().trim().equals("*")) {
+      config.setProperty(configKey, "*");
+    } else {
```
efeg, line 196: Does this private function ever expected to get `*` as input for a replica in `replicas`? Isn't this dead code?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -173,15 +179,28 @@ private void setFollowerThrottledReplicas(String topic, Set<String> replicas) {
   }
 
   private void setThrottledReplicas(String topic, Set<String> replicas, String configKey) {
```
efeg, line 181: Don't we need a similar special treatment when removing throttled replicas from a topic? If the existing throttle is `*`, it should be guaranteed that it is not set by CC; hence, shouldn't we leave it as it is (i.e. untouched)?

efeg (commented):
@Lincong Thanks for the PR!
Left some comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
Lincong: 1. I will remove the superfluous parentheses.

2. I will keep this check. I don't think of it as a sanity check since it's not checking input parameters. Instead, it's basically saying "if at this point of the execution, the `_throttleRate ` field is null, that means we have an illegal state". Therefore, it has nothing to do with whether this method is public/protected/private.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
Lincong: As for `If the existing throttle is *, it should be guaranteed that it is not set by CC`, I think my change ensures that. Did I misunderstood your point or miss anything?

```
    if (oldThrottledReplicas != null && oldThrottledReplicas.trim().equals("*")) {
      // The existing setup throttles all replica. So, nothing needs to be changed
      return;
    }
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
Lincong: Considering the current state of the codebase, this private function does not take `*` as input. But just looking at this private method itself, we do not know what is going to change on its caller side in the future. Due to my defensive style, I prefer not to make assumptions regarding what callers will or never pass to a method even if the method is private. It might be dead code. But it's safe and handles a very valid case.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
efeg: > I will keep this check. I don't think of it as a sanity check since it's not checking input parameters. Instead, it's basically saying "if at this point of the execution, the _throttleRate field is null, that means we have an illegal state". Therefore, it has nothing to do with whether this method is public/protected/private.

My comment is about `configKey`, which is an input to this private function, -- i.e. not about `_throttleRate`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
efeg: > But it's safe and handles a very valid case.

The key issue is that this is _not_ a valid case.
This code can generate inconsistent / incorrect behavior -- i.e. replication throttler shall never set this config to `*`.

This is not making the code defensive, on the contrary, it makes it more vulnerable to errors.

* We also seem to be missing tests for the added fix -- can we add unit tests please?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
efeg: Does `#removeThrottledReplicasFromTopic` take the right action (i.e. leave the value as `*`) when the relevant config was already `*`? I suspect that it doesn't.

* We are also missing a unit test for this scenario, can we add one?

efeg (commented):
@Lincong thanks for the update.
Replied to some earlier comments and left more comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -173,15 +179,28 @@ private void setFollowerThrottledReplicas(String topic, Set<String> replicas) {
...
+    }
+
     Properties config = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
-    // Merge new throttled replicas with existing configuration values.
-    Set<String> newThrottledReplicas = new TreeSet<>(replicas);
     String oldThrottledReplicas = config.getProperty(configKey);
-    if (oldThrottledReplicas != null) {
-      newThrottledReplicas.addAll(Arrays.asList(oldThrottledReplicas.split(",")));
+    if (oldThrottledReplicas != null && oldThrottledReplicas.trim().equals("*")) {
+      // The existing setup throttles all replica. So, nothing needs to be changed
```
robinp-tw, line 190: nit - end-of-line dot missing
```suggestion
      // The existing setup throttles all replica. So, nothing needs to be changed.
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -173,15 +179,28 @@ private void setFollowerThrottledReplicas(String topic, Set<String> replicas) {
...
+      return;
+    }
+
+    if (replicas.size() == 1 && replicas.iterator().next().trim().equals("*")) {
```
robinp-tw, line 194: nit: magic value `"*"` to a constant?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
robinp-tw: I'm not following exactly, but given the bug happened where a `*` and a `kv-list` were joined (thus this PR), it well might be that this method gets called with a `*` member, isn't it?

https://kafka.apache.org/documentation/#leader.replication.throttled.replicas and the Kafka validation code certainly allows for `*`. If CC should not allow for a `*` in this method, what is the codepath that still smuggled it? Or otherwise how could the bug happen?

Agree the unit-tests would be welcome nevertheless.

robinp-tw (commented):
Random comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
Lincong: I see. OK, I will remove this sanity check.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
Lincong: That's a good point. I will modify `removeThrottledReplicasFromTopic ` to handle the `*` case and add unit test.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -173,15 +179,28 @@ private void setFollowerThrottledReplicas(String topic, Set<String> replicas) {
   }
 
   private void setThrottledReplicas(String topic, Set<String> replicas, String configKey) {
-    assert (configKey.equals(LEADER_THROTTLED_REPLICAS) || configKey.equals(FOLLOWER_THROTTLED_REPLICAS));
+    if (!configKey.equals(LEADER_THROTTLED_REPLICAS) && !configKey.equals(FOLLOWER_THROTTLED_REPLICAS)) {
```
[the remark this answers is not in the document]
Lincong: I will change the method to be the following so that we do not need to worry about illegal input arguments:

```
  private void setThrottledReplicas(String topic, Set<String> replicas, boolean throttleLeaderReplica) {
    String replicaThrottleConfigKey = throttleLeaderReplica ? LEADER_THROTTLED_REPLICAS : FOLLOWER_THROTTLED_REPLICAS;
 // ...
```

Lincong:
Hi @efeg @robinp-tw , I just added a unit test. I hope to get another round of review. Thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -28,7 +28,7 @@
  */
 class ReplicationThrottleHelper {
   private static final Logger LOG = LoggerFactory.getLogger(ReplicationThrottleHelper.class);
-
+  static final String WILD_CARD_ASTERISK = "*";
```
efeg, line 31: Nit: `WILD_CARD_ASTERISK` -> `WILDCARD_ASTERISK`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -191,65 +199,101 @@ static String removeReplicasFromConfig(String throttleConfig, Set<String> replic
...
+   * @param config configuration properties
+   * @param topic name of topic which contains <code>replicas</code>
+   * @param replicas replicas to remove from the configuration properties
+   * @return true if the given configuration properties are modified and false otherwise
+   */
+  private boolean removeLeaderThrottledReplicasFromTopic(Properties config, String topic, Set<String> replicas) {
+    String currLeaderThrottledReplicas = config.getProperty(LEADER_THROTTLED_REPLICAS);
+    if (currLeaderThrottledReplicas != null) {
+      if (currLeaderThrottledReplicas.equals(WILD_CARD_ASTERISK)) {
+        LOG.debug("Existing config throttles all leader replicas. So, not remove any leader replica throttle");
```
efeg, line 216: Nit: `So, not remove ...` -> `So, do not remove ...`
- multiple occurrences, please see `#removeThrottledRateFromBroker `

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -61,6 +61,15 @@ private void createTopics() {
     }
   }
 
+  private void setWildcardThrottleReplicaForTopic(KafkaZkClient kafkaZkClient, String topicName) {
+    for (String replicaThrottleProp :
+        Arrays.asList(ReplicationThrottleHelper.LEADER_THROTTLED_REPLICAS, ReplicationThrottleHelper.FOLLOWER_THROTTLED_REPLICAS)) {
```
efeg, line 66: Nit: Indentation seems off -- can we use:
```
    for (String replicaThrottleProp : Arrays.asList(ReplicationThrottleHelper.LEADER_THROTTLED_REPLICAS,
                                                    ReplicationThrottleHelper.FOLLOWER_THROTTLED_REPLICAS)) {
```

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -61,6 +61,15 @@ private void createTopics() {
     }
   }
 
+  private void setWildcardThrottleReplicaForTopic(KafkaZkClient kafkaZkClient, String topicName) {
```
efeg, line 64: Can we make this function `static`? This would make it clear that it does not modify the state of the object and just a utility function.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -206,6 +215,81 @@ public void testAddingThrottlesWithPreExistingThrottles() throws InterruptedExce
     assertExpectedThrottledReplicas(kafkaZkClient, TOPIC1, "1:1");
   }
 
+  @Test
+  public void testDoNotModifyExistingWildcardReplicaThrottles() {
+    createTopics();
+
+    KafkaZkClient kafkaZkClient = KafkaCruiseControlUtils.createKafkaZkClient(zookeeper().connectionString(),
+        "ReplicationThrottleHelperTestMetricGroup", "AddingThrottlesWithNoPreExistingThrottles", false);
```
efeg, line 223: Nit: Bad indentation.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -206,6 +215,81 @@ public void testAddingThrottlesWithPreExistingThrottles() throws InterruptedExce
...
+        new ReplicaPlacementInfo(0),
+        Arrays.asList(new ReplicaPlacementInfo(0), new ReplicaPlacementInfo(1)),
+        Arrays.asList(new ReplicaPlacementInfo(0), new ReplicaPlacementInfo(2)));
+
+    ExecutionProposal proposal2 = new ExecutionProposal(
+        new TopicPartition(TOPIC0, 1),
+        100,
+        new ReplicaPlacementInfo(0),
+        Arrays.asList(new ReplicaPlacementInfo(0), new ReplicaPlacementInfo(3)),
+        Arrays.asList(new ReplicaPlacementInfo(0), new ReplicaPlacementInfo(2)));
```
efeg, line 244: Nit: Bad indentation -- please align after `(` and refrain from putting the first argument to `ExecutionProposal` on a different line -- looks like the rest of this test (independent of this PR) also has issues with those.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -206,6 +215,81 @@ public void testAddingThrottlesWithPreExistingThrottles() throws InterruptedExce
     assertExpectedThrottledReplicas(kafkaZkClient, TOPIC1, "1:1");
   }
 
+  @Test
+  public void testDoNotModifyExistingWildcardReplicaThrottles() {
```
efeg, line 219: This test and `testDoNotRemoveThrottlesForInProgressTasks` seems to have some duplicate parts. I am curious if we can extract those common code to a function?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -206,6 +215,81 @@ public void testAddingThrottlesWithPreExistingThrottles() throws InterruptedExce
...
+    // we expect broker 1 to be null since all replica movement related to it has completed.
+    assertExpectedThrottledRateForBroker(kafkaZkClient, 1, null);
+    assertExpectedThrottledRateForBroker(kafkaZkClient, 2, throttleRate);
+    // We expect broker 3 to have a throttle on it because there is an in-progress replica being moved
+    assertExpectedThrottledRateForBroker(kafkaZkClient, 3, throttleRate);
+    // Topic-level throttled replica config value should remain as "*"
+    assertExpectedThrottledReplicas(kafkaZkClient, TOPIC0, ReplicationThrottleHelper.WILD_CARD_ASTERISK);
+    assertExpectedThrottledReplicas(kafkaZkClient, TOPIC1, ReplicationThrottleHelper.WILD_CARD_ASTERISK);
+
+    // Completing the in-progress task and clearing the throttles should clean everything up.
```
efeg, line 282: This comment is not correct right? -- i.e. it should _not_ clean everything as you have demonstrated below with the topic-level throttled replica config remaining as `WILD_CARD_ASTERISK`.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -206,6 +215,81 @@ public void testAddingThrottlesWithPreExistingThrottles() throws InterruptedExce
...
+    inProgressTask.completed(3);
+    throttleHelper.clearThrottles(Arrays.asList(completedTask, inProgressTask), Collections.emptyList());
+
+    Arrays.asList(0, 1, 2, 3).forEach((i) -> assertExpectedThrottledRateForBroker(kafkaZkClient, i, null));
+    // Topic-level throttled replica config value should remain as "*"
+    assertExpectedThrottledReplicas(kafkaZkClient, TOPIC0, ReplicationThrottleHelper.WILD_CARD_ASTERISK);
+    assertExpectedThrottledReplicas(kafkaZkClient, TOPIC1, ReplicationThrottleHelper.WILD_CARD_ASTERISK);
+  }
+
```
efeg, line 291: Nit: Redundant new line

efeg (commented):
@Lincong Thanks for the update!
Left some comments.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
Lincong: Correct

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
Lincong: Fixed

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
Lincong: OK. Btw, I used to add `static` to methods that are "stateless" in some other projects as well. But I was told by code reviewers not to do that and their argument is that `static` should, in general, be avoided as much as possible.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
Lincong: It's totally possible to extract common code to a function. However, IMO, that would decrease the readability of both test cases. I would suggest considering refactoring out the common part when we have more unit tests testing the same workflow since currently there are only 2 such unit tests (not too much redundancy) and each of them is readable and easy to follow the test workflow.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
efeg: I typically choose `static` to improve clarity if at all possible.
There are some cases, where mocking a method is needed and it is either not possible or tricky to do so with `static` methods (e.g. due to limitations of `EasyMock`). For those cases, making the method non-static would be understandable.

efeg (approved):
@Lincong Thanks for the updates -- LGTM.

PR #1385, opened 2020-11-12
Ensure that resource distribution goals compute balance constraints properly

### LINKED ISSUE #1384: Ensure that resource distribution goals compute balance constraints properly
Resource distribution goals suffer from the same issue as https://github.com/linkedin/cruise-control/issues/1382.

Ensure that resource distribution goals do not include brokers excluded for replica moves while computing balance constraints:
1. CpuUsageDistributionGoal
2. DiskUsageDistributionGoal
3. NetworkInboundUsageDistributionGoal
4. NetworkOutboundUsageDistributionGoal
5. PotentialNwOutGoal

efeg:
* Ensure that `CpuUsageDistributionGoal`, `DiskUsageDistributionGoal`, `NetworkInboundUsageDistributionGoal`, `NetworkOutboundUsageDistributionGoal`, `DiskCapacityGoal`, `NetworkInboundCapacityGoal`, `NetworkOutboundCapacityGoal`, `CpuCapacityGoal`, `LeaderBytesInDistributionGoal` do not include brokers excluded for replica moves while computing balance constraints (`PotentialNwOutGoal` does not need further changes).

This PR resolves #1384.

-- Selected Evaluation --
Selected evaluations on a cluster with 10 brokers with 3 brokers permanently marked as excluded -- i.e. preferred controllers of [LinkedIn Kafka](https://github.com/linkedin/kafka) -- please note standard deviations (lower is better).

Standard deviation **before** and **after**:

- CpuUsageDistributionGoal: 42.917 -> 1.919
- DiskUsageDistributionGoal: 69 ->  0.939
- NetworkInboundUsageDistributionGoal: 0.763 -> 0.039
- NetworkOutboundUsageDistributionGoal: 3.455 ->  0.127

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationOptions.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationOptions.java
@@ -73,6 +73,15 @@ public class OptimizationOptions {
                              boolean isTriggeredByGoalViolation,
                              Set<Integer> requestedDestinationBrokerIds,
                              boolean onlyMoveImmigrantReplicas) {
+    if (excludedTopics == null) {
+      throw new IllegalArgumentException("Excluded topics cannot be null.");
+    } else if (excludedBrokersForLeadership == null) {
+      throw new IllegalArgumentException("Excluded brokers for leadership cannot be null.");
+    } else if (excludedBrokersForReplicaMove == null) {
+      throw new IllegalArgumentException("Excluded brokers for replica move cannot be null.");
+    } else if (requestedDestinationBrokerIds == null) {
+      throw new IllegalArgumentException("Requested destination broker ids cannot be null.");
+    }
     _excludedTopics = excludedTopics;
     _excludedBrokersForLeadership = excludedBrokersForLeadership;
     _excludedBrokersForReplicaMove = excludedBrokersForReplicaMove;
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
@@ -15,7 +15,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
-import com.google.gson.Gson;
 import java.util.Set;
 import java.util.function.Function;
 import org.apache.kafka.common.TopicPartition;
@@ -37,11 +36,10 @@ public class ClusterModelStats {
   private final Map<Statistic, Number> _leaderReplicaStats;
   private final Map<Statistic, Number> _topicReplicaStats;
   private int _numBrokers;
-  private int _numAliveBrokers;
   private int _numReplicasInCluster;
   private int _numPartitionsWithOfflineReplicas;
   private int _numTopics;
-  private Map<Resource, Integer> _numBalancedBrokersByResource;
+  private final Map<Resource, Integer> _numBalancedBrokersByResource;
   private int _numBrokersUnderPotentialNwOut;
   private BalancingConstraint _balancingConstraint;
   private double[][] _utilizationMatrix;
@@ -83,12 +81,11 @@ public class ClusterModelStats {
    */
   ClusterModelStats populate(ClusterModel clusterModel, BalancingConstraint balancingConstraint, OptimizationOptions optimizationOptions) {
     _numBrokers = clusterModel.brokers().size();
-    _numAliveBrokers = clusterModel.aliveBrokers().size();
     _numTopics = clusterModel.topics().size();
     _balancingConstraint = balancingConstraint;
     _brokersAllowedReplicaMove = GoalUtils.aliveBrokersNotExcludedForReplicaMove(clusterModel, optimizationOptions);
-    utilizationForResources(clusterModel);
-    utilizationForPotentialNwOut(clusterModel);
+    utilizationForResources(clusterModel, optimizationOptions);
+    utilizationForPotentialNwOut(clusterModel, optimizationOptions);
     numForReplicas(clusterModel);
     numForLeaderReplicas(clusterModel);
     numForAvgTopicReplicas(clusterModel);
@@ -163,14 +160,16 @@ public class ClusterModelStats {
   }
 
   /**
-   * @return The number of balanced brokers by resource for the cluster instance that the object was populated with.
+   * @return The number of balanced brokers that are not excluded for replica moves by resource for the cluster instance
+   * that the object was populated with.
    */
   public Map<Resource, Integer> numBalancedBrokersByResource() {
     return _numBalancedBrokersByResource;
   }
 
   /**
-   * @return The number of brokers under potential nw out for the cluster instance that the object was populated with.
+   * @return The number of brokers that are not excluded for replica moves under potential nw out for the cluster instance
+   * that the object was populated with.
    */
   public int numBrokersUnderPotentialNwOut() {
     return _numBrokersUnderPotentialNwOut;
@@ -218,14 +217,6 @@ public class ClusterModelStats {
   }
 
   /**
-   * @return A valid JSON encoded string.
-   */
-  public String getJSONString() {
-    Gson gson = new Gson();
-    return gson.toJson(getJsonStructure());
-  }
-
-  /**
    * @return An object that can be further used to encode into JSON.
    */
   public Map<String, Object> getJsonStructure() {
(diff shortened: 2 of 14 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationOptions.java
```diff
@@ -73,6 +73,15 @@ public OptimizationOptions(Set<String> excludedTopics,
                              boolean isTriggeredByGoalViolation,
                              Set<Integer> requestedDestinationBrokerIds,
                              boolean onlyMoveImmigrantReplicas) {
+    if (excludedTopics == null) {
```
Lincong, line 76: Nit: in another project of mine, I have the following util method since doing the null check is quite common.

```
  public static <T> T validateNotNull(T obj, String objName) {
    Objects.requireNonNull(obj, objName + " cannot be null.");
    return obj;
  }
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
```diff
@@ -257,42 +248,48 @@ public String toString() {
...
-    Map<Resource, Double> minUtilizationByResource = new HashMap<>();
-    Map<Resource, Double> stDevUtilizationByResource = new HashMap<>();
+    Map<Resource, Double> avgUtilizationByResource = new HashMap<>(Resource.cachedValues().size());
+    Map<Resource, Double> maxUtilizationByResource = new HashMap<>(Resource.cachedValues().size());
+    Map<Resource, Double> minUtilizationByResource = new HashMap<>(Resource.cachedValues().size());
+    Map<Resource, Double> stDevUtilizationByResource = new HashMap<>(Resource.cachedValues().size());
     for (Resource resource : Resource.cachedValues()) {
-      double avgUtilizationPercentage = clusterModel.load().expectedUtilizationFor(resource) / clusterModel.capacityFor(resource);
+      double resourceUtilization = clusterModel.load().expectedUtilizationFor(resource);
+      double avgUtilizationPercentage = resourceUtilization / clusterModel.capacityWithAllowedReplicaMovesFor(resource, optimizationOptions);
```
Lincong, line 264: Is there any chance for the denominator (return value of the below method) to be 0?
```
clusterModel.capacityWithAllowedReplicaMovesFor(resource, optimizationOptions);
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
```diff
@@ -257,42 +248,48 @@ public String toString() {
...
+   * @param optimizationOptions Options to take into account while retrieving cluster capacity.
    */
-  private void utilizationForResources(ClusterModel clusterModel) {
+  private void utilizationForResources(ClusterModel clusterModel, OptimizationOptions optimizationOptions) {
     // Average, maximum, and standard deviation of utilization by resource.
-    Map<Resource, Double> avgUtilizationByResource = new HashMap<>();
-    Map<Resource, Double> maxUtilizationByResource = new HashMap<>();
-    Map<Resource, Double> minUtilizationByResource = new HashMap<>();
-    Map<Resource, Double> stDevUtilizationByResource = new HashMap<>();
+    Map<Resource, Double> avgUtilizationByResource = new HashMap<>(Resource.cachedValues().size());
```
Lincong, line 258: Nit: I'd do:
```
final int resourceCachedValueSize = Resource.cachedValues().size();
Map<Resource, Double> maxUtilizationByResource = new HashMap<>(resourceCachedValueSize);
...
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
```diff
@@ -257,42 +248,48 @@ public String toString() {
...
+      double resourceUtilization = clusterModel.load().expectedUtilizationFor(resource);
+      double avgUtilizationPercentage = resourceUtilization / clusterModel.capacityWithAllowedReplicaMovesFor(resource, optimizationOptions);
       double balanceUpperThreshold = avgUtilizationPercentage * _balancingConstraint.resourceBalancePercentage(resource);
       double balanceLowerThreshold = avgUtilizationPercentage * Math.max(0, (2 - _balancingConstraint.resourceBalancePercentage(resource)));
       // Maximum, minimum, and standard deviation utilization for the resource.
       double hottestBrokerUtilization = 0.0;
       double coldestBrokerUtilization = Double.MAX_VALUE;
       double varianceSum = 0.0;
-      int numBalancedBrokers = 0;
+      int numBalancedBrokersInBrokersAllowedReplicaMove = 0;
```
Lincong, line 271: Nit: Maybe it's a personal preference. But I'd name this variable `numBalancedAllowedReplocaMoveBrokers`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ExcludedBrokersForReplicaMoveTest.java
```diff
@@ -162,30 +165,30 @@
                                                          CpuUsageDistributionGoal.class)) {
       // Test: With single excluded broker, balance not satisfiable cluster, no dead broker (No exception, No proposal
       // for excluded broker, Not expected to look optimized)
```
Lincong, line 167: I think we should change the comment to be `Expected to look optimized` (drop the "Not")

Lincong (changes requested):
Some thoughts irrelevant to this PR:

Can we consider using the `final` keyword for immutable numerical values when they are used for non-trivial calculation? I noticed that CC often does non-trivial calculations with numerical values. If some variables are marked as `final`, it's safer, easier to interpret, and even improve performance since the Java compiler can apply additional optimization when it knows something (primitive types in this case) is immutable.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationOptions.java
efeg: Sure, that sounds good. Created https://github.com/linkedin/cruise-control/issues/1387 to make this case globally rather than case by case.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
efeg: TL;DR. Yes, and that is expected and acceptable for cluster model stats.

--
Model stats for resource utilization presents `min`, `max`, `std`, and `average`. Division by zero makes the average and the std infinity. This can happen when the cluster has no load to balance. `ExcludedBrokersForReplicaMoveTest` has test cases that triggers this scenario (e.g. please see `tid=2` for resource distribution goals).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
efeg: I'd prefer not using the proposed approach -- please see my answer on when I think we should use the `final` keyword.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
efeg: Since we are counting the number of balanced brokers in `_brokersAllowedReplicaMove`, I'd prefer to keep the current name.

efeg:
> Some thoughts irrelevant to this PR:

> Can we consider using the `final` keyword for immutable numerical values when they are used for non-trivial calculation? I noticed that CC often does non-trivial calculations with numerical values. If some variables are marked as `final`, it's safer, easier to interpret, and even improve performance since the Java compiler can apply additional optimization when it knows something (primitive types in this case) is immutable. See SO discussion below.

> https://stackoverflow.com/questions/4279420/does-use-of-final-keyword-in-java-improve-the-performance#:~:text=Short%20answer%3A%20don't%20worry,end%20result%20in%20faster%20code.

Thanks for sharing your thoughts @Lincong!
I'd prefer not making a blanket statement about the use of `final` for what you propose. I think such an approach tends to yield poor readability w/o much performance gain. In the SO link you shared, the accepted answer by [Jon Skeet](https://stackoverflow.com/users/22656/jon-skeet) aligns with my opinion on this matter -- in particular,

> To my mind you should use final based on clear design and readability rather than for performance reasons. If you want to change anything for performance reasons, you should perform appropriate measurements before bending the clearest code out of shape - that way you can decide whether any extra performance achieved is worth the poorer readability/design. (In my experience it's almost never worth it; YMMV.)

And just as Jon Skeet, I also support the use of final for (1) fields and (2) classes.

Lincong:
> > Some thoughts irrelevant to this PR:
> 
> > Can we consider using the `final` keyword for immutable numerical values when they are used for non-trivial calculation? I noticed that CC often does non-trivial calculations with numerical values. If some variables are marked as `final`, it's safer, easier to interpret, and even improve performance since the Java compiler can apply additional optimization when it knows something (primitive types in this case) is immutable. See SO discussion below.
> 
> > https://stackoverflow.com/questions/4279420/does-use-of-final-keyword-in-java-improve-the-performance#:~:text=Short%20answer%3A%20don't%20worry,end%20result%20in%20faster%20code.
> 
> Thanks for sharing your thoughts @Lincong!
> I'd prefer not making a blanket statement about the use of `final` for what you propose. I think such an approach tends to yield poor readability w/o much performance gain. In the SO link you shared, the accepted answer by [Jon Skeet](https://stackoverflow.com/users/22656/jon-skeet) aligns with my opinion on this matter -- in particular,
> 
> > To my mind you should use final based on clear design and readability rather than for performance reasons. If you want to change anything for performance reasons, you should perform appropriate measurements before bending the clearest code out of shape - that way you can decide whether any extra performance achieved is worth the poorer readability/design. (In my experience it's almost never worth it; YMMV.)
> 
> And just as Jon Skeet, I also support the use of final for (1) fields and (2) classes.

The performance factor only constitutes ~5% of my argument for using the `final` keyword. The more important 95% part is actually about `reasonability`. Since CC codebase does lots of numerical computations and I think a clear pattern could be:
```
final a = ...
final b = ...
final c = ...
...
d = someOp(a, b, ...)
e = someOp(c, ...)
...
return someOp(d, e, ...)
```
In the above example, all immutable variables are created upfront, followed by mutable variables that could be calculated from these immutable ones as intermediate results. In this pattern, readers know which variables are immutable and which ones are immutable. To me, knowing which variables are immutable is HUGE.
Another thing is about intermediate results (e.g. `d`, `e` in the above example). I have 2 thoughts:

1. Intermediate results should be stored in their own variables so that the calculating process can be easy to understand. I personally don't like long, chained, even nested computation. For example, `a * b + c / (e + f - g)` scares me away.
2. Immutable intermediate results should be marked as `final`.

As for readability, could you please explain how exactly readability gets hurt by adding the `final` keyword? The only reason I can think of is that it requires slightly more brainpower to process and realize "oh, this variable (or this object's reference) is not mutable". But I suggest that we only do it when that realization itself is helpful.
Looking at a larger picture, `final` in Java is similar to C++`const`, Scala `val` (I know they have differences) which are used so commonly in their own (C++ or Scala) codebases and that means proves this concept has been widely accepted and applied. I believe most people would feel the same way as I do when they see a `final` keyword indicating immutability, that feeling is safe and happy.

Lastly, commenting on your statement `I also support the use of final for (1) fields and (2) classes`.
1. Using final for fields makes perfect sense and that's how it should be done IMO
2. Using `final` for classes??? I know we can do that and I have seen other codebases doing it. But why? Why conveying the message that a class cannot be inherited? I cannot think of a very strong use case that a class will and should never be extended. The `final class` prevents many software design patterns from being applied and it violates the `open-closed principle` since it makes the class NOT open for extension to some degree.

Lincong (approved):

efeg:
@Lincong This discussion is orthogonal to this PR -- if you have further questions, let's continue on a separate channel.

PR #1386, opened 2020-11-15
Create a new producer if metrics report gets an SSL authentication exception when it tries to send metrics

### LINKED ISSUE #1148: Dynamically reload metricsreporter keystore (and truststore)
We use short lived certificates to authenticate to Kafka. Is it possible to tell the metricsreporter to refresh it's certificates (keystore/truststore) without stopping the Kafka process?
We have tried the following:
```
/opt/kafka/bin/kafka-configs.sh --bootstrap-server ${bootstrap-server} --command-config /opt/kafka/config/kafka-client.properties --entity-type brokers --entity-name $(grep broker.id /data/kafka/meta.properties | awk -F "=" '{print$2}') --alter --add-config listener.name.ssl.truststore.location=/opt/kafka/config/kafka-keystore.jks,listener.name.ssl.keystore.location=/opt/kafka/config/kafka-keystore.jks,cruise.control.metrics.reporter.ssl.keystore.location=/opt/kafka/config/kafka-keystore.jks,cruise.control.metrics.reporter.ssl.truststore.location=/opt/kafka/config/kafka-keystore.jks
```
Logs: 
```
[2020-03-17 10:10:35,126] ERROR [Producer clientId=CruiseControlMetricsReporter] Connection to node 1026 (${broker}/${broker_ip}:9093) failed authentication due to: SSL handshake failed (org.apache.kafka.clients.NetworkClient)
```
This does work for the Kafka broker, but not for the metricsreporter.
Any idea on how to do this?

Lincong:
This PR resolves #1148.

### DIFF
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -38,15 +38,14 @@ import org.apache.kafka.clients.admin.DescribeConfigsResult;
 import org.apache.kafka.clients.admin.NewPartitions;
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.admin.TopicDescription;
-import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
-import org.apache.kafka.clients.producer.RecordMetadata;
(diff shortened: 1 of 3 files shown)


efeg (changes requested):
Ideally Kafka producers/consumers should provide a cleaner way to refresh certificates. Destroying the client and creating a new one does not seem very elegant. From what I can tell, the broker-side of this issue has already been addressed (see [KAFKA-8336](https://issues.apache.org/jira/browse/KAFKA-8336)).

1. Can we create a Kafka ticket about the client side (if one does not exist already)?
2. I might be missing something, but why do we expect this PR to resolve the linked issue? Looks like if there is an `SslAuthenticationException`, the new producer is created using the old/existing `keystore` and `truststore` values cached in memory via `_producerProps`?

efeg:
@Lincong Closing this PR as it does not solve the linked issue -- and solving it cleanly requires Kafka-side support.

PR #1390, opened 2020-11-17
Convert traditional null checks to use Utils.validateNotNull

### LINKED ISSUE #1387: Convert traditional null checks to use Objects.requireNonNull
The current codebase uses traditional null checks:
```
    if (object == null) {
        throw new IllegalArgumentException("object cannot be null!");
    }
```

Adopt the following whenever appropriate, which would make the code easier to read:
```
  public static <T> T validateNotNull(T obj, String objName) {
    Objects.requireNonNull(obj, objName + " cannot be null.");
    return obj;
  }
```

aminburhan:
This PR resolves #1387.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/BalancingAction.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/BalancingAction.java
@@ -129,7 +129,7 @@ public class BalancingAction {
   private void validate() {
     switch (_actionType) {
       case REPLICA_ADDITION:
-        validateNotNull(_destinationBrokerId, "The destination broker cannot be null for balancing action " + this);
+        validateNotNull(_destinationBrokerId, () -> "The destination broker cannot be null for balancing action " + this);
         if (_sourceBrokerId != null) {
           throw new IllegalArgumentException("The source broker should be null for balancing action " + this);
         }
@@ -138,20 +138,20 @@ public class BalancingAction {
         if (_destinationBrokerId != null) {
           throw new IllegalArgumentException("The destination broker should be null for balancing action " + this);
         }
-        validateNotNull(_sourceBrokerId, "The source broker cannot be null for balancing action " + this);
+        validateNotNull(_sourceBrokerId, () -> "The source broker cannot be null for balancing action " + this);
         break;
       case INTER_BROKER_REPLICA_MOVEMENT:
       case LEADERSHIP_MOVEMENT:
       case INTER_BROKER_REPLICA_SWAP:
-        validateNotNull(_destinationBrokerId, "The destination broker cannot be null for balancing action " + this);
-        validateNotNull(_sourceBrokerId, "The source broker cannot be null for balancing action " + this);
+        validateNotNull(_destinationBrokerId, () -> "The destination broker cannot be null for balancing action " + this);
+        validateNotNull(_sourceBrokerId, () -> "The source broker cannot be null for balancing action " + this);
         break;
       case INTRA_BROKER_REPLICA_MOVEMENT:
       case INTRA_BROKER_REPLICA_SWAP:
-        validateNotNull(_destinationBrokerId, "The destination broker cannot be null for balancing action " + this);
-        validateNotNull(_sourceBrokerId, "The source broker cannot be null for balancing action " + this);
-        validateNotNull(_sourceBrokerLogdir, "The source disk cannot be null for balancing action " + this);
-        validateNotNull(_destinationBrokerLogdir, "The destination disk cannot be null for balancing action " + this);
+        validateNotNull(_destinationBrokerId, () -> "The destination broker cannot be null for balancing action " + this);
+        validateNotNull(_sourceBrokerId, () -> "The source broker cannot be null for balancing action " + this);
+        validateNotNull(_sourceBrokerLogdir, () -> "The source disk cannot be null for balancing action " + this);
+        validateNotNull(_destinationBrokerLogdir, () -> "The destination disk cannot be null for balancing action " + this);
         if (!_sourceBrokerId.equals(_destinationBrokerId)) {
           throw new IllegalArgumentException("Replica movement between disks across broker is not supported "
                                              + "for balancing action " + this);
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/BalancingAction.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/BalancingAction.java
@@ -10,6 +10,8 @@ import org.apache.kafka.common.TopicPartition;
 import java.util.HashMap;
 import java.util.Map;
 
+import static com.linkedin.cruisecontrol.common.utils.Utils.validateNotNull;
+
 
 /**
  * Represents the load balancing operation over a replica for Kafka Load GoalOptimizer.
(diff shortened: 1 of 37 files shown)


Lincong (changes requested):
Thanks for the PR. How about considering adding the below static methods to the `Utils` class.

```
  public static <T> T validateNotNull(T obj, String errorMsg) {
    Objects.requireNonNull(obj, errorMsg);
    return obj;
  }

  public static <T> T validateNotNull(T obj, Supplier<String> errorMsgSupplier) {
    Objects.requireNonNull(obj, errorMsgSupplier.get());
    return obj;
  }
```

With the above two methods, we can do:

```
Utils.validateNotNull(someObj, "Hey! This object cannot be null");
```
or with lazy evaluation
```
Utils.validateNotNull(someObj, () -> "This error message is generated only when the exception needs to be thrown");
```

Another thing to note is that I think there are more cases that can be replaced with the above methods. I tried to search `cannot be null` in the codebase.

Let me know your opinions. Thanks!

aminburhan:
Thanks for your feedback. My apologies for misunderstanding the issue description, where it's clearly stated that a static method should be adopted. 

I agree with your suggested changes, which will make the code easier to read.  I'll implement the changes  and see if there's more cases that can be replaced. Thanks!

Lincong:
> Thanks for your feedback. My apologies for misunderstanding the issue description, where it's clearly stated that a static method should be adopted.
> 
> I agree with your suggested changes, which will make the code easier to read. I'll implement the changes and see if there's more cases that can be replaced. Thanks!

Thanks. Another thing to note is that most null check currently throws `IllegalArguementException`. When we start using `Objects.requireNotNull()`, it throws `NullPointException` instead. In theory, if there's any logic that catches `IllegalArguementException` and handles it in some ways, that logic will not work with `NullPointException` anymore since it does not catch it. It's very rare (I personally have not seen any) to try and catch `IllegalArguementException`. Still want to point it out.

aminburhan:
The previous traditional null checks before change throws an `IllegalArgumentExpection` if the checked object reference is null.

```
    if (object == null) {
        throw new IllegalArgumentException("object cannot be null!");
    }
```

However the `Objects.requireNonNull` throws a `NullPointerException`.  In order not to break any previous code, in case there are tests that expects an `IllegalArgumentException` to be thrown, or any code that specifically catches `IllegalArgumentException` (however unlikely), I have modified the static methods to throw a `IllegalArgumentException` instead?
 
```
public static <T> T validateNotNull(T obj, String errorMsg) {
    if (obj == null)
      throw new IllegalArgumentException(errorMsg);
    return obj;
   }

 public static <T> T validateNotNull(T obj, Supplier<String> errorMsgSupplier) {
    if (obj == null)
      throw new IllegalArgumentException(errorMsgSupplier.get());
    return obj;
  }

```

The alternative is to use the `Objects.requireNonNull` method, catch the `NullPointerException` and then rethrow a IllegalArgumentException. But I believe that's unnecessary.

```
 public static <T> T validateNotNull(T obj, String errorMsg) {
    try {
      Objects.requireNonNull(obj, errorMsg);
    } catch (NullPointerException npe) {
      throw new IllegalArgumentException(npe.getMessage());
    }
    return obj;
  }

  public static <T> T validateNotNull(T obj, Supplier<String> errorMsgSupplier) {
    try {
      Objects.requireNonNull(obj, errorMsgSupplier.get());
    } catch (NullPointerException npe) {
      throw new IllegalArgumentException(npe.getMessage());
    }
    return obj;
  }
```

aminburhan:
> It's very rare (I personally have not seen any) to try and catch `IllegalArgu`

Thank you. I was just drafting a comment on this..

aminburhan:
> > Thanks for your feedback. My apologies for misunderstanding the issue description, where it's clearly stated that a static method should be adopted.
> > I agree with your suggested changes, which will make the code easier to read. I'll implement the changes and see if there's more cases that can be replaced. Thanks!
> 
> Thanks. Another thing to note is that most null check currently throws `IllegalArguementException`. When we start using `Objects.requireNotNull()`, it throws `NullPointException` instead. In theory, if there's any logic that catches `IllegalArguementException` and handles it in some ways, that logic will not work with `NullPointException` anymore since it does not catch it. It's very rare (I personally have not seen any) to try and catch `IllegalArguementException`. Still want to point it out.

Thank you. I was just drafting a comment on this..

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/BalancingAction.java
```diff
@@ -127,41 +129,29 @@ private BalancingAction(TopicPartition sourceTp,
   private void validate() {
     switch (_actionType) {
       case REPLICA_ADDITION:
-        if (_destinationBrokerId == null) {
-          throw new IllegalArgumentException("The destination broker cannot be null for balancing action " + this);
-        } else if (_sourceBrokerId != null) {
+        validateNotNull(_destinationBrokerId, "The destination broker cannot be null for balancing action " + this);
```
Lincong, line 132: I think this is exactly the case where the lazy-evaluation `validateNotNull(T obj, Supplier<String> errorMsgSupplier)` method needs to be used since in the previous version, `"The destination broker cannot be null for balancing action " + this` does not get executed unless `_destinationBrokerId == null` is true. So, the change should be:

validateNotNull(_destinationBrokerId, () -> "The destination broker cannot be null for balancing action " + this);

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/BalancingAction.java
```diff
@@ -127,41 +129,29 @@ private BalancingAction(TopicPartition sourceTp,
...
           throw new IllegalArgumentException("The source broker should be null for balancing action " + this);
         }
         break;
       case REPLICA_DELETION:
         if (_destinationBrokerId != null) {
           throw new IllegalArgumentException("The destination broker should be null for balancing action " + this);
-        } else if (_sourceBrokerId == null) {
-      throw new IllegalArgumentException("The source broker cannot be null for balancing action " + this);
         }
+        validateNotNull(_sourceBrokerId, "The source broker cannot be null for balancing action " + this);
```
Lincong, line 141: Same issue. Use lazy evaluation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/BalancingAction.java
```diff
@@ -127,41 +129,29 @@ private BalancingAction(TopicPartition sourceTp,
...
         break;
       case INTER_BROKER_REPLICA_MOVEMENT:
       case LEADERSHIP_MOVEMENT:
       case INTER_BROKER_REPLICA_SWAP:
-        if (_destinationBrokerId == null) {
-          throw new IllegalArgumentException("The destination broker cannot be null for balancing action " + this);
-        } else if (_sourceBrokerId == null) {
-          throw new IllegalArgumentException("The source broker cannot be null for balancing action " + this);
-        }
+        validateNotNull(_destinationBrokerId, "The destination broker cannot be null for balancing action " + this);
```
Lincong, line 146: Same issue. Use lazy evaluation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationOptions.java
```diff
@@ -73,15 +75,10 @@ public OptimizationOptions(Set<String> excludedTopics,
...
-    if (excludedTopics == null) {
-      throw new IllegalArgumentException("Excluded topics cannot be null.");
-    } else if (excludedBrokersForLeadership == null) {
-      throw new IllegalArgumentException("Excluded brokers for leadership cannot be null.");
-    } else if (excludedBrokersForReplicaMove == null) {
-      throw new IllegalArgumentException("Excluded brokers for replica move cannot be null.");
-    } else if (requestedDestinationBrokerIds == null) {
-      throw new IllegalArgumentException("Requested destination broker ids cannot be null.");
-    }
+    validateNotNull(excludedTopics, "Excluded topics cannot be null.");
```
Lincong, line 78: Since `validateNotNull ` returns the object if it's not null, you can do:

_excludedTopics = validateNotNull(excludedTopics, "Excluded topics cannot be null.");
_excludedBrokersForLeadership = validateNotNull(excludedBrokersForLeadership, "Excluded brokers for leadership cannot be null.");
...

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyUtils.java
```diff
@@ -35,10 +36,8 @@ private AnomalyUtils() {
   public static KafkaCruiseControl extractKafkaCruiseControlObjectFromConfig(Map<String, ?> configs,
                                                                             AnomalyType anomalyType) {
     KafkaCruiseControl kafkaCruiseControl = (KafkaCruiseControl) configs.get(KAFKA_CRUISE_CONTROL_OBJECT_CONFIG);
-    if (kafkaCruiseControl == null) {
-      throw new IllegalArgumentException(String.format("Missing %s when creating anomaly of type %s.",
-                                                       KAFKA_CRUISE_CONTROL_OBJECT_CONFIG, anomalyType));
-    }
+    validateNotNull(kafkaCruiseControl, String.format("Missing %s when creating anomaly of type %s.",
```
Lincong, line 39: We can do `return validateNotNull(...)`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -739,12 +740,8 @@ public synchronized void setGeneratingProposalsForExecution(String uuid, Supplie
     if (currentExecutorState != NO_TASK_IN_PROGRESS) {
       throw new OngoingExecutionException(String.format("Cannot generate proposals while the executor is in %s state.", currentExecutorState));
     }
-    if (uuid == null) {
-      throw new IllegalArgumentException("UUID of the execution cannot be null.");
-    }
-    if (reasonSupplier == null) {
-      throw new IllegalArgumentException("Reason supplier cannot be null.");
-    }
+    validateNotNull(uuid, "UUID of the execution cannot be null.");
```
Lincong, line 743: Same issue: `_uuid  = validateNotNull(uuid, "UUID of the execution cannot be null.");`

Lincong (changes requested):
There are 2 things to improve.

1. Use lazy evaluation when the error message is dynamically generated
2. Use the return value from the `validateNotNull` directly

Could you please fix all cases to which the above two issues apply? Thanks!

aminburhan:
> There are 2 things to improve.
> 
> 1. Use lazy evaluation when the error message is dynamically generated
> 2. Use the return value from the `validateNotNull` directly
> 
> Could you please fix all cases to which the above two issues apply? Thanks!

I have implemented the suggested improvements. 

There are more cases that can be replaced, however these null checks throws other types of exceptions.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
```diff
@@ -354,11 +355,9 @@ public static void sanityCheckOfflineReplicaPresence(ClusterModel clusterModel)
     private final Set<Integer> _recentlyDemotedBrokers;
 
     public RecentBrokers(Set<Integer> recentlyRemovedBrokers, Set<Integer> recentlyDemotedBrokers) {
-      if (recentlyRemovedBrokers == null || recentlyDemotedBrokers == null) {
-        throw new IllegalArgumentException("Attempt to set a null value for recent brokers.");
-      }
-      _recentlyRemovedBrokers = recentlyRemovedBrokers;
-      _recentlyDemotedBrokers = recentlyDemotedBrokers;
+      String errorMsg = "Attempt to set a null value for recent brokers.";
```
Lincong, line 358: Instead of reusing the error message, I think we can be more specific by:

```
_recentlyRemovedBrokers = validateNotNull(recentlyRemovedBrokers, "Attempt to set a null value for recent removed brokers.");
_recentlyDemotedBrokers = validateNotNull(recentlyDemotedBrokers, "Attempt to set a null value for recent demoted brokers.");
```

Lincong (changes requested):
One last nit issue

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RunnableUtils.java
aminburhan: Thanks for your input. I've made the error message more specific.

Lincong (approved):
LGTM. Thanks for your PR

aminburhan:
> LGTM. Thanks for your PR. Could you please squash and merge this PR to master after modifying the commit message to reflect the actual change. The current commit message `Converts traditional null checks to use Objects.requireNonNull` is not accurate.

I don't have write access to this repository.

Lincong:
> > LGTM. Thanks for your PR. Could you please squash and merge this PR to master after modifying the commit message to reflect the actual change. The current commit message `Converts traditional null checks to use Objects.requireNonNull` is not accurate.
> 
> I don't have write access to this repository.

I just merged it to master. Thanks for your contribution!

PR #1391, opened 2020-11-20
Calculate balance lower bound for resource distribution lower bound with low utilization threshold

### LINKED ISSUE #1147: DISK, CPU and NW_OUT low utilization thresholds are not used

Lincong:
This PR resolves #1147 .

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
@@ -5,6 +5,7 @@
 
 package com.linkedin.kafka.cruisecontrol.analyzer.goals;
 
+import com.linkedin.kafka.cruisecontrol.analyzer.BalancingConstraint;
 import com.linkedin.kafka.cruisecontrol.analyzer.OptimizationOptions;
 import com.linkedin.kafka.cruisecontrol.analyzer.ActionType;
 import com.linkedin.kafka.cruisecontrol.common.Resource;
@@ -485,5 +486,71 @@ public class GoalUtils {
     }
 
     return sb.toString();
+  }
+
+  /**
+   * Compute the utilization upper/lower threshold in percent for the given type of resource
+   *
+   * @param avgUtilizationPercentage Average cluster utilization that excludes the capacity of brokers excluded for replica moves.
+   * @param resource {@link Resource}
+   * @param balancingConstraint balancing contraints
+   * @param isTriggeredByGoalViolation Options to adjust balance percentage with margin in case goal optimization is triggered
+   * by goal violation detector.
+   * @param balanceMargin resource distribution goal balance margin
+   * @param isLowerThreshold whether this method calculates resource utilization threshold upper bound or lower bound
+   * @return The utilization upper/lower threshold in percent for the given type of resource
+   */
+  public static double computeResourceUtilizationBalanceThreshold(double avgUtilizationPercentage,
+                                                                  Resource resource,
+                                                                  BalancingConstraint balancingConstraint,
+                                                                  boolean isTriggeredByGoalViolation,
+                                                                  double balanceMargin,
+                                                                  boolean isLowerThreshold) {
+
+    if (avgUtilizationPercentage >= 1) {
+      throw new IllegalArgumentException("Average resource utilization percentage must be less than 1.0. Got: " + avgUtilizationPercentage);
+    }
+    if (balanceMargin >= 1) {
+      throw new IllegalArgumentException("Balance margin must be less than 1.0. Got: " + balanceMargin);
+    }
+    boolean isLowUtilization = avgUtilizationPercentage <= balancingConstraint.lowUtilizationThreshold(resource);
+
+    if (isLowerThreshold) {
+      if (isLowUtilization) {
+        return 0.0;
+      }
+      double balancePercentageWithMargin =
+          balancePercentageWithMargin(isTriggeredByGoalViolation, balancingConstraint, resource, balanceMargin);
+      return avgUtilizationPercentage * Math.max(0, balancePercentageWithMargin);
+
+    } else {
+      if (isLowUtilization) {
+        return balancingConstraint.lowUtilizationThreshold(resource) * balanceMargin;
+      }
+      double balancePercentageWithMargin =
+          balancePercentageWithMargin(isTriggeredByGoalViolation, balancingConstraint, resource, balanceMargin);
+      return avgUtilizationPercentage * (1 + balancePercentageWithMargin);
+    }
+  }
+
+  /**
+   * To avoid churns, we add a balance margin to the user specified rebalance threshold. e.g. when user sets the
+   * threshold to be resourceBalancePercentage, we use (resourceBalancePercentage-1)*balanceMargin instead.
+   *
+   * @param isTriggeredByGoalViolation Options to adjust balance percentage with margin in case goal optimization is triggered
+   * by goal violation detector.
+   * @return The rebalance threshold with a margin.
+   */
+  private static double balancePercentageWithMargin(boolean isTriggeredByGoalViolation,
+      BalancingConstraint balancingConstraint,
+      Resource resource,
+      double balanceMargin) {
+
+    double balancePercentage = balancingConstraint.resourceBalancePercentage(resource);
+    if (isTriggeredByGoalViolation) {
+      // Make the balance threshold percentage more lenient in this case
+      balancePercentage *= balancingConstraint.goalViolationDistributionThresholdMultiplier();
+    }
+    return (balancePercentage - 1) * balanceMargin;
   }
 }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
@@ -507,9 +507,6 @@ public class GoalUtils {
                                                                   double balanceMargin,
                                                                   boolean isLowerThreshold) {
 
-    if (avgUtilizationPercentage >= 1) {
-      throw new IllegalArgumentException("Average resource utilization percentage must be less than 1.0. Got: " + avgUtilizationPercentage);
-    }
     if (balanceMargin >= 1) {
       throw new IllegalArgumentException("Balance margin must be less than 1.0. Got: " + balanceMargin);
     }
@@ -521,7 +518,7 @@ public class GoalUtils {
       }
       double balancePercentageWithMargin =
           balancePercentageWithMargin(isTriggeredByGoalViolation, balancingConstraint, resource, balanceMargin);
-      return avgUtilizationPercentage * Math.max(0, balancePercentageWithMargin);
+      return avgUtilizationPercentage * Math.max(0, (1 - balancePercentageWithMargin));
 
     } else {
       if (isLowUtilization) {
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
@@ -489,13 +489,21 @@ public class GoalUtils {
   }
 
   /**
-   * Compute the utilization upper/lower threshold in percent for the given type of resource
+   * Compute the utilization upper/lower threshold in percent for the given type of resource.
+   *  There are two cases described below (note that this description demonstrates the general idea):
+   *
+   *    1. Low utilization case (the average utilization percentage is not greater than the low utilization threshold
+   *      Threshold lower bound = zero
+   *      Threshold upper bound = the low utilization threshold multiplied with the balance margin
+   *
+   *    2. Not low utilization case
+   *      Threshold lower bound = average utilization * (1 - balance percentage * balance margin)
+   *      Threshold upper bound = average utilization * (1 + balance percentage * balance margin)
    *
    * @param avgUtilizationPercentage Average cluster utilization that excludes the capacity of brokers excluded for replica moves.
    * @param resource {@link Resource}
-   * @param balancingConstraint balancing contraints
-   * @param isTriggeredByGoalViolation Options to adjust balance percentage with margin in case goal optimization is triggered
-   * by goal violation detector.
+   * @param balancingConstraint Balancing constraint
+   * @param isTriggeredByGoalViolation whether this computation is triggered by goal violation.
    * @param balanceMargin resource distribution goal balance margin
    * @param isLowerThreshold whether this method calculates resource utilization threshold upper bound or lower bound
    * @return The utilization upper/lower threshold in percent for the given type of resource
@@ -534,14 +542,16 @@ public class GoalUtils {
    * To avoid churns, we add a balance margin to the user specified rebalance threshold. e.g. when user sets the
    * threshold to be resourceBalancePercentage, we use (resourceBalancePercentage-1)*balanceMargin instead.
    *
-   * @param isTriggeredByGoalViolation Options to adjust balance percentage with margin in case goal optimization is triggered
-   * by goal violation detector.
+   * @param isTriggeredByGoalViolation whether this computation is triggered by goal violation.
+   * @param balancingConstraint Balancing constraint
+   * @param resource Resource for which the balance percentage with margin gets calculated
+   * @param balanceMargin resource distribution goal balance margin
    * @return The rebalance threshold with a margin.
    */
   private static double balancePercentageWithMargin(boolean isTriggeredByGoalViolation,
-      BalancingConstraint balancingConstraint,
-      Resource resource,
-      double balanceMargin) {
+                                                    BalancingConstraint balancingConstraint,
+                                                    Resource resource,
+                                                    double balanceMargin) {
 
     double balancePercentage = balancingConstraint.resourceBalancePercentage(resource);
     if (isTriggeredByGoalViolation) {
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
@@ -829,14 +829,6 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
     }
   }
 
-  private boolean isLoadUnderLowUtilizationThreshold(Broker broker) {
-    double utilization = resource().isHostResource() ? broker.host().load().expectedUtilizationFor(resource())
-                                                     : broker.load().expectedUtilizationFor(resource());
-    double resourceCapacity = resource().isHostResource() ? broker.host().capacityFor(resource())
-                                                          : broker.capacityFor(resource());
-    return utilization / resourceCapacity < _balancingConstraint.lowUtilizationThreshold(resource());
-  }
-
   private boolean isLoadUnderBalanceUpperLimitAfterChange(Load load,
                                                           Broker broker,
                                                           ChangeType changeType,
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
@@ -969,7 +969,8 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
    * @return The utilization lower threshold in percent for the {@link #resource()}
    */
   private double computeBalanceLowerThreshold(double clusterUtilization, OptimizationOptions optimizationOptions) {
-    return clusterUtilization * Math.max(0, (1 - balancePercentageWithMargin(optimizationOptions)));
+    return Math.max(clusterUtilization * Math.max(0, (1 - balancePercentageWithMargin(optimizationOptions))),
+                    _balancingConstraint.lowUtilizationThreshold(resource()));
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
@@ -969,8 +969,7 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
    * @return The utilization lower threshold in percent for the {@link #resource()}
    */
   private double computeBalanceLowerThreshold(double clusterUtilization, OptimizationOptions optimizationOptions) {
-    return Math.max(clusterUtilization * Math.max(0, (1 - balancePercentageWithMargin(optimizationOptions))),
-                    _balancingConstraint.lowUtilizationThreshold(resource()));
+    return clusterUtilization * Math.max(0, (1 - balancePercentageWithMargin(optimizationOptions)));
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
@@ -255,15 +255,17 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
   @Override
   protected void updateGoalState(ClusterModel clusterModel, OptimizationOptions optimizationOptions)
       throws OptimizationFailureException {
+
+    if (clusterModel.aliveBrokers().stream().allMatch(this::isLoadUnderLowUtilizationThreshold)) {
+      finish();
+      return;
+    }
+
     Set<Integer> brokerIdsAboveBalanceUpperLimit = new HashSet<>();
     Set<Integer> brokerIdsUnderBalanceLowerLimit = new HashSet<>();
     // Log broker Ids over balancing limit.
     // While proposals exclude the excludedTopics, the balance still considers utilization of the excludedTopic replicas.
     for (Broker broker : clusterModel.aliveBrokers()) {
-      if (isLoadUnderLowUtilizationThreshold(broker)) {
-        continue;
-      }
-
       if (!isLoadUnderBalanceUpperLimit(broker)) {
         brokerIdsAboveBalanceUpperLimit.add(broker.id());
       }
@@ -334,7 +336,9 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
                                     Set<Goal> optimizedGoals,
                                     OptimizationOptions optimizationOptions) {
 
-    if (clusterModel.deadBrokers().isEmpty() && allAliveBrokersBalanced(clusterModel, optimizationOptions)) {
+    if (clusterModel.deadBrokers().isEmpty()
+        && allAliveBrokersBalanced(clusterModel, optimizationOptions)
+        && !hasBadDisksOnBrokers(clusterModel.aliveBrokers())) {
       return;
     }
 
@@ -389,6 +393,10 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
     if (!unbalanced) {
       LOG.debug("Successfully balanced {} for broker {} by moving leaders and replicas.", resource(), broker.id());
     }
+  }
+
+  private boolean hasBadDisksOnBrokers(Set<Broker> brokers) {
+    return brokers.stream().anyMatch(Broker::hasBadDisks);
   }
 
   private boolean allAliveBrokersBalanced(ClusterModel clusterModel, OptimizationOptions optimizationOptions) {
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
@@ -260,6 +260,10 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
     // Log broker Ids over balancing limit.
     // While proposals exclude the excludedTopics, the balance still considers utilization of the excludedTopic replicas.
     for (Broker broker : clusterModel.aliveBrokers()) {
+      if (isLoadUnderLowUtilizationThreshold(broker)) {
+        continue;
+      }
+
       if (!isLoadUnderBalanceUpperLimit(broker)) {
         brokerIdsAboveBalanceUpperLimit.add(broker.id());
       }
@@ -329,6 +333,11 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
                                     ClusterModel clusterModel,
                                     Set<Goal> optimizedGoals,
                                     OptimizationOptions optimizationOptions) {
+
+    if (clusterModel.deadBrokers().isEmpty() && allAliveBrokersBalanced(clusterModel, optimizationOptions)) {
+      return;
+    }
+
     int numOfflineReplicas = broker.currentOfflineReplicas().size();
 
     boolean isExcludedForReplicaMove = isExcludedForReplicaMove(broker);
@@ -380,6 +389,14 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
     if (!unbalanced) {
       LOG.debug("Successfully balanced {} for broker {} by moving leaders and replicas.", resource(), broker.id());
     }
+  }
+
+  private boolean allAliveBrokersBalanced(ClusterModel clusterModel, OptimizationOptions optimizationOptions) {
+    int balancedBrokers = clusterModel
+                          .getClusterStats(_balancingConstraint, optimizationOptions)
+                          .numBalancedBrokersByResource()
+                          .get(resource());
+    return balancedBrokers == clusterModel.aliveBrokers().size();
   }
 
   private boolean rebalanceByMovingLoadIn(Broker broker,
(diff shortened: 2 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -260,6 +260,10 @@ protected void updateGoalState(ClusterModel clusterModel, OptimizationOptions op
     // Log broker Ids over balancing limit.
```
efeg, line 260: Looks like `actionAcceptance` implementation above does not take low utilization threshold into account. It should ensure that if all brokers were already under the low threshold, anything under low threshold is acceptable. Am I missing something?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -260,6 +260,10 @@ protected void updateGoalState(ClusterModel clusterModel, OptimizationOptions op
...
     // While proposals exclude the excludedTopics, the balance still considers utilization of the excludedTopic replicas.
```
efeg, line 261: If all brokers are under low utilization threshold, shouldn't we update the following in `initGoalState` function to make sure that balancing limits are set properly?

```
  private double _balanceUpperThreshold;
  private double _balanceLowerThreshold;
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -329,6 +333,11 @@ protected void rebalanceForBroker(Broker broker,
                                     ClusterModel clusterModel,
                                     Set<Goal> optimizedGoals,
                                     OptimizationOptions optimizationOptions) {
+
+    if (clusterModel.deadBrokers().isEmpty() && allAliveBrokersBalanced(clusterModel, optimizationOptions)) {
+      return;
+    }
+
```
efeg, line 340: What if there are dead disks -- i.e. broker is alive, but there are dead disks on alive brokers?
Wouldn't this cause us to drop JBOD support when there are offline disks?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
```diff
@@ -279,13 +282,23 @@ private void utilizationForResources(ClusterModel clusterModel, OptimizationOpti
...
             numBalancedBrokersInBrokersAllowedReplicaMove++;
           }
           varianceSum += Math.pow(utilization - avgUtilizationPercentage * capacity, 2);
         }
       }
-      _numBalancedBrokersByResource.put(resource, numBalancedBrokersInBrokersAllowedReplicaMove);
+
+      if (allBrokersBelowLowUtilizationThreshold) {
+        // All alive brokers are balanced
+        _numBalancedBrokersByResource.put(resource, clusterModel.aliveBrokers().size());
```
efeg, line 298: Above you are checking whether `_brokersAllowedReplicaMove` are having utilization over `lowUtilizationThreshold` or not. Here you are claiming that _all_ alive brokers are balanced. Which one is it?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
```diff
@@ -279,13 +282,23 @@ private void utilizationForResources(ClusterModel clusterModel, OptimizationOpti
           double capacity = resource.isHostResource() ? broker.host().capacityFor(resource)
                                                       : broker.capacityFor(resource);
           double utilizationPercentage = utilization / capacity;
+          if (utilizationPercentage > lowUtilizationThreshold) {
```
efeg, line 285: What if the `lowUtilizationThreshold` and `utilizationPercentage` are both 0? Wouldn't `>=` be a sufficient condition?

efeg (changes requested):
@Lincong Thanks for the patch.
Left some comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
Lincong: What exactly do you mean by `It should ensure that if all brokers were already under the low threshold, anything under low threshold is acceptable`?

I think it should be that after the proposed action, if the load is not within the range of [`lowerLimit`, `upperLimit`], instead of rejecting the action right away, we check whether the load is under the low threshold. If it is, we accept the action and vice versa. Is it correct?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
Lincong: Do you mean that if there is any alive broker having bad disks, we should continue with the normal rebalance process instead of returning here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
Lincong: Right

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: > What exactly do you mean by `It should ensure that if all brokers were already under the low threshold, anything under low threshold is acceptable`?

When `actionAcceptance` is called, suppose the cluster has 3 brokers: broker `B1` has a resource (e.g. `disk` resource) utilization under `LowUtilizationThreshold`, and brokers `B2` and `B3` are above `LowUtilizationThreshold`. This is an example cluster, where not all brokers were already under the low threshold -- i.e. `B2` and `B3` are above it. Hence, `LowUtilizationThreshold` does not apply to _any_ broker in the cluster because of how this threshold is defined in its documentation:

https://github.com/linkedin/cruise-control/blob/5ed774eb5c64f6e57a6a001c1306c520170416c5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/AnalyzerConfig.java#L141-L147

> I think it should be that after the proposed action, if the load is not within the range of [`lowerLimit`, `upperLimit`], instead of rejecting the action right away, we check whether the load is under the low threshold. If it is, we accept the action and vice versa. Is it correct?

No, because we should apply only the relevant limits. If the cluster has brokers above the lower threshold, then LowUtilizationThreshold is irrelevant. Similarly, if all brokers are under the threshold, we use [0, `LowUtilizationThreshold`] as the acceptable limits as described above.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: Yes. If we terminate early due to this new if statement, after the goal optimization we would end up with offline replicas on dead disks, don't we?
But adding this if statement causes other issues. In particular, brokers excluded for replica move no longer give away their replicas to other available brokers. Let's discuss a potential approach to resolve this issues offline.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
Lincong: After some offline discussion, we decided to only look at the overall cluster resource utilization over brokers that are not excluded for replica move instead of whether all brokers are under the low utilization threshold. The reason is that suppose we have a cluster of 3 broker A, B, and C where A and B have no resource utilization and C has some resource utilization that is slightly over the low utilization threshold. In this case, we adjust the lower/upper thresholds basing on the low utilization threshold instead of using the thresholds calculated from the original resource utilization percentage given in the config. This way, the rebalance might involve fewer operations.

In general, the gist is to set more lenient and reasonable values to the balance upper/lower threshold parameters instead of avoiding rebalance completely.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModelStats.java
Lincong: After offline discussion, this logic is not needed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: We discussed offline -- see https://github.com/linkedin/cruise-control/pull/1391#discussion_r530608090

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
```diff
@@ -0,0 +1,186 @@
...
+    // Expect rebalance on CPU usage distribution goal since one broker's CPU usage is above the low CPU utilization threshold
+    ClusterModel clusterModel = createSmallClusterModel();
+    Map<String, Object> goalConfigs = getDefaultGoalConfigs();
+    goalConfigs.put(AnalyzerConfig.CPU_LOW_UTILIZATION_THRESHOLD_CONFIG, 0.3474);
+    p.add(params(clusterModel, new CpuUsageDistributionGoal(), goalConfigs, true));
+
+    // Expect no rebalance on CPU usage distribution goal since all brokers' CPU usage is below the low CPU utilization threshold
+    clusterModel = createSmallClusterModel();
+    goalConfigs = getDefaultGoalConfigs();
+    goalConfigs.put(AnalyzerConfig.CPU_LOW_UTILIZATION_THRESHOLD_CONFIG, 0.3862);
```
efeg, line 81: Can we explain in comments how we came up with these numbers? Looks like `0.3862` is much higher than the highest broker utilization (i.e. `34.75`). Why do we add this additional buffer?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
```diff
@@ -0,0 +1,91 @@
...
+                                                                  double goalViolationDistributionThresholdMultiplier,
+                                                                  double expectedComputedBalanceThreshold) {
+
+    BalancingConstraint mockBalanceConstraint = EasyMock.mock(BalancingConstraint.class);
+    EasyMock.expect(mockBalanceConstraint.lowUtilizationThreshold(resource)).andReturn(lowUtilizationThreshold).anyTimes();
+    EasyMock.expect(mockBalanceConstraint.resourceBalancePercentage(resource)).andReturn(resourceBalancePercentage).anyTimes();
+    EasyMock.expect(mockBalanceConstraint.goalViolationDistributionThresholdMultiplier()).
+        andReturn(goalViolationDistributionThresholdMultiplier).anyTimes();
+
+    EasyMock.replay(mockBalanceConstraint);
```
efeg, line 81: Is EasyMock.verify(XXX) missing intentionally?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private void verifyComputingResourceUtilizationBalanceThreshold(double avgUtilizationPercentage,
+                                                                  Resource resource,
+                                                                  double balanceMargin,
+                                                                  boolean isLowerThreshold,
+                                                                  boolean isTriggeredByGoalViolation,
+                                                                  double lowUtilizationThreshold,
+                                                                  double resourceBalancePercentage,
+                                                                  double goalViolationDistributionThresholdMultiplier,
+                                                                  double expectedComputedBalanceThreshold) {
```
efeg, line 73: We have a lot of parameters, but seems like we only test a few. Can we either drop the constant parameters and assign the value within this function or add test cases that varies the tested parameter.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    BalancingConstraint mockBalanceConstraint = EasyMock.mock(BalancingConstraint.class);
+    EasyMock.expect(mockBalanceConstraint.lowUtilizationThreshold(resource)).andReturn(lowUtilizationThreshold).anyTimes();
+    EasyMock.expect(mockBalanceConstraint.resourceBalancePercentage(resource)).andReturn(resourceBalancePercentage).anyTimes();
+    EasyMock.expect(mockBalanceConstraint.goalViolationDistributionThresholdMultiplier()).
+        andReturn(goalViolationDistributionThresholdMultiplier).anyTimes();
```
efeg, line 79: Do we really expect this functions to be called `anyTimes()`? Can't we be more specific?

efeg (changes requested):
Thanks for the updates -- left some comments

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
```diff
@@ -486,4 +487,67 @@ public static String mitigationForOptimizationFailures(OptimizationOptions optim
...
+      return avgUtilizationPercentage * (1 + balancePercentageWithMargin);
+    }
+  }
+
+  /**
+   * To avoid churns, we add a balance margin to the user specified rebalance threshold. e.g. when user sets the
+   * threshold to be resourceBalancePercentage, we use (resourceBalancePercentage-1)*balanceMargin instead.
+   *
+   * @param isTriggeredByGoalViolation Options to adjust balance percentage with margin in case goal optimization is triggered
+   * by goal violation detector.
```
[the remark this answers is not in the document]
Lincong: Thanks for pointing it out. I'll change it to `whether this computation is triggered by goal violation`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+  private final ClusterModel _clusterModel;
+  private final ResourceDistributionGoal _resourceDistributionGoal;
+  private final Map<String, Object> _goalConfigs;
+  private final Boolean _expectRebalance;
+
+  public LowResourceUtilizationTest(ClusterModel clusterModel,
+                                    ResourceDistributionGoal resourceDistributionGoal,
+                                    Map<String, Object> goalConfigs,
+                                    Boolean expectRebalance) {
```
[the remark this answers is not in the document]
Lincong: It is not expected to be `null`. I will change it to `boolean`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
Lincong: It was missed. I will add it back

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
Lincong: I used `anyTimes()` intentionally since ensuring specific methods on this `BalancingConstraint` object gets invoked exactly N number of times is meaningless IMO.  The `BalancingConstraint` object serves as a group of constant values, as long as `computeResourceUtilizationBalanceThreshold` computes the correct value, the user of this method should not care how many times the implementation of this method tries to access constant values in the `BalancingConstraint` object.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
```diff
@@ -0,0 +1,186 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+
+/**
+ * Unit test for testing low utilization threshold taking effect to prevent rebalance for resource distribution goals
+ */
+@RunWith(Parameterized.class)
+public class LowResourceUtilizationTest {
+
+  private static final Map<Resource, Double> BROKER_CAPACITY = new HashMap<Resource, Double>() {{
```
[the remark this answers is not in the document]
Lincong: There would be a syntax error `cannot use "<>" with anonymous inner classes` if we do not specify types in `<>` in this form

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
```diff
@@ -0,0 +1,186 @@
...
+  private final ResourceDistributionGoal _resourceDistributionGoal;
+  private final Map<String, Object> _goalConfigs;
+  private final Boolean _expectRebalance;
+
+  public LowResourceUtilizationTest(ClusterModel clusterModel,
+                                    ResourceDistributionGoal resourceDistributionGoal,
+                                    Map<String, Object> goalConfigs,
+                                    Boolean expectRebalance) {
+
+    _clusterModel = clusterModel;
```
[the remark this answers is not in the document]
Lincong: Good idea

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  @Test
+  public void test() throws OptimizationFailureException {
+    Map<TopicPartition, List<ReplicaPlacementInfo>> initReplicaDistribution = _clusterModel.getReplicaDistribution();
+    Map<TopicPartition, ReplicaPlacementInfo> initLeaderDistribution = _clusterModel.getLeaderDistribution();
+
+    _resourceDistributionGoal.configure(_goalConfigs);
+    assertTrue("Failed to optimize " + _resourceDistributionGoal.name(),
+        _resourceDistributionGoal.optimize(_clusterModel, Collections.emptySet(), new OptimizationOptions(Collections.emptySet())));
```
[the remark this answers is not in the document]
Lincong: We cannot do that since `_resourceDistributionGoal.optimize(_clusterModel, ..)` changed the `_clusterModel` and the values of `initReplicaDistribution` and `initLeaderDistribution` would be different if we calculate them after the optimization

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
```diff
@@ -0,0 +1,186 @@
...
+    goalConfigs = getDefaultGoalConfigs();
+    goalConfigs.put(AnalyzerConfig.NETWORK_OUTBOUND_LOW_UTILIZATION_THRESHOLD_CONFIG, 0.14);
+    p.add(params(clusterModel, new NetworkOutboundUsageDistributionGoal(), goalConfigs, true));
+
+    // Expect no rebalance on network outbound usage distribution goal since all brokers' network outbound usage is
+    // below the low network outbound utilization threshold
+    clusterModel = createSmallClusterModel();
+    goalConfigs = getDefaultGoalConfigs();
+    goalConfigs.put(AnalyzerConfig.NETWORK_OUTBOUND_LOW_UTILIZATION_THRESHOLD_CONFIG, 0.1639);
+    p.add(params(clusterModel, new NetworkOutboundUsageDistributionGoal(), goalConfigs, false));
```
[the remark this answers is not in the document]
Lincong: No, I don't see how that can work and what that would buy us. We need a new set of default configs with relevant configs overridden for each test case.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
Lincong: I will `drop the constant parameters and assign the value within this function`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
Lincong: I changed the implementation so that the calculation process of these numbers are clear

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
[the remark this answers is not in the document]
efeg: This is supported in Java 9 and above [source](https://java2blog.com/java-9-anonymous-inner-classes-and-diamond-operator/). Since, CC is compiled with Java 11, diamond notation should work.

@Lincong I assume you were building with Java 8? I am ok to leave it as is to maintain build compatibility with Java 8.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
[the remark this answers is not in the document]
efeg: @Lincong Here is what I mean:

```
/*
 * Copyright 2020 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
 */
package com.linkedin.kafka.cruisecontrol.analyzer;

import com.linkedin.kafka.cruisecontrol.analyzer.goals.CpuUsageDistributionGoal;
import com.linkedin.kafka.cruisecontrol.analyzer.goals.DiskUsageDistributionGoal;
import com.linkedin.kafka.cruisecontrol.analyzer.goals.Goal;
import com.linkedin.kafka.cruisecontrol.analyzer.goals.NetworkInboundUsageDistributionGoal;
import com.linkedin.kafka.cruisecontrol.analyzer.goals.NetworkOutboundUsageDistributionGoal;
import com.linkedin.kafka.cruisecontrol.analyzer.goals.ResourceDistributionGoal;
import com.linkedin.kafka.cruisecontrol.common.DeterministicCluster;
import com.linkedin.kafka.cruisecontrol.common.Resource;
import com.linkedin.kafka.cruisecontrol.config.constants.AnalyzerConfig;
import com.linkedin.kafka.cruisecontrol.config.constants.ExecutorConfig;
import com.linkedin.kafka.cruisecontrol.config.constants.MonitorConfig;
import com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException;
import com.linkedin.kafka.cruisecontrol.executor.ExecutionProposal;
import com.linkedin.kafka.cruisecontrol.model.ClusterModel;
import com.linkedin.kafka.cruisecontrol.model.ReplicaPlacementInfo;
import java.util.ArrayList;
import java.util.Collection;
import java.util.Collections;
import java.util.HashMap;
import java.util.List;
import java.util.Map;
import java.util.Properties;
import java.util.Set;
import org.apache.kafka.common.TopicPartition;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.junit.runners.Parameterized;
import static org.junit.Assert.assertTrue;
import static org.junit.Assert.assertFalse;
import static com.linkedin.kafka.cruisecontrol.analyzer.AnalyzerUnitTestUtils.goal;

/**
 * Unit test for testing low utilization threshold taking effect to prevent rebalance for resource distribution goals
 */
@RunWith(Parameterized.class)
public class LowResourceUtilizationTest {

  private static final Map<Resource, Double> BROKER_CAPACITY = new HashMap<Resource, Double>() {{
    put(Resource.CPU, 200.0);
    put(Resource.DISK, 1000.0);
    put(Resource.NW_IN, 2000.0);
    put(Resource.NW_OUT, 2000.0);
  }};

  private final ResourceDistributionGoal _resourceDistributionGoal;
  private final boolean _expectRebalance;

  public LowResourceUtilizationTest(ResourceDistributionGoal resourceDistributionGoal, boolean expectRebalance) {
    _resourceDistributionGoal = resourceDistributionGoal;
    _expectRebalance = expectRebalance;
  }

  /**
   * Populate parameters to test rebalance with low utilization threshold. All brokers are alive.
   *
   * @return Parameters to test rebalance with low utilization threshold.
   */
  @Parameterized.Parameters
  public static Collection<Object[]> data() throws Exception {
    Collection<Object[]> p = new ArrayList<>();
    final double maxCpuUtilizationRatio = 0.3475;
    final double maxDiskUtilizationRatio = 0.28;
    final double maxNetworkInUtilizationRatio = 0.13;
    final double maxNetworkOutUtilizationRatio = 0.1475;
    final double balanceMargin = 0.9;

    // Expect rebalance on CPU usage distribution goal since one broker's CPU usage is above the low CPU utilization threshold
    Properties configOverrides = new Properties();
    configOverrides.put(AnalyzerConfig.CPU_LOW_UTILIZATION_THRESHOLD_CONFIG, onePercentSmaller(maxCpuUtilizationRatio));
    p.add(params(goal(CpuUsageDistributionGoal.class, configOverrides), true));

    // Expect no rebalance on CPU usage distribution goal since all brokers' CPU usage is below the low CPU utilization threshold
    configOverrides.put(AnalyzerConfig.CPU_LOW_UTILIZATION_THRESHOLD_CONFIG,
                    onePercentGreater(maxCpuUtilizationRatio / balanceMargin));
    p.add(params(goal(CpuUsageDistributionGoal.class, configOverrides), false));

    // Expect rebalance on disk usage distribution goal since one broker's disk usage is above the low disk utilization threshold
    configOverrides.put(AnalyzerConfig.DISK_LOW_UTILIZATION_THRESHOLD_CONFIG, onePercentSmaller(maxDiskUtilizationRatio));
    p.add(params(goal(DiskUsageDistributionGoal.class, configOverrides), true));

    // Expect no rebalance on disk usage distribution goal since all brokers' disk usage is below the low disk utilization threshold
    configOverrides.put(AnalyzerConfig.DISK_LOW_UTILIZATION_THRESHOLD_CONFIG,
                    onePercentGreater(maxDiskUtilizationRatio / balanceMargin));
    p.add(params(goal(DiskUsageDistributionGoal.class, configOverrides), false));

    // Expect rebalance on network inbound usage distribution goal since one broker's network inbound usage is
    // above the low network inbound utilization threshold
    configOverrides.put(AnalyzerConfig.NETWORK_INBOUND_LOW_UTILIZATION_THRESHOLD_CONFIG,
                    onePercentSmaller(maxNetworkInUtilizationRatio));
    p.add(params(goal(NetworkInboundUsageDistributionGoal.class, configOverrides), true));

    // Expect no rebalance on network inbound usage distribution goal since all brokers' network inbound usage is
    // below the low network inbound utilization threshold
    configOverrides.put(AnalyzerConfig.NETWORK_INBOUND_LOW_UTILIZATION_THRESHOLD_CONFIG,
                    onePercentGreater(maxNetworkInUtilizationRatio / balanceMargin));
    p.add(params(goal(NetworkInboundUsageDistributionGoal.class, configOverrides), false));

    // Expect rebalance on network outbound usage distribution goal since one broker's network outbound usage is
    // above the low network outbound utilization threshold
    configOverrides.put(AnalyzerConfig.NETWORK_OUTBOUND_LOW_UTILIZATION_THRESHOLD_CONFIG, onePercentSmaller(maxNetworkOutUtilizationRatio));
    p.add(params(goal(NetworkOutboundUsageDistributionGoal.class, configOverrides), true));

    // Expect no rebalance on network outbound usage distribution goal since all brokers' network outbound usage is
    // below the low network outbound utilization threshold
    configOverrides.put(AnalyzerConfig.NETWORK_OUTBOUND_LOW_UTILIZATION_THRESHOLD_CONFIG,
                    onePercentGreater(maxNetworkOutUtilizationRatio / balanceMargin));
    p.add(params(goal(NetworkOutboundUsageDistributionGoal.class, configOverrides), false));

    return p;
  }

  private static Map<String, Object> getDefaultGoalConfigs() {
    return new HashMap<String, Object>() {{
      put(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG, "bootstrap.servers");
      put(ExecutorConfig.ZOOKEEPER_CONNECT_CONFIG, "connect:1234");
    }};
  }

  private static String onePercentGreater(double value) {
    return Double.toString(value * 1.01);
  }

  private static String onePercentSmaller(double value) {
    return Double.toString(value * 0.99);
  }

  /**
   * Creates a small cluster model with broker resource utilization percentage shown below:
   *    Broker-0:
   *      CPU:    34.75%
   *      Disk:   28%
   *      NW_IN:  13%
   *      NW_OUT: 14.75%
   *
   *    Broker-1:
   *      CPU:    14%
   *      Disk:   15.5%
   *      NW_IN:  7%
   *      NW_OUT: 5.8%
   *
   *    Broker-2:
   *      CPU:    9.75%
   *      Disk:   13.5%
   *      NW_IN:  6.5%
   *      NW_OUT: 0%
   *
   * @return cluster model
   */
  private static ClusterModel createSmallClusterModel() {
    return DeterministicCluster.smallClusterModel(BROKER_CAPACITY);
  }

  private static Object[] params(Goal resourceDistributionGoal, Boolean expectRebalance) {
    return new Object[]{resourceDistributionGoal, expectRebalance};
  }

  @Test
  public void test() throws OptimizationFailureException {
    ClusterModel clusterModel = createSmallClusterModel();
    Map<TopicPartition, List<ReplicaPlacementInfo>> initReplicaDistribution = clusterModel.getReplicaDistribution();
    Map<TopicPartition, ReplicaPlacementInfo> initLeaderDistribution = clusterModel.getLeaderDistribution();

    assertTrue("Failed to optimize " + _resourceDistributionGoal.name(),
        _resourceDistributionGoal.optimize(clusterModel, Collections.emptySet(), new OptimizationOptions(Collections.emptySet())));

    Set<ExecutionProposal> goalProposals =
        AnalyzerUtils.getDiff(initReplicaDistribution, initLeaderDistribution, clusterModel);

    if (_expectRebalance) {
      assertFalse(goalProposals.isEmpty());
    } else {
      assertTrue(goalProposals.isEmpty());
    }
  }
}

```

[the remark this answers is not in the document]
efeg: @Lincong Now you can also drop the following from the above as it is no longer used:

```
  private static Map<String, Object> getDefaultGoalConfigs() {
    return new HashMap<String, Object>() {{
      put(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG, "bootstrap.servers");
      put(ExecutorConfig.ZOOKEEPER_CONNECT_CONFIG, "connect:1234");
    }};
  }
```

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
efeg: I don't have a strong opinion -- feel free to leave it as is.

But, we as developers (not users) may care about the number of times a function is called to improve/preserve efficiency of our implementations. For example, an implementation that calls a function within a loop and just before a loop may achieve the same outcome, but the latter implementation would be more efficient/preferable.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
```diff
@@ -17,75 +17,35 @@ public void testComputeResourceUtilizationBalanceThreshold() {
...
     EasyMock.expect(mockBalanceConstraint.lowUtilizationThreshold(resource)).andReturn(lowUtilizationThreshold).anyTimes();
-    EasyMock.expect(mockBalanceConstraint.resourceBalancePercentage(resource)).andReturn(resourceBalancePercentage).anyTimes();
+    EasyMock.expect(mockBalanceConstraint.resourceBalancePercentage(resource)).andReturn(1.3).anyTimes();
     EasyMock.expect(mockBalanceConstraint.goalViolationDistributionThresholdMultiplier()).
-        andReturn(goalViolationDistributionThresholdMultiplier).anyTimes();
+        andReturn(1.2).anyTimes();
 
     EasyMock.replay(mockBalanceConstraint);
-    double computedBalanceThreshold = GoalUtils.computeResourceUtilizationBalanceThreshold(avgUtilizationPercentage,
+    double computedBalanceThreshold = GoalUtils.computeResourceUtilizationBalanceThreshold(0.3,
```
efeg: Nit: Can we move the hardcoded constants to static variables?

efeg (commented):
@Lincong Thanks for the updates! Replied earlier comments and added one more comment.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
Lincong: I see your point. I agree that it worth verifying the number of invocations of a method if we need to measure/ensure efficiency. However, that is not necessary in this case. So, I will leave it as is.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtilsTest.java
Lincong: Sure

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/LowResourceUtilizationTest.java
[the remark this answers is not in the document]
Lincong: I see. I did not realize that using `com.linkedin.kafka.cruisecontrol.analyzer.AnalyzerUnitTestUtils.goal` with config overrides does not require specifying default goal configs. Thanks

efeg (approved):
Thanks for the updates -- LGTM.

PR #1396, opened 2020-11-23
fix typo in GoalUtils.java

zornhsu:
Under onlyMoveImmigrantReplicas check


efeg (approved):
@zornhsu Thanks for the fix! LGTM.

PR #1400, opened 2020-11-24
Add gap-based balance limits for TopicReplicaDistributionGoal

### LINKED ISSUE #1399: Add gap-based balance limits for TopicReplicaDistributionGoal
**Background**: `TopicReplicaDistributionGoal` is a soft goal to balance collocations of replicas of the same topic over alive brokers not excluded for replica moves. Having a balanced distribution of replicas of a topic ensures that (1) produce / consume to a single topic does not overload a subset of the cluster, and (2) clients do not hit quota limits due to aggregation of replicas of a given topic in a small set of brokers.

Consider the following scenarios in a cluster with `29` brokers and when `topic.replica.count.balance.threshold` is `2.0`.

*Scenario-1*: A topic with a high replica count:

Suppose the topic has `3072` replicas -- i.e. `1024` partitions with replication factor (RF) `3`.
Based on these values, the average topic replicas on a broker is `~105` (i.e. `3072/29`). 
Hence, for a broker to be considered balanced, its replica count from this topic should be in [0, 210].

**Issue**: The limits are too lenient. With these limits, it is possible to distribute all replicas of this topic on 15 brokers without any replicas on the remaining 14 brokers.

*Scenario-2*: A topic with a low replica count:

Suppose the topic has `24` replicas -- i.e. `8` partitions with replication factor (RF) `3`.
Based on these values, the average topic replicas on a broker is `~1` (i.e. `24/29`). 
Hence, for a broker to be considered balanced, its replica count from this topic should be in [0, 2].

**Issue**: The limits are too strict. With these limits, unless each brokers has either 0 or 2 replicas from the topic, the broker will be considered unbalanced.

**Proposal**: Add the following gap-based balance limits for `TopicReplicaDistributionGoal`:
* `topic.replica.count.balance.min.gap`: The minimum allowed gap between a balance limit and the average replica count for each topic. A balance limit is set via topic.replica.count.balance.threshold config. If the difference between the computed limit and the average replica count for the relevant topic is smaller than the value specified by this config, the limit is adjusted accordingly.
* `topic.replica.count.balance.max.gap`: The maximum allowed gap between a balance limit and the average replica count for each topic. A balance limit is set via topic.replica.count.balance.threshold config. If the difference between the computed limit and the average replica count for the relevant topic is greater than the value specified by this config, the limit is adjusted accordingly.

efeg:
This PR resolves #1399.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -287,6 +287,23 @@ void sanityCheckConcurrency() {
...
 
+  /**
+   * Sanity check to ensure that
+   * <ul>
+   *   <li>{@link AnalyzerConfig#TOPIC_REPLICA_COUNT_BALANCE_MIN_GAP_CONFIG} <=
+   *   {@link AnalyzerConfig#TOPIC_REPLICA_COUNT_BALANCE_MIN_GAP_CONFIG}</li>
+   * </ul>
+   */
+  private void sanityCheckBalancingConstraints() {
+    long topicReplicaBalanceMinGap = getInt(AnalyzerConfig.TOPIC_REPLICA_COUNT_BALANCE_MIN_GAP_CONFIG);
```
Lincong, line 298: Nit: shouldn't the type be `int` since `getInt` is used?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/TopicReplicaDistributionGoal.java
```diff
@@ -103,15 +104,40 @@ private double balancePercentageWithMargin(OptimizationOptions optimizationOptio
...
+   *
+   * @param computedLimit Computed balance upper or lower limit
+   * @param average Average topic replicas on broker.
+   * @return A balance limit that falls into [minGap, maxGap] for topic replica balance.
+   */
+  private int gapBasedBalanceLimit(int computedLimit, double average, boolean isLowerLimit) {
+    int minLimit;
+    int maxLimit;
+    if (isLowerLimit) {
+      minLimit = Math.max(0, (int) (Math.floor(average) - _balancingConstraint.topicReplicaBalanceMinGap()));
```
Lincong, line 122: Shouldn't it be (swapping `maxLimit` and `minLimit`):
```
maxLimit = Math.max(0, (int) (Math.floor(average) - _balancingConstraint.topicReplicaBalanceMinGap()));
minLimit = Math.max(0, (int) (Math.floor(average) - _balancingConstraint.topicReplicaBalanceMaxGap()));
```
In the above way, `maxLimit` >= `minLimit`. Otherwise, it would be the other way (`maxLimit` <= `minLimit`), and the below return statement will always return `minLimit`

```
return Math.max(minLimit, Math.min(computedLimit, maxLimit));
```

Lincong (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/TopicReplicaDistributionGoal.java
efeg: Absolutely -- great catch!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -287,6 +287,23 @@ void sanityCheckConcurrency() {
...
+   * Sanity check to ensure that
+   * <ul>
+   *   <li>{@link AnalyzerConfig#TOPIC_REPLICA_COUNT_BALANCE_MIN_GAP_CONFIG} <=
+   *   {@link AnalyzerConfig#TOPIC_REPLICA_COUNT_BALANCE_MIN_GAP_CONFIG}</li>
+   * </ul>
+   */
+  private void sanityCheckBalancingConstraints() {
+    int topicReplicaBalanceMinGap = getInt(AnalyzerConfig.TOPIC_REPLICA_COUNT_BALANCE_MIN_GAP_CONFIG);
+    int topicReplicaBalanceMaxGap = getInt(AnalyzerConfig.TOPIC_REPLICA_COUNT_BALANCE_MAX_GAP_CONFIG);
+    if (topicReplicaBalanceMinGap > topicReplicaBalanceMaxGap) {
```
Lincong, line 300: Should we allow `topicReplicaBalanceMinGap  == topicReplicaBalanceMaxGap`? In this case, the `computedLimit` is essentially ignored in the `gapBasedBalanceLimit` method since the return value of this method does not depend on `computedLimit`

Lincong (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
efeg: We intentionally allow equality of min/max gaps.

As per their discretion, this enables users to configure their goal such that each topic has deterministic limits on each broker that is independent of the percentage-based `computedLimit`.

Lincong: ok

Lincong (approved):

PR #1401, opened 2020-11-24
Honor webserver.api.urlprefix config

### LINKED ISSUE #1336: `webserver.api.urlprefix` configuration invalid
Hi,

## Problem description

I noticed that even though the `webserver.api.urlprefix` is documented and configurable, anything different from the default `/kafkacruisecontrol/*` is as a matter of fact producing an error, eg:

with the configuration:
```
webserver.api.urlprefix=/mycluster/kafkacruisecontrol/*
```
```
curl  http://cruisecontrol.example:9090/mycluster/kafkacruisecontrol/state
```
will result in:
```
unrecognized endpoint in request '/state'
* Connection #0 to host 10.231.3.123 left intact
Supported GET endpoints: [BOOTSTRAP, TRAIN, LOAD, PARTITION_LOAD, PROPOSALS, STATE, KAFKA_CLUSTER_STATE, USER_TASKS, REVIEW_BOARD]
```

I looked a bit into the code and it seems that the culprit for this behaviour is to be found in this code: https://github.com/linkedin/cruise-control/blob/5ed774eb5c64f6e57a6a001c1306c520170416c5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java#L38-L39

and in the following validation attempt: https://github.com/linkedin/cruise-control/blob/fbc1b1dfac27768193d046ce852a7deaad7f7bd9/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java#L179-L186

## Expected behaviour

I would expect this configuration to behave as documented

Thank you!

Lincong:
This PR resolves #1336 .


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -30,6 +30,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.linkedin.kafka.cruisecontrol.servlet.KafkaCruiseControlServletUtils.*;
```
efeg, line 33: Can we avoid wildcard imports?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
```diff
@@ -193,7 +193,7 @@ static String urlEncode(String s) throws UnsupportedEncodingException {
    */
   static CruiseControlEndPoint getValidEndpoint(HttpServletRequest request, HttpServletResponse response, KafkaCruiseControlConfig config)
       throws IOException {
-    CruiseControlEndPoint endPoint = endPoint(request);
+    CruiseControlEndPoint endPoint = endPoint(request, config.getString(WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG));
     if (endPoint == null) {
       String method = request.getMethod();
       String errorMessage = String.format("Unrecognized endpoint in request '%s'%nSupported %s endpoints: %s",
```
efeg, line 199: If this PR honors configurable request URI, can we drop this hardcoded value?

https://github.com/linkedin/cruise-control/blob/5ed774eb5c64f6e57a6a001c1306c520170416c5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java#L38-L39

efeg, line 199: According to the Java Servlet specification, the config `WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG` must end with `/*` suffix.
1. The following comment seems incomplete: 

https://github.com/linkedin/cruise-control/blob/5ed774eb5c64f6e57a6a001c1306c520170416c5/config/cruisecontrol.properties#L325-L327

Can we update it to say:
```
# REST API default prefix (dont forget the ending /*)
```

2. We don't have any sanity check to enforce this suffix upon configuration time. Can we add a sanity check for `WEBSERVER_API_URLPREFIX_CONFIG` under `KafkaCruiseControlConfig`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -177,7 +177,11 @@ public static CruiseControlEndPoint endPoint(HttpServletRequest request) {
         throw new UserRequestException("Unsupported request method: " + request.getMethod() + ".");
     }
 
-    String path = request.getRequestURI().toUpperCase().replace(REQUEST_URI, "");
+    if (apiUrlPrefix.endsWith("/*")) {
```
efeg, line 180: `apiUrlPrefix` must end with `/*` -- i.e. failure of this condition must result in a configuration time failure.
Hence, this if check is redundant.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -177,7 +177,11 @@ public static CruiseControlEndPoint endPoint(HttpServletRequest request) {
...
+      // Ignore the last character '*'
+      apiUrlPrefix = apiUrlPrefix.substring(0, apiUrlPrefix.length() - 1);
+    }
+    String path = request.getRequestURI().toUpperCase().replace(apiUrlPrefix.toUpperCase(), "");
```
efeg, line 184: Do we have to do this pruning and upper case conversion for `apiUrlPrefix` for every single request that hits a CC endpoint?

`apiUrlPrefix` is already known at configuration time -- i.e. it won't change throughout the lifetime of this CC instance. So can't we initialize this static util class with the well-formatted `REQUEST_URI` (which is constructed by pruning and upper case conversion of `WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG` once) and use that here as before?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
```diff
@@ -193,7 +193,7 @@ static String urlEncode(String s) throws UnsupportedEncodingException {
    */
   static CruiseControlEndPoint getValidEndpoint(HttpServletRequest request, HttpServletResponse response, KafkaCruiseControlConfig config)
       throws IOException {
-    CruiseControlEndPoint endPoint = endPoint(request);
+    CruiseControlEndPoint endPoint = endPoint(request, config.getString(WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG));
```
efeg, line 196: Passing `config.getString(WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG)` to endpoint seems redundant (please see my comment on `ParameterUtils#endPoint `)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
```diff
@@ -136,6 +138,7 @@ protected boolean removeEldestEntry(Map.Entry<UUID, UserTaskInfo> eldest) {
     }
     _sessionExpiryMs = sessionExpiryMs;
     _maxActiveUserTasks = maxActiveUserTasks;
+    _apiUrlPrefix = KafkaCruiseControlServletUtils.REQUEST_URI;
```
efeg, line 141: `REQUEST_URI` is a malformed `_apiUrlPrefix` -- i.e. it does not end with `*` -- hence, it shall not be used interchangeably here.

With the proposed change in `ParameterUtils` (please see earlier comment), we can drop these changes.

efeg (changes requested):
@Lincong Thanks for the patch!
I think we are making things a little more complicated than necessary -- please see my comments.

Other comments:
1. Should we rename this PR to `Honor webserver.api.urlprefix config`? The current title seems a little implementation specific.
2. Would you assign the linked issue to yourself please?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
Lincong: Yes. I agree that we can drop this value and (indirectly) use the default web server API URL prefix value `DEFAULT_WEBSERVER_API_URLPREFIX`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
Lincong: It's not a good practice to use a static class to "cache" any value. A static method should be 100% stateless. So I strongly prefer to just stick with the current implementation with the blow change:
```
    if (!apiUrlPrefix.endsWith("/*")) {
      throw new IllegalArgumentException("API URL prefix should end with \"/*\". Got: " + apiUrlPrefix));
    }
```
In this approach, we need to do this pruning and upper case conversion for every single request. It's not a big deal. If we really want to save this "overhead". We will need to think of some way that does not use the `static` method/class. However, that might over-complicate the whole thing. So, I prefer to stick with this approach.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
Lincong: Replied to the previous comment

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManager.java
Lincong: I changed it to:
```
_apiUrlPrefix = config.getString(WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG);
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
Lincong: Discussed offline and we decided to stick with the current implementation for now. In the future, we can consider using [Dagger 2](https://dagger.dev/)

efeg (changes requested):
@Lincong Thanks for the updates!
The PR is almost ready, left some comments.

Removal of `SessionManager.java` and `SessionManagerTest.java` seems irrelevant to this PR, which also requires updates on `Sensors.md`, but are missing here. Can we please do it in its own PR?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -464,6 +464,25 @@ void sanityCheckSecurity() { // visible for testing
     }
   }
 
+  /**
+   * Sanity check to ensure that webserver URL prefix is set up correctly. This means the following:
+   *  <code>webserver.api.urlprefix</code> and <code>webserver.ui.urlprefix</code> must end with "/*"
+   */
+  void sanityCheckWebServerUrlPrefix() { // visible for testing
+    String expectedUrlPrefixEnd = "/*";
```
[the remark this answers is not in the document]
Lincong: Good idea

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletUtils.java
```diff
@@ -283,4 +281,10 @@ static void ensureHeaderNotPresent(HttpServletRequest request, String headerName
   public static String httpServletRequestToString(HttpServletRequest request) {
     return String.format("%s %s", request.getMethod(), request.getRequestURI());
   }
+
+  public static String getDefaultWebServerApiUrlPrefix() {
+    // Strip off the '*' from the end of the default web server API URL prefix
+    return WebServerConfig.DEFAULT_WEBSERVER_API_URLPREFIX
+        .substring(0, WebServerConfig.DEFAULT_WEBSERVER_API_URLPREFIX.length() - 1);
+  }
```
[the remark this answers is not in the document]
Lincong: After adding back the `SessionManager` class, it will use this method. So, I will leave it here for now and move it to some util test class when we remove `SessionManager`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/WebserverUrlPrefixConfigTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class WebserverUrlPrefixConfigTest {
+
+  @Rule
+  public ExpectedException _expectedException = ExpectedException.none();
+
+  @Test
+  public void testWebserverUrlPrefixConfigsCorrectSetup() {
+    KafkaCruiseControlConfig config = partialMockBuilder(KafkaCruiseControlConfig.class)
+        .addMockedMethod("getString")
```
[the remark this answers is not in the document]
Lincong: Fixed

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -136,4 +140,40 @@ public void testParseExecutionProgressCheckIntervalMsWithValue() {
...
+    verifyGetEndpoint("/kafkacruisecontrol/", WebServerConfig.DEFAULT_WEBSERVER_API_URLPREFIX);
+  }
+
+  /**
+   * Verifies that {@link ParameterUtils#endPoint} can parse out all endpoints
+   *
+   * @param apiUrlPrefix API URL prefix used to concatenate with "/endpoint_name" to mock the URL in a request
+   * @param apiUrlPrefixParam API URL prefix parameter passed to the {@link ParameterUtils#endPoint}
+   */
+  private void verifyGetEndpoint(String apiUrlPrefix, String apiUrlPrefixParam) {
```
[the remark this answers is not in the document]
Lincong: Yes, good idea

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/WebserverUrlPrefixConfigTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    expect(config.getString(WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG)).andReturn("a/b/c/d/*");
+    expect(config.getString(WebServerConfig.WEBSERVER_UI_URLPREFIX_CONFIG)).andReturn("a/b/c/*");
+    replay(config);
+    config.sanityCheckWebServerUrlPrefix();
+    verify(config);
+  }
+
+  @Test
+  public void testWebserverUrlPrefixConfigCorrectSetupOnDefault() {
```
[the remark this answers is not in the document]
Lincong: I think the value of this test is to ensure that if those default values ever change to something that is not valid, this test will catch it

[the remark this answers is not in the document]
efeg: Don't we have sanity checks in `KafkaCruiseControlConfig.java` to ensure validity of config values in production regardless of whether they are the default ones or overridden by users?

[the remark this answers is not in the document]
Lincong: Yes, the sanity checks will catch invalid web server URL prefix. However, invalid URL prefix can only be caught during runtime whereas this test fails the build process if the default webserver URL is invalid.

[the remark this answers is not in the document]
efeg: The existing unit tests already verify (as they should) the behavior of the tested unit with all possible good and bad inputs. The default value is just one such input, whose behavior is known -- because we already tested what such an input would generate. We know that if a future programmer changes the default value, CC will fail to start and provide a meaningful message to help resolve the configuration issue, because it is tested.

I am against adding individual tests specifically for the default values as it eventually leads to bloating, making a build take longer at CircleCI with no additional testing coverage.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
```diff
@@ -51,8 +52,8 @@
   private final Time _time;
   private final ScheduledExecutorService _sessionCleaner =
       Executors.newSingleThreadScheduledExecutor(new KafkaCruiseControlThreadFactory("SessionCleaner",
-                                                                                     true,
```
efeg, line 54: Looks like this change modifies and breaks the indentation formatting. Can we avoid such changes?
Importing the project style file on IntelliJ may help: https://github.com/linkedin/cruise-control/blob/master/docs/code-style.xml

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManagerTest.java
```diff
@@ -36,8 +36,8 @@ public void testCreateAndCloseSession() {
     SessionManager sessionManager = new SessionManager(1, 1000, context.time(), new MetricRegistry(), null);
 
     sessionManager.getAndCreateSessionIfNotExist(context.request(0),
-                                                 () -> new OperationFuture("testCreateSession"),
```
efeg, line 39: Bad indentation -- please see the earlier comment.

efeg (commented):
@Lincong Thanks for the updates -- left some comments.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/WebserverUrlPrefixConfigTest.java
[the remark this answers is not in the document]
Lincong: Makes sense. I'll remove it

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/SessionManager.java
Lincong: Will do. Thanks

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
```diff
@@ -74,6 +74,10 @@ public void destroy() {
     }
   }
 
+  public String apiUrlPrefix() {
+    return _config.getString(WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG);
+  }
+
```
efeg, line 80: Was this added intentionally?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/AbstractConfigTest.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright 2020 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+package com.linkedin.kafka.cruisecontrol.config;
+
+abstract class AbstractConfigTest {
```
efeg, line 6: I am not sure if extending an abstract class like this by `SecurityAndSslConfigTest` and `WebserverUrlPrefixConfigTest` is the best approach. 
* `SecurityAndSslConfigTest` does not use `GET_CLASS_METHOD_NAME` anywhere, and 
* `WebserverUrlPrefixConfigTest` does not use `GET_BOOLEAN_METHOD_NAME` and `HASH_CODE_METHOD_NAME` anywhere.

Inheritance should imply that everything in the abstract class applies to the classes that extend from it. If the goal here is to share common constants across classes, can we move these constants to a utility class within `config` package?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/SecurityAndSslConfigTest.java
```diff
@@ -16,17 +16,17 @@
...
   @Rule
   public ExpectedException _expectedException = ExpectedException.none();
 
   @Test
   public void testSslConfigsSetUpThrowsExceptionOnNullKeystoreLocation() {
     KafkaCruiseControlConfig config = partialMockBuilder(KafkaCruiseControlConfig.class)
-        .addMockedMethod("getBoolean")
-        .addMockedMethod("getString")
-        .addMockedMethod("hashCode")
+        .addMockedMethod(GET_BOOLEAN_METHOD_NAME)
```
efeg, line 27: There are some missing cases below -- see:

```
        .addMockedMethod("getBoolean")
        .addMockedMethod("getClass")
        .addMockedMethod("hashCode")
```

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletTestUtils.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2020 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+package com.linkedin.kafka.cruisecontrol.servlet;
+
+import com.linkedin.kafka.cruisecontrol.config.constants.WebServerConfig;
+
+
+public class KafkaCruiseControlServletTestUtils {
```
efeg, line 9: Nit: It is common practice / convention to make utility classes `final`.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -136,4 +141,39 @@ public void testParseExecutionProgressCheckIntervalMsWithValue() {
...
+      EasyMock.replay(mockRequest);
+      CruiseControlEndPoint endPoint = ParameterUtils.endPoint(mockRequest, apiUrlPrefixParam);
+      Assert.assertEquals(getEndPoint, endPoint);
+      EasyMock.reset(mockRequest);
+    }
+
+    for (CruiseControlEndPoint postEndPoint : CruiseControlEndPoint.postEndpoints()) {
+      String mockRequestUri = apiUrlPrefix + postEndPoint;
+      EasyMock.expect(mockRequest.getMethod()).andReturn(POST_METHOD).times(1);
+      EasyMock.expect(mockRequest.getRequestURI()).andReturn(mockRequestUri).times(1);
```
efeg, line 172: Nit: `times(1)` -> `once()`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -136,4 +141,39 @@ public void testParseExecutionProgressCheckIntervalMsWithValue() {
     EasyMock.verify(mockRequest);
     Assert.assertEquals(Long.valueOf(EXECUTION_PROGRESS_CHECK_INTERVAL_STRING), executionProgressCheckIntervalMs);
   }
+
+  @Test
+  public void testGetEndpoint() {
+    verifyGetEndpoint("/a/*");
+    verifyGetEndpoint(WebServerConfig.DEFAULT_WEBSERVER_API_URLPREFIX);
```
efeg, line 148: I'd recommend dropping the test for the default -- i.e.
```
verifyGetEndpoint(WebServerConfig.DEFAULT_WEBSERVER_API_URLPREFIX);
```

and moving the content of the private function `verifyGetEndpoint` inside this test function as it is not used anywhere else and contains all the logic of this test.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
```diff
@@ -136,4 +141,39 @@ public void testParseExecutionProgressCheckIntervalMsWithValue() {
...
+    }
+
+    for (CruiseControlEndPoint postEndPoint : CruiseControlEndPoint.postEndpoints()) {
+      String mockRequestUri = apiUrlPrefix + postEndPoint;
+      EasyMock.expect(mockRequest.getMethod()).andReturn(POST_METHOD).times(1);
+      EasyMock.expect(mockRequest.getRequestURI()).andReturn(mockRequestUri).times(1);
+      EasyMock.replay(mockRequest);
+      CruiseControlEndPoint endPoint = ParameterUtils.endPoint(mockRequest, apiUrlPrefixParam);
+      Assert.assertEquals(postEndPoint, endPoint);
+      EasyMock.reset(mockRequest);
```
efeg, line 176: Does `reset` call also `verify` the mock? -- i.e. if not do we need `EasyMock.verify()`? (similar use above)

efeg (commented):

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
Lincong: I have changed the implementation and the `verifyGetEndpoint` method is not used any more in test

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtilsTest.java
Lincong: I will call `verify` before `reset`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletTestUtils.java
Lincong: Thanks for letting me know

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServlet.java
Lincong: I was added for testing purposes. I'll remove it.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/AbstractConfigTest.java
Lincong: Sure

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
```diff
@@ -260,8 +261,10 @@ private HttpServletRequest prepareRequest(HttpSession session, String userTaskId
     EasyMock.expect(request.getSession()).andReturn(session).anyTimes();
     EasyMock.expect(request.getSession(false)).andReturn(session).anyTimes();
     EasyMock.expect(request.getMethod()).andReturn(method).anyTimes();
-    EasyMock.expect(request.getRequestURI()).andReturn(KafkaCruiseControlServletUtils.REQUEST_URI + resource).anyTimes();
+    EasyMock.expect(request.getRequestURI())
+        .andReturn(KafkaCruiseControlServletTestUtils.getDefaultWebServerApiUrlPrefix() + resource).anyTimes();
     EasyMock.expect(request.getParameterMap()).andReturn(params).anyTimes();
+    EasyMock.expect(request.getPathInfo()).andReturn("/" + resource).anyTimes();
```
efeg, line 267: (Not introduced in this patch) Looks like this class has no `EasyMock.verify(XXX)` calls; hence, it is hard to verify if the expected calls are made as many as required. Should we add verify before appropriate EasyMock.reset() calls? -- e.g. for the first one in `testUserTaskParameters`

```
    EasyMock.verify(answerQueryRequest1, parameters1, _mockUUIDGenerator, _mockHttpSession, _mockHttpServletResponse);
```

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
```diff
@@ -296,6 +296,7 @@ private HttpServletRequest prepareRequest(HttpSession session, String userTaskId
     EasyMock.expect(request.getSession(false)).andReturn(session).anyTimes();
     EasyMock.expect(request.getMethod()).andReturn(GET_METHOD).anyTimes();
     EasyMock.expect(request.getRequestURI()).andReturn(resource).anyTimes();
+    EasyMock.expect(request.getPathInfo()).andReturn(resource).anyTimes();
```
efeg, line 299: (Not introduced in this PR) Can we add `verify` for mocks? (please see earlier similar comment)

efeg (commented):
Thanks for the updates -- left a few comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java
```diff
@@ -177,7 +176,7 @@ public static CruiseControlEndPoint endPoint(HttpServletRequest request) {
         throw new UserRequestException("Unsupported request method: " + request.getMethod() + ".");
     }
 
-    String path = request.getRequestURI().toUpperCase().replace(REQUEST_URI, "");
+    String path = request.getPathInfo().substring(1); // Skip the first character '/'
```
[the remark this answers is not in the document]
Lincong: Good point!

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
```diff
@@ -234,7 +234,8 @@ private static UserTasksParameters mockUserTasksParameters(HttpServletRequest an
     EasyMock.expect(parameters.userTaskIds()).andReturn(ParameterUtils.userTaskIds(answerQueryRequest)).anyTimes();
     EasyMock.expect(parameters.clientIds()).andReturn(ParameterUtils.clientIds(answerQueryRequest)).anyTimes();
     EasyMock.expect(parameters.endPoints()).andReturn(ParameterUtils.endPoints(answerQueryRequest)).anyTimes();
-    EasyMock.expect(parameters.endPoint()).andReturn(ParameterUtils.endPoint(answerQueryRequest)).anyTimes();
+    EasyMock.expect(parameters.endPoint()).andReturn(ParameterUtils
+        .endPoint(answerQueryRequest)).anyTimes();
```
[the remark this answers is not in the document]
Lincong: Right. No need to break the line here and no code change here

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
Lincong: I tried to add `verify` to `KafkaCruiseControlServletEndpointTest.java` and `UserTaskManagerTest.java` and I realized that the test cases need to be changed to pass the `verify` method and it requires some relatively non-trivial change or even re-write some of the unit tests.

So, I suggest not doing it in this PR since it's irrelevant and creating an issue for this task.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
Lincong: Same comment as above

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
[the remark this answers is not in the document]
efeg: Nit: This line still has the line break even after this issue is marked as resolved.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
efeg: Sure, we can address them in a separate PR -- can we create an issue about them?

efeg (approved):
@Lincong Thanks for the updates!
Replied on 2 earlier conversations -- one of them is a nit, the other is a request to create a followup issue. Otherwise LGTM!

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
Lincong: Issue created: https://github.com/linkedin/cruise-control/issues/1408

Lincong:
Created another issue to improve unit tests https://github.com/linkedin/cruise-control/issues/1408

PR #1406, opened 2020-11-30
Remove unused SessionManager class

### LINKED ISSUE #1405: Remove the SessionManager class
The `SessionManager` class is no longer used and it can be safely removed. `SessionManager` was replaced by the `UserTaskManager` class in the [this](https://github.com/linkedin/cruise-control/pull/304) PR.

Lincong:
This PR resolves #1405.


efeg (approved):
Thanks for the patch! LGTM.

PR #1407, opened 2020-12-03
Provide idempotency support to handle duplicate maintenance events

### LINKED ISSUE #1347: Provide idempotency support to handle duplicate maintenance events
https://github.com/linkedin/cruise-control/pull/1332 added support for handling planned maintenance events submitted via a topic.
However, in case duplicate maintenance events are submitted, maintenance event handler can process the same event multiple times -- e.g. demote broker `123` twice. Since processing the same event is not intentional and can be disruptive, we should ensure that CC can reasonably filter out duplicate events.

efeg:
This PR resolves #1347.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/KafkaCruiseControlThreadFactory.java
```diff
@@ -17,9 +17,11 @@
...
   private final Logger _logger;
 
-  public KafkaCruiseControlThreadFactory(String name,
-                                         boolean daemon,
-                                         Logger logger) {
+  public KafkaCruiseControlThreadFactory(String name) {
+    this(name, true, null);
+  }
+
+  public KafkaCruiseControlThreadFactory(String name, boolean daemon, Logger logger) {
```
Lincong, line 24: Just curious: in what situation(s) CC needs to use a non-daemon thread?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/AnomalyDetectorConfig.java
```diff
@@ -227,6 +228,31 @@ private AnomalyDetectorConfig() {
   public static final String DEFAULT_MAINTENANCE_EVENT_CLASS = MaintenanceEvent.class.getName();
```
Lincong, line 228: Is it reasonable to think about creating another static class `MaintenanceEventConfig` for all configs related to maintenance event?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/AnomalyDetectorConfig.java
```diff
@@ -227,6 +228,31 @@ private AnomalyDetectorConfig() {
...
+  public static final String MAINTENANCE_EVENT_IDEMPOTENCE_RETENTION_MS_CONFIG = "maintenance.event.idempotence.retention.ms";
+  public static final long DEFAULT_MAINTENANCE_EVENT_IDEMPOTENCE_RETENTION_MS = TimeUnit.MINUTES.toMillis(3);
+  public static final String MAINTENANCE_EVENT_IDEMPOTENCE_RETENTION_MS_DOC = "The maximum time in ms to store events retrieved from the"
+      + " MaintenanceEventReader. Relevant only if idempotency is enabled (see " + MAINTENANCE_EVENT_ENABLE_IDEMPOTENCE_CONFIG + ").";
+
+  /**
+   * <code>maintenance.event.max.idempotence.cache.size</code>
+   */
+  public static final String MAINTENANCE_EVENT_MAX_IDEMPOTENCE_CACHE_SIZE_CONFIG = "maintenance.event.max.idempotence.cache.size";
+  public static final int DEFAULT_MAINTENANCE_EVENT_MAX_IDEMPOTENCE_CACHE_SIZE = 25;
```
Lincong, line 251: Even though 25 as default is probably good enough. But why not make the mac cache size larger (e.g. 1000)? I assume the cache entry is not too big in size.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventDetector.java
```diff
@@ -6,42 +6,112 @@
...
     _shutdown = false;
     Map<String, Object> configWithCruiseControlObject = Collections.singletonMap(KAFKA_CRUISE_CONTROL_OBJECT_CONFIG,
                                                                                  kafkaCruiseControl);
-    _maintenanceEventReader = config.getConfiguredInstance(MAINTENANCE_EVENT_READER_CLASS_CONFIG,
+    _maintenanceEventReader = config.getConfiguredInstance(AnomalyDetectorConfig.MAINTENANCE_EVENT_READER_CLASS_CONFIG,
                                                            MaintenanceEventReader.class,
                                                            configWithCruiseControlObject);
+    _idempotenceEnabled = config.getBoolean(AnomalyDetectorConfig.MAINTENANCE_EVENT_ENABLE_IDEMPOTENCE_CONFIG);
+    _idempotenceRetention = Duration.ofMillis(config.getLong(AnomalyDetectorConfig.MAINTENANCE_EVENT_IDEMPOTENCE_RETENTION_MS_CONFIG));
+    if (_idempotenceEnabled) {
```
Lincong, line 53: Nit: I usually prefer to put a shorter branch first. For example, in this case:

```
if (! _idempotenceEnabled) {
    _maintenanceEventByTime = null;
    _idempotenceCacheCleaner = null;
} else{
    ...
}
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventDetector.java
```diff
@@ -6,42 +6,112 @@
...
+                                                   IDEMPOTENCE_CACHE_CLEANER_PERIOD_SECONDS,
+                                                   TimeUnit.SECONDS);
+    } else {
+      _maintenanceEventByTime = null;
+      _idempotenceCacheCleaner = null;
+    }
+  }
+
+  private synchronized void removeOldEventsFromIdempotenceCache() {
+    LOG.debug("Remove old events from idempotence cache.");
```
Lincong, line 73: Maybe we should log something after we remove some entries and include the number of entries we have removed in this method

Lincong (changes requested):
The overall implementation looks good. But I am wondering whether we can consider the following two options.

1. Use `[caffeine cache](https://github.com/ben-manes/caffeine)`

2. Create a separate class `MaintenanceEventCache` that includes the cache cleaner, cache size/ time retention, etc. There could be several advantages for doing it.

  - Improve simplicity/readability/modularity of the `MaintenanceEventDetector` class
  - The concurrency control can happen on the `MaintenanceEventCache` class level instead of `MaintenanceEventDetector` level (currently, `removeOldEventsFromIdempotenceCache` and `ensureIdempotency` methods are synchronized on the `MaintenanceEventDetector` level). It makes more sense to make `MaintenanceEventCache` handle its own concurrent data access.
  - Improve testability since it's more modular

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventDetector.java
```diff
@@ -6,42 +6,112 @@
...
+  public static final long IDEMPOTENCE_CACHE_CLEANER_PERIOD_SECONDS = 5;
+  public static final long IDEMPOTENCE_CACHE_CLEANER_INITIAL_DELAY_SECONDS = 0;
   // TODO: Make this configurable.
   public static final Duration READ_EVENTS_TIMEOUT = Duration.ofSeconds(5);
   private volatile boolean _shutdown;
   private final MaintenanceEventReader _maintenanceEventReader;
+  private final boolean _idempotenceEnabled;
+  private final Duration _idempotenceRetention;
+  private final ScheduledExecutorService _idempotenceCacheCleaner;
+  private final Map<Long, MaintenanceEvent> _maintenanceEventByTime;
```
Lincong, line 40: Instead of storing reference of `MaintenanceEvent` as values in the map, can we just add `hashCode()` and `equals()` methods to the `MaintenanceEvent` class and store hash code of `MaintenanceEvent` object in this map? I think we can save memory and increase the max cache size by a lot

Lincong (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/KafkaCruiseControlThreadFactory.java
efeg: A daemon thread provides a general service in the background as long as the program is running. It does not prevent the JVM from exiting when the program finishes but the thread is still running.

For example, when CC is executing a proposal, we may not want to let JVM exit until we make sure that the execution cleanup is handled properly -- i.e. stopped/cancelled tasks. Hence, if execution is still in progress, it might take slightly longer to shutdown the executor. That's why we have a non-daemon proposal executor.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/AnomalyDetectorConfig.java
efeg: `MaintenanceEvent` is an anomaly -- it extends from it -- and all anomaly configs are located in `AnomalyDetectorConfig`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/AnomalyDetectorConfig.java
efeg: Maintenance events (e.g. planned maintenance events in Azure) are expected to be rare. In particular, a burst of up to `25` _distinct_ (i.e. not duplicate) maintenance events within the default retention period must be extremely rare. CC's anomaly detector manager cannot (and is not expected to) handle this many maintenance events in such a short period.

Presumably, pushing the limit to a higher value would be an exceptional case that would better handled based on user needs.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventDetector.java
efeg: This became irrelevant with the latest update.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventDetector.java
efeg: I agree that logging the `idempotence cache` before and after to see if there are any issues with the cache or cleanup mechanism might be useful for debugging purposes.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventDetector.java
efeg: (_This patch already adds `hashCode()` and `equals()` methods to the `MaintenanceEvent` class_).
I see the reasoning, but as discussed earlier this cache is not expected to contain many entries. Hence, I don't think the additional programming complexity of storing hashcode rather than the event itself would worth it.

efeg:
Thanks for the review.

1. Correct -- adding dependency on something like `Caffeine` (or `Guava`) for this simple need would be an overkill.
2. It makes sense to have a separate class -- please see `IdempotenceCache`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/KafkaCruiseControlThreadFactory.java
Lincong: Makes sense

Lincong (approved):

PR #1410, opened 2020-12-03
Add a sensor to emit topic count in cluster

### LINKED ISSUE #1321: Add a sensor to emit topic count in cluster
This would be useful to keep track of the change in terms of created/deleted topics.

Lincong:
This PR resolves #1321.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
@@ -707,11 +707,7 @@ public class LoadMonitor {
   }
 
   private int numTopics() {
-    // Get topic count from the metadata cache in metadata client first. If it's 0 that means the metadata cache has not
-    // bootstrapped yet. In this case, refresh the metadata cache and get topic count. In other words, refreshing
-    // metadata cache only when it is necessary.
-    int topicCount = _metadataClient.metadata().fetch().topics().size();
-    return topicCount == 0 ? _metadataClient.refreshMetadata().cluster().topics().size() : topicCount;
+    return _metadataClient.metadata().fetch().topics().size();
   }
 
   private double getMonitoredPartitionsPercentage() {
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
@@ -82,6 +82,7 @@ public class LoadMonitor {
   // Metadata TTL is set based on experience -- i.e. a short TTL with large metadata may cause excessive load on brokers.
   private static final long METADATA_TTL = TimeUnit.SECONDS.toMillis(10);
   private static final long METADATA_REFRESH_BACKOFF = TimeUnit.SECONDS.toMillis(5);
+  private static final String LOAD_MONITOR_METRICS_NAME_PREFIX = "LoadMonitor";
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -703,6 +706,14 @@ private int numPartitionsWithExtrapolations() {
     return _latestStateUpdateMs + _monitorStateUpdateTimeoutMs > _time.milliseconds() ? _numPartitionsWithExtrapolations : -1;
   }
 
+  private int numTopics() {
+    // Get topic count from the metadata cache in metadata client first. If it's 0 that means the metadata cache has not
+    // bootstrapped yet. In this case, refresh the metadata cache and get topic count. In other words, refreshing
+    // metadata cache only when it is necessary.
+    int topicCount = _metadataClient.metadata().fetch().topics().size();
+    return topicCount == 0 ? _metadataClient.refreshMetadata().cluster().topics().size() : topicCount;
```
efeg, line 714: Fetching metadata to update this might be expensive. I wonder if we can use the metadata that is periodically updated within `PartitionMetricSampleAggregatorCleaner` for updating this sensor?

It is ok for it to be stale for a couple seconds as long is it keeps the metadata fetches at minimum.

efeg (changes requested):
Thanks for the PR -- left a comment.

Can we also verify the expected topic count via `jconsole` using a local deployment?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
Lincong: In this implementation, the `numTopics` method will fetch metadata at most once when the metadata cache has not finished bootstrapping. Once the metadata cache has something, it will never return an empty set of topics (`topicCount == 0`).

The `_metadataClient` which has the metadata cache is also used by other methods as well. For example, `getMonitoredPartitionsPercentage`

efeg: Thanks for the clarification.

Does `0` topic count always imply that the that the metadata cache has not bootstrapped -- i.e. however unlikely, if the cluster indeed has no topics, wouldn't this approach keep refreshing metadata?

I am curious what you think about simply returning `_metadataClient.metadata().fetch().topics().size()` rather than having the extra logic to handle the one time bootstrapping scenario. We know that `PartitionMetricSampleAggregatorCleaner` starts refreshing metadata instantaneously upon startup, so the topic count can only be reported as `0` during the bootstrapping for a brief period of time.

Lincong: Thanks for your comment.

In my implementation, I tried to minimize the occurrences of returning `0` since we are using `Gauge` as the sensor and a `0` value is a sudden drop in the plotted metrics line.

If the cluster has no topic, such a cluster is very likely to be OK with CC fetching metadata once every minute. In other words, the zero-topic clusters should be extremely rare, and even though they do exist, they are extremely unlikely to be sensitive to the overhead of fetching metadata since there should be little load on them anyway.

Lincong: Discussed offline and I will change it to always fetch from the metadata cache in the admin client.

Lincong:
Locally deployed and verified that the MBean works using VisualVM

image

efeg (approved):
@Lincong Thanks for the updates and verification! LGTM.

PR #1419, opened 2020-12-10
Handle non-existent topic while setting/removing throttled replicas for a topic

### LINKED ISSUE #1318: Deleted topic shouldn't stop the executor
Hi, 
While running Cruise Control for the first time, a topic that was part of a reassignment got deleted, making the executor stop the current rebalance. 

```
[2020-08-31 14:09:38] ERROR ProposalExecutor-0 Executor:1028 - Executor got exception during execution
kafka.admin.AdminOperationException: Topic "topic_name_here" does not exist.
	at kafka.zk.AdminZkClient.validateTopicConfig(AdminZkClient.scala:326)
	at kafka.zk.AdminZkClient.changeTopicConfig(AdminZkClient.scala:340)
	at com.linkedin.kafka.cruisecontrol.executor.ExecutorUtils$.changeTopicConfig(ExecutorUtils.scala:133)
	at com.linkedin.kafka.cruisecontrol.executor.ExecutorUtils.changeTopicConfig(ExecutorUtils.scala)
	at com.linkedin.kafka.cruisecontrol.executor.ReplicationThrottleHelper.setThrottledReplicas(ReplicationThrottleHelper.java:185)
	at com.linkedin.kafka.cruisecontrol.executor.ReplicationThrottleHelper.setLeaderThrottledReplicas(ReplicationThrottleHelper.java:168)
	at java.util.HashMap.forEach(HashMap.java:1289)
	at com.linkedin.kafka.cruisecontrol.executor.ReplicationThrottleHelper.setThrottles(ReplicationThrottleHelper.java:54)
	at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.interBrokerMoveReplicas(Executor.java:1156)
	at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.execute(Executor.java:985)
	at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.run(Executor.java:911)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
[2020-08-31 14:09:38] WARN  ProposalExecutor-0 operationLogger:1063 - Task [fde0e0e0-698a-4720-ac31-da31e5ac377e] userPOST /kafkacruisecontrol/REBALANCE execution is interrupted with exception Topic "topic_name_here" does not exist..
[2020-08-31 14:09:38] INFO  ProposalExecutor-0 Executor:912 - Execution finished.
```

Wouldn't be beneficial if the executor just skipped this topic and continue the rebalance tasks instead of stopping it ? 
Thanks in advance !

Lincong:
This PR resolves #1318.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
@@ -178,8 +178,14 @@ class ReplicationThrottleHelper {
     String replicaThrottleConfigKey = throttleLeaderReplica ? LEADER_THROTTLED_REPLICAS : FOLLOWER_THROTTLED_REPLICAS;
     Properties topicConfigs = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
     if (topicConfigs == null || topicConfigs.isEmpty()) {
-      LOG.debug("Skip setting throttled replicas {} for topic {} since no configs can be read", String.join(",", replicas), topic);
-      return;
+      // Not assume that no topic config means means topic does not exist. Always check topic existence in this situation
+      if (!_kafkaZkClient.topicExists(topic)) {
+        LOG.debug("Skip setting throttled replicas {} for topic {} since topic does not exist", String.join(",", replicas), topic);
+        return;
+      }
+      if (topicConfigs == null) {
+        topicConfigs = new Properties();
+      }
     }
     String currThrottledReplicas = topicConfigs.getProperty(replicaThrottleConfigKey);
     if (currThrottledReplicas != null && currThrottledReplicas.trim().equals(WILDCARD_ASTERISK)) {
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
@@ -6,7 +6,6 @@ package com.linkedin.kafka.cruisecontrol.executor;
 
 import com.linkedin.kafka.cruisecontrol.model.ReplicaPlacementInfo;
 import java.util.Optional;
-import kafka.admin.AdminOperationException;
 import kafka.log.LogConfig;
 import kafka.server.ConfigType;
 import kafka.zk.AdminZkClient;
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
@@ -194,16 +194,40 @@ class ReplicationThrottleHelper {
       newThrottledReplicas.addAll(Arrays.asList(currThrottledReplicas.split(",")));
     }
     topicConfigs.setProperty(replicaThrottleConfigKey, String.join(",", newThrottledReplicas));
-    ExecutorUtils.changeTopicConfig(_adminZkClient, topic, topicConfigs);
+    changeTopicConfigs(topic, topicConfigs);
   }
 
-  private Optional<Properties> readTopicConfigs(String topic) {
+  /**
+   * Read topic configuration properties from Zookeeper
+   *
+   * @param topicName topic name
+   * @return Empty optional if topic has no config properties or topic does not exist
+   */
+  private Optional<Properties> readTopicConfigs(String topicName) {
     try {
-      return Optional.ofNullable(_kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic));
+      return Optional.ofNullable(_kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topicName));
     } catch (Exception e) {
-      if (!_kafkaZkClient.topicExists(topic)) {
-        LOG.warn("Got no config for topic {} since it does not exist", topic);
+      if (!_kafkaZkClient.topicExists(topicName)) {
+        LOG.warn("Got no config for topic {} since it does not exist", topicName);
         return Optional.empty();
+      }
+      throw e;
+    }
+  }
+
+  /**
+   * Update topic configuration properties if a topic exist. No-op for non-existent topic
+   *
+   * @param topicName topic name
+   * @param topicConfigs configuration properties to update
+   */
+  private void changeTopicConfigs(String topicName, Properties topicConfigs) {
+    try {
+      ExecutorUtils.changeTopicConfig(_adminZkClient, topicName, topicConfigs);
+    } catch (Exception e) {
+      if (!_kafkaZkClient.topicExists(topicName)) {
+        LOG.warn("Change no configs for topic {} since it does not exist", topicName);
+        return;
       }
       throw e;
     }
@@ -285,7 +309,7 @@ class ReplicationThrottleHelper {
     boolean removedFollowerThrottle = removeFollowerThrottledReplicasFromTopic(topicConfigs, topic, replicas);
 
     if (removedLeaderThrottle || removedFollowerThrottle) {
-      ExecutorUtils.changeTopicConfig(_adminZkClient, topic, topicConfigs);
+      changeTopicConfigs(topic, topicConfigs);
     }
   }
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
@@ -5,7 +5,6 @@
 package com.linkedin.kafka.cruisecontrol.executor;
 
 import com.linkedin.kafka.cruisecontrol.model.ReplicaPlacementInfo;
-import java.util.Optional;
 import kafka.log.LogConfig;
 import kafka.server.ConfigType;
 import kafka.zk.AdminZkClient;
@@ -177,8 +176,8 @@ class ReplicationThrottleHelper {
 
   private void setThrottledReplicas(String topic, Set<String> replicas, boolean throttleLeaderReplica) {
     String replicaThrottleConfigKey = throttleLeaderReplica ? LEADER_THROTTLED_REPLICAS : FOLLOWER_THROTTLED_REPLICAS;
-    Properties topicConfigs = readTopicConfigs(topic).orElse(null);
-    if (topicConfigs == null) {
+    Properties topicConfigs = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
+    if (topicConfigs == null || topicConfigs.isEmpty()) {
       LOG.debug("Skip setting throttled replicas {} for topic {} since no configs can be read", String.join(",", replicas), topic);
       return;
     }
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -260,12 +287,16 @@ private boolean removeFollowerThrottledReplicasFromTopic(Properties config, Stri
   }
 
   private void removeThrottledReplicasFromTopic(String topic, Set<String> replicas) {
-    Properties config = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
-    boolean removedLeaderThrottle = removeLeaderThrottledReplicasFromTopic(config, topic, replicas);
-    boolean removedFollowerThrottle = removeFollowerThrottledReplicasFromTopic(config, topic, replicas);
+    Properties topicConfigs = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
+    if (topicConfigs == null || topicConfigs.isEmpty()) {
+      LOG.warn("Skip removing throttled replicas {} from topic {} since no configs can be read", String.join(",", replicas), topic);
```
efeg, line 292: When would this be expected to happen? 
Here are some scenarios that I can think of:
  (1) a topic is deleted and then shortly after recreated with the same name with no config,
  (2) someone externally overrides the configs,
  (3) a programming error

If it happens due to (1) or (2), are users expected to do anything if they see this warning log? Does the way we handle it suffice to provide the right behavior?
Assuming that we have proper testing so that (3) is unlikely, should we make this a `debug` level log so that we don't cause extra logs that are not actionable?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -101,6 +101,57 @@ public void testIsNoOpWhenThrottleIsNull() {
...
+    EasyMock.expect(mockKafkaZkClient.topicExists(TOPIC0)).andReturn(false).times(4);
+    EasyMock.replay(mockKafkaZkClient);
+
+    throttleHelper.setThrottles(Collections.singletonList(proposal));
+    EasyMock.verify(mockKafkaZkClient);
+  }
+
+  private void prepareKafkaZkClientMockWithBrokerConfigs(KafkaZkClient mockKafkaZkClient) {
+    // All participating brokers have throttled rate set already
+    Properties brokerThrottledRateProps = new Properties();
```
efeg, line 146: Nit: ... -> `new Properties(2);`?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -101,6 +101,57 @@ public void testIsNoOpWhenThrottleIsNull() {
...
+    EasyMock.replay(mockKafkaZkClient);
+    ReplicationThrottleHelper throttleHelper = new ReplicationThrottleHelper(mockKafkaZkClient, throttleRate);
+
+    throttleHelper.setThrottles(Collections.singletonList(proposal));
+    EasyMock.verify(mockKafkaZkClient);
+
+    // Case 2: a situation where Topic0 gets deleted after its configs were read. Change configs should not fail.
+    EasyMock.reset(mockKafkaZkClient);
+    prepareKafkaZkClientMockWithBrokerConfigs(mockKafkaZkClient);
+    Properties topicConfigProps = new Properties();
```
efeg, line 132: Nit: ... -> `new Properties(2);`?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -101,6 +101,57 @@ public void testIsNoOpWhenThrottleIsNull() {
...
+    String throttledReplicas = brokerId0 + "," + brokerId1;
+    topicConfigProps.put(ReplicationThrottleHelper.LEADER_THROTTLED_REPLICAS, throttledReplicas);
+    topicConfigProps.put(ReplicationThrottleHelper.FOLLOWER_THROTTLED_REPLICAS, throttledReplicas);
+    EasyMock.expect(mockKafkaZkClient.getEntityConfigs(ConfigType.Topic(), TOPIC0)).andReturn(topicConfigProps).times(2);
+    EasyMock.expect(mockKafkaZkClient.topicExists(TOPIC0)).andReturn(false).times(4);
+    EasyMock.replay(mockKafkaZkClient);
+
+    throttleHelper.setThrottles(Collections.singletonList(proposal));
```
efeg, line 140: Should we test the changes in `removeThrottledReplicasFromTopic`, too?

efeg (changes requested):
Thanks for the patch! Left some comments

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -260,12 +287,16 @@ private boolean removeFollowerThrottledReplicasFromTopic(Properties config, Stri
   }
 
   private void removeThrottledReplicasFromTopic(String topic, Set<String> replicas) {
-    Properties config = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
-    boolean removedLeaderThrottle = removeLeaderThrottledReplicasFromTopic(config, topic, replicas);
-    boolean removedFollowerThrottle = removeFollowerThrottledReplicasFromTopic(config, topic, replicas);
+    Properties topicConfigs = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
+    if (topicConfigs == null || topicConfigs.isEmpty()) {
```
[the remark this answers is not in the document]
Lincong: The current implementation of the `getEntityConfigs` method does not return `null`. However, it's not safe to assume it will never return `null`. We can assume it will not return null only if the API clearly claims its behavior.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
Lincong: I don't think `java.util. Properties` class has a constructor that takes a size parameter.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
Lincong: I don't think java.util. Properties class has a constructor that takes a size parameter.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -189,8 +198,26 @@ private void setThrottledReplicas(String topic, Set<String> replicas, boolean th
...
+  /**
+   * Update topic configuration properties if a topic exist. No-op for non-existent topic
+   *
+   * @param topicName topic name
+   * @param topicConfigs configuration properties to update
+   */
+  private void changeTopicConfigs(String topicName, Properties topicConfigs) {
+    try {
+      ExecutorUtils.changeTopicConfig(_adminZkClient, topicName, topicConfigs);
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
Lincong: The `AdminZkClient.changeTopicConfig` API does not claim that it might throw `AdminOperationException`. Even though we know that is the kind of exception it might throw, we should not make this assumption unless it's part of the public contract.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
Lincong: OK, I agree that we can change it to `debug` level

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
Lincong: Yes, test added

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -176,9 +176,18 @@ private void setFollowerThrottledReplicas(String topic, Set<String> replicas) {
...
+    if (topicConfigs == null || topicConfigs.isEmpty()) {
+      // Not assume that no topic config means means topic does not exist. Always check topic existence in this situation
+      if (!_kafkaZkClient.topicExists(topic)) {
+        LOG.debug("Skip setting throttled replicas {} for topic {} since topic does not exist", String.join(",", replicas), topic);
+        return;
+      }
+      if (topicConfigs == null) {
+        topicConfigs = new Properties();
+      }
+    }
```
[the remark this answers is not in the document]
Lincong: 1. It's safer to assume `getEntityConfigs` might return `null` since it does not claim it never returns null as its public contract
2. You are right. No need to check the topic existence here.
3. Fixed

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
efeg: It does in Java 11 -- but it is fine to keep it as is for supporting Java 8+. https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#%3Cinit%3E(int)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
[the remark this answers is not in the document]
efeg: -- Discussed offline.
We will keep the existing logic because API contract leaves the behavior unspecified and the implementation does not return `null`. Hence, we cannot assume that if a future version happens to start returning `null` it would imply that the topic does not exist (i.e. it can return null for some other irrelevant reason). Thus, It is safer to fail if a behavior change causes a `null` response.

efeg (approved):
Thanks for the updates -- LGTM.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -260,27 +277,31 @@ private boolean removeFollowerThrottledReplicasFromTopic(Properties config, Stri
   }
 
   private void removeThrottledReplicasFromTopic(String topic, Set<String> replicas) {
-    Properties config = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
-    boolean removedLeaderThrottle = removeLeaderThrottledReplicasFromTopic(config, topic, replicas);
-    boolean removedFollowerThrottle = removeFollowerThrottledReplicasFromTopic(config, topic, replicas);
+    Properties topicConfigs = _kafkaZkClient.getEntityConfigs(ConfigType.Topic(), topic);
+    if (topicConfigs.isEmpty()) {
```
jiao-zhangS, line 281: Question: should also call "_kafkaZkClient.isTopicMarkedForDeletion()" to exclude the topic that is marked as deletion? @Lincong

PR #1422, opened 2020-12-13
Add a component to allow waiting on KafkaFutures in parallel

### LINKED ISSUE #1421: Waiting on KafkaFutures' results in serial
In `ExecutionUtils`, results of `KafkaFuture` objects are retrieved in a sequential way (see method `processAlterPartitionReassignmentsResult`). A more optimal way is to wait on all `KafkaFuture` objects in parallel.

Lincong:
This PR resolves #1421.

### DIFF
--- a/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/common/utils/AutoCloseableLock.java
+++ b/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/common/utils/AutoCloseableLock.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2020 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+
+package com.linkedin.cruisecontrol.common.utils;
+
+import java.util.concurrent.locks.Lock;
+
+
+public class AutoCloseableLock implements AutoCloseable {
+
+  private final Lock _lock;
+
+  public AutoCloseableLock(Lock lock) {
+    _lock = Utils.validateNotNull(lock, "Lock cannot be null");
+    _lock.lock();
+  }
+
+  @Override
+  public void close() {
+    try {
+      _lock.unlock();
+    } catch (Exception e) {
+      throw new IllegalStateException("while invoking close action", e);
+    }
+  }
+}
--- a/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/common/utils/AutoCloseableLock.java
+++ b/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/common/utils/AutoCloseableLock.java
@@ -4,24 +4,34 @@
 
 package com.linkedin.cruisecontrol.common.utils;
 
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
 
 
 public class AutoCloseableLock implements AutoCloseable {
 
   private final Lock _lock;
+  private final AtomicBoolean _closed;
 
   public AutoCloseableLock(Lock lock) {
     _lock = Utils.validateNotNull(lock, "Lock cannot be null");
+    _closed = new AtomicBoolean(false);
     _lock.lock();
   }
 
   @Override
   public void close() {
-    try {
-      _lock.unlock();
-    } catch (Exception e) {
-      throw new IllegalStateException("while invoking close action", e);
+    if (_closed.compareAndSet(false, true)) {
+      try {
+        _lock.unlock();
+      } catch (Exception e) {
+        throw new IllegalStateException("while invoking close action", e);
+      }
     }
+  }
+
+  // Visible for testing purpose
+  boolean isClosed() {
+    return _closed.get();
   }
 }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
@@ -353,33 +353,65 @@ public final class ExecutionUtils {
                                                               Set<TopicPartition> deleted,
                                                               Set<TopicPartition> dead,
                                                               Set<TopicPartition> noReassignmentToCancel) {
-    if (result != null) {
-      for (Map.Entry<TopicPartition, KafkaFuture<Void>> entry: result.values().entrySet()) {
-        TopicPartition tp = entry.getKey();
-        try {
-          entry.getValue().get(EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
-          LOG.debug("Replica reassignment for {} has been accepted.", tp);
-        } catch (ExecutionException ee) {
-          if (Errors.INVALID_REPLICA_ASSIGNMENT.exception().getClass() == ee.getCause().getClass()) {
-            dead.add(tp);
-            LOG.debug("Replica reassignment failed for {} due to dead destination broker(s).", tp);
-          } else if (Errors.UNKNOWN_TOPIC_OR_PARTITION.exception().getClass() == ee.getCause().getClass()) {
-            deleted.add(tp);
-            LOG.debug("Replica reassignment failed for {} due to its topic deletion.", tp);
-          } else if (Errors.NO_REASSIGNMENT_IN_PROGRESS.exception().getClass() == ee.getCause().getClass()) {
-            // Attempt to cancel/rollback a reassignment that does not exist.
-            noReassignmentToCancel.add(tp);
-            LOG.debug("Rollback failed for {} due to lack of corresponding ongoing replica reassignment.", tp);
-          } else {
-            // Not expected to happen.
-            throw new IllegalStateException(String.format("%s encountered an unknown execution exception.", tp), ee);
-          }
-        } catch (TimeoutException | InterruptedException e) {
-          // May indicate transient (e.g. network) issues and might require a task re-execution -- i.e. handled in Executor.
-          // If this is observed frequently, we may need to bump up EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS.
-          LOG.warn("Failed to process AlterPartitionReassignmentsResult of {}.", tp, e);
-        }
+    if (result == null) {
+      return;
+    }
+    Set<KafkaFuture<Void>> waitingToFinishKafkaFutures = new HashSet<>(result.values().values());
+    KafkaFutureResultsIterator<Void> kafkaFutureResultsIterator = new KafkaFutureResultsIterator<>(waitingToFinishKafkaFutures);
+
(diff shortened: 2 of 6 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/KafkaFutureResultsIterator.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  @Override
+  public FutureResult<T> next() {
+    if (!hasNext()) {
+      throw new IllegalStateException("All Kafka future(s) have finished");
+    }
+    try {
+      return _finishedKafkaFutures.take();
```
radai-rosenblatt, line 93: on multiple concurrent calls to next() this could easily return null (or fail)

Lincong: In my original intention, the `next()` method is not supposed to be invoked concurrently. Since we cannot prevent users from concurrently accessing this method, synchronizing access to this method seems to be a reasonable option.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/common/utils/AutoCloseableLock.java
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  @Override
+  public void close() {
+    try {
+      _lock.unlock();
+    } catch (Exception e) {
+      throw new IllegalStateException("while invoking close action", e);
+    }
+  }
```
efeg, line 26: According to the documentation of this method:
```
...
     * However, implementers of this interface are strongly encouraged
     * to make their {@code close} methods idempotent.
...
```

But the current implementation can throw an exception if the lock has already been unlocked. Can we make this idempotent by adding ` private final AtomicBoolean _closed = new AtomicBoolean(false);` to this class and updating the close method as follows:
```
    if (_closed.compareAndSet(false, true)) {
      _lock.unlock();
    }
```

efeg (changes requested):
@Lincong thanks for the PR! Left an initial comment.
Will continue the review after your updates.

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/common/utils/AutoCloseableLock.java
Lincong: Good idea. Fixed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -353,36 +353,68 @@ public static void processAlterPartitionReassignmentsResult(AlterPartitionReassi
...
+      return;
+    }
+    Set<KafkaFuture<Void>> waitingToFinishKafkaFutures = new HashSet<>(result.values().values());
+    KafkaFutureResultsIterator<Void> kafkaFutureResultsIterator = new KafkaFutureResultsIterator<>(waitingToFinishKafkaFutures);
+
+    while (kafkaFutureResultsIterator.hasNext()) {
+      KafkaFutureResultsIterator.FutureResult<Void> kafkaFutureResult = kafkaFutureResultsIterator.next();
+      if (kafkaFutureResult.exceptional()) {
+        handleAlterPartitionReassignmentsException(kafkaFutureResult.exception(),
+                                                   kafkaFutureResult.finishedKafkaFuture(),
```
efeg, line 366: To have fewer method arguments, can we pass `kafkaFutureResult` and let the retrieval of `exception()` and `finishedKafkaFuture()` happen inside the called method?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -353,36 +353,68 @@ public static void processAlterPartitionReassignmentsResult(AlterPartitionReassi
...
+                                                   deleted,
+                                                   dead,
+                                                   noReassignmentToCancel,
+                                                   transformToTopicPartitionByKafkaFuture(result.values()));
```
efeg, line 370: Looks like if the exception is due to `TimeoutException` or `InterruptedException`, then the call to `transformToTopicPartitionByKafkaFuture` would be redundant. 

Can we ensure that we don't call this function redundantly?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -353,36 +353,68 @@ public static void processAlterPartitionReassignmentsResult(AlterPartitionReassi
...
+
+    if (exception instanceof TimeoutException || exception instanceof InterruptedException) {
+      // May indicate transient (e.g. network) issues and might require a task re-execution -- i.e. handled in Executor.
+      // If this is observed frequently, we may need to bump up EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS.
+      LOG.warn("Failed to process AlterPartitionReassignmentsResult.", exception);
+      return;
+    }
+    if (!(exception instanceof ExecutionException)) {
+      return;
+    }
```
efeg, line 400: Shouldn't we throw the exception rather than silently returning here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
efeg, line 370: Are we doing busy waiting here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
efeg, line 400: These are untested -- please use `jacoco` to verify coverage.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -353,36 +353,68 @@ public static void processAlterPartitionReassignmentsResult(AlterPartitionReassi
...
+    ExecutionException ee = (ExecutionException) exception;
+    if (Errors.INVALID_REPLICA_ASSIGNMENT.exception().getClass() == ee.getCause().getClass()) {
+      dead.addAll(topicPartitionByKafkaFuture.get(finishedKafkaFuture));
+      LOG.debug("Replica reassignment failed due to dead destination broker(s).", ee);
+    } else if (Errors.UNKNOWN_TOPIC_OR_PARTITION.exception().getClass() == ee.getCause().getClass()) {
+      deleted.addAll(topicPartitionByKafkaFuture.get(finishedKafkaFuture));
+      LOG.debug("Replica reassignment failed due to its topic deletion.", ee);
```
efeg, line 407: Untested -- please use `jacoco` to verify coverage.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -353,36 +353,68 @@ public static void processAlterPartitionReassignmentsResult(AlterPartitionReassi
...
+    } else if (Errors.NO_REASSIGNMENT_IN_PROGRESS.exception().getClass() == ee.getCause().getClass()) {
+      // Attempt to cancel/rollback a reassignment that does not exist.
+      noReassignmentToCancel.addAll(topicPartitionByKafkaFuture.get(finishedKafkaFuture));
+      LOG.debug("Rollback failed due to lack of corresponding ongoing replica reassignment.", ee);
+    } else {
+      // Not expected to happen.
+      throw new IllegalStateException("Encountered an unknown execution exception.", ee);
+    }
```
efeg, line 415: Untested -- please use `jacoco` to verify coverage.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -353,36 +353,68 @@ public static void processAlterPartitionReassignmentsResult(AlterPartitionReassi
...
+  }
```
efeg, line 416: This patch drops the name of the affected partitions from the exception logs. Can we add those?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -353,36 +353,68 @@ public static void processAlterPartitionReassignmentsResult(AlterPartitionReassi
...
       }
     }
   }
 
+  private static Map<KafkaFuture<Void>, Set<TopicPartition>> transformToTopicPartitionByKafkaFuture(
+      Map<TopicPartition, KafkaFuture<Void>> kafkaFutureByTopicPartition) {
+
+    Map<KafkaFuture<Void>, Set<TopicPartition>> topicPartitionByKafkaFuture = new HashMap<>();
+    kafkaFutureByTopicPartition.forEach((tp, kafkaFuture) -> {
+      topicPartitionByKafkaFuture.computeIfAbsent(kafkaFuture, kf -> new HashSet<>()).add(tp);
```
efeg, line 380: I am confused -- don't we have 1-1 mapping between kafka future and partitions? Can different `tp`s point to the same `KafkaFuture` value? -- i.e. why do we have a set of `TopicPartition` for each kafka future?

efeg (changes requested):
@Lincong Left more comments.

efeg (changes requested):
@Lincong Left some more comments

-- but before we further look into this, can we take a step back and check whether there are any cases, where the `AlterPartitionReassignmentsResult` can return futures that are partially done?

For example, if 5 replica reassignments were started via a single `alterPartitionReassignments` request, would the response ever has anything other than (1) 5 futures done or (2) 5 futures not done?

Lincong:
It turns out that futures in an `AlterPartitionReassignmentsResult` object get completed all at the same time. Therefore, there will not be such a case where some futures have been completed whereas others in the `AlterPartitionReassignmentsResult` object are still waiting.

More specifically, [this](https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/controller/KafkaController.scala#L816) method in the Kafka controller takes a reassignment map from an `AlterPartitionReassignmentsRequest` and process all topic partition reassignments and then return a map from topic partitions to their corresponding processing results.

On the `KafkaAdminClient` side, [this method](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L3585) takes an `AlterPartitionReassignmentsResponse` which contains all topic partition reassignment results generated by the controller (described above) and complete all waiting futures. Hence, this is the point where all waiting futures in `AlterPartitionReassignmentsResult` get completed.

PR #1424, opened 2020-12-15
Exclude stuck topics

### LINKED ISSUE #1423: Ignore unfixable partitions during rebalance
Hello, 
I have a cluster with 3 brokers and some topics with replication factor=3. I also have some topics with replication factor=2. When I bring down a broker, cruisecontrol gets stuck with one of the topics where the rf=3 and doesn’t try to fix even the fixable rf=2 topics. Is there some way around this? Ideally, cruisecontrol would want to fix what it can fix and leave the rest

mavemuri:
This PR resolves #1423

### DIFF
--- a/config/cruisecontrol.properties
+++ b/config/cruisecontrol.properties
@@ -8,7 +8,8 @@
 # =======================================
 
 # The Kafka cluster to control.
-bootstrap.servers=localhost:9092
+
+bootstrap.servers=localhost:9091,localhost:9092,localhost:9093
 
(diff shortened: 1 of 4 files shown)


efeg (changes requested):
@mavemuri Thanks for the PR.
Replied to the linked issue regarding the recommended approach (please see https://github.com/linkedin/cruise-control/issues/1423#issuecomment-751137145).

efeg:
Closing this issue as the recommended approach differs from the proposed PR.

Issue #325, opened 2018-09-13
Issue with v0.1.4


jmarkan:
Hello,
We're on v0.1.4 and just spun up a cruise control machine and see this warning all over the logs:

`[2018-09-13 18:31:36,539] WARN Goal violation detector received exception (com.linkedin.kafka.cruisecontrol.detector.GoalViolationDetector)
com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException: Insufficient healthy cluster capacity for resource:disk existing cluster utilization 2215449.25 allowed capacity 720000.0
	at com.linkedin.kafka.cruisecontrol.analyzer.goals.CapacityGoal.initGoalState(CapacityGoal.java:173)
	at com.linkedin.kafka.cruisecontrol.analyzer.goals.AbstractGoal.optimize(AbstractGoal.java:81)
	at com.linkedin.kafka.cruisecontrol.detector.GoalViolationDetector.optimizeForGoal(GoalViolationDetector.java:172)
	at com.linkedin.kafka.cruisecontrol.detector.GoalViolationDetector.run(GoalViolationDetector.java:125)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)`

In my capacity.json, the Disk capacity is set as "DISK": "500000", so not sure from where it is getting the value 720000.0.

Could you please suggest @becketqin ?

efeg:
Hi @jmarkan 
-- Sorry for the late response!
### Explanation of the Exception
`720000.0` in the exception message corresponds to the `[sum of all broker DISK capacities]` * `capacityThreshold for DISK`.
For example, if 
(1) the cluster has `5` brokers,
[...]

PR #1446, opened 2021-01-14
Add information about `capacity.config.file` to README.md

### LINKED ISSUE #1052: Irrelevant DiskCapacityGoal error ?
Hello,

I'm getting the following error when trying to add a broker using CC UI.
`ERROR: Error processing POST request '/add_broker' due to: 'com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException: [DiskCapacityGoal] Insufficient healthy cluster capacity for resource: disk existing cluster utilization 177247.171875 allowed capacity 30000.000000 (capacity threshold: 0.500000).'.`

Though here is the disk capacity of my cluster
Capture d’écran 2019-12-26 à 15 19 50

It would be helpful if someone  could explain how CC is calculating the disk utilization of 177247.171875 and the capacity of 30000.000000
Thank you !

qz-fordham:
Add information about `capacity.config.file` to README.md. 

Cruise Control server can start with the default file, but the capacity information may not be correct. This is because the default resolver of Cruise Control uses `capacity.config.file` as the source of truth to determine the capacity of brokers. 

Thus, `capacity.config.file` should be specified by the user. Without doing so, the default file may not show the correct capacity of brokers, which might fail the proposal process in further.

This PR resolves #1052 #325.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,10 @@ Control). The metrics reporter periodically samples the Kafka raw metrics on the
 2. Start ZooKeeper and Kafka server ([See tutorial](https://kafka.apache.org/quickstart)).
 3. Modify `config/cruisecontrol.properties` of Cruise Control:
     * (Required) fill in `bootstrap.servers` and `zookeeper.connect` to the Kafka cluster to be monitored.
+    * (Required) update `capacity.config.file` to the path of your capacity file.  
+      * Capacity file is a JSON file that provides the capacity of the brokers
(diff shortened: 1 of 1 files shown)


efeg (approved):
@qz-fordham Thanks for the PR -- LGTM!

qz-fordham:
@efeg Wow! You have no idea what this means to me. My first ever PR for an enterprise-level open source project. Thank you. Glad I can help!

PR #1453, opened 2021-01-20
Prevent optimization failures with excluded brokers for replica move with replicas

### LINKED ISSUE #1452: Prevent optimization failures with excluded brokers for replica move with replicas
Goals that are extending `AbstractGoal` has a sanity check to ensure that the optimization cannot make goal stats worse unless the cluster has broken brokers -- i.e. either dead or has offline disks (relevant to JBOD). However, this check misses the scenario where the cluster has no broken brokers, but has alive brokers that are excluded from replica moves and contain replicas on them. In such cases, goals (e.g. `ReplicaDistributionGoal`) can still satisfy their requirements, while making stats seemingly worse due to replica movements from alive brokers that are excluded from replica moves. This can show itself during goal violation detection as "unfixable soft goals".

We should prevent optimization failures with excluded brokers for replica move with replicas due to making stats seemingly worse for goals.

efeg:
This PR resolves #1452.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
@@ -70,6 +70,11 @@ public abstract class AbstractGoal implements Goal {
     _minMonitoredPartitionPercentage = parsedConfig.getDouble(MonitorConfig.MIN_VALID_PARTITION_RATIO_CONFIG);
   }
 
+  private static boolean hasNoExcludedBrokersForReplicaMoveWithReplicas(ClusterModel clusterModel, OptimizationOptions optimizationOptions) {
+    Set<Integer> excludedBrokers = optimizationOptions.excludedBrokersForReplicaMove();
+    return clusterModel.aliveBrokers().stream().noneMatch(broker -> excludedBrokers.contains(broker.id()) && !broker.replicas().isEmpty());
+  }
+
   @Override
   public boolean optimize(ClusterModel clusterModel, Set<Goal> optimizedGoals, OptimizationOptions optimizationOptions)
       throws OptimizationFailureException {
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
```diff
@@ -70,6 +70,11 @@ public void configure(Map<String, ?> configs) {
     _minMonitoredPartitionPercentage = parsedConfig.getDouble(MonitorConfig.MIN_VALID_PARTITION_RATIO_CONFIG);
   }
 
+  private static boolean hasNoExcludedBrokersForReplicaMoveWithReplicas(ClusterModel clusterModel, OptimizationOptions optimizationOptions) {
```
Lincong, line 73: Nit: My personal preference is to make this method `hasExcludedBrokersForReplicaMoveWithReplicas` instead of `hasNoExcludedBrokersForReplicaMoveWithReplicas`. I think that removing this negation in the method name (and meaning) improves the readability a little.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/AbstractGoal.java
```diff
@@ -85,6 +90,7 @@ public boolean optimize(ClusterModel clusterModel, Set<Goal> optimizedGoals, Opt
       long goalStartTime = System.currentTimeMillis();
       initGoalState(clusterModel, optimizationOptions);
       SortedSet<Broker> brokenBrokers = clusterModel.brokenBrokers();
+      boolean hasNoExcludedBrokersForReplicaMoveWithReplicas = hasNoExcludedBrokersForReplicaMoveWithReplicas(clusterModel, optimizationOptions);
```
Lincong, line 93: Nit: What do you think if we rename this variable to be `originallyHasNoExcludedBrokersForReplicaMoveWithReplicas`? IIUC, this name is more accurate since this boolean is used is after rebalancing all brokers, and when this variable is used, readers can be reminded that this variable reflects the `original` state of the cluster model. It makes this logic easier to understand IMO.

Lincong (approved):
Overall LGTM. Left 2 comments.

Lincong (approved):

PR #1456, opened 2021-01-23
Update slow broker detection sensitivity and reporting details

### LINKED ISSUE #1455: Update slow broker detection sensitivity and logging details
* Expose broker slowness score in slow broker description, along with its configured cap (i.e. `slow.broker.decommission.score`)
* Set the default value of `slow.broker.log.flush.time.threshold.ms` to 1 second -- the earlier default value was too low, causing false positives.
* Avoid repeated anomaly type and redundant content in KafkaMetricAnomaly `toString()`, which is logged for every metric anomaly.
* Refactor SlowBrokerFinder to eliminate redundant parameter passing and provide other minor fixes.

efeg:
This PR resolves #1455.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/KafkaMetricAnomaly.java
```diff
@@ -65,7 +65,7 @@ public boolean fix() throws KafkaCruiseControlException {
 
   @Override
   public String toString() {
-    return String.format("%s anomaly with id: %s. Anomaly description: %s", METRIC_ANOMALY, anomalyId(), _description);
+    return String.format("[%s] %s", anomalyId(), _description);
```
Lincong, line 68: Nit: Can we consider formatting it as `format("[anomaly ID=%s] %s", anomalyId(), _description);`? I think this way is slightly more descriptive.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
```diff
@@ -230,17 +230,17 @@ private void collectLogFlushTimeMetric(BrokerEntity broker,
...
-                                                Map<BrokerEntity, ValuesAndExtrapolations> metricsHistoryByBroker,
-                                                Map<BrokerEntity, ValuesAndExtrapolations> currentMetricsByBroker,
+                                                ValuesAndExtrapolations metricsHistory,
+                                                ValuesAndExtrapolations currentMetrics,
                                                 Map<BrokerEntity, List<Double>> historicalPerByteLogFlushTimeMetricValues,
                                                 Map<BrokerEntity, Double> currentPerByteLogFlushTimeMetricValues) {
-    AggregatedMetricValues aggregatedMetricValues = currentMetricsByBroker.get(broker).metricValues();
+    AggregatedMetricValues aggregatedMetricValues = currentMetrics.metricValues();
     double latestLogFlushTime = aggregatedMetricValues.valuesFor(BROKER_LOG_FLUSH_TIME_MS_999TH_ID).latest();
     double latestTotalBytesIn = aggregatedMetricValues.valuesFor(LEADER_BYTES_IN_ID).latest() +
```
Lincong, line 239: Question: Is it possible for this `latestTotalBytesIn` variable to be zero? If it's zero,  `latestLogFlushTime / latestTotalBytesIn` throws an exception

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/KafkaMetricAnomaly.java
efeg: Anomaly ID is a UUID (i.e. `d1c17515-8cb6-4ecc-8079-e57425273f6b`) -- from the format, it should be obvious that this is an anomaly id. I prefer conciseness to to help with readability in a single page.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/SlowBrokerFinder.java
efeg: No -- there is an earlier sanity check to ensure that it is a positive value.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/KafkaMetricAnomaly.java
Lincong: My concern is that when users look at this ID in the log file, they have no idea what ID is it since there are other types of IDs that are also in this UUID format right? I guess users can figure out that it is an anomaly ID by reading the description. IMO, add several characters `anomaly ID=` does not harm the conciseness that much since it is much shorter than the UUID itself.

Feel free to drop this comment if you still think the change is not necessary.

Lincong (approved):

PR #1457, opened 2021-01-23
Remove the future-level timeout for the processAlterPartitionReassignmentsResult method

### LINKED ISSUE #1449: EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS is not necessary
In CC 2.4, the `processAlterPartitionReassignmentsResult` method waits on a collection of futures to get results of the partition reassignment request of a set of topic partitions. This method waits on futures with a timeout hard-coded as `EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS`. However, this timeout is not necessary and should be removed.

Firstly, let's clarify the behavior of the `KafkaAdminClient` and `KafkaController` regarding how the `alterPartitionReassignments` request was sent and processed.

- **Behavior of the KafkaAdminClient**
   1. The [`alterPartitionReassignments`](https://github.com/apache/kafka/blob/3ebd32b2a563440645de45ecb578d8a8dccbee95/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L3519) method sends the request to a Kafka broker which may forward the request to the active Kafka controller if it itself is not the controller. A set of futures are created and each future represents a waiting state for each topic partition which has an ongoing partition reassignment.
   2. This [`handleResponse`](https://github.com/apache/kafka/blob/3ebd32b2a563440645de45ecb578d8a8dccbee95/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L3585) method receives a response from the Kafka controller and this response **must** contain results to all topic partitions' reassignment execution. Otherwise, an exception gets thrown ([code](https://github.com/apache/kafka/blob/3ebd32b2a563440645de45ecb578d8a8dccbee95/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L3612)). Then this method completes all futures either with the exception received from the Kafka controller or successfully with no error/exception.
   3. If the Kafka server does not respond with any result after some timeout ([code](https://github.com/apache/kafka/blob/3ebd32b2a563440645de45ecb578d8a8dccbee95/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L1293)). This timeout exception is used to complete all futures and it is defined by [`request.timeout.ms`](REQUEST_TIMEOUT_MS_CONFIG).

- **Behavior of the KafkaController**
   After receiving the request, the Kafka controller tries to reassign replicas for each topic partition in the request. When all topic partitions have been handled, it returns a map from the topic partition to its replica reassignment result. ([code](https://github.com/apache/kafka/blob/4e7c789118fed28ee99275e3e84a626fe912c79a/core/src/main/scala/kafka/controller/KafkaController.scala#L816)). Then the controller sends by this map of results for all topic partitions back to the client to complete the request ([code](https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/controller/KafkaController.scala#L1815))

**It is important to note 2 things:**
1. All futures are always completed together since the controller always processes all topic partition reassignments and then replies with one response or the admin client times out all requested topic partition reassignments in the request.
2. There is **only one** timeout that could happen. That is when the admin client has not heard any response back from the Kafka server for an excessive amount of time.

**Therefore, I propose the following changes.**
1. Remove `EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS` (future level timeout) and let users specify the admin client level request timeout explicitly.
2. Let the `processAlterPartitionReassignmentsResult` method throws the `org.apache.kafka.common.errors.TimeoutException` explicitly and make the caller handles this exception. Since this timeout exception is due to the Kafka server not responding to the admin client request, it might indicate Kafka server unavailability in some degrees. This method should not silently swallow this exception and move on as if nothing happened. Instead, it should notify its caller since the whole executor component (or even the whole CC instance) needs to make a decision regarding how to handle this situation (unavailable or extremely slow Kafka server/controller), fail? re-try? etc

Lincong:
This PR resolves #1449.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -53,7 +53,6 @@ import org.apache.kafka.common.utils.SystemTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import scala.Option;
-
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
@@ -468,12 +468,21 @@ public class ExecutorConfig {
                             DEFAULT_EXECUTOR_NOTIFIER_CLASS,
                             ConfigDef.Importance.LOW,
                             EXECUTOR_NOTIFIER_CLASS_DOC)
+<<<<<<< HEAD
                     .define(ADMIN_CLIENT_REQUEST_TIMEOUT_MS_CONFIG,
                             ConfigDef.Type.INT,
                             DEFAULT_ADMIN_CLIENT_REQUEST_TIMEOUT_MS,
                             atLeast(0),
                             ConfigDef.Importance.MEDIUM,
                             ADMIN_CLIENT_REQUEST_TIMEOUT_MS_DOC)
+=======
+                    .define(KAFKA_REQUEST_TIMEOUT_MS_CONFIG,
+                            ConfigDef.Type.INT,
+                            DEFAULT_KAFKA_REQUEST_TIMEOUT_MS,
+                            atLeast(0),
+                            ConfigDef.Importance.MEDIUM,
+                            KAFKA_REQUEST_TIMEOUT_MS_DOC)
+>>>>>>> Handle issue 1449
                     .define(LEADER_MOVEMENT_TIMEOUT_MS_CONFIG,
                             ConfigDef.Type.LONG,
                             DEFAULT_LEADER_MOVEMENT_TIMEOUT_MS,
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
@@ -468,21 +468,12 @@ public class ExecutorConfig {
                             DEFAULT_EXECUTOR_NOTIFIER_CLASS,
                             ConfigDef.Importance.LOW,
                             EXECUTOR_NOTIFIER_CLASS_DOC)
-<<<<<<< HEAD
                     .define(ADMIN_CLIENT_REQUEST_TIMEOUT_MS_CONFIG,
                             ConfigDef.Type.INT,
                             DEFAULT_ADMIN_CLIENT_REQUEST_TIMEOUT_MS,
                             atLeast(0),
                             ConfigDef.Importance.MEDIUM,
                             ADMIN_CLIENT_REQUEST_TIMEOUT_MS_DOC)
-=======
-                    .define(KAFKA_REQUEST_TIMEOUT_MS_CONFIG,
-                            ConfigDef.Type.INT,
-                            DEFAULT_KAFKA_REQUEST_TIMEOUT_MS,
-                            atLeast(0),
-                            ConfigDef.Importance.MEDIUM,
-                            KAFKA_REQUEST_TIMEOUT_MS_DOC)
->>>>>>> Handle issue 1449
                     .define(LEADER_MOVEMENT_TIMEOUT_MS_CONFIG,
                             ConfigDef.Type.LONG,
                             DEFAULT_LEADER_MOVEMENT_TIMEOUT_MS,
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
@@ -70,7 +70,6 @@ public final class ExecutionUtils {
   static final Map<ConcurrencyType, Integer> MAX_CONCURRENCY = new HashMap<>(ConcurrencyType.cachedValues().size());
   static final Map<ConcurrencyType, Integer> MIN_CONCURRENCY = new HashMap<>(ConcurrencyType.cachedValues().size());
   static final Map<String, Double> CONCURRENCY_ADJUSTER_LIMIT_BY_METRIC_NAME = new HashMap<>(5);
-  public static final long EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS = 10000L;
   static long LIST_PARTITION_REASSIGNMENTS_TIMEOUT_MS;
   static int LIST_PARTITION_REASSIGNMENTS_MAX_ATTEMPTS;
 
@@ -360,7 +359,7 @@ public final class ExecutionUtils {
     for (Map.Entry<TopicPartition, KafkaFuture<Void>> entry: result.values().entrySet()) {
       TopicPartition tp = entry.getKey();
       try {
-        entry.getValue().get(EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+        entry.getValue().get();
         LOG.debug("Replica reassignment for {} has been accepted.", tp);
       } catch (ExecutionException ee) {
         if (Errors.INVALID_REPLICA_ASSIGNMENT.exception().getClass() == ee.getCause().getClass()) {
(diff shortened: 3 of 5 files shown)


docs/wiki/User Guide/Configurations.md
```diff
@@ -137,6 +137,7 @@ The following configurations are inherited from the open source Kafka client con
...
 | logdir.response.timeout.ms                                        | Long	  | N	      | 10000	                                                                                                                                                                                                                                                                                                                                                                                                               | Timeout in ms for broker logdir to respond	                                                                                                                                                                                                                                                                                                                                    |
+| kafka.request.timeout.ms                                          | Integer | N         | 180000
```
efeg, line 140: Broken table entry -- i.e. where is the documentation?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -360,7 +359,7 @@ public static void processAlterPartitionReassignmentsResult(AlterPartitionReassi
     for (Map.Entry<TopicPartition, KafkaFuture<Void>> entry: result.values().entrySet()) {
       TopicPartition tp = entry.getKey();
       try {
-        entry.getValue().get(EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+        entry.getValue().get();
```
efeg, line 362: With this change, an  `ExecutionException ee`, where `org.apache.kafka.common.errors.TimeoutException.class == ee.getCause().getClass()` is`true` should not be quietly handled with a `warn` as is.

Instead, in such a case, the code should throw an `IllegalStateException`, urging the user to increase the value of this new AdminClient timeout config (i.e. it should refer to static variable with the config name) if needed.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -187,7 +187,7 @@ public void testBrokerDiesBeforeMovingPartition() throws Exception {
 
   @Test
   public void testSubmitReplicaReassignmentTasksWithDeadTaskAndNoReassignmentInProgress()
-      throws InterruptedException, TimeoutException {
+      throws InterruptedException {
```
efeg, line 190: Nit: This can be shifted the line above.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -213,7 +213,7 @@ public void testSubmitReplicaReassignmentTasksWithDeadTaskAndNoReassignmentInPro
 
   @Test
   public void testSubmitReplicaReassignmentTasksWithInProgressTaskAndNonExistingTopic()
-      throws InterruptedException, TimeoutException {
+      throws InterruptedException {
```
efeg, line 216: Nit: This can be shifted the line above.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -288,13 +288,12 @@ public void testSubmitReplicaReassignmentTasksWithEmptyNullOrAllCompletedTasks()
...
 
-  private static boolean verifyFutureError(Future<?> future, Class<? extends Throwable> exceptionClass)
-      throws TimeoutException, InterruptedException {
+  private static boolean verifyFutureError(Future<?> future, Class<? extends Throwable> exceptionClass) throws InterruptedException {
     if (future == null) {
       return false;
     }
     try {
-      future.get(EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      future.get();
```
efeg, line 296: There is an unused import in this class after this change -- please see

```
import static com.linkedin.kafka.cruisecontrol.executor.ExecutionUtils.*;
```

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
```diff
@@ -113,10 +102,10 @@ public void testProcessAlterPartitionReassignmentResult() throws Exception {
     Map<TopicPartition, KafkaFuture<Void>> futureByTopicPartition = new HashMap<>(1);
     KafkaFuture<Void> kafkaFuture = EasyMock.mock(KafkaFuture.class);
     if (futureException == null) {
-      EasyMock.expect(kafkaFuture.get(ExecutionUtils.EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS, TimeUnit.MILLISECONDS))
+      EasyMock.expect(kafkaFuture.get())
```
efeg, line 105: Nit: no need for a line break after `EasyMock.expect(kafkaFuture.get())`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
```diff
@@ -113,10 +102,10 @@ public void testProcessAlterPartitionReassignmentResult() throws Exception {
...
               .andReturn(null).once();
     } else {
-      EasyMock.expect(kafkaFuture.get(ExecutionUtils.EXECUTION_TASK_FUTURE_ERROR_VERIFICATION_TIMEOUT_MS, TimeUnit.MILLISECONDS))
+      EasyMock.expect(kafkaFuture.get())
```
efeg, line 108: Nit: no need for a line break after `EasyMock.expect(kafkaFuture.get())`

efeg (changes requested):
@Lincong Thanks for the PR!
Left some comments.

docs/wiki/User Guide/Configurations.md
Lincong: This  issue is resolved in another [PR](https://github.com/linkedin/cruise-control/pull/1460)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
```diff
@@ -123,6 +123,14 @@ private ExecutorConfig() {
...
       + "execution finishes or is stopped (by a user or by Cruise Control).";
 
+  /**
+   * <code>kafka.request.timeout.ms</code>
+   */
+  public static final String KAFKA_REQUEST_TIMEOUT_MS_CONFIG = "kafka.request.timeout.ms";
+  public static final int DEFAULT_KAFKA_REQUEST_TIMEOUT_MS = (int) TimeUnit.MINUTES.toMillis(3);
+  public static final String KAFKA_REQUEST_TIMEOUT_MS_DOC = "The maximum time to wait for the response of a request to Kafka server. "
+      + "If the response is not received before this timeout elapses the client will fail the request.";
+
```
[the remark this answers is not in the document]
Lincong: This issue is fixed in another [PR](https://github.com/linkedin/cruise-control/pull/1460)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -576,6 +577,7 @@ public static void closeAdminClientWithTimeout(AdminClient adminClient, long tim
                                                     .replace("[", "")
                                                     .replace("]", "");
     adminClientConfigs.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServersString);
+    adminClientConfigs.put(AdminClientConfig.REQUEST_TIMEOUT_MS_CONFIG, configs.getInt(ExecutorConfig.KAFKA_REQUEST_TIMEOUT_MS_CONFIG));
```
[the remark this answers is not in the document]
Lincong: Another [PR](https://github.com/linkedin/cruise-control/pull/1460) is created

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -374,15 +373,17 @@ public static void processAlterPartitionReassignmentsResult(AlterPartitionReassi
           noReassignmentToCancelTopicPartitions.add(tp);
           LOG.debug("Rollback failed for {} due to lack of corresponding ongoing replica reassignment.", tp);
         } else if (org.apache.kafka.common.errors.TimeoutException.class == ee.getCause().getClass()) {
-          LOG.warn("Failed to process AlterPartitionReassignmentsResult of {}.", tp, ee.getCause());
+          throw new IllegalStateException(String.format("An alterPartitionReassignments request timed out. This timeout is specified by"
```
efeg, line 376: Can we be more succinct? -- maybe something like:

```
alterPartitionReassignments request timed out for partitions: %s. Check for Kafka broker- or controller-side issues and consider increasing the configured timeout (see %s)."
```

efeg (commented):
@Lincong Thanks for the updates!
Left some comments.

efeg (approved):
LGTM.

PR #1458, opened 2021-01-24
Add 2-step verification information to REST-API docs

### LINKED ISSUE #1451: Insufficient number of racks to distribute each replica
Trying to understand how rack-awareness goal works in Cruise Control.

I have a cluster with 8 nodes and 4 racks (2 nodes each). Sometimes, when I try to rebalance based on all hard goals (including RackAwareGoal) and maybe an optional Disk Usage Distribution goal, I will get 

> Insufficient number of racks to distribute each replica (Current: 4, Needed: 6)
> Insufficient number of racks to distribute each replica (Current: 4, Needed: 7) `Needed` is not always the same.

From the source code, it shows the error msg will happen when `clusterModel.maxReplicationFactor() > numAliveRacks`. I wonder why this is happening. One thing I can ensure is that all my topics have at most 4 replicas. It doesn't make sense to me if 6 or 7 is needed. 

Also, this often happens when there are producers & consumers actively processing messages (message processing & rebalance happening the same time).

Appreciate any pointer or advice. Thank you in advance.

idontknoooo:
This PR resolves #1451.

The issue is not related to the title but mentioned in the discussion.

### DIFF
--- a/docs/wiki/User Guide/REST-APIs.md
+++ b/docs/wiki/User Guide/REST-APIs.md
@@ -21,6 +21,7 @@
     * [Resume metrics load sampling](#resume-metrics-load-sampling)
     * [Change Kafka topic configuration](#change-kafka-topic-configuration)
     * [Change Cruise Control configuration](#change-cruise-control-configuration)
+    * [2-step Verification](#2-step-verification)
 
 ## Asynchronous Endpoints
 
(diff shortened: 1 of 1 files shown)


efeg (approved):
Thanks for the PR-- LGTM!

PR #1470, opened 2021-02-13
Provide recommendations on the estimated resource requirements

### LINKED ISSUE #1435: [Right-Sizing] Provide recommendations on the estimated resource requirements
A Kafka cluster might be _overprovisioned_ or _underprovisioned_ (see #1434). Once such a cluster is detected, the current approach to determine the number of brokers to add/remove relies on human decisions based on “rule of thumb”.

Adopt a systematic way in right-sizing decisions to save time, effort, and cost.

efeg:
This PR resolves #1435.

This patch also:
- updates the default value of `overprovisioned.max.replicas.per.broker` from `800` to `1500`,
- deprecates `Goal#provisionStatus()`, and
- drops redundant `KafkaCruiseControlException` from `OptimizationOptionsGenerator` methods


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ExcludedBrokersForReplicaMoveTest.java
```diff
@@ -148,8 +148,8 @@
     p.add(params(0, ReplicaCapacityGoal.class, excludeB1, null, unbalanced(), noDeadBroker, true, false));
     // Test: With single excluded broker, satisfiable cluster, one dead broker (No exception, Generates proposals, Expected to look optimized)
     p.add(params(1, ReplicaCapacityGoal.class, excludeB1, null, unbalanced(), deadBroker0, true, true));
-    // Test: With all brokers excluded, no dead brokers, not satisfiable (No exception, No proposal, Expected to look optimized)
```
zornhsu, line 151: Why do we not test this case anymore?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionResponseTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    // Case-3: Aggregating {@link ProvisionStatus#RIGHT_SIZED} with {@link ProvisionStatus#RIGHT_SIZED} or
+    // {@link ProvisionStatus#OVER_PROVISIONED} is {@link ProvisionStatus#RIGHT_SIZED}
+    ProvisionResponse rightSized = new ProvisionResponse(ProvisionStatus.RIGHT_SIZED);
+    rightSized.aggregate(generateProvisionResponse(ProvisionStatus.RIGHT_SIZED));
+    assertEquals(ProvisionStatus.RIGHT_SIZED, rightSized.status());
+    rightSized.aggregate(generateProvisionResponse(ProvisionStatus.OVER_PROVISIONED));
+    assertEquals(ProvisionStatus.RIGHT_SIZED, rightSized.status());
+    assertTrue(rightSized.recommendation().isEmpty());
+
```
zornhsu, line 67: Should we add another test for aggregating OVER_PROVISIONED with RIGHT_SIZED? This way we can test the recommendation has been cleared.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionResponse.java
```diff
@@ -0,0 +1,121 @@
...
+ * Recommendations are only relevant to {@link ProvisionStatus#UNDER_PROVISIONED} and {@link ProvisionStatus#OVER_PROVISIONED}.
+ */
+public class ProvisionResponse {
+  private ProvisionStatus _status;
+  private final StringBuilder _recommendation;
+
+  /**
+   * Constructor to be used for provision statuses, for which the recommendations are relevant.
+   */
+  public ProvisionResponse(ProvisionStatus status, String recommendation) {
```
smccauliff, line 21: Is recommendation going to be a human readable string or something else?  Can this be documented?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionResponse.java
```diff
@@ -0,0 +1,121 @@
...
+        aggregateRecommendations(other);
+      }
+    } else {
+      switch (other.status()) {
+        case UNDER_PROVISIONED:
+          _status = ProvisionStatus.UNDER_PROVISIONED;
+          clearRecommendation();
+          aggregateRecommendations(other);
+          break;
+        case RIGHT_SIZED:
```
smccauliff, line 82: This term RIGHT_SIZED is probably ok, but right-sized usually refers to laying off employees.  I've not see it used when referring to computer systems.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionResponse.java
```diff
@@ -0,0 +1,121 @@
...
+import static com.linkedin.cruisecontrol.common.utils.Utils.validateNotNull;
+
+
+/**
+ * Indicates the {@link ProvisionStatus} along with the recommended actions regarding the relevant status.
+ * Recommendations are only relevant to {@link ProvisionStatus#UNDER_PROVISIONED} and {@link ProvisionStatus#OVER_PROVISIONED}.
+ */
+public class ProvisionResponse {
+  private ProvisionStatus _status;
+  private final StringBuilder _recommendation;
```
smccauliff, line 16: Can the recommendations ever be duplicated?  In which case maybe this should be Set<String>?  If this a human readable string then the different recommendation may need to be formatted.  For example adding new lines in the right places; that implies some collections of String.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
```diff
@@ -153,10 +153,23 @@ protected void initGoalState(ClusterModel clusterModel, OptimizationOptions opti
     double allowedCapacity = capacity * _balancingConstraint.capacityThreshold(resource());
 
     if (allowedCapacity < existingUtilization) {
+      Set<Integer> brokersAllowedReplicaMove = GoalUtils.aliveBrokersNotExcludedForReplicaMove(clusterModel, optimizationOptions);
+      if (brokersAllowedReplicaMove.isEmpty()) {
+        // Handle the case when all alive brokers are excluded from replica moves.
+        throw new OptimizationFailureException(String.format("[%s] All alive brokers are excluded from replica moves.", name()),
+                                               String.format("Add at least %d brokers.", clusterModel.maxReplicationFactor()));
+      }
```
smccauliff, line 161: Should CC still generate a recommendation in this case?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
```diff
@@ -153,10 +153,23 @@ protected void initGoalState(ClusterModel clusterModel, OptimizationOptions opti
...
+
+      // Identify a typical broker capacity to be used in recommendations in case the cluster is under-provisioned.
+      int typicalBrokerId = brokersAllowedReplicaMove.iterator().next();
+      double typicalCapacity = clusterModel.broker(typicalBrokerId).capacityFor(resource());
+      double missingCapacity = existingUtilization - allowedCapacity;
+      int numBrokersToAdd = (int) (missingCapacity / (typicalCapacity * _balancingConstraint.capacityThreshold(resource())));
+
       throw new OptimizationFailureException(
```
smccauliff, line 169: Exceptions are very, very low performance and should only be used for exceptional circumstances.   It seems like having a recommendation here is a normal outcome.  A normal, large scale Kafka deployment will probably have some clusters that are not optimally sized.  If that's the case then may the return type needs to be changed.  In c++ I might used a variant type that can either be boolean or String (the recommendation).  A quick search for Java variant doesn't turn up anything interesting. 
```java
class Optimization {
  private Boolean opt;
  private String recommendation;

 // boolean constructor
 // String constructor
 void complete(Function<Boolean> onOptimization, Function<String> onRecommendation) {
   if (opt != null) {
     onOptimization(opt);
   } else if (recommendation != null) {
     onRecommendation(recommendation);
   } else {
    //constructor should prevent this from happening
  }
 } 
}
```

efeg: Thanks for the detailed explanation and the example case! I agree that exceptions are expensive and should be avoided when possible -- so that is a great general advice.

That being said, an `OptimizationFailureException` is expected to be a rare event -- i.e. unless it is fixed, Cruise Control cannot function. For example, it cannot manage the cluster to perform operations including rebalance, broker removal or addition, fixing offline replicas, and changing replication factor. 

      // An OptimizationFailureException indicates (1) a hard goal violation that cannot be fixed typically due to
      // lack of physical hardware (e.g. insufficient number of racks to satisfy rack awareness, insufficient number
      // of brokers to satisfy Replica Capacity Goal, or insufficient number of resources to satisfy resource
      // capacity goals), or (2) a failure to move offline replicas away from dead brokers/disks.

I see that a part of this confusion might be due to what CC considers as "under-provisioned". The current iteration of CC is quite conservative on what it considers as under-provisioned -- i.e. CC considers only the clusters that lead to an `OptimizationFailureException` for a goal as under-provisioned. So just a non-optimally sized cluster is unlikely to be reported as under-provisioned.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionResponse.java
efeg: Recommendations are generated by goals. In theory, they can be duplicated (i.e. different goals might recommend the same action), but in practice recommendations include identifiers (i.e. a goal name prefix), which makes them unique across goals.

I updated the constructor with a new `recommender` field to make it more clear that the recommender is required in addition to recommendation.

> If this a human readable string then the different recommendation may need to be formatted. For example adding new lines in the right places; that implies some collections of String.

`ProvisionResponse #aggregateRecommendations(ProvisionResponse)` does this formatting using a space as the delimiter between recommendations -- I don't use "new line" as the delimiter, because I feel that it makes the overall response harder to read / parse -- e.g. when used in json response field `provisionRecommendation` of `OptimizerResult`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionResponse.java
smccauliff: Seems good then.  Thanks for checking.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
smccauliff: Hmm.  I'm not really convinced and I think this will lead to poor performance.  I guess we will find out.

smccauliff: Does the exception throw cause termination of the goal?  Will I miss other optimizations that might be generated if there was a normal return?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
```diff
@@ -73,18 +74,33 @@ int numInterestedReplicas(ClusterModel clusterModel) {
     return _balancingConstraint.replicaBalancePercentage();
   }
 
+  /**
+   * Get the number of brokers that can be dropped if the given cluster is over-provisioned, {@code null} otherwise.
+   *
+   * @param clusterModel The state of the cluster.
+   * @return Number of brokers that can be dropped if the given cluster is over-provisioned, {@code null} otherwise.
+   */
+  private Integer numBrokersToDrop(ClusterModel clusterModel) {
```
smccauliff, line 83: Shouldn't this just return int?

smccauliff (commented):
I'll finish the last 1/3 tomorrow.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/CapacityGoal.java
efeg: > Does the exception throw cause termination of the goal?

Yes, the optimization for the goal will terminate with this exception message.

> Will I miss other optimizations that might be generated if there was a normal return?

There are 2 parts to the response:
1. Requests via REST API, which require an optimization, will (as before) receive the exception message with the optimization failure details for the first goal with `OptimizationFailureException` having the highest priority.
2. Anomaly Detector Substate > Goal violation will show the information regarding provision recommendations for each goal with optimization failure exception (if any).

> Hmm. I'm not really convinced and I think this will lead to poor performance. I guess we will find out.

This patch does not introduce or increase the frequency of these exceptions -- these exceptions were already in use before. So it is not clear why we would find out a performance difference due to this patch.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
efeg: No, JavaDoc indicates why it may return `null`:

```
Get the number of brokers that can be dropped if the given cluster is over-provisioned, {@code null} otherwise.
```

So if the cluster is not over-provisioned, it returns null.

smccauliff (approved):

PR #1474, opened 2021-02-22
Fix intra broker rebalance to work if there are brokers without JBOD

andrey-klochkov-liftoff:
This PR fixes the intra broker rebalance feature to work even if the cluster contains some brokers that don't use the JBOD disk configuration. We're currently experimenting with JBOD on some brokers in the cluster, and we still want CC to balance the data between disks of these specific brokers.


andrey-klochkov-liftoff:
Sorry, I filed it against a wrong repo. I didn't intend to submit it to the upstream yet.

PR #1476, opened 2021-02-23
Fix miscalculated recommendation for the number of racks to drop for clusters over-provisioned wrt rack count.

### LINKED ISSUE #1475: Fix miscalculated recommendation for the number of racks to drop for clusters over-provisioned wrt rack count
Due to a basic miscalculation (i.e. subtraction of `1` rather than addition `1`), the recommended number of racks to drop for clusters over-provisioned with respect to rack count is `2` fewer than it should be. For example, in a cluster, where dropping `1` rack should be recommended, the current recommendation says:

```
Provision Recommendation: [RackAwareDistributionGoal] Reduce rack diversity by at least -1 racks.
```

Fix this miscalculation in `RackAwareDistributionGoal` and `RackAwareGoal`.

efeg:
This PR resolves #1475.
* It also removes an extra `dot` from provision response.


zornhsu (approved):

zornhsu:
LGTM. I took a bit time to understand overprovisionedMinExtraRacks is a number we do not want to be at. 
Thanks for the patch, Efe!

PR #1484, opened 2021-03-02
Add a sensor to indicate the metadata factor of the managed Kafka cluster

### LINKED ISSUE #1075: Add sensors to indicate metadata factor of the managed Kafka cluster
The `metadata factor` is represented with the following formula:
```
metadata factor = (number of replicas) * (number of nodes) ^ X
```

It is an indication of growing scale and pressure on the centralized Kafka controller. Tracking the historical change of this metric helps quantifying the impact of the growing scale.

Ensure that the metadata factor is exposed via the configured JMX endpoint with selected `X` values.

efeg:
This PR resolves #1075.

* Also minor simplification: `metadataClient.metadata().fetch()` -> `metadataClient.cluster()` -- i.e. calls the same functions.


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtilsTest.java
```diff
@@ -42,30 +43,51 @@ public void testMetadataChanged() {
     Cluster cluster1 = new Cluster("cluster", Arrays.asList(node0, node1, node2), partitions1,
                                    Collections.emptySet(), Collections.emptySet());
 
+    // Verify number of replicas in the cluster
+    assertEquals(8, MonitorUtils.numReplicas(cluster1));
+
```
zornhsu, line 48: nit: There's no test for brokersWithReplicas(). Maybe we can add the tests here as well?

zornhsu (approved):
LGTM. Left a nit which can be split into another issue.

PR #1485, opened 2021-03-03
Add a sensor to indicate if the cluster has partitions with RF > the number of eligible racks

### LINKED ISSUE #1365: Add a sensor to identify a cluster containing partitions with replication factor > number of racks
To enhance the fault tolerance of clusters with insufficient number of racks to provide a full rack-awareness (i.e. each replica of each partition resides on a different rack), we added a new hard goal, called `RackAwareDistributionGoal` (see https://github.com/linkedin/cruise-control/pull/1345). This goal is a relaxed version of `RackAwareGoal`, and evenly distributes replicas over racks. Contrary to `RackAwareGoal`, if replicas of each partition can be evenly distributed across the racks, this goal lets placement of multiple replicas of a partition into a single rack. 

To help identification of clusters containing partitions with RF > number of racks, add a sensor that monitoring systems can alert on if needed.

efeg:
This PR resolves #1365.

Relevant Notes
* An _eligible_ rack is an alive rack that is allowed replica moves. A rack is alive and allowed replica moves as long as it has at least one alive host, which is not excluded for replica moves. A host is alive and allowed replica moves if it has at least one alive broker, which is not excluded for replica moves.
* An ongoing execution may modify the replication factor of partitions; hence, the detector avoids reporting potential false positives.
* The sensor is configured to be refreshed once in every `goal.violation.detection.interval.ms`; hence, it could be temporarily stale at most as long as this interval if during this time
    (1) new topics would be created with a replication factor > number of eligible racks in a cluster for the first time, or
    (2) the replication factor of existing topics are changed in either direction.
* Making this check based on cluster model rather than raw cluster metadata enables CC to (1) have fewer false positives by making sure that there are no ongoing executions, (2) take "rack exclusion" into account -- e.g. if a rack contains only excluded brokers (e.g. preferred controllers or brokers under-maintenance), then it is not an eligible rack, and (3) has negligible overhead compared to parsing raw cluster metadata to compute maximum replication factor and rack count.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorManager.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorManager.java
@@ -197,6 +197,9 @@ public class AnomalyDetectorManager {
                                                             String.format("num-%s-metric-anomalies", type.toString().toLowerCase())),
                                         (Gauge<Integer>) () -> _metricAnomalyDetector.numAnomaliesOfType(type));
     }
+    // The cluster has partitions with RF > the number of eligible racks (0: No such partitions, 1: Has such partitions)
+    dropwizardMetricRegistry.register(MetricRegistry.name(METRIC_REGISTRY_NAME, "has-partitions-with-replication-factor-greater-than-num-racks"),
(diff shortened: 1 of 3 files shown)


zornhsu (approved):
LGTM. Thanks for the patch and thanks for simplifying `new HashSet()`, Efe.

PR #1494, opened 2021-03-17
Add provisioner interface

### LINKED ISSUE #1436: [Right-Sizing] Provide a generic interface for auto-sizing
Detecting clusters to expand or shrink (see #1434) and providing recommendations on the estimated resource requirements (see #1435) would provide the building blocks to enable CC to auto-size clusters. This would let adoption of systematic and automated decision making in right sizing clusters; hence, saves time, effort, and cost.

As a prerequisite of elasticity on cloud (and in on-prem data centers with this support), provide a generic interface for auto-sizing.

efeg:
This PR resolves #1436.

This patch also...
1. Fixes miscalculated  `numBrokersToAdd` in `UNDER_PROVISIONED` cluster recommendation  of `CapacityGoal` -- i.e. missing `Math.ceil(...)` in computation would have caused a number of broker recommendation that is one less than it should be.
2. Fixes miscalculated  `requiredCapacity`  in `UNDER_PROVISIONED` cluster recommendation  of `IntraBrokerDiskCapacityGoal` -- i.e. the use of multiplication should be replaced with a division.
3. Adds `provisioner.class` config to let users plug their own provisioner implementation. This would allow users on different platforms to implement their own classes -- e.g. specific to `Azure`, `AWS`, or `on-prem`.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionResponse.java
```diff
@@ -12,25 +15,33 @@
...
    * Constructor to be used for provision statuses, for which the recommendations are relevant.
-   * Recommendation and recommender are expected to be human-readable strings.
+   * Recommender is expected to be a human-readable string.
    *
    * @param status The current provision status.
    * @param recommendation Recommended action regarding the given provision status.
    * @param recommender The source of the recommended action to be used in aggregate recommendation.
    */
-  public ProvisionResponse(ProvisionStatus status, String recommendation, String recommender) {
+  public ProvisionResponse(ProvisionStatus status, ProvisionRecommendation recommendation, String recommender) {
```
zornhsu, line 31:
```suggestion
  public ProvisionResponse(ProvisionStatus status, ProvisionRecommendation provisionRecommendation, String recommender) {
```
I find it a bit hard to follow this `ProvisionRecommendation recommendation` when we have another `StringBuilder _recommendation` in the same file. Maybe we can change this name?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionRecommendation.java
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    /**
+     * (Optional) Set the topic name
+     * @param topic Topic name for which to add partitions
+     * @return this builder.
+     */
+    public Builder topic(String topic) {
+      if (topic == null || topic.isEmpty()) {
+        throw new IllegalArgumentException(String.format("Number of partitions must be positive (%s).", topic));
```
zornhsu, line 136:
```suggestion
        throw new IllegalArgumentException("The name of the topic must be specified.");
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
```diff
@@ -201,11 +203,12 @@ private void validateEnoughLeaderToDistribute(ClusterModel clusterModel, Optimiz
 
     for (Map.Entry<String, Integer> numLeadersPerTopic : numLeadersByTopicNames.entrySet()) {
       if (numLeadersPerTopic.getValue() < totalMinimumLeaderCount) {
+        ProvisionRecommendation recommendation = new ProvisionRecommendation.Builder(ProvisionStatus.UNDER_PROVISIONED)
+            .numPartitions(totalMinimumLeaderCount - numLeadersPerTopic.getValue()).topic(numLeadersPerTopic.getKey()).build();
         throw new OptimizationFailureException(
             String.format("[%s] Cannot distribute %d leaders over %d broker(s) with minimum required per broker leader count %d for topic %s.",
                           name(), numLeadersPerTopic.getValue(), eligibleBrokersForLeadership.size(), minTopicLeadersPerBroker(),
-                          numLeadersPerTopic.getKey()),
-            String.format("Increase the partition count of topic %s to at least %d.", numLeadersPerTopic.getKey(), totalMinimumLeaderCount));
```
zornhsu, line 208: nit: I would like to keep this format `count of topic to at least %d`, but I admit that I can't come up with an elegant way to do so. 
Reason: To manually increase the partition count through CLI, operators give the count parameter as total count instead of how many more partitions to add. So displaying the total `totalMinimumLeaderCount` directly can help them avoiding the risk of miss calculating the addition of `numLeadersPerTopic` + `_numPartitions-to-add`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/Provisioner.java
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Rightsize the cluster using the given constraints. All given recommendations are expected to share the same {@link ProvisionStatus}.
+   * Implementations of this function are expected to be non-blocking -- i.e. starts the rightsizing, but does not block until the completion.
+   *
+   * <ul>
+   *   <li>For {@link ProvisionStatus#UNDER_PROVISIONED} clusters, each recommender (e.g. goal name) indicates requested resources
+   *   (e.g. number of brokers) along with relevant constraints (e.g. racks for which brokers should not be added). Typically, aggregating
+   *   different recommendations for the same resource type requires considering the maximum value over all recommendations.</li>
+   *   <li>For {@link ProvisionStatus#UNDER_PROVISIONED} clusters, each recommender (e.g. goal name) indicates resources that can be
```
zornhsu, line 28:
```suggestion
   *   <li>For {@link ProvisionStatus#OVER_PROVISIONED} clusters, each recommender (e.g. goal name) indicates resources that can be
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
```diff
@@ -208,6 +213,11 @@ public void run() {
         provisionResponse.aggregate(goal.provisionResponse());
       }
       _provisionResponse = provisionResponse;
+      // Right-size the cluster (if needed)
+      boolean isRightsized = _provisioner.rightsize(_provisionResponse.recommendationByRecommender());
+      if (isRightsized) {
+        LOG.info("Actions have been taken on the cluster towards rightsizing.");
+      }
```
zornhsu, line 220: Maybe we want to skip calling `rightsize()` if the status is `UNDECIDED or `RIGHT_SIZED`? 
```suggestion
      if (_provisionResponse.status() == ProvisionStatus.UNDER_PROVISIONED 
              || _provisionResponse.status() == ProvisionStatus.OVER_PROVISIONED) {
        boolean isRightsized = _provisioner.rightsize(_provisionResponse.recommendationByRecommender());
        if (isRightsized) {
          LOG.info("Actions have been taken on the cluster towards rightsizing.");
        }
      }
```

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionResponseTest.java
```diff
@@ -13,72 +15,179 @@
...
+
+    assertEquals(String.format("[%s] %s [%s] %s [%s] %s", recommenderUp, ProvisionResponse.DEFAULT_RECOMMENDATION, recommenderUp2,
+                               ProvisionResponse.DEFAULT_RECOMMENDATION, RECOMMENDER_UP, UNDER_PROV_REC_STR), nullRecommendationUp.recommendation());
+    assertEquals(String.format("[%s] %s [%s] %s [%s] %s", recommenderOp, ProvisionResponse.DEFAULT_RECOMMENDATION, recommenderOp2,
+                               ProvisionResponse.DEFAULT_RECOMMENDATION, RECOMMENDER_OP, OVER_PROV_REC_STR), nullRecommendationOp.recommendation());
+
+    // Verify recommendationByRecommender is no longer empty and has the expected value
+    assertEquals(1, nullRecommendationUp.recommendationByRecommender().size());
+    assertEquals(1, nullRecommendationOp.recommendationByRecommender().size());
+    assertEquals(NUM_BROKERS_UP, nullRecommendationUp.recommendationByRecommender().get(RECOMMENDER_UP).numBrokers());
```
zornhsu, line 103: Is it still relevant to verify `OVER_PROVISIONED` scenario here? Just wondering because we omit the counter part and verify only the `UNDER_PROVISIONED` from here on.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareDistributionGoal.java
```diff
@@ -357,8 +375,9 @@ private void ensureRackAwareDistribution(ClusterModel clusterModel, Optimization
       _numAliveRacksAllowedReplicaMoves = clusterModel.numAliveRacksAllowedReplicaMoves(optimizationOptions);
       if (_numAliveRacksAllowedReplicaMoves == 0) {
         // Handle the case when all alive racks are excluded from replica moves.
-        throw new OptimizationFailureException("All alive racks are excluded from replica moves.",
-                                               String.format("Add at least %d brokers to different racks.", clusterModel.maxReplicationFactor()));
+        ProvisionRecommendation recommendation = new ProvisionRecommendation.Builder(ProvisionStatus.UNDER_PROVISIONED)
+            .numBrokers(clusterModel.maxReplicationFactor()).build();
+        throw new OptimizationFailureException("All alive racks are excluded from replica moves.", recommendation);
```
zornhsu, line 380: IIUC, the patch would only prompt users to add X brokers. But adding all X brokers to the same rack would not be enough, they must add brokers on different racks. Maybe we can add more instructions to make it more clear?
```suggestion
        throw new OptimizationFailureException("All alive racks are excluded from replica moves, please add more brokers to different racks.", recommendation);
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionRecommendation.java
```diff
@@ -0,0 +1,397 @@
...
+      if (numDisks <= 0) {
+        throw new IllegalArgumentException(String.format("Number of disks must be positive (%d).", numDisks));
+      }
+      _numDisks = numDisks;
+      return this;
+    }
+
+    /**
+     * (Optional) Set number of partitions
+     * @param numPartitions Number of partitions to add or remove
```
zornhsu, line 118:
```suggestion
     * @param numPartitions The recommended number of partitions.
```
Can we align the definition of this field with [`admin.NewTopic`](https://kafka.apache.org/26/javadoc/org/apache/kafka/clients/admin/NewTopic.html#numPartitions--)?
>  The number of partitions for the new topic or -1 if a replica assignment has been specified.

### Reason
By aligning the two, users can pass this value directly into `KafkaCruiseControlUtils.wrapTopic(int partitionCount)` when updating the partition count. 
Users would save the step to query the current partition count to add this number to pass this parameter.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
zornhsu: Please see my related [change request](https://github.com/linkedin/cruise-control/pull/1494#pullrequestreview-622401143) below.

zornhsu (dismissed):
Thanks Efe for bringing us this new feature! Left some feedback and would appreciate your perspective:
1. [Call `rightsize()` only if `UNDER_PROVISIONED` or `OVER_PROVISIONED`.](https://github.com/linkedin/cruise-control/pull/1494#discussion_r601880544)
2. [Explicitly asking users to add brokers to different racks.](https://github.com/linkedin/cruise-control/pull/1494#discussion_r601907289)
3. [Recommend number of partitions to add to, instead of how many more to add.](https://github.com/linkedin/cruise-control/pull/1494#discussion_r602484675)

zornhsu (commented):
Requested myself to review again by mistake.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/RackAwareDistributionGoal.java
efeg: Thanks for pointing this out -- the recommendation should indeed ask for `Adding racks with brokers` rather than `Adding brokers`. Fixed the relevant provision recommendation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionRecommendation.java
efeg: Makes sense.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/ProvisionResponseTest.java
efeg: Absolutely, thanks for pointing this out!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
efeg: I feel this check might be redundant, because if the status is `UNDECIDED` or `RIGHT_SIZED`, the input to `rightsize()` function will be an empty map, which should be a no-op.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/Provisioner.java
efeg: Good catch!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/GoalViolationDetector.java
zornhsu: Got it. I wasn't thinking in functional programming terms. I agree that every implementation of `rightsize()` should first check for empty map and return `false` if it's empty. 
I feel like half of the future contributors would return `true` for this case. We might need to keep an eye out while merging in PR the future.

zornhsu (approved):
Awesome. Looks good to me. Thanks for the new feature, Efe!

PR #1495, opened 2021-03-18
Handle topics with dot for Prometheus Sampler

### LINKED ISSUE #1468: Prometheus metrics sampler not pickup up on topics with dots
I've been testing out the Prometheus metrics sampler and so far it seems to work as expected, but noticed it's failing to pick up a subset of topics. I suspect it has to do with use of dots (`.`) in the topic names, thus probably related to https://github.com/linkedin/cruise-control/pull/443. Sounds like the default sampler may be doing some implicit conversion of these dots that requires this logic, but I don't think that's happening in the Prometheus sampler, so there's a mismatch later on?

This is the warning I was seeing that clued me in, since all the topics listed actually had dots instead of underscores.
```
WARN Broker 2 is missing 2/9 topics metrics and 2/53 leader partition metrics. Missing leader topics: [a_topic, another_topic]. (com.linkedin.kafka.cruisecontrol.monitor.sampling.holder.BrokerLoad)
```

I may also be able to help with fixing or testing, but at least wanted to make sure I'm on the right track here before taking a deeper look.

hd-rk:
This PR resolves #1468

### DIFF
--- a/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/prometheus/PrometheusMetricSamplerTest.java
+++ b/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/prometheus/PrometheusMetricSamplerTest.java
@@ -61,6 +61,7 @@ public class PrometheusMetricSamplerTest {
     private static final int TOTAL_PARTITIONS = 3;
 
     private static final String TEST_TOPIC = "test-topic";
+    private static final String TEST_TOPIC_WITH_DOT = "test.topic";
 
     private PrometheusMetricSampler _prometheusMetricSampler;
     private PrometheusAdapter _prometheusAdapter;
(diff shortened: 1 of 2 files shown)


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/prometheus/PrometheusMetricSamplerTest.java
```diff
@@ -184,18 +186,40 @@ public void testGetSamplesSuccess() throws Exception {
...
+
+        replay(_prometheusAdapter);
+        MetricSampler.Samples samples = _prometheusMetricSampler.getSamples(metricSamplerOptions);
+
+        assertSamplesValid(samples, TEST_TOPIC);
+        verify(_prometheusAdapter);
+    }
+
+    @Test
+    public void testGetSamplesDotHandledTopic() throws Exception {
```
efeg, line 205: This seems a little repetitive and I suspect that it might be useful to test dot handled topic with other cases.

* Would it be possible to keep `TEST_TOPIC` and `TEST_TOPIC_WITH_DOT` in a  `Set`, and then run each test in this class that requires a `topic` for each topic in our set?

efeg (changes requested):
@hd-rk Thanks for the patch!
Left a comment.

efeg (approved):
@hd-rk Thanks for the updates -- LGTM.

PR #1498, opened 2021-03-24
Fixed network.outbound.low.utilization.threshold

tomek-cz:
Fixed description of network.outbound.low.utilization.threshold


efeg (approved):
@tomek-cz Thanks or the fix!

PR #1514, opened 2021-04-11
Prevent ResourceDistributionGoal from generating provision recommendations with a negative number of brokers to remove.

### LINKED ISSUE #1513: Prevent ResourceDistributionGoal from generating provision recommendations with a negative number of brokers to remove
In heterogeneous clusters with brokers having different resource capacities, `ResourceDistributionGoal` may generate provision recommendations with a negative number of brokers to remove. This causes the optimization to fail with an `IllegalArgumentException` such as the following:

```
Error processing POST request '/rebalance' due to: 'com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException: java.lang.IllegalArgumentException: Number of brokers must be positive (-176).'
```

Ensure that the ResourceDistributionGoal does not generate provision recommendations with a negative number of brokers to remove.

efeg:
This PR resolves #1513.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
@@ -268,16 +268,33 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
     _isLowUtilization = avgUtilizationPercentage <= _balancingConstraint.lowUtilizationThreshold(resource());
     if (_isLowUtilization) {
       // Identify a typical broker capacity to be used in recommendations in case the cluster is over-provisioned.
-      int typicalBrokerId = _brokersAllowedReplicaMove.iterator().next();
+      // To make sure that the recommended brokers to drop in an heterogeneous cluster is a positive number, the typical broker must have
+      // the max capacity.
+      int typicalBrokerId = brokerIdWithMaxCapacity(clusterModel);
       double typicalCapacity = clusterModel.broker(typicalBrokerId).capacityFor(resource());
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -268,18 +268,35 @@ protected void initGoalState(ClusterModel clusterModel, OptimizationOptions opti
...
+      // To make sure that the recommended brokers to drop in an heterogeneous cluster is a positive number, the typical broker must have
+      // the max capacity.
+      int typicalBrokerId = brokerIdWithMaxCapacity(clusterModel);
       double typicalCapacity = clusterModel.broker(typicalBrokerId).capacityFor(resource());
 
       // Any capacity greater than the allowed capacity may yield over-provisioning.
       double allowedCapacity = resourceUtilization / _balancingConstraint.lowUtilizationThreshold(resource());
       int allowedNumBrokers = (int) (allowedCapacity / typicalCapacity);
-      int numBrokersToDrop = _brokersAllowedReplicaMove.size() - allowedNumBrokers;
+      int numBrokersToDrop = Math.max(_brokersAllowedReplicaMove.size() - allowedNumBrokers, 1);
```
bachmanity1, line 279: @efeg 

What if `_brokersAllowedReplicaMove.size()`  equals `allowedNumBrokers`? Wouldn't this suggest removing a broker even when cluster size doesn't need any change?

Wouldn't it be safer to make this line look like this? 

```java
int numBrokersToDrop = Math.max(_brokersAllowedReplicaMove.size() - allowedNumBrokers, 0);
```

Issue #1531, opened 2021-04-23
Stopping cruise control will kill kafka accidentally


wenbingshen:
Because the metrics reporter of cruise control needs to be integrated into the kafka dependency, this will cause ps ax | grep -i'cruise-control' to also filter out the kafka process.

PR #1532, opened 2021-04-23
Consolidate the stop script to avoid stopping the kafka service

wenbingshen:
This PR resolves [1531](https://github.com/linkedin/cruise-control/issues/1531).

For cruise control, we should use the main class to filter the process id in the stop step, so as to avoid accidentally killing other processes that rely on cruise control.

### DIFF
--- a/kafka-cruise-control-stop.sh
+++ b/kafka-cruise-control-stop.sh
@@ -3,7 +3,7 @@
 # See License in the project root for license information.
 
 SIGNAL=${SIGNAL:-TERM}
-PIDS=$(ps ax | grep -i 'cruise-control' | grep java | grep -v grep | awk '{print $1}')
+PIDS=$(ps ax | grep -i 'KafkaCruiseControlMain' | grep java | grep -v grep | awk '{print $1}')
 
 if [ -z "$PIDS" ]; then
   echo "No cruise-control to stop"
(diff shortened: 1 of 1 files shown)


kafka-cruise-control-stop.sh
```diff
@@ -3,7 +3,7 @@
 # See License in the project root for license information.
 
 SIGNAL=${SIGNAL:-TERM}
-PIDS=$(ps ax | grep -i 'cruise-control' | grep java | grep -v grep | awk '{print $1}')
+PIDS=$(ps ax | grep -i 'KafkaCruiseControlMain' | grep java | grep -v grep | awk '{print $1}')
```
efeg, line 6: Nits:
1. Is case insensitive matching really needed? If not can we drop `-i` option?
2. According to `kafka-cruise-control-start.sh` `java` always precedes `KafkaCruiseControlMain`. Can we enforce that in our command?

 -- i.e. would the following work?
```
PIDS=$(ps ax | grep 'java.*KafkaCruiseControlMain' | grep -v grep | awk '{print $1}')
```

efeg (commented):
@wenbingshen Thanks for the PR!
Left a comment.

kafka-cruise-control-stop.sh
wenbingshen: Thanks for you comment. Good suggestion. I submitted your feedback.

efeg (approved):
@wenbingshen thanks for the update! LGTM.

PR #1555, opened 2021-05-12
Add Known Compatibility Issues section to README

### LINKED ISSUE #1535: Cc failed to restart due to the number of kafka nodes being less than the number of sample topic replicas
First, the sample topic has been created in Kafka, and the number of replicas is the number of kafka nodes. When one of the kafka nodes is stopped and cc is restarted, cc will report that the creation of the sample topic failed because the number of kafka nodes is less than the number of replicas, but the actual Above, the sample topic of cc has been created in kafka, so we should not block the startup of cc.

I analyzed the kafka code. This is because kafka will first assign replicas for the created topic, and then verify whether the topic exists. When the copy allocation cannot be completed, an InvalidReplicationFactorException will be thrown.

image

image

zornhsu:
This PR resolves #1535. Additionally we could add more details to the troubleshooting guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Cruise Control for Apache Kafka
     * Fix offline replicas
     * Adjust replication factor
 
-### Environment Requirements
+### Environment Requirements ###
 * The current `master` branch of Cruise Control is compatible with Apache Kafka `2.0`, `2.1`, `2.2`, and `2.3` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.0.*`)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,7 +46,7 @@ Cruise Control for Apache Kafka
     * Fix offline replicas
     * Adjust replication factor
 
-### Environment Requirements
+### Environment Requirements ###
```
efeg, line 49: Other than their cosmetic use, the only other use of closing hashes that I know is to escape `#` sign in the header. Given that all the other headers have the closing hashes, this change is good for consistency -- thanks for this change!

efeg (approved):
Thanks for adding this to the documentation -- LGTM!

PR #1559, opened 2021-05-15
Add a ReplicaMovementStrategy that prioritizes (At/Under)MinISR partitions with offline replicas

### LINKED ISSUE #1558: Add a ReplicaMovementStrategy that prioritizes (At/Under)MinISR partitions with offline replicas
**Background**: During a recent issue, multiple brokers became offline and caused a lot of URPs. A small subset of these URPs became `AtMinISR` -- i.e. were at the risk of being unavailable for producers with `acks=all` in case of an additional broker failure hosting a replica of these partitions. Cruise Control detected the failed brokers and started a self-healing. However, although the total URPs in the cluster has been dropping over time, the total `AtMinISR` count was not.

When multiple brokers are offline, (At/Under)MinISR partitions are at a higher risk -- i.e. further failures can make them unavailable  for produce and consume. In particular, Under MinISR partitions are already unavailable for producers with `acks=all`. Cruise Control can avoid such a risky state by prioritizing the fix for `UnderMinISR` and then `AtMinISRs` partitions with offline replicas before proceeding to fix the general URPs with offline replicas.

Introduce a new replica movement strategy to help fixing such partitions faster.

efeg:
This PR resolves #1558.

* Make `TopicMinIsrCache#minIsrWithTimeByTopic()` return a snapshot within a `synchronized` block to ensure proper synchronization.
* Create `StrategyOptions` to move any options intended to be used during application of a replica movement strategy. The existing code was limited to `Cluster` as the sole option, and didn't allow adding new options to a replica movement strategy without an API change in `ReplicaMovementStrategy` interface. With `StrategyOptions` new options can be added w/o changing the API.
* Drop redundant chaining of `BaseReplicaMovementStrategy` to the default replica movement task strategy while creating an `ExecutionTaskPlanner` object. The redundancy was due to lack of check for the configured `default.replica.movement.strategies`. If the config had a list of entries containing `BaseReplicaMovementStrategy` (i.e. this is the default config), then the existing was still chaining an extra `BaseReplicaMovementStrategy`, causing inefficiency for no reason.
* Add an option to `ExecutionUtils#populateMinIsrState` function to control whether to retrieve (At/Under)MinISR partitions each containing at least an offline replica, or to retrieve (At/Under)MinISR partitions without any offline replicas.
* Current PR assumes that if users intend to use the new `PrioritizeMinIsrWithOfflineReplicasStrategy` strategy, then minISR-based concurrency adjustment is enabled. Otherwise, the `PrioritizeMinIsrWithOfflineReplicasStrategy` strategy is just a no-op.
* Remove duplicate config entry for `default.replica.movement.strategies` from `Analyzer Configurations` -- it is already documented in the correct section (i.e. `Executor Configurations`)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PostponeUrpReplicaMovementStrategy.java
```diff
@@ -15,11 +15,16 @@
  */
 public class PostponeUrpReplicaMovementStrategy extends AbstractReplicaMovementStrategy {
 
+  @Override
+  public Comparator<ExecutionTask> taskComparator(StrategyOptions strategyOptions) {
+    return (task1, task2) -> isPartitionUnderReplicated(strategyOptions.cluster(), task1.proposal().topicPartition())
+                             ? (isPartitionUnderReplicated(strategyOptions.cluster(), task2.proposal().topicPartition()) ? 0 : 1)
```
zornhsu, line 21:
```suggestion
                             ? (isPartitionUnderReplicated(strategyOptions.cluster(), task2.proposal().topicPartition()) ? PRIORITIZE_NONE : PRIORITIZE_TASK_2)
```
I like your `int PRIORITIZE_TASK_2 = 1` a lot. Is refactoring them here a good idea?

efeg: Sure -- applied the proposed change.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -122,60 +126,110 @@ public void testGetInterBrokerPartitionMovementTasks() {
...
     partitionMovementTasks = prioritizeSmallMovementPlanner.getInterBrokerReplicaMovementTasks(readyBrokers, Collections.emptySet());
     assertEquals("First task should be partitionMovement1", _partitionMovement1, partitionMovementTasks.get(0).proposal());
     assertEquals("Second task should be _partitionMovement3", _partitionMovement3, partitionMovementTasks.get(1).proposal());
     assertEquals("Third task should be partitionMovement4", _partitionMovement4, partitionMovementTasks.get(2).proposal());
     assertEquals("Fourth task should be partitionMovement2", _partitionMovement2, partitionMovementTasks.get(3).proposal());
+
+    smallUrpMovementPlanner.addExecutionProposals(proposals, strategyOptions, null);
+    partitionMovementTasks = smallUrpMovementPlanner.getInterBrokerReplicaMovementTasks(readyBrokers, Collections.emptySet());
+    assertEquals("First task should be partitionMovement4", _partitionMovement4, partitionMovementTasks.get(0).proposal());
+    assertEquals("Second task should be _partitionMovement2", _partitionMovement2, partitionMovementTasks.get(1).proposal());
```
zornhsu, line 216:
```suggestion
    assertEquals("Second task should be partitionMovement2", _partitionMovement2, partitionMovementTasks.get(1).proposal());
```

zornhsu: Or, maybe adding `_` to every line.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -122,60 +126,110 @@ public void testGetInterBrokerPartitionMovementTasks() {
...
     Set<PartitionInfo> partitions = new HashSet<>();
-    partitions.add(generatePartitionInfo(_partitionMovement1, true));
+    partitions.add(generatePartitionInfoWithUrpHavingOfflineReplica(_partitionMovement1, true));
     partitions.add(generatePartitionInfo(_partitionMovement2, false));
-    partitions.add(generatePartitionInfo(_partitionMovement3, true));
+    partitions.add(generatePartitionInfoWithUrpHavingOfflineReplica(_partitionMovement3, true));
     partitions.add(generatePartitionInfo(_partitionMovement4, false));
 
     Cluster expectedCluster = new Cluster(null, _expectedNodes, partitions, Collections.<String>emptySet(), Collections.<String>emptySet());
+    // This ensures that the _partitionMovement1 and _partitionMovement3 are AtMinISR, while the other partitions are not.
```
zornhsu, line 178: Maybe we want to add a test for UnderMinISR is prioritized before AtMinISR partitions?
I am a little hesitant because I don't see a quick way to add this without involving 3 replicas in the replica set.

zornhsu (approved):
LGTM. Thanks for the feature, Efe! 
I left some comments, but shouldn't be merge blockers.

PR #1569, opened 2021-05-19
Switch from ZK to Kafka Admin Client

### LINKED ISSUE #1415: Replace AdminZkClient with AdminClient
Motivations:

1. `AdminZkClient` gets information from Zookeeper which will be removed from Kafka as a dependency ([KIP-500](https://cwiki.apache.org/confluence/display/KAFKA/KIP-500%3A+Replace+ZooKeeper+with+a+Self-Managed+Metadata+Quorum)).

2. Java doc of `AdminZkClient` class says `This is an internal class and no compatibility guarantees are provided, see org.apache.kafka.clients.admin.AdminClient for publicly supported APIs.`

tomncooper:
This PR partially resolves #1415. 

To keep code review easier I am proposing to make changes in stages, starting with the `KafkaTopicConfigProvider` and then moving on to the other classes which use `AdminZkClient`.


tomncooper:
@Lincong @efeg I have started the work on removing the `ZkAdminClient`, currently (to keep the size of the code review down) this just deals with the `TopicConfigProvider`. If you are happy with the general approach I can either add to this PR or open separate ones?

tomncooper:
This converts the default `KafkaTopicConfigProvider` to use the Kafka Admin Client and creates a new `TopicConfigProvider` implementation that uses the `ZkAdminClient` so a user could switch back to the old Provider logic if they need to?

efeg:
@tomncooper Thanks for the PR! I will try to review it within this week.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaTopicConfigProvider.java
```diff
@@ -88,8 +146,10 @@ private void loadClusterConfigs(String clusterConfigsFile) throws FileNotFoundEx
 
   @Override
   public void configure(Map<String, ?> configs) {
-    _connectString = (String) configs.get(ExecutorConfig.ZOOKEEPER_CONNECT_CONFIG);
-    _zkSecurityEnabled = (Boolean) configs.get(ExecutorConfig.ZOOKEEPER_SECURITY_ENABLED_CONFIG);
+
+    KafkaCruiseControlConfig ccConfig = new KafkaCruiseControlConfig(configs);
+    _adminClient = KafkaCruiseControlUtils.createAdminClient(KafkaCruiseControlUtils.parseAdminClientConfigs(ccConfig));
+
```
efeg, line 152: Rather than creating a new admin client here, can we use the existing admin client of `KafkaCruiseControl`? We can pass the existing admin client to the TopicConfigProvider upon configuration.
-- please see example usage in other pluggable component implementations.

efeg (commented):
@tomncooper Thanks for the PR!
Made a first pass and left a few comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -90,6 +90,7 @@
   public static final int ZK_CONNECTION_TIMEOUT = (int) TimeUnit.MINUTES.toMillis(2);
   public static final long KAFKA_ZK_CLIENT_CLOSE_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(10);
   public static final long ADMIN_CLIENT_CLOSE_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(10);
+  public static final long ADMIN_CLIENT_REQUEST_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(10);
```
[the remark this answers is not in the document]
tomncooper: Good point I will use the configured `admin.client.request.timeout.ms`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaTopicConfigProvider.java
tomncooper: So that was my original intention but I thought that would break compatibility with the `TopicConfigProvider` and `CruiseControlConfigurable` interfaces. Are you suggesting we add a new `configure` method signiture to the `CruiseControlConfigurable` interface or somehow add the Admin client into the `Map<String,?> given to the existing `configure` method?

Also I was worried that as `TopicConfigProvider` is auto-closable and we are using a single admin client this may cause issues on shutdown (unless we make the `close()` method no-op in the new topic config provider)? But maybe that is not such a concern?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaZkTopicConfigProvider.java
```diff
@@ -0,0 +1,105 @@
...
+ * The Kafka topic config provider implementation based on files. The format of the file is JSON, listing properties:
+ * <pre>
+ *   {
+ *     "min.insync.replicas": 1,
+ *     "an.example.cluster.config": false
+ *   }
+ * </pre>
+ *
+ */
+public class KafkaZkTopicConfigProvider implements TopicConfigProvider {
```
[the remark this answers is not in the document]
tomncooper: I have added deprecated to the ZK provider and switched the names around. 

I have made the Kafka Admin provider the default? Is that how you want to proceed or do you want to have a release or two with the ZK provider as the primary still and then switch over?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  @Override
+  public Map<String, Properties> allTopicConfigs() {
+    Map<ConfigResource, Config> topicConfigs = null;
+    try {
+      LOG.debug("Requesting configurations for all topics");
+      topicConfigs = _adminClient
+              .listTopics()
+              .names()
```
efeg, line 95: It is unfortunate that Kafka does not provide a direct API to retrieve all topic configs.
Given this change, we double the number of adminClient requests and roundtrip time -- i.e. once to receive topic names and another time to get their configs.

The following item doesn't have to be in this patch, but I think it should definitely be before switching to this class as the default topic config provider:
* Can we add a new method to `TopicConfigProvider` interface with the following signature?
```
Map<String, Properties> topicConfigs(Set<String> topics);
```

Adding this method would let us retrieve given topics with a single network roundtrip -- i.e. single Admin Client request. This new method would be useful to retrieve configs efficiently in `KafkaClusterState`, where we know the topics beforehand.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
```diff
@@ -0,0 +1,168 @@
...
+              .thenApply(
+                      topicNameSet -> _adminClient.describeConfigs(
+                              topicNameSet.stream().map(name -> new ConfigResource(ConfigResource.Type.TOPIC, name)).collect(Collectors.toList())
+                      ).all()
```
efeg, line 99: `all()` method returns a future which succeeds _only if all the config descriptions succeed_.
To prevent individual failures causing the whole request to fail, can we instead use `values()` response and handle individual failures separately?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaTopicConfigProvider.java
efeg: Regarding the first point, we can achieve this w/o any changes in the API of `TopicConfigProvider` or `CruiseControlConfigurable` interfaces. I left a comment on how to do so with an existing example -- e.g. `AnomalyDetectorUtils#KAFKA_CRUISE_CONTROL_OBJECT_CONFIG`.

Regarding the second point, we should indeed make this a no-op close() as the shared admin client is already being closed last at `KafkaCruiseControl#shutdown()`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaZkTopicConfigProvider.java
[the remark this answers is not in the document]
efeg: It would indeed be useful to have at least one release before we change the default value. So this patch would add support and a followup patch would make it the default.

efeg (changes requested):
Thanks for the changes! Left more comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
tomncooper: Good point, I think I have changed this to something that should work but it might be a bit _over_ engineered. Let me know what you think.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
```diff
@@ -0,0 +1,141 @@
...
+ *     "an.example.cluster.config": false
+ *   }
+ * </pre>
+ *
+ */
+public class KafkaAdminTopicConfigProvider extends JsonFileTopicConfigProvider {
+
+  private static final Logger LOG = LoggerFactory.getLogger(KafkaAdminTopicConfigProvider.class);
+
+  public static final String CLUSTER_CONFIGS_FILE = "cluster.configs.file";
```
efeg, line 40: Nit: Should we move this to `JsonFileTopicConfigProvider`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaTopicConfigProvider.java
```diff
@@ -30,14 +23,20 @@
...
+ * This class uses the Zookeeper based admin client that will be removed in Kafka 3.0. Therefore this class has been
+ * deprecated and will be removed in a future Cruise Control release. A new {@link TopicConfigProvider} implementation
+ * using the Kafka Admin Client has been created ({@link KafkaAdminTopicConfigProvider}) and can be set using the
+ * {@code topic.config.provider.class} configuration setting.
+ *
  */
-public class KafkaTopicConfigProvider implements TopicConfigProvider {
+@Deprecated
+public class KafkaTopicConfigProvider extends JsonFileTopicConfigProvider {
   public static final String CLUSTER_CONFIGS_FILE = "cluster.configs.file";
```
efeg, line 34: Nit: Should we move this to `JsonFileTopicConfigProvider`?

efeg (commented):
Thanks for the update!
Left a few initial comments, will take a closer look at `KafkaAdminTopicConfigProvider.java` later.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
tomncooper: I agree this would be a good thing to add but was leaving this until you are happy with the `KafkaAdminTopicProvider` implementation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
tomncooper: Good point, I will move it there and remove from both implementations.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
tomncooper: I am implementing this change. However, I had a question about what these methods should return if an error is encountered. Under the Zookeeper client implementations the `kafka.zk.KafkaZkClient#getEntityConfigs` method will return an empty `Properties` instance if the ZK node doesn't exist and throw an exception for any other case. 

So I was wondering what the expected behavior would be for the Kafka Admin Client version. Should I throw an exception if any error is encountered or are their cases where I should return an empty `Map`/`Properties`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
```diff
@@ -0,0 +1,140 @@
...
+
+      }
+
+      //Block on all the config futures completing
+      try {
+        KafkaFuture.allOf(topicConfigs.values().toArray(new KafkaFuture[0])).get();
+      } catch (InterruptedException | ExecutionException e) {
+        LOG.warn("Config check for all topics failed due to failure to describe their configs.", e);
+      }
+    }
```
efeg, line 113: I feel we might be able to simplify this logic a bit using a single loop that iterates over `topicConfigs.entrySet()` -- similar to [TopicMinIsrCache#putTopicMinIsr(DescribeConfigsResult)](https://github.com/linkedin/cruise-control/blob/master/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/TopicMinIsrCache.java#L101 ).

efeg (commented):
Thanks for the updates! Left comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
tomncooper: I have simplified this method in line with the method you suggested. Let me know what you think?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
tomncooper: I added the multi-topic fetch method to the interface and made it so that empty properties/maps are returned in the case of most errors. Let me know what you think.

efeg:
@tomncooper the build seems to have [failed due to style issues](https://app.circleci.com/pipelines/github/linkedin/cruise-control/1119/workflows/6d84a1f8-0de1-44d4-813c-c210ddfe22a8/jobs/4283):
```
> Task :cruise-control:checkstyleMain
[ant:checkstyle] [ERROR] /home/circleci/workspace/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/JsonFileTopicConfigProvider.java:52: Multiple empty lines after this line [RegexpMultiline]
[ant:checkstyle] [ERROR] /home/circleci/workspace/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/TopicConfigProvider.java:12:1: Extra separation in import group before 'org.apache.kafka.common.annotation.InterfaceStability' [CustomImportOrder]
```

tomncooper:
> @tomncooper the build seems to have [failed due to style issues](https://app.circleci.com/pipelines/github/linkedin/cruise-control/1119/workflows/6d84a1f8-0de1-44d4-813c-c210ddfe22a8/jobs/4283):

Sorry about that. I keep hitting checkstyle issues (I need to tweak my IDE). I have made a pre-commit hook to run checkstyle before each commit. Would that be something upstream would be interested in (optionally obviously)?.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
```diff
@@ -0,0 +1,171 @@
...
+    LOG.debug("Requesting configurations for all topics");
+    Set<String> topicNames = null;
+    try {
+      topicNames = _adminClient.listTopics().names().get();
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.warn("Unable to obtain list of all topic names from the Kafka Cluster");
+    }
+
+    if (topicNames == null) {
+      return new HashMap<>();
```
efeg, line 106: Nit: `return new HashMap<>();` -> `return Collections.emptyMap();` -- i.e. no need to create an object.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaAdminTopicConfigProvider.java
```diff
@@ -0,0 +1,171 @@
...
+    if (topicConfigs != null) {
+      for (Map.Entry<ConfigResource, KafkaFuture<Config>> entry : topicConfigs.entrySet()) {
+        try {
+          Config config = entry.getValue().get();
+          propsMap.put(entry.getKey().name(), convertTopicConfigToProperties(config));
+        } catch (ExecutionException ee) {
+          if (org.apache.kafka.common.errors.TimeoutException.class == ee.getCause().getClass()) {
+            LOG.warn("Failed to retrieve config for topic '{}' due to describeConfigs request time out. Check for Kafka-side issues"
+                            + " and consider increasing the configured timeout.", entry.getKey().name());
+          } else {
```
efeg, line 137: I believe if one of the futures would returne a `org.apache.kafka.common.errors.TimeoutException`, then all of the futures are expected to return a `org.apache.kafka.common.errors.TimeoutException`. Hence, to avoid logging a `warn` for individual topics, maybe we can log a single `warn` message to indicate failure for all topics and then return?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaTopicConfigProvider.java
```diff
@@ -58,10 +62,33 @@ public Properties topicConfigs(String topic) {
...
+
+    KafkaCruiseControlUtils.closeKafkaZkClientWithTimeout(kafkaZkClient);
+    return topicConfigs;
+  }
+
   @Override
   public Map<String, Properties> allTopicConfigs() {
     KafkaZkClient kafkaZkClient = KafkaCruiseControlUtils.createKafkaZkClient(_connectString,
-                                                                              ZK_KAFKA_TOPIC_CONFIG_PROVIDER_METRIC_GROUP,
+            ZK_KAFKA_TOPIC_CONFIG_PROVIDER_METRIC_GROUP,
```
efeg, line 91: Nit: Broken indentation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaTopicConfigProvider.java
```diff
@@ -58,10 +62,33 @@ public Properties topicConfigs(String topic) {
     }
   }
 
+  @Override
+  public Map<String, Properties> topicConfigs(Set<String> topics) {
+    KafkaZkClient kafkaZkClient = KafkaCruiseControlUtils.createKafkaZkClient(_connectString,
+            ZK_KAFKA_TOPIC_CONFIG_PROVIDER_METRIC_GROUP,
+            ZK_KAFKA_TOPIC_CONFIG_PROVIDER_METRIC_TYPE,
+            _zkSecurityEnabled);
```
efeg, line 70: Nit: Broken indentation

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaTopicConfigProvider.java
```diff
@@ -58,10 +62,33 @@ public Properties topicConfigs(String topic) {
...
+
+    AdminZkClient adminZkClient = new AdminZkClient(kafkaZkClient);
+
+    Map<String, Properties> topicConfigs = new HashMap<>();
```
efeg, line 74: Nit: `new HashMap<>();` -> `new HashMap<>(topics.size());`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaTopicConfigProvider.java
```diff
@@ -58,10 +62,33 @@ public Properties topicConfigs(String topic) {
...
   }
 
+  @Override
+  public Map<String, Properties> topicConfigs(Set<String> topics) {
+    KafkaZkClient kafkaZkClient = KafkaCruiseControlUtils.createKafkaZkClient(_connectString,
+            ZK_KAFKA_TOPIC_CONFIG_PROVIDER_METRIC_GROUP,
+            ZK_KAFKA_TOPIC_CONFIG_PROVIDER_METRIC_TYPE,
+            _zkSecurityEnabled);
+
+    AdminZkClient adminZkClient = new AdminZkClient(kafkaZkClient);
```
efeg, line 72: Nit: Just for consistency -- can we keep (1) creation of AdminZkClient and (2) closing the AdminZkClient in a `try-finally` block?

efeg (commented):
Left a few nits. Once they are addressed, the PR looks good to me.

efeg:
Hi @tomncooper -- do you think we can address the remaining items this week to merge this PR?

tomncooper:
Hi @efeg, yes, I got sidetracked with some other things. I will try and address these this week.

tomncooper:
Hi @efeg, I think I addressed all your comments. Thanks.

efeg (approved):
Thanks for the updates! LGTM.

PR #1571, opened 2021-05-21
Copy through configuration for sasl.client.callback.handler.class in AdminClient

### LINKED ISSUE #1570: Admin Client does not pass through sasl.client.callback.handler.class
I'm currently running into an issue where I'm changing the SASL mechanism between our components in the kafka cluster. Unfortunately, the AdminClient created by cruise control does not pass in the sasl.client.callback.handler.class class, causing the calls to fail due to a `javax.security.auth.callback.UnsupportedCallbackException: Unrecognized SASL ClientCallback` error.

After a good amount of investigating, I determined that the issue is that the issue is with KafkaCruiseControlUtils, specifically [parseAdminClientConfigs](https://github.com/linkedin/cruise-control/blob/22113c884e6fb13ab2aaf068b835cca666a52242/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java#L523). The parser never sets `SASL_CLIENT_CALLBACK_HANDLER_CLASS`, meaning we can never set the callback handler class to something other then the default `SaslClientCallbackHandler`. 

If possible, I would like to merge in a fix that simply adds 1 line to the parser and allows for `sasl.client.callback.handler.class` to be passed through to the AdminClient.

Sylvain-Royer:
This PR resolves #1570.


efeg (approved):
@Sylvain-Royer Thanks for the fix! LGTM.

PR #1572, opened 2021-05-21
Remove unused skip.sample.store.topic.rack.awareness.check config

### LINKED ISSUE #1548: Drop unused skip.sample.store.topic.rack.awareness.check config from KafkaSampleStore configurations
This config was added in https://github.com/linkedin/cruise-control/pull/508, but over time, self-healing operations for the violations of desired replication factor are left to `TOPIC_ANOMALY` self-healing.

So effectively, this is an unused config with no affect. We should drop this config.

CalvinPVIII:
This PR resolves #1548
I think that this was what was wanted but not exactly sure


efeg (approved):
@CalvinPVIII Thanks for the patch! LGTM.

PR #1576, opened 2021-06-03
Change skip_urp_demotion parameter default to true

### LINKED ISSUE #1575: Change skip_urp_demotion parameter default to true
During a recent demote operation to free up the request handler on a broker, CC stopped my operation due to my demoted broker had offline replicas:
```
2021/04/28 16:44:59.444 INFO [Executor] [ProposalExecutor-0] [kafka-cruise-control] [] Stop the execution due to 9 dead tasks: ...
```
However, users usually don't care if there's one or two leader partitions left on the target broker. In this case, they might just want to free the request handler ASAP. The workaround was to use `skip_urp_demotion=true`, where it is default to false currently.

We should change this parameter default to `true`, so that CC has a higher chance to complete the demote operation without specifying explicit parameter.

zornhsu:
This PR resolves #1575.


efeg (commented):
@zornhsu Thanks for the PR!

Can we also update the `default` value of `skip_urp_demotion` in `yaml/endpoints/demoteBroker.yaml`?

efeg (approved):
Thanks for the updates - LGTM.

PR #1581, opened 2021-06-09
Identify and fix inefficient code paths in goal optimization

### LINKED ISSUE #1580: Identify and fix inefficient code paths in goal optimization
Goal optimization is a core process in Cruise Control, which is in the critical path for many key maintenance operations, including rebalance, broker removal/addition/demotion, fixing offline replicas, changing replication factor, and detecting goal violations. Hence, it is important to ensure that the process is free from inefficient code paths that cause unnecessary delays in the process.

Identify and fix (selected) inefficient code paths in goal optimization.

efeg:
This PR resolves #1580.

The selected inefficiencies addressed in this PR have been identified based on CPU profiling during optimization of goals. This patch does not cover _all_ potential inefficient code paths, but only addresses key inefficiencies that stand out in profiling results.

As a result of these improvements, clean build time of Cruise Control with all unit tests and static analysis dropped from
* [on Local box] `~16 minutes` to `~14.5 minutes` -- i.e. local CC build is `9.4%` faster

In particular, local run of the `RandomClusterExpDistNewBrokerTest` (i.e. one of the most expensive CC test that performs multiple goal optimizations with different parameters) has dropped from 
* [on Local box] `3minutes 50s` to `3minutes 10s`, which means `~17.4%` improvement. _Note: Performance numbers are collected from running the test 3 times before and after the change and comparing the average._

Notes -- notable changes:
* Avoid cloning sortedReplicas in `ResourceDistributionGoal#rebalanceByMovingLoadIn` as this operation is expensive and has been performed frequently in this function.
* Add priority function `ReplicaSortFunctionFactory.prioritizeOfflineReplicas()` only if the cluster has offline replicas
* Add selection function `ReplicaSortFunctionFactory.selectReplicasBasedOnExcludedTopics(excludedTopics)` only if the cluster has excluded topics.
* Cache number of replicas by topic rather than recomputing it every time based on topics on each brokers.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
@@ -27,7 +27,6 @@ import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.PriorityQueue;
 import java.util.Set;
 import java.util.SortedSet;
@@ -461,6 +460,7 @@ public abstract class ResourceDistributionGoal extends AbstractGoal {
       // Get candidate replicas on candidate broker to try moving load from -- sorted in the order of trial (descending load).
       if (utilization(candidate, resource()) > (isExcludedForReplicaMove(candidate) ? 0.0 : _balanceLowerThreshold)) {
         replicaSortName = sortedCandidateReplicas(candidate,
+                                                  clusterModel,
                                                   excludedTopics,
                                                   0.0,
                                                   false,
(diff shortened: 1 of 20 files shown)


zornhsu (commented):
Thanks for the patch, Efe! 
I have finished reviewing the PR except for `ClusterModelStats.java` and `ResourceDistributionGoal.java`. 
I'll keep reviewing them tomorrow. The rest of the PR looks good to me.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -476,31 +476,47 @@ private boolean rebalanceByMovingLoadIn(Broker broker,
...
+      while (!readyToPollNextCb) {
+        // Track indices that have already been iterated in candidate replicas to receive.
+        int iteratedIndices = indicesToSkip;
+        int maxIndicesToIterate = candidateReplicasToReceive.size();
+        for (Replica replica : candidateReplicasToReceive) {
+          // Skip indices of candidate replicas that have already been checked before.
+          if (indicesToSkip > 0) {
+            indicesToSkip--;
+            continue;
           }
```
zornhsu, line 494: Where do we remove source broker's replica in the new PR? Or do we still need to remove it?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
```diff
@@ -476,31 +476,47 @@ private boolean rebalanceByMovingLoadIn(Broker broker,
...
-        // Only need to check status if the action is taken. This will also handle the case that the source broker
-        // has nothing to move in. In that case we will never reenqueue that source broker.
-        if (b != null) {
-          if (isLoadAboveBalanceLowerLimit(broker)) {
-            clusterModel.untrackSortedReplicas(replicaSortName);
-            return false;
-          }
-          // Remove the replica from its source broker if it was a replica movement.
-          if (actionType == INTER_BROKER_REPLICA_MOVEMENT) {
-            iterator.remove();
```
zornhsu, line 495: Where do we remove source broker's replica in this PR? Or, maybe we don't need to remove the replica in the beginning?

zornhsu, line 495: Where do we remove source broker's replica in this PR? Or, maybe we don't need to remove the replica in the original code?

zornhsu (approved):
Looks good to me as long as we are sure we didn't miss removing source broker's replica.
Thanks for speeding the optimizations up, Efe!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -493,8 +495,8 @@ public boolean relocateLeadership(TopicPartition tp, int sourceBrokerId, int des
   /**
    * @return Brokers containing offline replicas residing on bad disks in the current cluster model.
    */
-  public Set<Broker> brokersHavingOfflineReplicasOnBadDisks() {
-    Set<Broker> brokersWithOfflineReplicasOnBadDisks = new HashSet<>();
```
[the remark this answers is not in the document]
efeg: This is mainly to ensure consistency -- i.e. both (1) consistency with the other methods and (2) to make sure that the response is always providing the same order, which prevents potential future goals from generating different proposals due to processing `brokersHavingOfflineReplicasOnBadDisks` in a different order.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ResourceDistributionGoal.java
efeg: The original remove was non-harmful, but redundant. The actual removal happens once `maybeApplyBalancingAction` is called above -- i.e. it calls `clusterModel#relocateReplica` or `clusterModel#relocateLeadership` to achieve that.

PR #1584, opened 2021-06-10
Add (optional) checkstyle pre-commit hook script

tomncooper:
I often forget to run CheckStyle before pushing a commit, which then fails the CI build. Therefore I added a pre-commit hook to run checkstyle (via the gradle checkstyle plugin). This is obviously optional, it does increase the time to do a commit, but it reduces the wasted effort of running CI on something that has style issues.


efeg (approved):
Thanks for the patch! LGTM.

PR #1589, opened 2021-06-15
Add parameters to relevant endpoints to control the speed of proposal generation

### LINKED ISSUE #589: Add parameters to relevant endpoints to control the speed of proposal generation
Endpoints that rely on proposal generation (e.g. `rebalance`, `add_broker`, `remove_broker`, `proposals`, `fix_offline_replicas`) may take long / unpredictable time to generate proposals due to the implementation logic in goals.

Cruise Control would enable users to control the speed at which proposals are generated. While the fast proposal generation may reduce the quality of proposals for soft goals, it may improve the user experience to retrieve a quick solution under time-critical cases.

efeg:
This PR resolves #589.

**Notes.**
* New parameter `fast_mode=[true/false]` under `GoalBasedOptimizationParameters` -- i.e. `add_broker`, `fix_offline_replicas`, `proposals`, `rebalance`, `remove_broker`, and `topic_configuration`.
* New Analyzer config `fast.mode.per.broker.move.timeout.ms` -- The per broker move timeout in fast mode in milliseconds. Users can run goal optimizations in fast mode by setting the `fast_mode` parameter to `true` in relevant endpoints. This mode intends to provide a more predictable runtime for goal optimizations.
* Cleanup `OptimizationOptions` constructors with 1 and 2 parameters.
* Make `PER_BROKER_SWAP_TIMEOUT_MS` use the configurable `2*_fastModePerBrokerMoveTimeoutMs` (i.e. multiplied by 2 to take bidirectional move into account).


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/OptimizationOptions.java
```diff
@@ -20,20 +20,7 @@
...
-  public OptimizationOptions(Set<String> excludedTopics) {
-    this(excludedTopics, Collections.emptySet());
-  }
-
-  /**
-   * Default value for {@link #_excludedBrokersForReplicaMove} is an empty set.
-   */
-  public OptimizationOptions(Set<String> excludedTopics, Set<Integer> excludedBrokersForLeadership) {
-    this(excludedTopics, excludedBrokersForLeadership, Collections.emptySet());
-  }
```
zornhsu, line 36: Just curious, is the reason to remove these two method because both the 2nd and 3rd parameters are Set<int>, which causes confusion?

efeg: This class uses _telescoping constructor pattern_.  It does not scale well with the growing number of parameters, because it gets harder to write client code and read it. As you correctly identified 2nd and 3rd parameters are providing a similar functionality (i.e. both `Set<Integer> excludedBrokersForX`). Hence, having a constructor with the first one, but not the other indeed causes confusion. That is why these two constructors were dropped.

That being said, there are alternatives to this design pattern. In particular _Builder pattern_ would be a good alternative (see `ProvisionRecommendation` for an example use of it). The challenge with builder pattern is that the inheritance becomes more messy -- i.e. requires all the non-leaf classes to be abstract -- and `OptimizationOptions` is a class that we would like to extend from (internally).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/DemoteBrokerRunnable.java
```diff
@@ -95,7 +96,7 @@ public DemoteBrokerRunnable(KafkaCruiseControl kafkaCruiseControl,
     super(kafkaCruiseControl, future, parameters.dryRun(), null,
           parameters.stopOngoingExecution(), null, true, null,
           parameters.allowCapacityEstimation(), parameters.excludeRecentlyDemotedBrokers(), false,
-          uuid, parameters::reason, true);
+          uuid, parameters::reason, true, false);
```
zornhsu, line 99: I noticed that we didn't include fastMode parameter in DemoteBrokerParameters. Is the reason being we seldom observed potential timeouts?
If we decide to expose this parameter, is setting demote fastMode default to true a good idea? The reason being the same as described in #1575. My only concern is that all other fastMode default would be false, so accepting my suggestion might lead to some confusion.
```suggestion
          uuid, parameters::reason, true, true);
```
BTW, I agree that all other fastMode defaults should be false.

zornhsu (approved):
LGTM. Thanks for granting us the option to not sit and wait, Efe!
I left some comments but none of them would block merging this PR.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/DemoteBrokerRunnable.java
efeg: Demote broker is a special case that uses a hardcoded goal (i.e. `PreferredLeaderElectionGoal`), which is expected to run fast. To verify my assertion, I ran a demote operation in a cluster with `270+` brokers. The optimization of `PreferredLeaderElectionGoal` in this cluster took `1402 ms`. Hence, given its special case, at least for the time being, I feel we can avoid exposing the parameter for this endpoint.

PR #1601, opened 2021-06-30
Add Java version compatibility to use

### LINKED ISSUE #1600: Add Java version compatibility to use
Cruise Control requires `Java 11+` compatibility when compiling Java sources. In such cases, builds of users with older Java versions will fail with an exception message that is not clear.

We should indicate the appropriate `sourceCompatibility` in `build.gradle` and add a note in the project wiki

efeg:
This PR resolves #1600.

**Notes.** It also drops
1. the deprecated `classifier` and adds  `archiveClassifier.set(xxx)` instead, and
2. `jcenter()` dependency, which shows the warning -- _Builds will no longer be able to resolve artifacts from JCenter after February 1st, 2022_


zornhsu (approved):
LGTM. Thanks for the pointing java 11 out, Efe!

PR #1604, opened 2021-07-01
Move broker failure detector away from using zNode-based failed broker persistence

### LINKED ISSUE #1603: Move broker failure detector away from using zNode-based failed broker persistence
Currently broker failure detector uses a zNode (default: `/CruiseControlBrokerList`) to let CC back up failed broker list in case it fails in the middle of the grace period that it waits before taking action on the failed brokers. So if CC dies, and comes back up, it reads this zNode to figure out "since when" the failed brokers have been dead.

Given that Zookeeper will be removed from Kafka as a dependency in the future ([KIP-500](https://cwiki.apache.org/confluence/display/KAFKA/KIP-500%3A+Replace+ZooKeeper+with+a+Self-Managed+Metadata+Quorum)), we should instead use an alternative store -- e.g. a file-based store.

efeg:
This PR resolves #1603.

In addition to unit tests, I verified the functionality on an actual test cluster.
* This PR also cleans up the deprecated `num.metric.fetchers` config.


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetectorTest.java
```diff
@@ -86,62 +95,81 @@ public void testFailureDetection() throws Exception {
...
-      detector = createBrokerFailureDetector(anomalies, mockTime);
+      detector = createBrokerFailureDetector(anomalies, mockTime, failedBrokersFile.getPath());
+      failedBrokerListString = detector.loadPersistedFailedBrokerList();
+      assertEquals(String.format("%d=%d", brokerId, anomalyTime), failedBrokerListString);
       // start the newly created detector and the broker down time should remain previous time.
       detector.startDetection();
-      assertEquals(Collections.singletonMap(brokerId, 100L), detector.failedBrokers());
+      assertEquals(Collections.singletonMap(brokerId, anomalyTime), detector.failedBrokers());
+      failedBrokerListString = detector.loadPersistedFailedBrokerList();
+      assertEquals(String.format("%d=%d", brokerId, anomalyTime), failedBrokerListString);
```
zornhsu, line 157: Can we add another test case for migrating CC to a different host, i.e., failedBrokersFile is not present in the new host?

zornhsu, line 157: Can we add another test case for migrating CC to a different host, i.e., failedBrokersFile is not present in the new host?

zornhsu (approved):
LGTM as long as we are sure migrating CC to a new host without carrying `_failedBrokersFile` does not do any fatal harm to internal CC cluster model. 
I added a comment below in the testing section.
Thanks for the patch, Efe!

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetectorTest.java
efeg: Migrating CC to a different host without carrying `_failedBrokersFile` would not do any fatal harm to internal CC cluster model. It would cause CC to not detect the initial failure time of broker(s), which failed in the middle of the grace period that CC waits before taking action on the failed brokers. As a result, during such a migration, CC might wait longer than intended as a side effect -- which should be a rare event.

PR #1609, opened 2021-07-01
Fix inconsistent use of disk unit in partition size anomalies

### LINKED ISSUE #1528: Inconsistent use of disk unit in partition size anomalies
1. `TopicPartitionSizeAnomaly#toString()` method says "bytes" rather than "MB"
2. Within `PartitionSizeAnomalyFinder#topicAnomalies()`, `partitionSize` (in MB) is compared with `_partitionSizeThreshold` (in bytes) without proper unit conversion, which leads to inability to detect partitions smaller than `500TB` using the default configs -- original config intends to detect partitions larger than 500MB.

efeg:
This PR resolves #1528.

Note: This PR sets the default self healing partition size threshold to `1Tib`.


zornhsu (approved):
LGTM. Thanks for making the unit consistent, Efe!

PR #1613, opened 2021-07-04
Parameterise the scala version in the gradle build

tomncooper:
This PR adds the scala version to the `gradle.properties` file and calculates the scala binary version so the correct kafka dependency can be requested.

Will this be ported to the migrate branches or should I open separate PRs for those?


efeg (approved):
Thanks for the PR -- I will take care of porting this to the other branches.

PR #1614, opened 2021-07-04
Upgrade to Scala 2.13

### LINKED ISSUE #1586: Upgrade to Scala 2.13
Currently CC uses Scala 2.12 (on its 2.5+ branch). With the passing of [KIP-751](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=181308218) which deprecates Scala 2.12 support in Kafka it seems prudent to update CC to a more modern (and supported) Scala version. 

Also, CC is often deployed alongside Kafka in the same infrastructure environment so it would be nice to have the CC and Kafka versions aligned.

tomncooper:
This PR resolves #1586 and upgrades to Scala 2.13.6.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 org.gradle.daemon=false
 org.gradle.parallel=false
 org.gradle.jvmargs=-Xms512m -Xmx512m
-scalaVersion=2.12.10
+scalaVersion=2.13.6
 kafkaVersion=2.7.0
 zookeeperVersion=3.5.8
 nettyVersion=4.1.63.Final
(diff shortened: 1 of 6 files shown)


efeg (changes requested):
@tomncooper Thanks for the PR!
Looks like there are some conflicts after merging recent PRs.- would you address them when you get a chance?

tomncooper:
@efeg I have fixed the conflicts. Could you have another pass at this?

gradle.properties
```diff
@@ -1,7 +1,7 @@
 org.gradle.daemon=false
 org.gradle.parallel=false
 org.gradle.jvmargs=-Xms512m -Xmx512m
-scalaVersion=2.12.10
+scalaVersion=2.13.6
```
efeg, line 4: Should we update the following version in `kafka-cruise-control-start.sh`?
```
if [ -z "$SCALA_VERSION" ]; then
  SCALA_VERSION=2.12.10
fi
```

efeg (commented):
@tomncooper Thanks for the updates -- left a comment.

gradle.properties
tomncooper: Yes we should! (`grep -r` FTW)!

efeg (approved):
@tomncooper Thanks for upgrading CC to a modern Scala version! LGTM.

PR #1622, opened 2021-07-20
Enable users to monitor the min.insync.replicas of all topics

### LINKED ISSUE #1621: Enable users to monitor the min.insync.replicas of all topics
`min.insync.replicas` is a config with the following documentation:

```
When a producer sets acks to "all" (or "-1"), min.insync.replicas specifies the minimum number of replicas that must acknowledge a write for the write to be considered successful. If this minimum cannot be met, then the producer will raise an exception (either NotEnoughReplicas or NotEnoughReplicasAfterAppend).
When used together, min.insync.replicas and acks allow you to enforce greater durability guarantees. A typical scenario would be to create a topic with a replication factor of 3, set min.insync.replicas to 2, and produce with acks of "all". This will ensure that the producer raises an exception if a majority of replicas do not receive a write.
```
While this config is critical in controlling write availability of topics, currently Kafka cluster operators lack a convenient way to retrieve the value of this config per topic.

Extend the response of `kafka_cluster_state` endpoint to include the value of `min.insync.replicas` is a config for each partition.

efeg:
This PR resolves #1621.

* Enable users to monitor the `min.insync.replicas` of all topics via `kafka_cluster_state` endpoint.
* Update the default value of `topic.config.provider.class` config to switch from `ZK-based client` to `AdminClient` by default (see [earlier PR](https://github.com/linkedin/cruise-control/pull/1569) that introduced admin client-based topic config provider).
* Make `KafkaAdminTopicConfigProvider` retrieve cluster configs via admin client rather than reading them from a file.
* Make private variables / methods of `KafkaTopicConfigProvider` and `KafkaAdminTopicConfigProvider` to let them be extended.

* This patch has been tested on a Kafka cluster containing topics that (1) override the cluster default and (2) use the cluster default. The response has been verified for correctness of the corresponding value for each topic with and without `json=true` parameter. Excerpt (i.e. showing only selected partition states) from the example plaintext response with `verbose=true`:

```
Summary: The cluster has XXX brokers, ...
...

All Partitions in the Cluster (verbose):
                                                         TOPIC PARTITION    LEADER                      REPLICAS                       IN-SYNC              OUT-OF-SYNC                  OFFLINE   MIN-ISR
Offline Partitions:
Partitions with Offline Replicas:
Under Replicated Partitions:
Under MinIsr Partitions:
Other Partitions:
                                ExampleTopicWithClusterDefault         0     10000           [10000, 2000, 1000]           [10000, 1000, 2000]                       []                       []         1
                                ExampleTopicWithClusterDefault         1     10000           [10000, 1000, 2000]           [10000, 1000, 2000]                       []                       []         1
                                 ...
                                  TopicThatOverridesTheDefault         0      1000           [1000, 2000, 30000]           [30000, 2000, 1000]                       []                       []         2
```


zornhsu (approved):
LGTM. Thanks for the deprecating more ZK clients, Efe! I left some comment below.

PR #1635, opened 2021-07-28
Add config to skip rack-awareness check while self-healing RF anomalies.

### LINKED ISSUE #1634: Add config to skip rack-awareness check while self-healing RF anomalies
Currently, when a replication factor violation is detected and the corresponding self-healing is triggered to fix that violation, the fix always tries to run a rack-awareness sanity check before changing the replication factor (RF) of topics with RF violation.

While this might be desirable in clusters with sufficient number of racks, in clusters with insufficient number of racks to place a single replica of each partition to a different rack, users may want to skip this sanity check.

To achieve that, we should add config to skip rack-awareness check while self-healing RF anomalies, and keep the default value of the config the same as the current value in use (i.e. `false`).

efeg:
This PR resolves #1634.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/TopicReplicationFactorAnomaly.java
```diff
@@ -16,9 +16,7 @@
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
 
-import static com.linkedin.kafka.cruisecontrol.config.constants.AnomalyDetectorConfig.ANOMALY_DETECTION_ALLOW_CAPACITY_ESTIMATION_CONFIG;
-import static com.linkedin.kafka.cruisecontrol.config.constants.AnomalyDetectorConfig.SELF_HEALING_EXCLUDE_RECENTLY_DEMOTED_BROKERS_CONFIG;
-import static com.linkedin.kafka.cruisecontrol.config.constants.AnomalyDetectorConfig.SELF_HEALING_EXCLUDE_RECENTLY_REMOVED_BROKERS_CONFIG;
+import static com.linkedin.kafka.cruisecontrol.config.constants.AnomalyDetectorConfig.*;
```
zornhsu, line 19: Good to avoid star import. 
(Well, my IDE also auto _correct_ me to star import really often.)

zornhsu (approved):
LGTM. Thanks for the feature to self heal RF reliably, Efe!

PR #1644, opened 2021-08-06
Move PLE / leadership movement requests from KafkaZkClient to AdminClient

### LINKED ISSUE #1643: Move PLE / leadership movement requests from KafkaZkClient to AdminClient
See #1415 for context on this transition.

efeg:
This PR resolves #1643.

* Adopt AdminClient-based leadership reassignment to replace KafkaZkClient
* Drop ZK-based calls to check existence of ongoing leadership reassignments (i.e. via `/admin/preferred_replica_election`) as they are no longer relevant under AdminClient-based leadership election
* Make force-stop requests function as the same as regular stop requests -- i.e. deletion of PLE znode and a following controller bounce is no longer relevant.
-- 
In addition to unit tests, I verified the functionality on a test cluster by performing leadership reassignment via demotion and rebalance.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -43,6 +43,7 @@ import org.apache.kafka.clients.Metadata;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AlterPartitionReassignmentsResult;
 import org.apache.kafka.clients.admin.DescribeConfigsResult;
+import org.apache.kafka.clients.admin.ElectLeadersResult;
 import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.TopicPartition;
@@ -92,7 +93,6 @@ public class Executor {
 
   private static final int NO_STOP_EXECUTION = 0;
   private static final int STOP_EXECUTION = 1;
-  private static final int FORCE_STOP_EXECUTION = 2;
 
   // Some state for external service to query
   private final AtomicInteger _stopSignal;
@@ -395,7 +395,7 @@ public class Executor {
             setRequestedMovementConcurrency(recommendedConcurrency, concurrencyType);
           } else {
             LOG.info("Stop the ongoing execution as per recommendation of Concurrency Adjuster.");
-            stopExecution(false);
+            stopExecution();
           }
         }
       }
@@ -813,7 +813,7 @@ public class Executor {
   }
 
   /**
-   * Sanity check whether there are ongoing (1) inter-broker or intra-broker replica movements or (2) leadership movements.
+   * Sanity check whether there are ongoing inter-broker or intra-broker replica movements.
    * This check ensures lack of ongoing movements started by external agents -- i.e. not started by this Executor.
    */
   private void sanityCheckOngoingMovement() throws OngoingExecutionException {
@@ -839,8 +839,6 @@ public class Executor {
       }
       if (hasOngoingIntraBrokerReplicaMovement) {
         throw new OngoingExecutionException("There are ongoing intra-broker partition movements.");
-      } else if (hasOngoingLeaderElection()) {
-        throw new OngoingExecutionException("There are ongoing leadership movements.");
       }
     }
   }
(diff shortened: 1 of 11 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1629,29 +1594,33 @@ private void handleProgressWithCompletion(ExecutionTask task, List<ExecutionTask
...
       do {
-        // If there is no finished tasks, we need to check if anything is blocked.
-        maybeReexecuteLeadershipTasks();
         Cluster cluster = getClusterForExecutionProgressCheck();
 
         List<ExecutionTask> slowTasksToReport = new ArrayList<>();
         boolean shouldReportSlowTasks = _time.milliseconds() - _lastSlowTaskReportingTimeMs > _slowTaskAlertingBackoffTimeMs;
         for (ExecutionTask task : inExecutionTasks()) {
           TopicPartition tp = task.proposal().topicPartition();
-          if (_stopSignal.get() == FORCE_STOP_EXECUTION) {
```
zornhsu, line 1647: This if statement was true only if the signal was `FORCE_STOP`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1629,29 +1594,33 @@ private void handleProgressWithCompletion(ExecutionTask task, List<ExecutionTask
...
-            // If the execution is force-stopped, the executor will mark all in progress tasks as dead, cleanup the
-            // relevant zNodes, and then bounce the controller (see ExecutionUtils#deleteZNodesToForceStopLeadershipMoves).
-            LOG.debug("Task {} is marked as dead to force-stop the execution with a controller bounce.", task);
+          if (_stopSignal.get() != NO_STOP_EXECUTION) {
```
zornhsu, line 1617: We made it evaluate to true for regular `STOP` here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1629,29 +1594,33 @@ private void handleProgressWithCompletion(ExecutionTask task, List<ExecutionTask
...
+            // If the execution is stopped, the executor will mark all in progress tasks as dead
+            LOG.debug("Task {} is marked as dead to stop the execution.", task);
             finishedTasks.add(task);
             stoppedTaskIds.add(task.executionId());
             _executionTaskManager.markTaskDead(task);
```
zornhsu, line 1622: So now even a regular `STOP` task will be marked as dead from now on, which is a different behavior. 
Just curious, would it 
  1. affect some other logic related to `_executionTaskManager` by surprise? 
  2. be better to remove this line?
  3. be better to `markTaskAborting()` instead?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: 1. We made it evaluate to `true` for regular `STOP`, because we want to break this loop faster to make CC available for a new execution in case the user intends to stop an execution during a leadership movement.
2. Alternatively, we can completely drop this branch, but dropping it has the drawback of the `waitForLeadershipTasksToFinish()` taking longer.

Note that both options are eventually going to stop the execution as soon as `waitForLeadershipTasksToFinish()` completes, so the decision to keep or remove this branch will not cause an eventual correctness issue. Contrary to replica reassignments, there is no way to gracefully cancel an ongoing leadership reassignment. A leadership reassignment is expected to be a quick operation as there is no data move involved in the process.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: (copied from the other comment regarding the same line)

1. We made it evaluate to `true` for regular `STOP`, because we want to break this loop faster to make CC available for a new execution in case the user intends to stop an execution during a leadership movement.
2. Alternatively, we can completely drop this branch, but dropping it has the drawback of the `waitForLeadershipTasksToFinish()` taking longer.

Note that both options are eventually going to stop the execution as soon as `waitForLeadershipTasksToFinish()` completes, so the decision to keep or remove this branch will not cause an eventual correctness issue. Contrary to replica reassignments, there is no way to gracefully cancel an ongoing leadership reassignment. A leadership reassignment is expected to be a quick operation as there is no data move involved in the process.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
efeg: 1. It may cause (1) the executor state to report some dead tasks and (2) relevant JMX sensor to report some dead leadership movement tasks. This has no other implication other than reporting purposes.
2. Please see the previous comment regarding pros/cons of removing it.
3. `ABORTING` is a transient state followed by `ABORTED` or `DEAD`. I don't have a strong opinion on this. We can mark it as `ABORTING` -> `ABORTED` or keep it as is. Let me know which one you'd prefer.

zornhsu: Got it. Let's keep the PR simple and leave it as is. Thanks for the explanations, Efe!

zornhsu (approved):
LGTM. Thanks for the patch, Efe!

PR #1651, opened 2021-08-11
Create ConfigProvider in Cruise-Control-Metrics-Reporter modul and optimise dependencies

egyedt:
This development enables parsing config values from environment variables. This is particularly useful if sensitive information can't be hardcoded in the config files but instead passed to Cruise Control via an environment variable.
Furthermore this change is optimise the dependencies by removing unused one and rearrange groups.


egyedt:
Won't merge.

PR #1652, opened 2021-08-11
Create ConfigProvider in Cruise-Control-Metrics-Reporter modul and optimise dependencies

### LINKED ISSUE #1649: Create ConfigProvider in Cruise-Control-Metrics-Reporter modul and optimise dependencies
Kafka provides ConfigProvider interface which is a provider of configuration data, which may optionally support subscriptions to configuration. With this feature the Cruise Control Metrics Reporter will be able to use environment variable references as configs.

There is an unused dependency in Cruise-Control-Metrics-Reporter modul (commons-math) which should be deleted. The junit package has to be moved to testCompile group because it is not used at compile time (only if tests are running).

egyedt:
This development enables parsing config values from environment variables. This is particularly useful if sensitive information can't be hardcoded in the config files but instead passed to Cruise Control via an environment variable.
Furthermore this change is optimise the dependencies by removing unused one and rearrange groups.

This PR resolves #1649.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -160,9 +160,11 @@ project(':cruise-control-core') {
     }
     compile "org.slf4j:slf4j-api:1.7.32"
     compile "org.apache.logging.log4j:log4j-slf4j-impl:2.14.1"
-    compile 'junit:junit:4.13.2'
     compile 'org.apache.commons:commons-math3:3.6.1'
     compile "org.eclipse.jetty:jetty-servlet:${jettyVersion}"
+
+    testCompile 'junit:junit:4.13.2'
+
     implementation group: 'com.google.code.findbugs', name: 'jsr305', version: '3.0.2'
     testOutput sourceSets.test.output
   }
@@ -245,7 +247,6 @@ project(':cruise-control') {
     compile "org.apache.kafka:kafka_$scalaBinaryVersion:$kafkaVersion"
     compile "org.apache.kafka:kafka-clients:$kafkaVersion"
     compile "org.scala-lang:scala-library:$scalaVersion"
-    compile 'junit:junit:4.13.2'
     compile 'org.apache.commons:commons-math3:3.6.1'
     compile 'org.apache.httpcomponents:httpclient:4.5.13'
     compile 'commons-codec:commons-codec:1.15'
@@ -260,6 +261,7 @@ project(':cruise-control') {
     testCompile project(path: ':cruise-control-metrics-reporter', configuration: 'testOutput')
     testCompile project(path: ':cruise-control-core', configuration: 'testOutput')
     testCompile "org.scala-lang:scala-library:$scalaVersion"
+    testCompile 'junit:junit:4.13.2'
     testCompile 'org.easymock:easymock:4.3'
     testCompile "org.apache.kafka:kafka_$scalaBinaryVersion:$kafkaVersion:test"
     testCompile "org.apache.kafka:kafka-clients:$kafkaVersion:test"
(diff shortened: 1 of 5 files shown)


cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/config/EnvConfigProvider.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.apache.kafka.common.config.ConfigData;
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.config.provider.ConfigProvider;
+
+/**
+ * Enables config properties of the form ${env:ENV_KEY}
+ */
+public class EnvConfigProvider implements ConfigProvider {
```
viktorsomogyi, line 18: Should we use this across the project (in `cruise-control` module) to avoid functionality duplication?
@efeg would you please comment on this? I've seen a couple of examples of using metrics reporter classes in the `cruise-control` module (mainly the `metrics` and `exception` packages) but I'm not sure if this module is intended to be used as a common utilility-like module.

efeg: Given the existing dependency, I agree that using this in `cruise-control` module is better than introducing functionality duplication.

That being said, the ideal solution, as you pointed out, is to have a utility-like module for such common Kafka-related logic, and use it from (1) `cruise-control` and (2) `cruise-control-metrics-reporter` modules.

I am ok with adopting the earlier approach and leaving the ideal solution to a future PR. Or if we would decide to go with the ideal solution and cleanup the dependency between `cruise-control` and `cruise-control-metrics-reporter` that is even better :)

efeg (commented):
Thanks for the PR!
Left some comments.

egyedt:
@efeg, @viktorsomogyi 
Thanks for the comments, I just introduced the new utility module and applied the testComplie group change for junit's every appearance. Could you please review it?

build.gradle
```diff
@@ -237,6 +239,7 @@ project(':cruise-control') {
     }
     compile project(':cruise-control-metrics-reporter')
     compile project(':cruise-control-core')
+    compile project(':cruise-control-util')
```
efeg, line 242: Now that we have `cruise-control-util`, should we extract the shared logic from `metrics-reporter` to `util` and drop `compile project(':cruise-control-metrics-reporter')`?

efeg (commented):
Thanks for the update!
Left a question / comment.

build.gradle
viktorsomogyi: We thought it through with @egyedt and we think it might not be possible to completely drop this dependency. The reason is most of the classes are used from the `cruise-control` module and some of them would look a bit weird in a util module (for instance `BrokerMetric` or `TopicMetric`). 
It might be a bit better to rename this util module to `cruise-control-metrics-reporter-core` (which might be better aligned with the classes inside it), move the common classes there and just depend on that in both `cruise-control` and `cruise-control-metrics-reporter`.
The drawback of this solution is that it wouldn't be too different than just having only `cruise-control-metrics-reporter`.

Also parts of `CruiseControlMetricsReporterConfig` are used in `cruise-control` (checking reporting interval against sampling interval) which is quite hard to refactor as it requires the reporting config's name. Is this check required on Cruise Control's side?

For these reasons I think for now we could just go with the original solution that is move `EnvConfigProvider` to `cruise-control-metrics-reporter` and consider the utils/core module in a later PR. What do you think?

efeg: Thanks for detailed explanation!

> For these reasons I think for now we could just go with the original solution that is move `EnvConfigProvider` to `cruise-control-metrics-reporter` and consider the utils/core module in a later PR. What do you think?

I agree with (1) using the original solution and (2) leaving the utils/core module to a later PR.

efeg (approved):
Thanks for the PR! LGTM.

PR #1669, opened 2021-08-25
Mention JMX in Sensors.md

EladLeev:
Hi,
I think that it worth mentioning in the Wiki that you can easily enable JMX in order to collect Cruise Control metrics.

Thanks, and keep up the good work! 🥇


efeg (approved):
Thanks for the PR! LGTM.

PR #1676, opened 2021-08-27
Fix EnvConfigProvider to work well if there is no pre configured env vars

### LINKED ISSUE #1675: EnvConfigProviders not works when no preconfigured configs set
There is two EnvConfigProvider in the Cruise Control. The `getEnv` method not works well when there is no `_preConfiguredEnvironmentVariables` since in this case the data field's value will be an empty map instead of null. The problematic code part is the following:
```java
  private Map<String, String> getenv() {
    if (_preConfiguredEnvironmentVariables == null) {
      return System.getenv();
    } else {
      return _preConfiguredEnvironmentVariables;
    }
```

egyedt:
This PR resolves #1675.


efeg (approved):
Thanks for the PR! LGTM.

PR #1683, opened 2021-09-04
Extend MinTopicLeadersPerBrokerGoal to compute min.topic.leaders.per broker dynamically for each topic

### LINKED ISSUE #1682: Extend MinTopicLeadersPerBrokerGoal to compute min.topic.leaders.per.broker dynamically for each topic
Currently MinTopicLeadersPerBrokerGoal accepts a list of topics (actually a regex) and a static `min.topic.leaders.per.broker`
It would be useful for optionally have the  min.topic.leaders.per.broker computed dynamically for each topic when this parameter is set to say `-1`
The formula can be as simple as `number-of-topic-leaders /  available-brokers`
This allows a better distribution of selected topics leaders across brokers when number of partitions differs for each topic.

amuraru:
When `min.topic.leaders.per.broker` is set to `-1` the value is computed
dynamically for each topic as `number-of-topic-leaders / available-brokers`

This PR resolves #1682


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
```diff
@@ -51,7 +52,7 @@
 public class MinTopicLeadersPerBrokerGoal extends AbstractGoal {
   private static final Logger LOG = LoggerFactory.getLogger(MinTopicLeadersPerBrokerGoal.class);
   private final String _replicaSortName = replicaSortName(this, true, false);
-  private Set<String> _mustHaveTopicLeadersPerBroker;
+  private Map<String, Integer> _mustHaveTopicMinLeadersPerBroker;
```
efeg, line 55: Is it possible to add some comments on when the value of this map is relevant -- i.e. only when `min.topic.leaders.per.broker` is non-positive -- and what it is used for?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
```diff
@@ -158,9 +160,16 @@ private boolean doesLeaderRemoveViolateOptimizedGoal(Replica replicaToBeRemoved)
...
+    Set<String> mustHaveTopicLeadersPerBroker = Collections.unmodifiableSet(
         Utils.getTopicNamesMatchedWithPattern(_balancingConstraint.topicsWithMinLeadersPerBrokerPattern(), clusterModel::topics));
-    if (_mustHaveTopicLeadersPerBroker.isEmpty()) {
+    Map<String, Integer> numLeadersByTopicNames = clusterModel.numLeadersPerTopic(mustHaveTopicLeadersPerBroker);
+    int eligibleBrokersForLeadership = eligibleBrokersForLeadership(clusterModel, optimizationOptions).size();
+    _mustHaveTopicMinLeadersPerBroker = new HashMap<>();
+    for (String topicName : mustHaveTopicLeadersPerBroker) {
+      int topicNumLeaders = numLeadersByTopicNames.get(topicName);
+      _mustHaveTopicMinLeadersPerBroker.put(topicName, eligibleBrokersForLeadership == 0 ? 0 : topicNumLeaders / eligibleBrokersForLeadership);
+    }
```
efeg, line 171: If `mustHaveTopicLeadersPerBroker` is empty, this method will still do a significant work before returning.
Similarly, calls to `clusterModel.numLeadersPerTopic(mustHaveTopicLeadersPerBroker)` and `eligibleBrokersForLeadership` are only relevant if the `min.topic.leaders.per.broker` is non-positive.

Should we avoid introducing extra work when possible?

efeg (commented):
Thanks for the PR -- left some initial comments.

amuraru:
thanks for your review @efeg - addressed the suggestions in the last commit

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/AnalyzerConfig.java
```diff
@@ -349,7 +349,9 @@
   public static final String MIN_TOPIC_LEADERS_PER_BROKER_CONFIG = "min.topic.leaders.per.broker";
   public static final int DEFAULT_MIN_TOPIC_LEADERS_PER_BROKER = 1;
   public static final String MIN_TOPIC_LEADERS_PER_BROKER_DOC = "The minimum required number of leaders per broker"
-      + " for topics that must have leader on brokers that are not excluded for replica move.";
+      + " for topics that must have leader on brokers that are not excluded for replica move."
+      + " When set to a value less than 1, the minimum number of leaders per broker is computed"
+      + " dynamically for each topic as no-of-topic-leaders / no-of-brokers";
```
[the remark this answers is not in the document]
amuraru: Good call, yes - updated

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/AnalyzerConfig.java
```diff
@@ -584,7 +586,6 @@ public static ConfigDef define(ConfigDef configDef) {
                     .define(MIN_TOPIC_LEADERS_PER_BROKER_CONFIG,
                             ConfigDef.Type.INT,
                             DEFAULT_MIN_TOPIC_LEADERS_PER_BROKER,
-                            atLeast(1),
```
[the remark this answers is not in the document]
amuraru: Makes sense. Updated

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
amuraru: Good call, added an inline comment

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
amuraru: Good call,  updated to compute the map conditionally

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
amuraru: To clarify, this `_mustHaveTopicMinLeadersPerBroker` is always used to retrieve the min leaders requested for each topic. When `min.topic.leaders.per.broker` is set to `0` the computed value is stored as key, when >0 the specified `min.topic.leaders.per.broker` value is stored for each topic in the map.
This has been done to keep the changes at minimum and replace the usage of `_mustHaveTopicLeadersPerBroker` which was a `Set` with the new `_mustHaveTopicMinLeadersPerBroker` `Map`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
```diff
@@ -158,24 +164,36 @@ private boolean doesLeaderRemoveViolateOptimizedGoal(Replica replicaToBeRemoved)
...
+    Set<String> mustHaveTopicLeadersPerBroker = Collections.unmodifiableSet(
         Utils.getTopicNamesMatchedWithPattern(_balancingConstraint.topicsWithMinLeadersPerBrokerPattern(), clusterModel::topics));
-    if (_mustHaveTopicLeadersPerBroker.isEmpty()) {
+    _mustHaveTopicMinLeadersPerBroker = new HashMap<>();
+    // populate min leaders per broker for each topic when dynamic computation is configured
+    if (mustHaveTopicLeadersPerBroker.isEmpty()) {
       return;
     }
+    Map<String, Integer> numLeadersByTopicNames = clusterModel.numLeadersPerTopic(mustHaveTopicLeadersPerBroker);
+    int eligibleBrokersForLeadership = eligibleBrokersForLeadership(clusterModel, optimizationOptions).size();
```
efeg, line 175: Nit: `#validateEnoughLeaderToDistribute(ClusterModel, OptimizationOptions)` function, which is only called from `#initGoalState`, seems to recompute the same values for
1. `numLeadersByTopicNames` and 
2. `eligibleBrokersForLeadership(clusterModel, optimizationOptions)`. 
 
Would it be possible to change the `#validateEnoughLeaderToDistribute(ClusterModel, OptimizationOptions)` to something like `#validateEnoughLeaderToDistribute(Map<String, Integer> numLeadersByTopicNames, Set<Broker> eligibleBrokersForLeadership)` and pass the parameters computed here?

efeg (commented):
Thanks for the updates!
Left just 2 nits, once they are addressed I will merge the PR.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
amuraru: Agree - this has been implemented

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
```diff
@@ -143,8 +148,9 @@ private boolean doesLeaderRemoveViolateOptimizedGoal(Replica replicaToBeRemoved)
       // Moving a follower replica does not violate/affect this goal
       return false;
     }
-    int topicLeaderCountOnSourceBroker = replicaToBeRemoved.broker().numLeadersFor(replicaToBeRemoved.topicPartition().topic());
-    return topicLeaderCountOnSourceBroker <= minTopicLeadersPerBroker();
+    final String topic = replicaToBeRemoved.topicPartition().topic();
+    int topicLeaderCountOnSourceBroker = replicaToBeRemoved.broker().numLeadersFor(topic);
+    return _mustHaveTopicMinLeadersPerBroker.containsKey(topic) && topicLeaderCountOnSourceBroker <= minTopicLeadersPerBroker(topic);
```
amuraru, line 153: this includes also a bug fix, previously the check 
```return topicLeaderCountOnSourceBroker <= minTopicLeadersPerBroker();``` 
didn't take into account that the goal may not be applicable to the topic

Comment (efeg) [reply cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java:153]: I am not sure if this check is needed. This function is used only after `#actionAffectsRelevantTopics(BalancingAction)` is called, which guarantees that the function `#doesLeaderRemoveViolateOptimizedGoal(Replica)` is called only for a replica from an applicable topic.

Maybe we can update the JavaDoc of `#doesLeaderRemoveViolateOptimizedGoal(Replica)` to make it more clear that this function is intended only for a replica from an applicable topic?

File: cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java [hunk 7]
```diff
@@ -166,24 +166,24 @@ protected void initGoalState(ClusterModel clusterModel, OptimizationOptions opti
       throws OptimizationFailureException {
     Set<String> mustHaveTopicLeadersPerBroker = Collections.unmodifiableSet(
         Utils.getTopicNamesMatchedWithPattern(_balancingConstraint.topicsWithMinLeadersPerBrokerPattern(), clusterModel::topics));
-    _mustHaveTopicMinLeadersPerBroker = new HashMap<>();
     // populate min leaders per broker for each topic when dynamic computation is configured
+    _mustHaveTopicMinLeadersPerBroker = new HashMap<>();
```
efeg, line 170: Nit: In corresponding comment line above, should we drop the "`when dynamic computation is configured`" part? -- i.e.`_mustHaveTopicMinLeadersPerBroker` is populated for each topic regardless of whether dynamic computation is configured or not.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java
amuraru: The issue I encountered (ie tests failing) is when a `INTER_BROKER_REPLICA_SWAP` is tried in https://github.com/linkedin/cruise-control/blob/fa13315ef00889f5ce119cdb2070ca5ec78d3027/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java#L110

`acceptReplicaSwap` calls `doesLeaderRemoveViolateOptimizedGoal` for both `src` and `dst` topics involved in the swap but only one (e.g src) may be defined as applicable for this goal and not the other one (e.g `dst` topic)
https://github.com/linkedin/cruise-control/blob/fa13315ef00889f5ce119cdb2070ca5ec78d3027/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/MinTopicLeadersPerBrokerGoal.java#L127

In that case `actionAffectsRelevantTopics` returns true correctly as the action involves a topic applicable (src)
 
that's the reason for double checking

amuraru: LE: It's worth mentioning that previously this bug was hidden as ```return topicLeaderCountOnSourceBroker <= minTopicLeadersPerBroker();``` didn't raise an exception but was incorrectly testing the leader count against min.leaders configuration for an incorrect topic

efeg: Ah got it, good catch! Thanks for the explanation.

Maybe as a final minor improvement, is it possible to skip the computation of `numLeadersFor` if this goal is not applicable to the `topic`?
```
    if (!_mustHaveTopicMinLeadersPerBroker.containsKey(topic)) {
      // Moving a replica from a non-applicable topic does not violate/affect this goal
      return false;
    }
    int topicLeaderCountOnSourceBroker = replicaToBeRemoved.broker().numLeadersFor(topic);
    return topicLeaderCountOnSourceBroker <= minTopicLeadersPerBroker(topic);
```

efeg:
@amuraru Thanks for all the updates -- LGTM.

PR #1684, opened 2021-09-08
Bump version for cruise-control-client changes

mgrubent:
#1671 made some changes to `cruise-control-client`, but I neglected to bump its version in the PR.

So, patch-version-bump here.


efeg (approved):
Thanks for the version bump Michael!

PR #1688, opened 2021-09-10
Upgrade build to gradle 7.2

### LINKED ISSUE #1687: Upgrade to gradle 7.x
Currently the project is using gradle 6 for build.
We should migrate to newer gradle 7.x to stay up to date.

amuraru:
Fixes #1687


build.gradle
```diff
@@ -237,38 +237,40 @@ project(':cruise-control') {
...
+    implementation "org.apache.kafka:kafka_$scalaBinaryVersion:$kafkaVersion"
+    implementation "org.apache.kafka:kafka-clients:$kafkaVersion"
+    implementation "org.scala-lang:scala-library:$scalaVersion"
+    implementation 'org.apache.commons:commons-math3:3.6.1'
+    implementation 'org.apache.httpcomponents:httpclient:4.5.13'
+    implementation 'commons-codec:commons-codec:1.15'
+    implementation 'com.google.code.gson:gson:2.8.8'
+    implementation "org.eclipse.jetty:jetty-server:${jettyVersion}"
+    implementation "org.eclipse.jetty:jetty-servlet:${jettyVersion}"
+    implementation "org.eclipse.jetty:jetty-security:${jettyVersion}"
```
efeg, line 256: It is interesting that we lost these dependencies after a Gradle version bump.

Do we actually need to add `implementation "org.eclipse.jetty:jetty-security:${jettyVersion}"` as `jetty-security` is already pulled by `jetty-servlet`?

build.gradle
```diff
@@ -97,7 +97,7 @@ subprojects {
   checkstyle {
     toolVersion = '8.45.1'
     ignoreFailures = false
-    configDir = file("$rootDir/checkstyle")
+    configDirectory = file("$rootDir/checkstyle")
```
efeg, line 100: Should we update the following?
```
//wrapper generation task
wrapper {
  gradleVersion = '6.9'
  distributionType = Wrapper.DistributionType.ALL
}
```

to 
```
//wrapper generation task
wrapper {
  gradleVersion = '7.2'
  distributionType = Wrapper.DistributionType.ALL
}
```

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-all.zip
```
efeg, line 3: Should we update the `minimumVersion` in `settings.gradle`? (some changes in `7.2` seem backwards incompatible)

```
def minimumVersion = GradleVersion.version("3.5")
if (gradleVer < minimumVersion) {
  throw new GradleScriptException("this build cannot be run with gradle < " + minimumVersion + ". current detection version is " +
      gradleVer + ". use newer gradle or (better yet) use the wrapper", null)
}
```

efeg (commented):

build.gradle
amuraru: missed that - thanks for spotting it

build.gradle
amuraru: > It is interesting that we lost these dependencies after a Gradle version bump.

Good point, debugged further and replaced `compile` with `api` instead of `implementation` to inherit depedencies from core and metrics sub-projects

gradle/wrapper/gradle-wrapper.properties
amuraru: Done

efeg (approved):

build.gradle
efeg: Thanks for making the migration using the the equivalent of the deprecated `compile` (i.e. `api`)! This is definitely cleaner for the migration purposes.

I read a little bit about the distinction between `implementation` and `api`. Maybe in the future we can revisit this part to explore the feasibility of using `implementation`, which claims to provide faster builds and prevent leaking the internal implementation.

amuraru: I agree - I did the same and while for the cruise-control-core and cruise-control-metrics-reporter it does makes sense to define them as `api` dependencies as they are more like libraries for the main project - it;s worth discussing if other should be defined using api or implementation.
There may be integration out there that use cruise-control as a library dep and in that case it makes sense to keep the transitive libraries

PR #1693, opened 2021-09-11
Remove initial size parameter for new HashMap()

### LINKED ISSUE #1692: We should not specify initial size when creating new HashMap
Most of our initial size does not consider load factor, causing confusion for users when reading the code.
We should use default as much as possible.

See https://stackoverflow.com/a/565255

zornhsu:
This PR resolves #1692.


efeg (approved):
Thanks for the PR!

PR #1698, opened 2021-09-20
Clarify the exception message in ReplicaCapacityGoal

### LINKED ISSUE #1697: Clarify the exception message in ReplicaCapacityGoal
When `ReplicaCapacityGoal` fails to bring the number of replicas at each broker under the configured maximum limit, it throws an exception as the following:
```
Replica count in broker XXX exceeds the maximum allowed number of replicas per broker: YYY
```

In the above exception message, `YYY` indicates the configured limit and `XXX` indicates the number of replicas on a broker after optimizing higher priority goals, but before being able to fully optimize `ReplicaCapacityGoal`. However, the value presented in `XXX` gives the impression that it is a broker id.

We should clarify the exception message to avoid confusion.

efeg:
This PR resolves #1697.


zornhsu (approved):
LGTM. Thanks for the patch, Efe!

PR #1703, opened 2021-09-23
Add ZK TLS support with properties file and modify broker failure detection

### LINKED ISSUE #1702: ZK TLS support with properties file
#1323 - related issue

Cruise Control is able to connect to Zookeeper with TLS when ZK configs are coming from "-D" system environment variables but cannot connect to ZK if the configs are coming from the "properties" file.

The BrokerFailureDetector class creates two different ZkClient in the current CC version (kafka and 101tec). 101tec client is only used to detect broker failures, but KafkaZkClient applied every other part of the source code. Unfortunately 101tec:zkclient not able to handle ZK TLS when configs are coming from the "properties" file, so this dependency has to be eliminated. This could happen with the refactor of the BrokerFailureDetector to use the already existing KafkaZkClient to be able to detect failures.

The tasks are the following:
- Modify broker failure detection to work with KafkaZkClient instead of 101tec:zkclient
- Remove 101tec:zkclient dependency from the project.

egyedt:
This PR resolves #1702.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
@@ -7,6 +7,7 @@ package com.linkedin.kafka.cruisecontrol.detector;
 import com.linkedin.cruisecontrol.detector.Anomaly;
 import com.linkedin.kafka.cruisecontrol.KafkaCruiseControl;
 import com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils;
+import com.linkedin.kafka.cruisecontrol.common.KafkaCruiseControlThreadFactory;
 import com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig;
 import com.linkedin.kafka.cruisecontrol.config.ZKConfigUtils;
 import com.linkedin.kafka.cruisecontrol.config.constants.AnomalyDetectorConfig;
@@ -69,7 +70,7 @@ public class BrokerFailureDetector extends AbstractAnomalyDetector {
     _failedBrokersFile = new File(config.getString(AnomalyDetectorConfig.FAILED_BROKERS_FILE_PATH_CONFIG));
     _fixableFailedBrokerCountThreshold = config.getShort(AnomalyDetectorConfig.FIXABLE_FAILED_BROKER_COUNT_THRESHOLD_CONFIG);
     _fixableFailedBrokerPercentageThreshold = config.getDouble(AnomalyDetectorConfig.FIXABLE_FAILED_BROKER_PERCENTAGE_THRESHOLD_CONFIG);
-    _detectionExecutor = Executors.newSingleThreadExecutor();
+    _detectionExecutor = Executors.newSingleThreadScheduledExecutor(new KafkaCruiseControlThreadFactory("BrokerFailureDetectorExecutor"));
   }
 
   /**
@@ -87,7 +88,7 @@ public class BrokerFailureDetector extends AbstractAnomalyDetector {
     // Detect broker failures.
     detectBrokerFailures(false);
     // Register ZNodeChildChangeHandler to ZK.
-    _kafkaZkClient.registerZNodeChildChangeHandler(new BrokerFailureHandler(BrokerIdsZNode.path()));
+    _kafkaZkClient.registerZNodeChildChangeHandler(new BrokerFailureHandler());
     _kafkaZkClient.getChildren(BrokerIdsZNode.path());
   }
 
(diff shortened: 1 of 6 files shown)


egyedt:
@efeg 
Could you please review it?

efeg:
@egyedt Thanks for the PR!
I will review it this week.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
```diff
@@ -55,35 +51,44 @@
...
-    ZkConnection zkConnection = new ZkConnection(zkUrl, ZK_SESSION_TIMEOUT);
     ZKClientConfig zkClientConfig = ZKConfigUtils.zkClientConfigFromKafkaConfig(config);
-    _zkClient = new ZkClient(zkConnection, ZK_CONNECTION_TIMEOUT, new ZkStringSerializer());
     _kafkaZkClient = KafkaCruiseControlUtils.createKafkaZkClient(zkUrl, ZK_BROKER_FAILURE_METRIC_GROUP, ZK_BROKER_FAILURE_METRIC_TYPE,
                                                                  zkSecurityEnabled, zkClientConfig);
     _failedBrokers = new HashMap<>();
     _failedBrokersFile = new File(config.getString(AnomalyDetectorConfig.FAILED_BROKERS_FILE_PATH_CONFIG));
     _fixableFailedBrokerCountThreshold = config.getShort(AnomalyDetectorConfig.FIXABLE_FAILED_BROKER_COUNT_THRESHOLD_CONFIG);
     _fixableFailedBrokerPercentageThreshold = config.getDouble(AnomalyDetectorConfig.FIXABLE_FAILED_BROKER_PERCENTAGE_THRESHOLD_CONFIG);
+    _detectionExecutor = Executors.newSingleThreadExecutor();
```
efeg, line 72: Can we pass the CC factory to use when creating the new thread?

```
_detectionExecutor = Executors.newSingleThreadScheduledExecutor(new KafkaCruiseControlThreadFactory("BrokerFailureDetectorExecutor"));
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
```diff
@@ -113,7 +118,15 @@ synchronized void detectBrokerFailures(boolean skipReportingIfNotUpdated) {
   }
 
   void shutdown() {
-    _zkClient.close();
+    _detectionExecutor.shutdown();
```
efeg, line 121: Should we call `_kafkaZkClient.unregisterZNodeChildChangeHandler` with `BrokerIdsZNode.path()` during shutdown?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
```diff
@@ -113,7 +118,15 @@ synchronized void detectBrokerFailures(boolean skipReportingIfNotUpdated) {
...
+    try {
+      _detectionExecutor.awaitTermination(2, TimeUnit.MINUTES);
```
efeg, line 123: Nit: `2 minutes` might be a little long for the timeout. Would it be possible to limit this to 30 seconds? If we are already shutting down CC, would it be ok to await for a shorter time to detect potential broker failures?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
```diff
@@ -113,7 +118,15 @@ synchronized void detectBrokerFailures(boolean skipReportingIfNotUpdated) {
...
+    } catch (InterruptedException e) {
+      LOG.warn("Unable to shut down BrokerFailureDetector narmally, the active detection may be terminated.", e);
```
efeg, line 125: Typo: 
1. `narmally` -> `normally`
2. `shut down` -> `shutdown`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
```diff
@@ -223,31 +236,27 @@ private void reportBrokerFailures() {
...
-  private class BrokerFailureListener implements IZkChildListener {
+  private final class BrokerFailureHandler implements ZNodeChildChangeHandler {
 
-    @Override
-    public void handleChildChange(String parentPath, List<String> currentChildren) {
-      // Ensure that broker failures are not reported if there are no updates in already known failed brokers.
-      // Anomaly Detector guarantees that a broker failure detection will not be lost. Skipping reporting if not updated
-      // ensures that the broker failure detector will not report superfluous broker failures due to flaky zNode.
-      detectBrokerFailures(currentChildren.stream().map(Integer::parseInt).collect(toSet()), true);
+    private String _path;
```
efeg, line 243: Nit: Can we make this `final`? Alternatively, maybe we can drop this variable and make `path()` always return `BrokerIdsZNode.path()`.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetectorTest.java
```diff
@@ -97,6 +97,13 @@ public void testFailureDetection() throws Exception {
       assertTrue(detector.failedBrokers().isEmpty());
       failedBrokerListString = detector.loadPersistedFailedBrokerList();
       assertTrue(failedBrokerListString.isEmpty());
+      // Kill the 2nd broker after 15 sec
+      Thread.sleep(15000);
```
efeg, line 101: This sleep seems redundant and costly -- can we drop this?

efeg (changes requested):

egyedt:
@efeg 
Thank you for your suggestions. I update the branch with a new commit where I fixed what you mentioned.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/BrokerFailureDetector.java
egyedt: In my opinion, what you suggest is more clearer than the current solution so I will add the unregister call to the source code.

efeg (approved):
Thanks for the updates - LGTM!

PR #1711, opened 2021-10-09
Add basic provisioner implementation for brokers.

### LINKED ISSUE #1710: Add basic provisioner implementation for brokers
Cruise Control has support for pluggable provisioners to add / remove resources to / from the cluster (see https://github.com/linkedin/cruise-control/pull/1494)

In addition to the automated provision recommendations created and submitted by the anomaly detector (i.e. if enabled to do so), a recent PR made it possible to submit provision requests using a new `rightsizing` endpoint (see https://github.com/linkedin/cruise-control/pull/1646).

The current default concrete implementation for the Provisioner interface, i.e. `PartitionProvisioner`, support only the addition of partition, but not brokers.

We should expand the support of the default provisioner to also support broker addition / removal to / from the cluster in addition to the existing partition increase support. The partition increase related logic should still remain platform-agnostic. 
Broker addition / removal related logic should be pluggable -- i.e. users should be able to override one method in their provisioner (e.g. method to override can be `BasicBrokerProvisioner#addOrRemoveBrokers(ProvisionRecommendation)`) and update a single config (e.g. the config could be called `broker.provisioner.class`) to use their platform-specific broker provisioner logic.

efeg:
This PR resolves #1710.


bachmanity1:
Hi @efeg @zornhsu 

I have a question regarding broker removal. In the `addOrRemoveBrokers` method when we want to remove a broker should we first run `RemoveBrokersRunnable` or is it already handled and we just need to tell the platform to remove the node on which broker is running?

PR #1720, opened 2021-10-25
Apply reconnect.backoff.ms configs to all contexts

### LINKED ISSUE #1694: IAM MSK support
Support for https://docs.aws.amazon.com/msk/latest/developerguide/iam-access-control.html

liko9:
Add line to use reconnect.backoff.ms when specified, without this line it is not being applied to the Admin Client which can prevent connections to a busy cluster

This PR resolves #1694 .


efeg (approved):
@liko9 Thanks for the contribution!

PR #1732, opened 2021-11-17
Add a ReplicaMovementStrategy that prioritizes OneAboveMinISR partitions with offline replicas

### LINKED ISSUE #1731: Add a ReplicaMovementStrategy that prioritizes OneAboveMinISR partitions with offline replicas
**Background**: During a recent issue, we had a broker which was detected with a hardware failure. We started removing the broker proactively, even though it serves traffic normally, and the cluster is healthy with no URP. 
Before the removal was completed, another broker dropped dead. Now we want to recover the URP first from the dead broker, while we remove both brokers.
Yet, this topic is set with replication-factor = 3, minISR = 1. With in-sync-replica = 2, this URP is OneAboveMinISR instead of (At/Under)minISR. Hence, CC does not prioritize fixing this URP with existing #1559.

zornhsu:
This PR resolves #1731.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/strategy/PrioritizeOneAboveMinIsrWithOfflineReplicasStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+
+
+/**
+ * The strategy, which tries to move replicas of OneAboveMinISR partitions with offline replicas.
+ * When multiple brokers are offline, OneAboveMinISR partitions are at a higher risk than partitions with no offline replicas.
+ * This strategy aims to help fixing such partitions faster.
+ * When using this strategy, always chain it after {@link PrioritizeMinIsrWithOfflineReplicasStrategy},
+ * which aims to help fixing higher risk (At/Under)MinISR.
```
efeg, line 21: Nit: `always chain` -> `it is recommended to chain`
Because if, for some reason, users want to use this strategy w/o chaining it, CC still supports it.

docs/wiki/User Guide/Configurations.md
```diff
@@ -137,7 +137,7 @@ The following configurations are inherited from the open source Kafka client con
...
+| replica.movement.strategies                                       | List    | N         | [com.linkedin.kafka.cruisecontrol.executor.strategy.PostponeUrpReplicaMovementStrategy, com.linkedin.kafka.cruisecontrol.executor.strategy.PrioritizeLargeReplicaMovementStrategy, com.linkedin.kafka.cruisecontrol.executor.strategy.PrioritizeSmallReplicaMovementStrategy, com.linkedin.kafka.cruisecontrol.executor.strategy.PrioritizeMinIsrWithOfflineReplicasStrategy, com.linkedin.kafka.cruisecontrol.executor.strategy.PrioritizeOneAboveMinIsrWithOfflineReplicasStrategy, com.linkedin.kafka.cruisecontrol.executor.strategy.BaseReplicaMovementStrategy]   | A list of supported strategies used to determine execution order for generated partition movement tasks.                                                                                                                                                               |
```
efeg, line 140: Nit: Can we update the spaces in the other rows to ensure that `|` is aligned for each column.

efeg (commented):
Thanks for the PR! Left a couple of nits.

efeg (approved):

PR #1733, opened 2021-11-17
Add disk failure test and fix detection

lkalina-cera:
In this pull request I introduced the disk failure test also with some configuration changes to speed up the tests a little bit. 
The DiskFailureIntegrationTest removes a  Kafka broker log folder and waits until the self healing fix the cluster.
I made a small change in the detector because under certain circumstances (from 10 test run 1 fails) the detector skips the detection. For example if the disk failure is detected but the self healing failed for this anomaly then the fix will not start again until there is no model change. That's why I modified the detector.

### DIFF
--- a/cruise-control/src/integrationTest/java/com/linkedin/kafka/cruisecontrol/DiskFailureIntegrationTest.java
+++ b/cruise-control/src/integrationTest/java/com/linkedin/kafka/cruisecontrol/DiskFailureIntegrationTest.java
@@ -1,0 +1,180 @@
+/*
+* Copyright 2020 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See
+ * License in the project root for license information.
+ */
+
+package com.linkedin.kafka.cruisecontrol;
+
+import java.io.File;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.StringJoiner;
+import com.jayway.jsonpath.Configuration;
+import com.jayway.jsonpath.JsonPath;
+import com.jayway.jsonpath.TypeRef;
+import com.linkedin.kafka.cruisecontrol.analyzer.goals.DiskCapacityGoal;
+import com.linkedin.kafka.cruisecontrol.analyzer.goals.ReplicaCapacityGoal;
+import com.linkedin.kafka.cruisecontrol.analyzer.goals.ReplicaDistributionGoal;
+import com.linkedin.kafka.cruisecontrol.config.constants.AnalyzerConfig;
+import com.linkedin.kafka.cruisecontrol.detector.TopicReplicationFactorAnomalyFinder;
+import com.linkedin.kafka.cruisecontrol.metricsreporter.utils.CCKafkaTestUtils;
+import com.linkedin.kafka.cruisecontrol.servlet.CruiseControlEndPoint;
+import kafka.server.KafkaConfig;
+import net.minidev.json.JSONArray;
+import org.apache.commons.io.FileUtils;
+import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.admin.AdminClientConfig;
+import org.apache.kafka.clients.admin.NewTopic;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.linkedin.kafka.cruisecontrol.common.TestConstants.TOPIC0;
+import static com.linkedin.kafka.cruisecontrol.servlet.CruiseControlEndPoint.KAFKA_CLUSTER_STATE;
+
+public class DiskFailureIntegrationTest extends CruiseControlIntegrationTestHarness {
+
+  private static final short TOPIC0_REPLICATION_FACTOR = (short) 2;
+  private static final int BROKER_ID_TO_CAUSE_DISK_FAILURE = 2;
+  private static final int PARTITION_COUNT = 3;
+  private static final int KAFKA_CLUSTER_SIZE = 3;
+  private static final String CRUISE_CONTROL_KAFKA_CLUSTER_STATE_ENDPOINT =
+      "kafkacruisecontrol/" + KAFKA_CLUSTER_STATE + "?verbose=true&json=true";
+  private static final String CRUISE_CONTROL_STATE_ENDPOINT =
+      "kafkacruisecontrol/" + CruiseControlEndPoint.STATE + "?substates=anomaly_detector&json=true";
+  private static final String CRUISE_CONTROL_ANALYZER_STATE_ENDPOINT =
+      "kafkacruisecontrol/" + CruiseControlEndPoint.STATE + "?substates=analyzer&json=true";
+  private final Configuration _gsonJsonConfig = KafkaCruiseControlIntegrationTestUtils.createJsonMappingConfig();
+  private static final Logger LOG = LoggerFactory.getLogger(DiskFailureIntegrationTest.class);
+  private List<Entry<File, File>> _brokerLogDirs = new ArrayList<>(KAFKA_CLUSTER_SIZE);
+  
+  @Before
+  public void setup() throws Exception {
+    super.start();
+  }
+
+  @Override
+  protected int clusterSize() {
+    return KAFKA_CLUSTER_SIZE;
+  }
+
+  @After
+  public void teardown() {
+    super.stop();
+  }
+
+  @Override
+  public Map<Object, Object> overridingProps() {
+    Map<Object, Object> props = KafkaCruiseControlIntegrationTestUtils.createBrokerProps();
+    Entry<File, File> logFolders = Map.entry(CCKafkaTestUtils.newTempDir(), CCKafkaTestUtils.newTempDir());
+    _brokerLogDirs.add(logFolders);
+    props.put(KafkaConfig.LogDirsProp(), logFolders.getKey().getAbsolutePath() + "," + logFolders.getValue().getAbsolutePath());
+    return props;
+  }
+
+  @Override
+  protected Map<String, Object> withConfigs() {
+    Map<String, Object> configs = KafkaCruiseControlIntegrationTestUtils.ccConfigOverrides();
+    configs.put(TopicReplicationFactorAnomalyFinder.SELF_HEALING_TARGET_TOPIC_REPLICATION_FACTOR_CONFIG, "2");
+    
+    configs.put(AnalyzerConfig.DEFAULT_GOALS_CONFIG, new StringJoiner(",")
+        .add(ReplicaDistributionGoal.class.getName())
+        .add(ReplicaCapacityGoal.class.getName())
+        .add(DiskCapacityGoal.class.getName()).toString());
+    
+    configs.put(AnalyzerConfig.HARD_GOALS_CONFIG, new StringJoiner(",")
+        .add(ReplicaCapacityGoal.class.getName())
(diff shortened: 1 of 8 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/DiskFailureDetector.java
```diff
@@ -60,15 +60,15 @@ public DiskFailureDetector(Queue<Anomaly> anomalies, KafkaCruiseControl kafkaCru
    * @return The {@link AnomalyDetectionStatus anomaly detection status}, indicating whether the anomaly detector is ready.
    */
   private AnomalyDetectionStatus getDiskFailureDetectionStatus() {
-    int currentClusterGeneration = _kafkaCruiseControl.loadMonitor().clusterModelGeneration().clusterGeneration();
-    if (currentClusterGeneration == _lastCheckedClusterGeneration) {
+    ModelGeneration currentClusterModelGeneration = _kafkaCruiseControl.loadMonitor().clusterModelGeneration();
+    if (currentClusterModelGeneration.equals(_lastCheckedModelGeneration)) {
```
egyedt, line 64: In my opinion, we should check here if the `currentClusterModelGeneration` variable is null since at the first run the `_lastCheckedModelGeneration` will be also null so there could be a non required state. What do you think about that @lkalina-cera?

Another option is to add the following line to the constructor:
`_lastCheckedModelGeneration = new ModelGeneration(clusterAndGeneration.generation(), -1L);`

egyedt (changes requested):
Thank you for your PR. It looks good to me overall but I have some questions and I wrote also some comments about code format.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/DiskFailureDetector.java
lkalina-cera: I would like to use `_lastCheckedModelGeneration = new ModelGeneration(0, -1L)`. What Do you think?

egyedt: That sounds good to me!

cruise-control/src/integrationTest/java/com/linkedin/kafka/cruisecontrol/DiskFailureIntegrationTest.java
```diff
@@ -0,0 +1,180 @@
...
+import static com.linkedin.kafka.cruisecontrol.servlet.CruiseControlEndPoint.KAFKA_CLUSTER_STATE;
+
+public class DiskFailureIntegrationTest extends CruiseControlIntegrationTestHarness {
+
+  private static final short TOPIC0_REPLICATION_FACTOR = (short) 2;
+  private static final int BROKER_ID_TO_CAUSE_DISK_FAILURE = 2;
+  private static final int PARTITION_COUNT = 3;
+  private static final int KAFKA_CLUSTER_SIZE = 3;
+  private static final String CRUISE_CONTROL_KAFKA_CLUSTER_STATE_ENDPOINT =
+      "kafkacruisecontrol/" + KAFKA_CLUSTER_STATE + "?verbose=true&json=true";
```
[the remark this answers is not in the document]
viktorsomogyi: "kafkacruisecontrol" prefix is configured by `WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG` so the most precise way would be to create a function that returns `_config.getString(WebServerConfig.WEBSERVER_API_URLPREFIX_CONFIG).replace('*', '');`
Would you please refactor it this way?

cruise-control/src/integrationTest/java/com/linkedin/kafka/cruisecontrol/DiskFailureIntegrationTest.java
```diff
@@ -0,0 +1,180 @@
...
+    }, 800, new AssertionError("There are still offline replica on broker"));
+  }
+
+  private void waitForDiskFailureDetection() {
+    KafkaCruiseControlIntegrationTestUtils.waitForConditionMeet(() -> {
+      String responseMessage = KafkaCruiseControlIntegrationTestUtils
+          .callCruiseControl(_app.serverUrl(), CRUISE_CONTROL_STATE_ENDPOINT);
+      JSONArray diskFailuresArray = JsonPath.read(responseMessage,
+            "$.AnomalyDetectorState.recentDiskFailures[*].anomalyId");
+      return diskFailuresArray.size() == 1;
```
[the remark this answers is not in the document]
lkalina-cera: I am going to fix this with filtering the FIX_STARTED disk failures only. I this test environment I didn't expect more than one disk failures fix to start.

cruise-control/src/integrationTest/java/com/linkedin/kafka/cruisecontrol/DiskFailureIntegrationTest.java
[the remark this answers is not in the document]
lkalina-cera: I think you want to use the WebServerConfig.DEFAULT_WEBSERVER_API_URLPREFIX in the tests. That means I can remove this data field from my class.

egyedt (approved):

viktorsomogyi (approved):

viktorsomogyi:

akatona84 (approved):
@efeg could u take a look pls?

cruise-control/src/integrationTest/java/com/linkedin/kafka/cruisecontrol/DiskFailureIntegrationTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+  private void waitForProposal() {
+    KafkaCruiseControlIntegrationTestUtils.waitForConditionMeet(() -> {
+      String responseMessage = KafkaCruiseControlIntegrationTestUtils
+          .callCruiseControl(_app.serverUrl(), CRUISE_CONTROL_ANALYZER_STATE_ENDPOINT);
+      return JsonPath.<Boolean>read(responseMessage, "AnalyzerState.isProposalReady");
+    }, Duration.ofSeconds(200), Duration.ofSeconds(15), new AssertionError("No proposal available"));
+  }
+
+  private void waitForMetadataPropogates() {
```
efeg, line 165: Typo in `waitForMetadataPropogates`

cruise-control/src/integrationTest/java/com/linkedin/kafka/cruisecontrol/DiskFailureIntegrationTest.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  private void waitForOfflineReplicasFixed() {
+    KafkaCruiseControlIntegrationTestUtils.waitForConditionMeet(() -> {
+      String responseMessage = KafkaCruiseControlIntegrationTestUtils
+          .callCruiseControl(_app.serverUrl(), CRUISE_CONTROL_KAFKA_CLUSTER_STATE_ENDPOINT);
+      Map<String, String> offlineReplicas = JsonPath.read(responseMessage,
+            "$.KafkaBrokerState.OfflineReplicaCountByBrokerId");
+      return offlineReplicas.isEmpty();
+    }, 800, new AssertionError("There are still offline replica on broker"));
```
efeg, line 134: (Applicable to the similar timeouts in integration tests) Why do we expect such large timeouts? -- i.e. what is taking so much time in these tests?

Due to integration tests taking too long, we have been observing test failures in our CI -- i.e. `Too long with no output (exceeded 10m0s): context deadline exceeded`. As a result, a recent PR had to bump up CI timeout as a temporary workaround (see https://github.com/linkedin/cruise-control/pull/1749).

Can we make sure that all aggregate timeouts (including retries) in integration tests is below `10 minutes`?

efeg (changes requested):
@lkalina-cera Thanks for the PR! Left some comments.
In particular, I think we should ensure that integration tests run fast to avoid CI timeouts.

cruise-control/src/integrationTest/java/com/linkedin/kafka/cruisecontrol/DiskFailureIntegrationTest.java
lkalina-cera: Hi @efeg,
An integration test execution time is about two and three minutes. With four tests the ITs run time between 8 and 12 mins. I can decrease the timeouts but the tests execution time can be over 10 mins easily. 
We can enable the tests logging to avoid no output issue during the ITs but I think this is not the best way to fix this. What is your opinion? Do you know any other solution?

egyedt: In my opinion, these tests could be useful for CC so I hope we will find a solution to merge these. :)

efeg: I think there are two actions to take:
1. We should figure out what is taking time in each integration test, and try to address that. For example, if the timeout is due to goal optimization times, we can configure the corresponding goal parameters to reduce that time.
2. Logging some success output after each integration test may provide a better view of what is going on during the overall integration testing process.

lkalina-cera: Thanks for your ideas. I am going to add minimal logging for ITs like the unit tests have. I reduced the test execution time as much I could.

viktorsomogyi: @efeg What gradle task are you using for testing? In a previous commit we created the `integrationTest` task to avoid running together unit tests and integration test. Perhaps we could hook this into a separate task or don't extend the `test` task even to avoid running them during a test phase. Would that work?

lkalina-cera: Hi @efeg 
Did you have time to check the changes?

efeg: @viktorsomogyi @lkalina-cera 
Sorry for the delayed response! It is a good idea to avoid running integration tests for each PR -- maybe we can run them as part of a separate task only when we publish a new release. What do you think?

efeg: @lkalina-cera @viktorsomogyi -- quick reminder on this.

viktorsomogyi: Thanks for the ping and sorry for the late reply :).
I think running it on a per release basis is good. I had a quick look on the circleci setup and I think the problem is that we're integrating into the `check` task but at least I can unlink it so it won't run when `build` is triggered.
The regarding the per release run: would it be enough if we defined an extra job in the circleci workflow (let's call it `release-test`) and integrated it into the `build-and-publish` workflow?

efeg: Thanks for the reply Viktor!
Yes, I think the proposed approach should work.

efeg: @viktorsomogyi Thanks for the commit -- I unsuccessfully tried making some naive changes to run integration-test only during publish. I suspect we might be able to figure out the right config, but didn't get a chance to dig into the documentation -- i.e.  https://circleci.com/docs/2.0/workflows/

efeg (commented):
Thanks for the updates -- LGTM!

PR #1737, opened 2021-11-18
Run cruise-control tests for s390x architecture

### LINKED ISSUE #1736: Run current cruise-control tests for Z (linux/s390x) architecture
My team is working now on [Strimzi](https://github.com/strimzi) enablement for Z(linux/s390x) architecture. Cruise-control project is used in strimzi and it's already practically verified that it works perfectly well for Z.
However we would like to extend current testing setup of cruise-control to run the tests on Z for each PR.
Following https://github.com/linkedin/cruise-control/issues/1486 the cruise-control team is positive to extend current CI/CD to run tests on non-Intel architectures. 
Our suggestion is to continue using Circle CI and just add new job to run the cruise-control tests inside the Docker container in hardware emulation mode in parallel with standard test/build job. This approach allows to run the tests on emulated Z hardware, using as host standard Circle Ci workers.

barthy1:
This PR resolves #1736.

The extension to the current Circle CI setup allows to run the cruise-control tests for non-Intel architectures. As first step
`linux/s390x` platform is used. New platforms can be added to the existing matrix, once there is interest for it. The tests for non-Intel architecture are executed in parallel with the standard build/test job, so no extra time is used for the whole test pipeline.

The idea is to run docker container on standard Intel host via standard Circle CI setup in hardware emulation mode via docker `--platform` parameter and do the tests inside the container, similar to how it can be executed on real ARM, Power or Z hardware. To make sure which architecture is emulated inside the container - `uname -a` command is executed before the actual test.

The proposed patch was tested with Circle CI - https://app.circleci.com/pipelines/github/barthy1/cruise-control/47/workflows/5f97d894-d3ae-4009-bd95-3b5861f05426

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,37 @@
-version: 2
+version: 2.1
 
 jobs:
+
+  test-multi-arch:
+    parameters:
+      platform:
+        type: string
+    environment:
+      _JAVA_OPTIONS: "-Xms512m -Xmx1g"
+      DOCKER_CLI_EXPERIMENTAL: "enabled"
+    working_directory: ~/workspace
+    machine:
+      image: ubuntu-2004:202107-02
+    steps:
+      - checkout
+      - run: |
+         # install required qemu libraries
+         docker run --rm --privileged tonistiigi/binfmt:latest --install all
+         # run docker container with qemu emulation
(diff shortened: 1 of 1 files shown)


efeg (approved):
Thanks for the patch -- LGTM.

PR #1715, opened 2021-10-14
Max cap on partition movement rate during executions

smtwilio:
This PR resolves https://github.com/linkedin/cruise-control/issues/1725

### Context:
At Twilio, we run hundreds of Kafka brokers spread across multiple clusters, environments, and regions and are an integral component to many of Twilio’s core products. 

This is an observation from the last production deployment. Right now, we artificially limit the disruption to clients on the fleet during deployment by doing, say, 20 brokers at a time, with the CC setting of # of concurrent partition movements per broker.

It might be better for us to extend CC so that we can set an overall cap of partition movements per reassignment so that we wouldn't need to do the 20 brokers at a time thing. A single CC operation (all brokers in the fleet) would be fine since the underlying step-by-step CC process would make sure that not too many OVERALL partitions were being moved at once.

### Changes
Changes are in 3 parts - 3 commits
1. Core change: Added the config `max.num.cluster.partition.movements` with default 1200 and logic to throttle the tasks being returned to the executor - 
`cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java`
2. Rest argument and dynamic update of parameter: This is 2 parts again. 
  a) Rest Parameters updated for the sync actions AddOrRemoveBrokers, Rebalance, FixOfflineReplicas, and TopicReplicationFactor change along with async action ChangeConcurrencyParameters
   b) Admin operation to update the value during runtime
3. Test cases updated 

#### Removing a broker
Find `max_partition_movements_in_cluster=8` in command below
```
/kafkacruisecontrol/remove_broker?brokerid=9006688&destination_broker_ids=8095530&concurrent_partition_movements_per_broker=30&max_partition_movements_in_cluster=8&concurrent_leader_movements=30&json=true&verbose=true&json=true&dryrun=true' 
```

#### Updating the value during runtime
Updating the value to 5 
```
$ curl -X POST '<SERVER>/kafkacruisecontrol/admin?max_partition_movements_in_cluster=5'
{
ongoingConcurrencyChangeRequest: Max Inter-broker partition movements is set to 5
}
```

PR #1746, opened 2021-11-30
Pick broker to submit task according to task priority and task set size

### LINKED ISSUE #1744: ExecutionTaskPlanner should pick tasks fairly from each broker
Currently `ExecutionTaskPlanner#getInterBrokerReplicaMovementTasks` always scans `readyBrokers` in the same sequence as the default hashMap sequence. This makes the broker at the tail of the sequence seldom gets the chance to submits its prioritized task `P`. As the result, the broker at the head of the sequence keeps getting chances to submit its tasks, even when the task priority is lower than `P`.

### LINKED ISSUE #1745: Each execution round should include at least one outstanding prioritized task
As mentioned in #1744, `ExecutionTaskPlanner#getInterBrokerReplicaMovementTasks` always scans readyBrokers in the same sequence as the default hashMap sequence. This makes the brokers at the head of the sequence drain their tasks faster.
Eventually the brokers at the head of the sequence finishes their prioritized tasks. Yet they keep getting chances to submit their low priority tasks in each execution round. 
As the result, outstanding prioritized tasks, which belongs to the brokers at the tail of the sequence, can't make any progress.

zornhsu:
This PR introduces a new broker `Comparator`.
Pick the broker with the prioritized task first resolves #1745. 
If tie, pick the broker with the larger task set first resolves #1744.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
@@ -117,12 +117,7 @@ public class ExecutionTaskPlanner {
         }
       }
 
-      if (!_defaultReplicaMovementTaskStrategy.name().contains(BaseReplicaMovementStrategy.class.getSimpleName())) {
-        // Unless the custom strategies are already chained with BaseReplicaMovementStrategy, chain it in the end to handle the scenario
-        // that provided custom strategy is unable to determine the order of two tasks.
-        // BaseReplicaMovementStrategy makes the task with smaller execution id to get executed first.
-        _defaultReplicaMovementTaskStrategy = _defaultReplicaMovementTaskStrategy.chain(new BaseReplicaMovementStrategy());
-      }
+      _defaultReplicaMovementTaskStrategy = _defaultReplicaMovementTaskStrategy.chainBaseReplicaMovementStrategyIfAbsent();
     }
   }
 
@@ -205,19 +200,9 @@ public class ExecutionTaskPlanner {
       }
     }
 
-    ReplicaMovementStrategy chosenReplicaMovementTaskStrategy;
-    if (replicaMovementStrategy == null) {
-      chosenReplicaMovementTaskStrategy = _defaultReplicaMovementTaskStrategy;
-    } else {
-      // Unless the custom strategies are already chained with BaseReplicaMovementStrategy,
-      // chain the generated composite strategy with BaseReplicaMovementStrategy in the end to ensure the returned
-      // strategy can always determine the order of two tasks.
-      if (!replicaMovementStrategy.name().contains(BaseReplicaMovementStrategy.class.getSimpleName())) {
-        chosenReplicaMovementTaskStrategy = replicaMovementStrategy.chain(new BaseReplicaMovementStrategy());
-      } else {
-        chosenReplicaMovementTaskStrategy = replicaMovementStrategy;
-      }
-    }
+    ReplicaMovementStrategy chosenReplicaMovementTaskStrategy = replicaMovementStrategy == null
+                                                                ? _defaultReplicaMovementTaskStrategy
+                                                                : replicaMovementStrategy.chainBaseReplicaMovementStrategyIfAbsent();
     _interPartMoveTasksByBrokerId = chosenReplicaMovementTaskStrategy.applyStrategy(_remainingInterBrokerReplicaMovements, strategyOptions);
     _interPartMoveBrokerIds = new TreeSet<>(brokerComparator(strategyOptions, chosenReplicaMovementTaskStrategy));
   }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
@@ -506,19 +506,16 @@ public class ExecutionTaskPlanner {
       int taskSet2Size = taskSet2.size();
 
       if (taskSet1Size == 0) {
+        // broker1 has no first task (broker1 has zero tasks). Let's check if broker2 has a first task by checking the task set size.
         return taskSet2Size == 0 ? PRIORITIZE_NONE : PRIORITIZE_BROKER_2;
-      } else {
-        if (taskSet2Size == 0) {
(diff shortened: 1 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
```diff
@@ -331,13 +338,16 @@ private void maybeAddLeaderChangeTasks(Collection<ExecutionProposal> proposals,
      * chances to make progress.
      */
     boolean newTaskAdded = true;
+    _interPartMoveBrokerId.addAll(_interPartMoveTaskByBrokerId.keySet());
```
zornhsu, line 341: Note: `readyBrokers` is replaced by `_interPartMoveTaskByBrokerId` to reduce iteration speed and avoid NPE in `brokerComparator`. 
`readyBroker` which is not in `_interPartMoveTaskByBrokerId` would never be involved in any execution round.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
```diff
@@ -446,10 +445,44 @@ private boolean isExecutableProposal(ExecutionProposal proposal, Map<Integer, In
...
     }
     _remainingInterBrokerReplicaMovements.remove(task);
   }
+
+  /**
+   * The comparing order:
+   * <ul>
+   *   <li>Priority of the the first task of each broker</li>
+   *   <li>The task set size of each broker, prioritize broker with the larger size</li>
+   *   <li>Broker ID integer</li>
```
efeg, line 460: Nit: Can we indicate if smaller or larger broker id is prioritized?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
```diff
@@ -331,55 +329,56 @@ private void maybeAddLeaderChangeTasks(Collection<ExecutionProposal> proposals,
      * chances to make progress.
      */
     boolean newTaskAdded = true;
+    _interPartMoveBrokerIds.addAll(_interPartMoveTasksByBrokerId.keySet());
     Set<Integer> brokerInvolved = new HashSet<>();
     Set<TopicPartition> partitionsInvolved = new HashSet<>();
     while (newTaskAdded) {
       newTaskAdded = false;
+      // The first task of each brokerInvolved must changed. Let's remove the brokers, then add them again by comparing their new first tasks.
```
efeg, line 337: Nit: `must changed` -> `might have changed`.
This is because, a broker B1 could be involved due to another broker B2's first task causing picking a non-first task from B1.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
```diff
@@ -418,8 +417,8 @@ private void maybeAddLeaderChangeTasks(Collection<ExecutionProposal> proposals,
    * Clear all the states.
    */
   public void clear() {
-    _intraPartMoveTaskByBrokerId.clear();
-    _interPartMoveTaskByBrokerId.clear();
+    _intraPartMoveTasksByBrokerId.clear();
+    _interPartMoveTasksByBrokerId.clear();
```
efeg, line 421: Should we also clear `_interPartMoveBrokerIds`?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -187,15 +187,15 @@ public void testGetInterBrokerPartitionMovementTasks() {
...
+    partitions.add(generatePartitionInfoWithUrpHavingOfflineReplica(_partitionMovement0, true));
+    partitions.add(generatePartitionInfo(_partitionMovement1, false));
+    partitions.add(generatePartitionInfoWithUrpHavingOfflineReplica(_partitionMovement2, true));
+    partitions.add(generatePartitionInfo(_partitionMovement3, false));
 
     Cluster expectedCluster = new Cluster(null, _expectedNodes, partitions, Collections.emptySet(), Collections.emptySet());
     // This ensures that the _partitionMovement1 and _partitionMovement3 are AtMinISR, while the other partitions are not.
     Map<String, MinIsrWithTime> minIsrWithTimeByTopic
-        = Collections.singletonMap(TOPIC2, new MinIsrWithTime((short) (_partitionMovement1.oldReplicas().size() - 1), 0));
+        = Collections.singletonMap(TOPIC2, new MinIsrWithTime((short) (_partitionMovement0.oldReplicas().size() - 1), 0));
```
efeg, line 198: Should we update the comment above this line?

efeg (commented):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
efeg: @zornhsu Would it be safer to call `.clear()` on `_interPartMoveBrokerIds` rather than setting the to `null`?
With the former option, a call to the public function `getInterBrokerReplicaMovementTasks(...)` would be safe, but the latter option can cause a NPE.

efeg (commented):
Thanks for the updates! Left a comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
zornhsu: Yeah, I was debating how we should do this. The difference was that `_interPartMoveBrokerIds` is initiated each time when we pass in a Strategy. 
Directly calling `.clear()` would fail `taskManager.clear()` in `ExecutionTaskManagerTest#testStateChangeSequences`.
In addition, A Strategy S2 which is passed in currently might not be the same as the previous Strategy S1. This was the main reason I ended up using null to force user to create a new `_interPartMoveBrokerIds ` with their new Strategy S2.

efeg (approved):
LGTM!

zornhsu:
Commits rebased from #1715.

PR #1752, opened 2021-12-09
Update ReplicationThrottleHelper to use AdminClient

mimaison:
This removes the dependency on ZooKeeper of this component.

After this, the last place Cruise Control requires ZooKeeper is `BrokerFailureDetector`. Unfortunately in KRaft mode, Kafka does not have a way to listen for broker changes. I can think of a couple of workarounds but has this been discussed already?


mimaison:
@efeg Did you get a chance to take a look? Thanks

efeg (approved):
@mimaison Thanks for the PR -- sorry for the delay in reviewing it!
LGTM.

mimaison:
Thanks @efeg !
Can you answer my question above regarding `BrokerFailureDetector`?

efeg:
@mimaison That is a good question and a valid point. We haven't had a discussion regarding the migration plan for `BrokerFailureDetector` off of ZooKeeper. I would be happy to discuss further if you have some proposals on it -- maybe we can discuss this under a relevant GitHub issue.

PR #1758, opened 2021-12-11
Fix request log to write to configured CruiseControlPublicAccessLogger

amuraru:
CruiseControlPublicAccessLogger was defined in [log4j properties](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_4/config/log4j.properties#L58-L60)
but not used in code.

Having jetty server logs routed through the same logging mechanism used for the other components is convenient as it can be controlled (path, retention policy, format, etc) in a single configuration file


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/WebServerConfig.java
```diff
@@ -114,21 +114,6 @@
...
-  public static final String DEFAULT_WEBSERVER_ACCESSLOG_PATH = "access.log";
-  public static final String WEBSERVER_ACCESSLOG_PATH_DOC = "HTTP Request log path";
-
-  /**
-   * <code>webserver.accesslog.retention.days</code>
-   */
-  public static final String WEBSERVER_ACCESSLOG_RETENTION_DAYS_CONFIG = "webserver.accesslog.retention.days";
-  public static final int DEFAULT_WEBSERVER_ACCESSLOG_RETENTION_DAYS = 7;
-  public static final String WEBSERVER_ACCESSLOG_RETENTION_DAYS_DOC = "HTTP Request log retention days";
-
```
efeg, line 131: We should ideally keep these configs and mark them as deprecated to be removed in a future release. Or we should ensure that the corresponding references to these configs should be removed from Configurations.md and cruisecontrol.properties files.

efeg (changes requested):
@amuraru Thanks for the PR -- left a comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/WebServerConfig.java
efeg: Hi @amuraru Are you still working on this PR?

amuraru: thanks for your reminder @efeg - let me review this in the upcoming days.

amuraru: @efeg Removed  corresponding references to these configs from Configurations.md and cruisecontrol.properties files.

Let me know if this looks good to you. I felt we shouldn't go through a deprecation,removal cycle for these

efeg (approved):
@amuraru Thanks for the PR -- LGTM.

PR #1765, opened 2021-12-16
Set operationId for swagger endpoints

viktorsomogyi:


viktorsomogyi:
@efeg would you please review that? I set the operationId in the swagger for these 3 endpoints as they aren't set yet compared to the rest.

efeg (approved):
@viktorsomogyi Thanks for fixing this -- LGTM.

PR #1769, opened 2022-01-03
Fix leaked interPartMoveBrokerIds

### LINKED ISSUE #1768: CC became unresponsive during rebalancing
After merging #1746, we noticed in some rebalancing, CC would take up to 5 minutes to response to inter-broker concurrency change request.

zornhsu:
This PR resolves #1768.

After profiling a running CC instance, we noticed that the majority of the time is spent on iterating through `SortedSet _interPartMoveBrokerIds`, which is introduced in #1746.
The reason for spending so much time is due to an element leak in this TreeSet.
We remove involved brokers in `_interPartMoveBrokerIds` after we change the broker comparison order (with changing first tasks on involved brokers potentially). This makes `remove()` can't find the element to remove by traversing the tree with the new comparison order. In some cases, TreeSet grows significantly with duplicate brokers, which increases overall iterating time.

I have eyeballed the test to make sure only add brokers with no tasks are skipped. Other add and remove operations always modify `_interPartMoveBrokerIds`. (https://github.com/zornhsu/cruise-control/commit/02e47459cdaf5e9f78959119692a413e9bdf4368)

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
@@ -68,6 +68,7 @@ public class ExecutionTaskPlanner {
   private static final Logger LOG = LoggerFactory.getLogger(ExecutionTaskPlanner.class);
   private Map<Integer, SortedSet<ExecutionTask>> _interPartMoveTasksByBrokerId;
   private SortedSet<Integer> _interPartMoveBrokerIds;
+  private List<Integer> _interPartMoveBrokerIdsList;
   private final Map<Integer, SortedSet<ExecutionTask>> _intraPartMoveTasksByBrokerId;
   private final Set<ExecutionTask> _remainingInterBrokerReplicaMovements;
   private final Set<ExecutionTask> _remainingIntraBrokerReplicaMovements;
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
```diff
@@ -382,11 +382,17 @@ private void maybeAddLeaderChangeTasks(Collection<ExecutionProposal> proposals,
               && !partitionsInvolved.contains(tp)) {
             partitionsInvolved.add(tp);
             executableReplicaMovements.add(task);
-            // Record the two brokers as involved in this round and stop involving them again in this round.
+            // Record the brokers as involved in this round and stop involving them again in this round.
             brokerInvolved.add(sourceBroker);
             brokerInvolved.addAll(destinationBrokers);
+            // The first task of each involved broker might have changed.
+            // Let's remove the brokers before the tasks change, then add them again later by comparing their new first tasks.
+            _interPartMoveBrokerIds.remove(sourceBroker);
```
gitlw, line 390: 1. Given _interPartMoveBrokerIds is a SortedSet<Integer>, I don't understand why we need to remove and add back the source and destinationBrokers. Is that not an no-op in the end?


2. It seems the _interPartMoveBrokerIds variable can be turned into a local variable inside this method from a field in this class. It's being set to an empty set inside the maybeAddInterBrokerReplicaMovementTasks method, but all we need to save in that method is the comparator.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
```diff
@@ -68,6 +68,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(ExecutionTaskPlanner.class);
   private Map<Integer, SortedSet<ExecutionTask>> _interPartMoveTasksByBrokerId;
   private SortedSet<Integer> _interPartMoveBrokerIds;
+  private List<Integer> _interPartMoveBrokerIdsList;
```
gitlw, line 71: This can be turned into a local variable inside the getInterBrokerReplicaMovementTasks method.

gitlw (approved):
LGTM, thanks!

PR #1779, opened 2022-01-31
Fix docs about throttling for adding and removing brokers

### LINKED ISSUE #1778: Incorrect default values in REST API documentation
Throttling when adding and removing brokers is `on` by default according to https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_4/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java but the documentation states that the default value is `false`.

chromy96:
This PR resolves #1778


efeg (approved):
@chromy96 Thanks for the fix! LGTM.

PR #1781, opened 2022-02-02
Update ReplicationThrottleHelper to use AdminClient

mimaison:
Second attempt at applying https://github.com/linkedin/cruise-control/pull/1752

I've added retry logic when checking broker throttle rates. With this change, I've not been able to replicate the test failures locally. Without this I only reproduce a couple of times over 100s of reruns!

I'm hoping we can rerun CI on this PR a few times to check whether it works before merging.


mimaison:
The failure yesterday was expected as there was still an issue with the retry logic. I forced pushed an update and CI passed now. @efeg Can you rerun CI a few times to see if it's better now?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -538,18 +541,26 @@ private void expectIncrementalBrokerConfigs(AdminClient adminClient, List<Intege
...
-    String expectedString = expectedRate == null ? null : String.valueOf(expectedRate);
-    assertNotNull(brokerConfig.get(cf));
-    if (expectedRate == null) {
-      assertNull(brokerConfig.get(cf).get(ReplicationThrottleHelper.LEADER_THROTTLED_RATE));
-      assertNull(brokerConfig.get(cf).get(ReplicationThrottleHelper.FOLLOWER_THROTTLED_RATE));
-    } else {
-      assertEquals(expectedString, brokerConfig.get(cf).get(ReplicationThrottleHelper.LEADER_THROTTLED_RATE).value());
-      assertEquals(expectedString, brokerConfig.get(cf).get(ReplicationThrottleHelper.FOLLOWER_THROTTLED_RATE).value());
-    }
+    waitUntilTrue(() -> {
```
efeg, line 546: We are adding this retry logic for the unit tests, which ensures that the eventual response will settle and provide the expected response. But IIUC, the `ReplicationThrottleHelper#getEntityConfigs` can still return an unexpected result, causing failure for the main code.

Am I missing something?

efeg (commented):
Thanks for the update -- left a question.
(unit tests pass with multiple CI runs)

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
mimaison: @efeg You're right, we ought do this in main instead of tests.

I initially thought it did not matter if we couldn't immediately read our writes but there are a number of scenarios where this is problematic. Also this keeps the existing semantics that were provided by ZooKeeper. So I agree with your feedback and I've added logic to ensure config changes are fully applied before we continue.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -36,13 +37,21 @@
   static final String FOLLOWER_THROTTLED_RATE = "follower.replication.throttled.rate";
   static final String LEADER_THROTTLED_REPLICAS = LogConfig.LeaderReplicationThrottledReplicasProp();
   static final String FOLLOWER_THROTTLED_REPLICAS = LogConfig.FollowerReplicationThrottledReplicasProp();
+  static final int RETRIES = 30;
+  static final long RETRY_INTERVAL_MS = 1000L;
```
efeg, line 41: Nit: Would it be possible to adopt retries with backoff? Please see `CruiseControlMetricsUtils#retry` for an example.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -139,180 +156,208 @@ private boolean throttlingEnabled() {
...
-  private void setLeaderThrottledReplicas(String topic, Set<String> replicas) {
-    setThrottledReplicas(topic, replicas, true);
+  private Config getBrokerConfigs(int brokerId) throws ExecutionException, InterruptedException {
+    return getEntityConfigs(new ConfigResource(ConfigResource.Type.BROKER, String.valueOf(brokerId)));
   }
 
-  private void setFollowerThrottledReplicas(String topic, Set<String> replicas) {
-    setThrottledReplicas(topic, replicas, false);
+  private Config getEntityConfigs(ConfigResource cf) throws ExecutionException, InterruptedException {
+    Map<ConfigResource, Config> configs = _adminClient.describeConfigs(Collections.singletonList(cf)).all().get();
```
efeg, line 196: Nit: Would it be possible to use `KafkaFuture#get(long timeout, TimeUnit unit)` rather than `KafkaFuture#get()`? This ensures that the request is blocked for at most the given time for this future to complete rather than waiting until the admin request times out.
Also applies to the other adminClient calls in this PR -- e.g. `incrementalAlterConfigs` and `listTopics`.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -139,180 +156,208 @@ private boolean throttlingEnabled() {
...
+    Map<ConfigResource, Collection<AlterConfigOp>> configs = Collections.singletonMap(cf, ops);
+    _adminClient.incrementalAlterConfigs(configs).all().get();
+    waitForConfigs(cf, ops);
+  }
+
+  boolean topicExists(String topic) {
+    try {
+      return _adminClient.listTopics().names().get().contains(topic);
+    } catch (InterruptedException | ExecutionException e) {
+      return false;
```
efeg, line 249: Would this function return `false` if the adminClient request times out? -- i.e. should we handle the following case separately inside the catch block?
```
if (Errors.REQUEST_TIMED_OUT.exception().getClass() == e.getCause().getClass()) { 
... 
}
````

**Review by efeg — COMMENTED** [review]
Thanks for the updates! Left some comments.

File: cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java (hunk 2 above)
Comment (mimaison) [reply cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java]: Since this module already depends on cruise-control-metrics-reporter (where `CruiseControlMetricsUtils#retry` is defined), I reused the method.

File: cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java [hunk 5]
```diff
@@ -139,180 +162,219 @@ private boolean throttlingEnabled() {
...
+    _adminClient.incrementalAlterConfigs(configs).all()
+        .get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+    waitForConfigs(cf, ops);
+  }
+
+  boolean topicExists(String topic) throws InterruptedException, TimeoutException {
+    try {
+      return _adminClient.listTopics().names().get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS).contains(topic);
+    } catch (ExecutionException ee) {
+      return false;
```
efeg, line 260: I am thinking about what the right response should be in case there is an `ExecutionException` while listing topics.
I wonder if it would be safer to log the error and throw the exception as we did for `InterruptedException` and `TimeoutException`?

efeg (commented):
Thanks for the updates!
Left just one more comment, o/w this PR looks good to me.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
mimaison: That's a good point, I've pushed an update

efeg (approved):
Thanks for all the updates and removing the ZK dependency from this component!
LGTM.

PR #1825, opened 2022-04-30
Remove execution tasks that have in-movement partitions from re-execution candidates

### LINKED ISSUE #1824: Improve repeated task reexecution for inter-broker replica reassignments
When partitions being reassigned (i.e. inter-broker replica reassignments) on Kafka are not a superset of `interBrokerReplicaTasksToReexecute`, CC reexecutes all the `interBrokerReplicaTasksToReexecute` regardless of whether there are some that are already being executed on Kafka-side:

https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_4/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L1866

This results in logs like the following on Controller side due to redundant executions:

```
[KafkaController] [Controller id=XXX] Leader YYY for partition topic-1 being reassigned, is already in the new list of replicas 1,2,3 and is alive
```

1. CC can improve this execution logic by only resubmitting tasks that are not in execution on Kafka-side, and
2. CC can improve the logging by logging the result of `ExecutionUtils.partitionsBeingReassigned(_adminClient)` in addition to the existing `LOG.info("Reexecuting tasks {}", interBrokerReplicaTasksToReexecute);`. So that we know what exactly is going on at the Controller-side.

CCisGG:
This PR resolves #1824 

Today, CruiseControl re-executes the inter-broker partition movement tasks in a batch if any task has partitions that's not being moved. The logic can cause unnecessary re-execution of partition movement.

 For example:

There is a list of under-execution tasks: [task1, task2, task3, ...]. Task1 is to move partition1, and task2 is to move partition2, etc.

If partition1 is not being moved, while all of the partition2, 3, ..., are already being moved in the kafka cluster, cruise-control will re-execute all tasks. 

This is not necessary, as ideally cruise-control should only re-execute task1. This PR is to address this issue and only re-execute the task that doesn't have in-movement partition.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
@@ -495,23 +495,28 @@ public final class ExecutionUtils {
   }
 
   /**
-   * Checks whether the topicPartitions of the execution tasks in the given subset is indeed a subset of the given set.
+   * Find the tasks that need to be re-executed from all candidate inter-broker replica movement tasks.
    *
-   * @param set The original set.
-   * @param subset The subset to validate whether it is indeed a subset of the given set.
-   * @return {@code true} if the topicPartitions of the given subset constitute a subset of the given set, {@code false} otherwise.
+   * @param partitionsInMovement The partitions that are being moved in kafka cluster
+   * @param candidateTasksToReexecute The candidate inter-broker replica movement tasks to re-execute.
+   * @return the set of inter-broker replica movement tasks to re-execute
    */
-  public static boolean isSubset(Set<TopicPartition> set, Collection<ExecutionTask> subset) {
-    boolean isSubset = true;
-    for (ExecutionTask executionTask : subset) {
+  public static List<ExecutionTask> getInterBrokerReplicaTasksToReexecute(Set<TopicPartition> partitionsInMovement,
+                                                                          Collection<ExecutionTask> candidateTasksToReexecute) {
+    List<ExecutionTask> tasksToReexecute = new ArrayList<>();
+
+    for (ExecutionTask executionTask : candidateTasksToReexecute) {
       TopicPartition tp = executionTask.proposal().topicPartition();
-      if (!set.contains(tp)) {
-        isSubset = false;
-        break;
+      if (!partitionsInMovement.contains(tp)) {
+        tasksToReexecute.add(executionTask);
       }
     }
 
-    return isSubset;
+    if (!tasksToReexecute.isEmpty()) {
+      LOG.info("Found tasks to re-execute: {} while detected in-movement partitions: {}.", tasksToReexecute, partitionsInMovement);
+    }
+
+    return tasksToReexecute;
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1859,18 +1859,19 @@ public class Executor {
      *             corresponding inter-broker replica reassignment tasks.
      */
     private void maybeReexecuteInterBrokerReplicaTasks(Set<TopicPartition> deleted, Set<TopicPartition> dead) {
-      List<ExecutionTask> interBrokerReplicaTasksToReexecute =
+      List<ExecutionTask> candidateInterBrokerReplicaTasksToReexecute =
           new ArrayList<>(_executionTaskManager.inExecutionTasks(Collections.singleton(INTER_BROKER_REPLICA_ACTION)));
-      boolean shouldReexecute = false;
+      List<ExecutionTask> tasksToReexecute;
       try {
-        shouldReexecute = !ExecutionUtils.isSubset(ExecutionUtils.partitionsBeingReassigned(_adminClient), interBrokerReplicaTasksToReexecute);
+        tasksToReexecute = ExecutionUtils.getInterBrokerReplicaTasksToReexecute(ExecutionUtils.partitionsBeingReassigned(_adminClient),
+            candidateInterBrokerReplicaTasksToReexecute);
       } catch (TimeoutException | InterruptedException | ExecutionException e) {
         // This may indicate transient (e.g. network) issues.
         LOG.warn("Failed to retrieve partitions being reassigned. Skipping reexecution check for inter-broker replica actions.", e);
+        tasksToReexecute = Collections.emptyList();
       }
-      if (shouldReexecute) {
-        LOG.info("Reexecuting tasks {}", interBrokerReplicaTasksToReexecute);
-        AlterPartitionReassignmentsResult result = ExecutionUtils.submitReplicaReassignmentTasks(_adminClient, interBrokerReplicaTasksToReexecute);
+      if (!tasksToReexecute.isEmpty()) {
+        AlterPartitionReassignmentsResult result = ExecutionUtils.submitReplicaReassignmentTasks(_adminClient, tasksToReexecute);
         // Process the partition reassignment result.
         Set<TopicPartition> noReassignmentToCancel = new HashSet<>();
         ExecutionUtils.processAlterPartitionReassignmentsResult(result, deleted, dead, noReassignmentToCancel);
--- a/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
+++ b/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
@@ -4,10 +4,12 @@
 
 package com.linkedin.kafka.cruisecontrol.executor;
 
+import com.linkedin.kafka.cruisecontrol.model.ReplicaPlacementInfo;
 import java.lang.reflect.Constructor;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -28,6 +30,7 @@ import org.easymock.EasyMock;
 import org.junit.Assert;
 import org.junit.Test;
 
+import static com.linkedin.kafka.cruisecontrol.executor.ExecutorTestUtils.*;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThrows;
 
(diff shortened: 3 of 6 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -495,23 +495,24 @@ public static AlterPartitionReassignmentsResult maybeStopPartitionReassignment(A
...
-      if (!set.contains(tp)) {
-        isSubset = false;
-        break;
+      if (!partitionsInMovement.contains(tp)) {
+        tasksToReexecute.add(executionTask);
       }
     }
 
-    return isSubset;
+    return tasksToReexecute;
```
efeg, line 515: Maybe we can move logging from `maybeReexecuteInterBrokerReplicaTasks` to here -- i.e. in case `tasksToReexecute` is not empty, we can log both (1) `tasksToReexecute` and (2) `partitionsInMovement` (which is the controller-side movements)?

efeg (commented):
@CCisGG Thanks for the PR!
Left some comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1859,18 +1859,22 @@ private boolean maybeMarkTaskAsDead(Cluster cluster,
...
+      List<ExecutionTask> candidateInterBrokerReplicaTasksToReexecute =
           new ArrayList<>(_executionTaskManager.inExecutionTasks(Collections.singleton(INTER_BROKER_REPLICA_ACTION)));
-      boolean shouldReexecute = false;
+      List<ExecutionTask> tasksToReexecute;
       try {
-        shouldReexecute = !ExecutionUtils.isSubset(ExecutionUtils.partitionsBeingReassigned(_adminClient), interBrokerReplicaTasksToReexecute);
+        tasksToReexecute = ExecutionUtils.getInterBrokerReplicaTasksToReexecute(
+            ExecutionUtils.partitionsBeingReassigned(_adminClient),
+            candidateInterBrokerReplicaTasksToReexecute
+        );
```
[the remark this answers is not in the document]
CCisGG: It doesn't enforce that specific format, but it does show red errors if the line is longer than the limit

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -495,23 +495,24 @@ public static AlterPartitionReassignmentsResult maybeStopPartitionReassignment(A
...
-   * @return {@code true} if the topicPartitions of the given subset constitute a subset of the given set, {@code false} otherwise.
+   * @param partitionsInMovement The partitions that are being moved in kafka cluster
+   * @param candidateTasksToReexecute The candidate inter-broker replica movement tasks to re-execute.
+   * @return the set of inter-broker replica movement tasks to re-execute
    */
-  public static boolean isSubset(Set<TopicPartition> set, Collection<ExecutionTask> subset) {
-    boolean isSubset = true;
-    for (ExecutionTask executionTask : subset) {
+  public static List<ExecutionTask> getInterBrokerReplicaTasksToReexecute(Set<TopicPartition> partitionsInMovement,
+                                                                                Collection<ExecutionTask> candidateTasksToReexecute) {
```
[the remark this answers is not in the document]
CCisGG: I thought the checkstyle file is automatically picked up by intellij, but seems not. I just enabled the checkstyle detection with the checkstyle file in the project. It should be working fine now.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
CCisGG: I added the logs in the latest commit. Just not sure if a WARN log makes more sense here, as it seems like not expected to happen if everything works fine?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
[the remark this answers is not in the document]
efeg: The indentation still seems broken -- let's sync up offline.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -495,23 +495,28 @@ public static AlterPartitionReassignmentsResult maybeStopPartitionReassignment(A
...
-        isSubset = false;
-        break;
+      if (!partitionsInMovement.contains(tp)) {
+        tasksToReexecute.add(executionTask);
       }
     }
 
-    return isSubset;
+    if (!tasksToReexecute.isEmpty()) {
+      LOG.info("Found tasks to re-execute: {} while detected in-movement replicas: {}.", tasksToReexecute, partitionsInMovement);
```
efeg, line 516: Nit: `replicas` -> `partitions`

efeg, line 516: Thanks for adding this log!
Can we drop `LOG.info("Re-executing tasks {}", tasksToReexecute);` from `maybeReexecuteInterBrokerReplicaTasks`?

Note: info-level log seems right, because this is an expected way of handling a Kafka-side edge-case on Cruise Control.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
```diff
@@ -127,6 +129,49 @@ public void testProcessAlterPartitionReassignmentResult() throws Exception {
     EasyMock.reset(result);
   }
 
+  @Test
+  public void testGetInterBrokerReplicaTasksToReexecute() {
+    Set<TopicPartition> partitionsInMovement = Set.of(P0, P1);
+
+    ExecutionProposal proposalForPartition0 = EasyMock.mock(ExecutionProposal.class);
+    ExecutionProposal proposalForPartition1 = EasyMock.mock(ExecutionProposal.class);
+    ExecutionProposal proposalForPartition2 = EasyMock.mock(ExecutionProposal.class);
```
efeg, line 138: Why do we need to mock this -- i.e. Can we create an actual `new ExecutionProposal` for each proposal rather than mocking it?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
```diff
@@ -127,6 +129,49 @@ public void testProcessAlterPartitionReassignmentResult() throws Exception {
...
+    EasyMock.expect(proposalForPartition0.topicPartition()).andReturn(P0).anyTimes();
+    EasyMock.expect(proposalForPartition1.topicPartition()).andReturn(P1).anyTimes();
+    EasyMock.expect(proposalForPartition2.topicPartition()).andReturn(P2).anyTimes();
+    EasyMock.replay(proposalForPartition0);
+    EasyMock.replay(proposalForPartition1);
+    EasyMock.replay(proposalForPartition2);
+
+    ExecutionTask executionTaskForPartition0 = EasyMock.mock(ExecutionTask.class);
+    ExecutionTask executionTaskForPartition1 = EasyMock.mock(ExecutionTask.class);
+    ExecutionTask executionTaskForPartition2 = EasyMock.mock(ExecutionTask.class);
```
efeg, line 149: Why do we need to mock this -- i.e. Can we create an actual object for each `ExecutionTask`?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
```diff
@@ -127,6 +129,49 @@ public void testProcessAlterPartitionReassignmentResult() throws Exception {
...
+    List<ExecutionTask> tasks2 = List.of(executionTaskForPartition0, executionTaskForPartition1, executionTaskForPartition2);
+    List<ExecutionTask> tasksToReexecute2 = ExecutionUtils.getInterBrokerReplicaTasksToReexecute(partitionsInMovement, tasks2);
+    Assert.assertEquals(1, tasksToReexecute2.size());
+    Assert.assertEquals(P2, tasksToReexecute2.get(0).proposal().topicPartition());
+
+    // Case 3: Partitions of candidate tasks is subset of partitions in movement. Expect nothing to re-execute
+    List<ExecutionTask> tasks3 = List.of(executionTaskForPartition0);
+    List<ExecutionTask> tasksToReexecute3 = ExecutionUtils.getInterBrokerReplicaTasksToReexecute(partitionsInMovement, tasks3);
+    Assert.assertTrue(tasksToReexecute3.isEmpty());
+  }
```
efeg, line 173: [Given that we may drop all mocks, this would irrelevant] When using mocks, it would be useful to `EasyMock.verify` them to make sure that all expected calls are made as many as the expected number of times.

efeg (commented):

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
CCisGG: To create an ExecutionProposal object,  there seems to be some other irrelevant/unnecessary parameters I need to provide, e.g. oldLeader, oldReplicas, newReplicas. Also I'm not sure what's a good value for those parameters.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
CCisGG: Same with above. I think those other parameters are not relevant, and not sure what values should I provide for those params.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtilsTest.java
CCisGG: Sometimes if the returning values pass the assertion, it should imply the function is called for a certain times. In those cases, do we still need those verify?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTestUtils.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
```
efeg, line 2: Nit: `2017` -> `2022`

efeg (commented):
Thanks for the updates! Left one more comment -- o/w lgtm.

PR #1831, opened 2022-05-16
Allow fraction CPU core capacity values in BrokerCapacityInfo

### LINKED ISSUE #1801: Feature Request: Allow fractional CPU core values in BrokerCapacityInfo
When running Kafka in container environments such as Kubernetes it's possible to configure fractional CPU capacities for the Kafka brokers. For example, Kubernetes allows users to specify CPU resources in "milliCPU" (suffixed "m"), where 1000m is equivalent to a single core CPU, so a user could specify 100m to get the equivalent of 0.1 of a single core CPU [1].  

It would be really cool if we allowed this CPU capacity precision in Cruise Control! It would be a really nice feature for Kubernetes users! From what I understand, it would be a non-intrusive change. It would mostly involve switching the of BrokerCapacityInfo's `_numCpuCores` [2] field type from a short to a double and then updating the related methods which interact with that field. The `_brokerCapacity` arrays [3] already store CPU capacity as double anyway!

Let me know what you think! If it sounds like a reasonable request, I would be happy to contribute this feature!

[1] https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#cpu-units
[2] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerCapacityInfo.java#L32
[3] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java#L75

kyguy:
This PR resolves #1801.


CCisGG (approved):
LGTM. It seems to be pretty reasonable and safe change. @efeg do you have other potential concerns?

efeg (approved):
@kyguy Thanks for the PR! LGTM.

PR #1839, opened 2022-05-25
Update the numCore type to double in BasicStats

### LINKED ISSUE #1838: Change numCores type from int to double in /load response.
Previously there was a PR to update the cpu cores to fractional: #1831 . 

The PR seems to forgot to update the type in /load response though. Now in the load response, CORE_NUM type is still an integer:

```
        HOST         BROKER         RACK         DISK_CAP(MB)            DISK(MB)/_(%)_            CORE_NUM         CPU(%)          NW_IN_CAP(KB/s)       LEADER_NW_IN(KB/s)     FOLLOWER_NW_IN(KB/s)         NW_OUT_CAP(KB/s)        NW_OUT(KB/s)       PNW_OUT(KB/s)    LEADERS/REPLICAS
 XXXhost.com,          1451,zone_4_of_10,        19070802.000,              0.000/00.00,                 32,         0.000,             1250000.000,                   0.000,                   0.000,             1250000.000,              0.000,              0.000,             0/0
```
 
To match the change in #1831 , we need to also update the CORE_NUM type in the /load response.

Ideally we still want to print without decimal point if the core_num is a whole number like 32, otherwise use a decimal point with a limit on number of digits like 32.235.

CCisGG:
This PR resolves #1838 .

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BrokerStats.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BrokerStats.java
@@ -192,11 +192,7 @@ public class BrokerStats extends AbstractCruiseControlResponse {
    * @param numCores the number of cores
    * @return StringFormatter String for the CORE_NUM
    */
-  private String coreNumStringFormatter(double numCores) {
-    if (numCores == (int) numCores) {
-      return "%19.0f";
-    } else {
-      return "%19.2f";
-    }
+  private static String coreNumStringFormatter(double numCores) {
+    return (numCores % 1) == 0 ? "%19.0f" : "%19.2f";
   }
 }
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BrokerStats.java
```diff
@@ -181,4 +186,17 @@ public String toString() {
...
     return sb.toString();
   }
+
+  /**
+   * Format the double type CORE_NUM to int if it is not a fractional value, otherwise format it to double with 2 digits after decimal point
+   * @param numCores the number of cores
+   * @return StringFormatter String for the CORE_NUM
+   */
+  private String coreNumStringFormatter(double numCores) {
+    if (numCores == (int) numCores) {
```
efeg, line 196: This works for values in the int range, which is the expected case for numCores -- but to make it even more generic regardless of the input size, would it make sense to use the following?
```
(numCores % 1) == 0
```

something like:
```
  private static String coreNumStringFormatter(double numCores) {
    return (numCores % 1) == 0 ? "%19.0f" : "%19.2f";
  }
```

CCisGG: Yeah I implement it that way exactly because coreNum should fit in int value, but I'm totally ok to change it to be more generic

efeg (approved):

PR #1848, opened 2022-06-15
Enable users to get remote-storage-enabled of topics in kafka_cluster_state

### LINKED ISSUE #1846: In kafka_cluster_state API response (verbose mode), indicate which partitions are tiered remotely
In the kafka_cluster_state REST API response (https://github.com/linkedin/cruise-control/wiki/REST-APIs#query-partition-and-replica-state) in verbose mode, cruise-control displays a list of partitions and its replicas. This view is often used by operators for gleaning information about a cluster. It will be helpful to know if a partition is tiered into a remote tier, by adding a new "Tiered" column that shows "yes" or "no" depending on whether `remote.storage.enable` topic configuration is set to `true`.

hshi2022:
This PR resolves #1846.

This PR enables users to view remote-storage-enabled (value: true or false) of topics via kafka_cluster_state endpoint. Only topic level config remote.storage.enable is considered. remote.log.storage.system.enable is not addressed in this PR.


CCisGG (commented):
Is that possible to add a test for this? If not, did you try locally to see if the output is expected?

CCisGG:
Also your commit message looks weird. You can squash them and push again if you want (not required, as we'll squash them anyways when merging the PR)

hshi2022:
> Is that possible to add a test for this? If not, did you try locally to see if the output is expected?

There is no existing test cover kafka_cluster_state in this repo. Will add tests in LiKafkaClusterStateTest in likafka-cruise-control.

hshi2022:
> Also your commit message looks weird. You can squash them and push again if you want (not required, as we'll squash them anyways when merging the PR)

Since we are going to squash commits during merge, I did not squash them here. I normally like to push commits frequently as a way to back up local changes, so the commit message is meaningless for now. PR description is well-written. I will update with a meaningful commit message when merging to origin.

CCisGG (approved):
LGTM

CCisGG (changes requested):
Actually I'm wondering if we can show this column only when remote.log.storage.system.enable is set to true. @efeg do you know whether CC can get broker/cluster config?

CCisGG:
I believe you can get the broker config from zookeeper:

`Properties brokerConfigs = _kafkaZkClient.getEntityConfigs(ConfigType.Broker(), String.valueOf(brokerId));`

efeg:
@hshi2022 @hshi2022  What is the behavior if topic-level config is enabled and broker-level config is disabled for some brokers and enabled for some other brokers on which the replicas of the topic reside?

hshi2022:
> @hshi2022 @hshi2022 What is the behavior if topic-level config is enabled and broker-level config is disabled for some brokers and enabled for some other brokers on which the replicas of the topic reside?

Based on @wyuka, remote storage is enabled only if both topic-level and broker-level configs are true. Think more about this, so we need to show remote-storage-enabled per topic per broken.

hshi2022:
> Actually I'm wondering if we can show this column only when remote.log.storage.system.enable is set to true. @efeg do you know whether CC can get broker/cluster config?

Since remote.log.storage.system.enable is a broker level config, we do not consider it in kafka_cluster_state api which is showing configs for partitions.

PR #1864, opened 2022-07-15
Make BrokerSetAwareGoal configurable to accept partition coloring

### LINKED ISSUE #1862: Make BrokerSetAwareGoal capable of doing partition coloring
Recently AWS has make contributions to CC and create this BrokerSetAwareGoal in: https://github.com/linkedin/cruise-control/pull/1809. This goal will ensure each topic (all its partitions and replicas) only lives within a single broker set.

Today we have a need to do partition level coloring, meaning a partition and all its replicas live within a broker set. Different partitions of a topic doesn't need to be in the same color.  With the current BrokerSetAwareGoal, it is not allowed. I'm thinking about to add partition coloring by either update the BrokerSetAwareGoal, or make BrokerSetAwareGoal easily extendable to achieve it.

Basically there are 2 things we need to change:
1. For "ensureBrokerSetAware" method, we need to make it accept partition level coloring.
2. We need to remove the _mustHaveTopicLeadersPerBroker affect. In topic coloring, the rebalance should ignore those topics; while in partition coloring, the rebalance should not ignore those topics.

CCisGG:
This PR resolves #1862 .

As in the issue described, there are 2 solutions:
1. Make this goal easily extendable, and users can create a new goal reusing the code of this goal
2. Make this goal configurable.

@efeg originally suggests that we'd better have a new goal, but after consideration, I think make it configurable is a cleaner solution. With this new configuration "allow.partition.level.broker.set.aware", only minimal code changes are needed.

@efeg Please feel free to comment on any benefits of creating a new goal if you think that's better.


mohitpali (commented):
Did an initial quick pass. I will give it another go again.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
```diff
@@ -178,28 +182,38 @@ protected void updateGoalState(ClusterModel clusterModel, OptimizationOptions op
...
         List<Partition> partitions = partitionsByTopic.getValue();
-        Set<Broker> allBrokersForTopic = partitions.stream()
-                                                   .map(partition -> partition.partitionBrokers())
-                                                   .flatMap(brokers -> brokers.stream())
-                                                   .collect(Collectors.toSet());
-        // Checks if a topic's brokers do not all live in a single brokerSet
-        if (_brokersByBrokerSet.values().stream().noneMatch(brokerSetBrokers -> brokerSetBrokers.containsAll(allBrokersForTopic))) {
-          throw new OptimizationFailureException(
-              String.format("[%s] Topic %s is not brokerSet-aware. brokers (%s).", name(), topicName, allBrokersForTopic));
+        if (_isPartitionLevelBrokerSetAware) {
```
mohitpali, line 199: Instead of doing this i would recommend introducing inheritance. 

1. Introduce a configuration - `broker.set.awareness.level=partition`
We can have only two choices - `partition` and `topic`
2. The Level can be picked from an ENUM so that no other level is allowed.
3. The default value of this configuration can be set to level `topic`.
4. Introduce an interface called `BrokerSetAwarenessValidator`
5. Implement two implementations `TopicLevelBrokerSetAwarenessValidator` and `PartitionLevelBrokerSetAwarenessValidator`
6. Based on configuration set, the instance of the BrokerSetAwarenessValidator will be chosen.
7. From here just call `brokerSetAwarenessValidator.validate(partitions)`

mohitpali: You could also just introduce the validate() method in th ENUMs (with values PARTITION and TOPIC )

```
    private enum BrokerSetAwarenessLevel {
        TOPIC {
            public void validate(Partitions p) {
                // do something
            }
        },
        PARTITION {
            public void validate(Partitions p) {
                // do something different
            }
        }
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
```diff
@@ -123,6 +125,8 @@ public boolean isHardGoal() {
    */
   @Override
   protected void initGoalState(ClusterModel clusterModel, OptimizationOptions optimizationOptions) throws OptimizationFailureException {
+    _isPartitionLevelBrokerSetAware = _balancingConstraint.allowPartitionColoring();
```
mohitpali, line 128: I would not use word `coloring` since it may be confusing. We have been referring to it as `brokerserAwareness`

PR #1875, opened 2022-08-03
Fix the BrokerFailureDetector bug

### LINKED ISSUE #1876: ZkBrokerFailureDetector doesn't work on migrate_to_kafka_2_5 branch
The broker failure detector doesn't work properly. 

The root cause is:
https://github.com/linkedin/cruise-control/blob/ee3d0e1c368c7c145c4ef1950924666b2c0e10a7/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/ZKBrokerFailureDetector.java#L93

This scala-to-java converter is throwing exception. 

By adding custom logs, the exception is like this:

```
2022/08/03 16:13:31.690 ERROR [AbstractBrokerFailureDetector] [AnomalyDetector-6] [kafka-cruise-control] [] Broker failure detector received exception:
java.lang.NoClassDefFoundError: scala/jdk/javaapi/CollectionConverters
        at com.linkedin.kafka.cruisecontrol.detector.ZKBrokerFailureDetector.aliveBrokers(ZKBrokerFailureDetector.java:102) ~[cruise-control-2.5.98-SNAPSHOT.jar:?]
        at com.linkedin.kafka.cruisecontrol.detector.AbstractBrokerFailureDetector.detectBrokerFailures(AbstractBrokerFailureDetector.java:63) [cruise-control-2.5.98-SNAPSHOT.jar:?]
        at com.linkedin.kafka.cruisecontrol.detector.ZKBrokerFailureDetector.run(ZKBrokerFailureDetector.java:71) [cruise-control-2.5.98-SNAPSHOT.jar:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) [?:?]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) [?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
        at java.lang.Thread.run(Thread.java:829) [?:?]
Caused by: java.lang.ClassNotFoundException: scala.jdk.javaapi.CollectionConverters
        at jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581) ~[?:?]
        at jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178) ~[?:?]
        at java.lang.ClassLoader.loadClass(ClassLoader.java:522) ~[?:?]
        at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:565) ~[?:?]
        at java.lang.ClassLoader.loadClass(ClassLoader.java:522) ~[?:?]
        ... 9 more
```

It seems like the project is using the scala 13 compiler, but the actual scala bytecode is still scala 12.

CCisGG:
This PR resolves #1876 .

**Description:**

In a previous PR, someone bumped the scala version from 2.12 to 2.13, and updated a bunch of APIs. For open source code, it doesn't have any issues. However, LinkedIn today still only supports scala 2.12. The API CollectionConverter that is supported by scala 2.13, is throwing exception in LinkedIn internal runtime.

This PR revert the API back to JavaConverter as a workaround. JavaConverter is an API in scala 2.12 and deprecated in scala 2.13 (still working though). With this change, it should fix the issue that LinkedIn is facing internally. For open source users, this PR shouldn't lead to any functional changes.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/ZKBrokerFailureDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/ZKBrokerFailureDetector.java
@@ -21,7 +21,7 @@ import kafka.zk.BrokerIdsZNode;
 import kafka.zk.KafkaZkClient;
 import org.apache.zookeeper.client.ZKClientConfig;
 import kafka.zookeeper.ZNodeChildChangeHandler;
-import scala.jdk.javaapi.CollectionConverters;
+import scala.collection.JavaConverters;
 
 import static java.util.stream.Collectors.toSet;
 
@@ -90,7 +90,7 @@ public class ZKBrokerFailureDetector extends AbstractBrokerFailureDetector {
   @Override
   Set<Integer> aliveBrokers() {
     // We get the alive brokers from ZK directly.
-    return CollectionConverters.asJava(_kafkaZkClient.getAllBrokersInCluster()).stream().map(Broker::id).collect(toSet());
+    return JavaConverters.asJavaCollection(_kafkaZkClient.getAllBrokersInCluster()).stream().map(Broker::id).collect(toSet());
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/ZKBrokerFailureDetector.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/ZKBrokerFailureDetector.java
@@ -90,6 +90,8 @@ public class ZKBrokerFailureDetector extends AbstractBrokerFailureDetector {
   @Override
   Set<Integer> aliveBrokers() {
     // We get the alive brokers from ZK directly.
+    // The JavaConverters API is deprecated in scala 2.13.0, and the new API is CollectionConverter. However, LinkedIn is still using
+    //  scala 2.12, which will cause it fail if it uses CollectionConverter.
     return JavaConverters.asJavaCollection(_kafkaZkClient.getAllBrokersInCluster()).stream().map(Broker::id).collect(toSet());
   }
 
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/ZKBrokerFailureDetector.java
```diff
@@ -90,7 +90,7 @@ public void shutdown() {
   @Override
   Set<Integer> aliveBrokers() {
     // We get the alive brokers from ZK directly.
-    return CollectionConverters.asJava(_kafkaZkClient.getAllBrokersInCluster()).stream().map(Broker::id).collect(toSet());
+    return JavaConverters.asJavaCollection(_kafkaZkClient.getAllBrokersInCluster()).stream().map(Broker::id).collect(toSet());
```
lmr3796, line 93: Should add a comment that it is deprecated in scala 2.13 but we change this as it's buggy?

To prevent future people from accidentally changing this again.

lmr3796: Also I suppose this is unit-test-able?
But since this is a site up we can have another PR for it.

lmr3796 (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/ZKBrokerFailureDetector.java
lmr3796: And also state the symptom of the error (under scala xxx version, it never returns).

CCisGG: I tried to create a unit test to use the same converter, and the test works. It seems to be a runtime only error.

wyuka (commented):
Please add an explanation of the issue and the fix to the PR description. I know that it is detailed in the issue, but it is a good idea to explain why your change should fix it in the PR description.

lmr3796 (approved):

PR #1879, opened 2022-08-10
Handle the inconsistency between clusterModel and kafka metadata during update topic configuration

### LINKED ISSUE #1878: TopicReplicationFactorAnomaly occasionally reduce RF of partition to less than the desired RF
At 7:55 the kafka controller is trying to changing replica assignment from  [45048, 60241, 41345] to [45058, 41914, 41244]:

 
`kafka-controller.log.2022-08-08-07.gz:2022/08/08 07:55:36.977 [KafkaController] [Controller id=45139] Successfully updated assignment of partition topic-5 to ReplicaAssignment(replicas=41914,41244,45048,60241,41345, addingReplicas=41914,41244, removingReplicas=60241,41345)`
 

Then at 7:56, cruise-control detected the TopicReplicationFactorAnomaly and trying to fix it. Before fixing the anomaly, CC has to terminate the ongoing assignment:

 

 
`
likafka-cruise-control.log.2022-08-08-07.gz:2022/08/08 07:56:02.689 INFO [KafkaCruiseControl] [AnomalyDetector-2] [kafka-cruise-control] [] External agent is reassigning partitions. The request to stop it is submitted successfully: [topic-5] `
 

 

As a result, the "addingReplicas=41914,41244" are terminated by cc and wasn't actually added to the replica set, but cc was thinking there are 5 replicas [41914, 41244, 45048, 60241, 41345] and tried to remove the replica [60241,41345].

 
It end up with the partition only have 1 replica [45058] while the desired RF is 3.

CCisGG:
This PR resolves #1878 .

**Description:**

CruiseControl holds 2 versions of kafka cluster metadata:
1. ClusterModel: comes from the load monitor and may have fresher information
2. Cluster: comes from the metadata response from a kafka bootstrap server. 

The Cluster object may contain stale data if the kafka server has the stale metadata (which is possible because it takes time for kafka controller to send out the latest metadata to all brokers).

In that case, the data in Cluster may differ from the data in ClusterModel. E.g. a broker A can exist in Cluster but not in ClusterModel.

This PR is to take care of the above inconsistency during topic configuration change.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
@@ -38,7 +38,6 @@ import org.apache.kafka.common.TopicPartition;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static com.linkedin.kafka.cruisecontrol.model.ModelUtils.*;
 import static com.linkedin.kafka.cruisecontrol.monitor.MonitorUtils.EMPTY_BROKER_CAPACITY;
 
 /**
@@ -992,7 +991,7 @@ public class ClusterModel implements Serializable {
 
           TopicPartition tp = new TopicPartition(topic, partitionInfo.partition());
           Partition partition = partition(tp);
-          if (!hasSameReplicasFor(partition, partitionInfo)) {
+          if (!ModelUtils.hasSameReplicasFor(partition, partitionInfo)) {
             LOG.warn("Detected partition info inconsistent with clusterModel: PartitionInfo: {}, Partition in ClusterModel: {}. "
                      + " Skip creating or deleting replicas for this partition.", partitionInfo, partition);
             continue;
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
@@ -18,7 +18,6 @@ import java.io.OutputStream;
 import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -47,7 +46,6 @@ import static com.linkedin.kafka.cruisecontrol.monitor.MonitorUtils.EMPTY_BROKER
  * the input of the analyzer to generate the proposals for load rebalance.
  */
 public class ClusterModel implements Serializable {
-
   private static final Logger LOG = LoggerFactory.getLogger(ClusterModel.class);
   private static final long serialVersionUID = -6840253566423285966L;
   // Hypothetical broker that indicates the original broker of replicas to be created in the existing cluster model.
@@ -987,9 +985,10 @@ public class ClusterModel implements Serializable {
         int rackCursor = 0;
         for (PartitionInfo partitionInfo : cluster.partitionsForTopic(topic)) {
 
-          if (!sanityCheckPartitionInfoConsistency(partitionInfo)) {
-            LOG.warn("Detected partition info {} inconsistent with clusterModel:"
-                     + " skip creating or deleting replicas for this partition.", partitionInfo);
+          Partition partition = partition(new TopicPartition(partitionInfo.topic(), partitionInfo.partition()));
+          if (!hasSameReplicasFor(partition, partitionInfo)) {
+            LOG.warn("Detected partition info inconsistent with clusterModel: PartitionInfo: {}, Partition in ClusterModel: {}. "
+                     + " Skip creating or deleting replicas for this partition.", partitionInfo, partition);
             continue;
           }
 
@@ -1404,15 +1403,23 @@ public class ClusterModel implements Serializable {
     }
   }
 
-  private boolean sanityCheckPartitionInfoConsistency(PartitionInfo partitionInfo) {
-    Partition partition = partition(new TopicPartition(partitionInfo.topic(), partitionInfo.partition()));
+  private boolean hasSameReplicasFor(Partition partition, PartitionInfo partitionInfo) {
     if (partition == null) {
       return false;
     }
 
-    Set<Integer> replicasFromPartition = partition.replicas().stream().map(Replica::broker).map(Broker::id).collect(Collectors.toSet());
-    Set<Integer> replicasFromPartitionInfo = Arrays.stream(partitionInfo.replicas()).map(Node::id).collect(Collectors.toSet());
-    return partition.leader().broker().id() == partitionInfo.leader().id()
-           && replicasFromPartition.equals(replicasFromPartitionInfo);
+    List<Replica> replicasFromPartition = partition.replicas();
+    Node[] replicasFromPartitionInfo = partitionInfo.replicas();
+    if (replicasFromPartition.size() != replicasFromPartitionInfo.length) {
+      return false;
+    }
+
+    for (int i = 0; i < replicasFromPartitionInfo.length; i++) {
+      if (replicasFromPartitionInfo[i].id() != replicasFromPartition.get(i).broker().id()) {
+        return false;
+      }
+    }
+
+    return true;
   }
 }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
@@ -18,6 +18,7 @@ import java.io.OutputStream;
 import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -1015,6 +1016,11 @@ public void createOrDeleteReplicas(Map<Short, Set<String>> topicsByReplicationFa
             // Make sure the leader replica is in new replica list.
             newAssignedReplica.add(partitionInfo.leader().id());
             for (Node node : partitionInfo.replicas()) {
+              // Current clusterModel may have inconsistency with the partitionInfo. In that case, we skip the replica.
+              if (!brokerIds.contains(node.id())) {
+                continue;
+              }
+
```
efeg, line 1023: I think this change might be missing some inconsistencies in case brokers are part of the cluster, but the cluster metadata is inconsistent with cluster model.
Why don't we compare the equality of replicas of the target partition in (1) `ClusterModel` and (2) `Cluster`?

In case there is a mismatch, we can log this as a warn, explain that the fix is skipped due to stale cluster metadata, and skip the fix for that partition. With this approach, ss long as the topic_anomaly self-healing is enabled or if the user retries, any inconsistency in replication factor should eventually be fixed and we would not block the progress of fixes for other partitions.

^ We should also document this potential behavior in the JavaDoc of `UpdateTopicConfigurationRunnable` and `LiUpdateTopicConfigurationRunnable`

efeg (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
CCisGG: That's actually a better idea. Let me make the change

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -981,6 +987,13 @@ public void createOrDeleteReplicas(Map<Short, Set<String>> topicsByReplicationFa
         int[] cursors = new int[racks.size()];
         int rackCursor = 0;
         for (PartitionInfo partitionInfo : cluster.partitionsForTopic(topic)) {
+
+          if (!sanityCheckPartitionInfoConsistency(partitionInfo)) {
+            LOG.warn("Detected partition info {} inconsistent with clusterModel:"
+                     + " skip creating or deleting replicas for this partition.", partitionInfo);
```
efeg, line 992: Can this log also show the replicas in the partition of the cluster model?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -1391,4 +1409,16 @@ private void refreshCapacity() {
       _clusterCapacity[r.id()] = capacity;
     }
   }
+
+  private boolean sanityCheckPartitionInfoConsistency(PartitionInfo partitionInfo) {
```
efeg, line 1407: The name of this function is not very clear -- i.e. looks like it checks whether the cluster model has the same partition having the same replicas with the passed partition info.

Can we rename this function to make that more clear? Maybe something like `hasSameReplicasFor(...)`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -1391,4 +1409,16 @@ private void refreshCapacity() {
...
+    Partition partition = partition(new TopicPartition(partitionInfo.topic(), partitionInfo.partition()));
+    if (partition == null) {
+      return false;
+    }
+
+    Set<Integer> replicasFromPartition = partition.replicas().stream().map(Replica::broker).map(Broker::id).collect(Collectors.toSet());
+    Set<Integer> replicasFromPartitionInfo = Arrays.stream(partitionInfo.replicas()).map(Node::id).collect(Collectors.toSet());
```
efeg, line 1414: Converting array to Set and List to Set introduces additional complexity. Under normal circumstances, we expect the order of the replica array and replica list to be the same. Can we first compare the sizes of the list and array and the use a simple loop to compare each element to make sure that they are equal?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -1391,4 +1409,16 @@ private void refreshCapacity() {
...
+    return partition.leader().broker().id() == partitionInfo.leader().id()
```
efeg, line 1415: Are there any side-effects if we omit the leader equality check?

efeg (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
CCisGG: If we omit the check, the real leader replica can be deleted. That case it may introduce extra leadership movement, which I think might be fine.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -981,6 +984,14 @@ public void createOrDeleteReplicas(Map<Short, Set<String>> topicsByReplicationFa
         int[] cursors = new int[racks.size()];
         int rackCursor = 0;
         for (PartitionInfo partitionInfo : cluster.partitionsForTopic(topic)) {
+
+          Partition partition = partition(new TopicPartition(partitionInfo.topic(), partitionInfo.partition()));
```
efeg, line 988: Not introduced in this PR: I noticed that the same `TopicPartition` is used below within this scope. Can we extract the following here:
```
TopicPartition tp = new TopicPartition(topic, partitionInfo.partition());
```

And use this tp anywhere else needed in this scope? Reusing the same immutable object is more efficient than recreating it multiple times below.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -981,6 +984,14 @@ public void createOrDeleteReplicas(Map<Short, Set<String>> topicsByReplicationFa
...
+          if (!hasSameReplicasFor(partition, partitionInfo)) {
+            LOG.warn("Detected partition info inconsistent with clusterModel: PartitionInfo: {}, Partition in ClusterModel: {}. "
+                     + " Skip creating or deleting replicas for this partition.", partitionInfo, partition);
+            continue;
+          }
+
```
efeg, line 994: Since this if check is more expensive compared to the following if check (i.e. `if (partitionInfo.replicas().length == replicationFactor) {`), can we move that one before this check?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -1391,4 +1402,24 @@ private void refreshCapacity() {
       _clusterCapacity[r.id()] = capacity;
     }
   }
+
+  private boolean hasSameReplicasFor(Partition partition, PartitionInfo partitionInfo) {
```
efeg, line 1406: Nit: After the latest commit (i.e. now we are passing the `partition` as a parameter), we can make this function `static` and move it to the utility class of this package (see `ModelUtils`). So that it can be reused by other relevant classes that may want to make a similar comparison.

efeg (changes requested):
Thanks for the updates -- left a few more comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ClusterModel.java
```diff
@@ -35,7 +35,10 @@
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import static com.linkedin.kafka.cruisecontrol.model.ModelUtils.*;
```
efeg, line 41: Nit: Can we avoid wildcard import? -- e.g. maybe we can directly use `ModelUtils.hasSameReplicasFor(...)` in the code w/o an import.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ModelUtils.java
```diff
@@ -166,4 +169,30 @@ public static double expectedUtilizationFor(Resource resource, AggregatedMetricV
     }
     return max(result, 0.0);
   }
+
+  /**
+   * Check the consistency between a Partition object and a PartitionInfo object
+   * @param partition the Partition object that contains partition info
+   * @param partitionInfo the PartitionInfo object that contains partition info
+   * @return true if both objects contains the same partition info
```
efeg, line 177: Nit: Can we use proper tags -- e.g. `{@code true}`, and `{@link }` for Partition and PartitionInfo.

efeg (commented):
Left a few nits -- o/w lgtm.

efeg (approved):
Thanks for the fix -- LGTM!

PR #1809, opened 2022-03-30
Add BrokerSetAwareGoal

mohitpali:
This PR resolves #1782

Currently, cruise-control does not have a way to limit the leader and replica movement plan within a Broker Group. We would like to make cruise-control aware of the Broker Groups and propose optimizations where each optimization does leader movement or replica movement within a Broker Group. In short, each optimization should be limited within the boundary of Broker Group.

In this PR :

1. I have added new `BrokerSetAwareGoal` (Hard Goal).
2. I have also added a `BrokerSetDataStore` interface.
3. I have provided a file based default implementation to fetch `BrokerSet` Information called `BrokerSetFileDataStore`

PR #1881, opened 2022-08-11
Mute KafkaCruiseControlConfig logs during anomaly detection

### LINKED ISSUE #1880: Verbose logs for KafkaCruiseControlConfig makes debug hard
There are a lots of config logs whenever anomaly detector runs:

```
2022/08/10 23:40:30.815 INFO [KafkaCruiseControlConfig] [AnomalyDetector-2] [kafka-cruise-control] [] KafkaCruiseControlConfig values: 
...
...
```

We should mute them.

CCisGG:
This PR resolves #1880.

@mohitpali FYI. I think this logging is not intentional brought in your PR #1809 . Please correct me if I'm wrong.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
@@ -4,6 +4,7 @@
 
 package com.linkedin.kafka.cruisecontrol.config;
 
+import com.linkedin.cruisecontrol.common.utils.Utils;
 import com.linkedin.kafka.cruisecontrol.config.constants.AnalyzerConfig;
 import com.linkedin.kafka.cruisecontrol.exception.BrokerSetResolutionException;
 import com.linkedin.kafka.cruisecontrol.model.ClusterModel;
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
```diff
@@ -44,7 +44,7 @@ public class BrokerSetFileResolver implements BrokerSetResolver {
 
   @Override
   public void configure(Map<String, ?> configs) {
-    KafkaCruiseControlConfig config = new KafkaCruiseControlConfig(configs);
+    KafkaCruiseControlConfig config = new KafkaCruiseControlConfig(configs, false);
```
efeg, line 47: Why do we even create a new `KafkaCruiseControlConfig` from the passed parameter. Can we avoid this?

efeg (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
CCisGG: The passed parameter is a Map. Several lines below it used the KafkaCruiseControlConfig object to create a instance:


`        config.getConfiguredInstance(AnalyzerConfig.BROKER_SET_ASSIGNMENT_POLICY_CLASS_CONFIG, BrokerSetAssignmentPolicy.class);`

So I think it's reasonable to use KafkaCruiseControlConfig here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
```diff
@@ -44,7 +44,7 @@ public class BrokerSetFileResolver implements BrokerSetResolver {
 
   @Override
```
mohitpali, line 45: not setting ``doLog`` was NOT intentional.
LGTM.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
mohitpali: @CCisGG is correct. This was done to load the instance and reuse the ``getConfiguredInstance()`` method. I was trying to emulate ``BrokerCapacityConfigFileResolver`` to keep up with the repo standards. 

However, ``BrokerCapacityConfigFileResolver`` just loaded the json file name from `` _configFile = KafkaCruiseControlUtils.getRequiredConfig(configs, CAPACITY_CONFIG_FILE);``.

``KafkaCruiseControlUtils`` did NOT have a method to get the instance. Hence I created an instance of ``KafkaCruiseControlConfig`` to reuse the ``getConfiguredInstance`` method.

efeg: Creating a new `KafkaCruiseControlConfig` is expensive and I suspect we can avoid it.
 
-- i.e. we can create the `BrokerSetAssignmentPolicy` instance before creating this class in. For examples of this usage, please search for the use of `parameterConfigOverrides` with `*_OBJECT_CONFIG;` for `getConfiguredInstance`.

CCisGG: @efeg I did a code search for parameterConfigOverrides and getConfiguredInstance, but failed to find an valid example for our case. All the examples are calling the getConfiguredInstance method on an KafkaCruiseControlConfig object. Would you mind provide a concrete example?

efeg: They are supposed to use KafkaCruiseControlConfig object. The `BrokerSetAssignmentPolicy` object should be created from within `BalancingConstraint` constructor using the parameter of the constructor,  which is a `KafkaCruiseControlConfig`. Then you should update `_brokerSetResolver = config.getConfiguredInstance(AnalyzerConfig.BROKER_SET_RESOLVER_CLASS_CONFIG, BrokerSetResolver.class);` to get the extra parameter (i.e. `Map<String, Object> configOverrides`).

CCisGG: Addressed

efeg (approved):
Thanks for the updates! LGTM.

PR #1884, opened 2022-08-11
Make sure the provision response is always RIGHT_SIZED if it is not OVER_PROVISIONED

### LINKED ISSUE #1877: Cruise-control falsely report OVER_PROVISIONED when the cluster is right_sized
There are multiple goals in cruise-control that could provide OVER_PROVISIONED as the ProvisionStatus. After all goals are optimized, there will be an aggregated ProvisionStatus. 

It is expected that the aggregated ProvisionStatus is OVER_PROVISIONED if and only if ALL relative goals provide OVER_PROVISIONED status.

Today there is a bug in cruise-control that violates the above rule. For example:

For ReplicaDistributionGoal,  even if this goal decides the cluster is NOT OVER_PROVISIONED, the final aggregated provision response could still be OVER_PROVISIONED. This happens when ReplicaDistributionGoal fails and returns an UNDECIDED provision status.

As a fix, we should make sure that all the goals that can possibly returns OVER_PROVISIONED, will never returns UNDECIDED as provision status. These goals include:

- BrokerSetAwareGoal
- RackAwareDistributionGoal
- RackAwareGoal
- ReplicaDistributionGoal
- ResourceDistributionGoal

CCisGG:
This PR resolves #1877 .

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
@@ -161,6 +161,7 @@ public class BrokerSetAwareGoal extends AbstractGoal {
     _brokerSetResolutionHelper = new BrokerSetResolutionHelper(clusterModel, brokerSetResolver);
     _brokersByBrokerSet = _brokerSetResolutionHelper.brokersByBrokerSetId();
     _replicaToBrokerSetMappingPolicy = _balancingConstraint.replicaToBrokerSetMappingPolicy();
+    _provisionResponse = new ProvisionResponse(ProvisionStatus.RIGHT_SIZED);
   }
 
   /**
@@ -179,9 +180,6 @@ public class BrokerSetAwareGoal extends AbstractGoal {
     GoalUtils.ensureReplicasMoveOffBrokersWithBadDisks(clusterModel, name());
     // Sanity check to confirm that the final distribution is broker set aware.
     ensureBrokerSetAware(clusterModel, optimizationOptions);
-    if (_provisionResponse.status() != ProvisionStatus.OVER_PROVISIONED) {
-      _provisionResponse = new ProvisionResponse(ProvisionStatus.RIGHT_SIZED);
-    }
     finish();
   }
 
@@ -236,6 +234,7 @@ public class BrokerSetAwareGoal extends AbstractGoal {
                                                        .collect(Collectors.toSet());
       if (maybeApplyBalancingAction(clusterModel, replica, eligibleBrokers, ActionType.INTER_BROKER_REPLICA_MOVEMENT, optimizedGoals,
                                     optimizationOptions) == null) {
+        _provisionResponse = new ProvisionResponse(ProvisionStatus.UNDER_PROVISIONED);
         // If balancing action can not be applied, provide recommendation to add new Brokers.
         ProvisionRecommendation recommendation =
             new ProvisionRecommendation.Builder(ProvisionStatus.UNDER_PROVISIONED).numBrokers(clusterModel.maxReplicationFactor()).build();
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
@@ -24,7 +24,6 @@ import com.linkedin.kafka.cruisecontrol.analyzer.BalancingAction;
 import com.linkedin.kafka.cruisecontrol.analyzer.BalancingConstraint;
 import com.linkedin.kafka.cruisecontrol.analyzer.OptimizationOptions;
 import com.linkedin.kafka.cruisecontrol.analyzer.ProvisionRecommendation;
-import com.linkedin.kafka.cruisecontrol.analyzer.ProvisionResponse;
 import com.linkedin.kafka.cruisecontrol.analyzer.ProvisionStatus;
 import com.linkedin.kafka.cruisecontrol.config.BrokerSetResolutionHelper;
 import com.linkedin.kafka.cruisecontrol.config.BrokerSetResolver;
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
```diff
@@ -236,6 +234,7 @@ protected void rebalanceForBroker(Broker broker, ClusterModel clusterModel, Set<
                                                        .collect(Collectors.toSet());
       if (maybeApplyBalancingAction(clusterModel, replica, eligibleBrokers, ActionType.INTER_BROKER_REPLICA_MOVEMENT, optimizedGoals,
                                     optimizationOptions) == null) {
+        _provisionResponse = new ProvisionResponse(ProvisionStatus.UNDER_PROVISIONED);
```
efeg, line 237: Why do we need this? Isn't this handled in `AbstractGoal#optimize`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
```diff
@@ -161,6 +161,7 @@ protected void initGoalState(ClusterModel clusterModel, OptimizationOptions opti
     _brokerSetResolutionHelper = new BrokerSetResolutionHelper(clusterModel, brokerSetResolver);
     _brokersByBrokerSet = _brokerSetResolutionHelper.brokersByBrokerSetId();
     _replicaToBrokerSetMappingPolicy = _balancingConstraint.replicaToBrokerSetMappingPolicy();
+    _provisionResponse = new ProvisionResponse(ProvisionStatus.RIGHT_SIZED);
```
efeg, line 164: Why do we set this to right sized? Wouldn't this cause the cluster never being reported as over provisioned if this goal is used?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
```diff
@@ -101,8 +101,7 @@ protected void updateGoalState(ClusterModel clusterModel, OptimizationOptions op
       ProvisionRecommendation recommendation = new ProvisionRecommendation.Builder(ProvisionStatus.OVER_PROVISIONED)
           .numBrokers(numBrokersToDrop).build();
       _provisionResponse = new ProvisionResponse(ProvisionStatus.OVER_PROVISIONED, recommendation, name());
-    } else if (_succeeded) {
-      // The cluster is not overprovisioned and all brokers are within the upper and lower balance limits.
+    } else {
```
efeg, line 104: Did we check the other goals, which set the _provisionResponse to `OVER_PROVISIONED` for the same bug?

efeg (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/ReplicaDistributionGoal.java
CCisGG: Yes, others will not have the same issue.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
CCisGG: I guess you're right. I'll remove this line. However this does cause confusion to me, and maybe to following developers. Essentially you don't return RIGHT_SIZED even if your goal proves the cluster is neither OVER_PROVISIONED nor UNDER_PROVISIONED. I think it is confusing.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
CCisGG: Got it. I just noticed it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
```diff
@@ -234,7 +232,6 @@ protected void rebalanceForBroker(Broker broker, ClusterModel clusterModel, Set<
                                                        .collect(Collectors.toSet());
       if (maybeApplyBalancingAction(clusterModel, replica, eligibleBrokers, ActionType.INTER_BROKER_REPLICA_MOVEMENT, optimizedGoals,
                                     optimizationOptions) == null) {
-        _provisionResponse = new ProvisionResponse(ProvisionStatus.UNDER_PROVISIONED);
```
mohitpali: I dont fully understand. If a brokerSetAware goal is violated and can not be fixed, the recommendation provided by BrokerSetAwareGoal would be UNDER_PROVISIONED, since we would need to add more brokers to the brokerSet. There could be other brokerSets that do not need to add any additional brokers.

efeg: @mohitpali Yes, it is going to be `UNDER_PROVISIONED`. Please check out how `AbstractGoal#optimize` handles `OptimizationFailureException` -- note that `BrokerSetAwareGoal` extends `AbstractGoal`.

efeg (approved):
Thanks for the updates - LGTM.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
efeg: For a goal to indicate whether a cluster is right sized, it must be able to gauge whether the cluster is over provisioned or not. Since the current implementation of this goal has no criteria to say whether the cluster is over provisioned or not, it cannot claim the cluster as right sized.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
mohitpali: Ok, makes sense.

PR #1888, opened 2022-08-19
Port BrokerSetResolver to KafkaClusterState

### LINKED ISSUE #1887: Kafka_cluster_state doesn't have access to broker set information
As recently @mohitpali added a BrokerSetAwareGoal, we need to visualize the information in kafka_cluster_state.

The first step is to port the BrokerSetResolver instance to the KafkaClusterState. Notice that we don't want to create a new instance of BrokerSetResolver, since it should be a singleton.

One side note of this change: this is an interface change. For users who adds wrappers on top of the KafkaClusterState, they will need to address this change to avoid compile issue.

This should be sub-issue of https://github.com/linkedin/cruise-control/issues/1860

CCisGG:
This PR resolves #1887.

This is the first step of visualizing broker set info in kafka_cluster_state endpoint. The BrokerSetResolver will be regarded as single source of truth to get the broker set info.


CCisGG:
Closed as we don't want to no-op changes. I'll make a single PR that includes all the necessary changes that are needed to visualize the broker set info.

PR #1889, opened 2022-08-24
Display the brokerSet id in kafka_cluster_state endpoint.

### LINKED ISSUE #1860: Add brokerSet id in kafka_cluster_state
As cruise control now supports BrokerSetAwareGoal (introduced in #1809), we may want to show the brokerSet id of each topic in the kafka_cluster_state.

Today, in kafka_cluster_state response, there is a section for broker state:

Screen Shot 2022-08-19 at 10 34 02 AM

What we would like to have, is a new column BROKER_SET which shows the broker set id of each broker. The new broker state should look like below:

Screen Shot 2022-08-19 at 10 34 59 AM

Preferably, this column should only show up when the BrokerSetAwareGoal is in the goals list.

This change will help the users to know what is the broker set id of each broker. Without this change, the broker set info will be opaque to users. E.g. after the rebalance happened, there is no way for users to know which broker set that each broker belongs to. This can introduce unnecessarily challenges for debugging cc and inspecting cluster status. Say if a user implements custom rules to assign broker set to brokers, the broker set arrangement would be unknown to the user after goal optimization.

CCisGG:
This PR resolves #1860.

**Description**:

This PR pass the BrokerSetResolver instance to KafkaClusterStateRequest, so that calling kafka_cluster_state endpoint would display the brokerSet information for each broker. E.g.

Screen Shot 2022-08-24 at 2 54 09 PM

Along with that change, this PR deprecates several BrokerSetAwareGoal related APIs and creates replacements. Reasons:

- BrokerSetResolver is now used in both BrokerSetAwareGoal and KafkaClusterStateRequest. The available/accessible info is different in those 2 classes.
- BrokerSetAwareGoal has access to ClusterModel object.
- KafkaClusterStateRequest has access to Cluster object.
- In order to make BrokerSetResolver usable for both places, we have to expose an interface that has input param of commonly available info. In our case, it's the "rackIdByBrokerId" map. The map contains all broker ids and their racks. which is available from both clusterModel and cluster.

Besides, there are multiple side-notes:
- Most importantly: BrokerSetResolver will be the single source of truth of broker-set assignment. There will be only one single instance of BrokerSetResolver exists in CC. In contrast, BrokerSetResolutionHelper is a wrapper on top of BrokerSetResolver, and there could be multiple instances of BrokerSetResolutionHelper.
- We use BrokerSetResolutionHelper as a facade, to convert ClusterModel or Cluster to the "rackIdByBrokerId" map.
- If the broker is not assigned with an BrokerSet, or the BrokerSetResolver fails, the kafka_cluster_state will show the BROKER_SET of the broker as "null".

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
@@ -9,6 +9,7 @@ import com.linkedin.kafka.cruisecontrol.common.Utils;
 import com.linkedin.kafka.cruisecontrol.config.ReplicaToBrokerSetMappingPolicy;
 import com.linkedin.kafka.cruisecontrol.exception.BrokerSetResolutionException;
 import com.linkedin.kafka.cruisecontrol.exception.ReplicaToBrokerSetMappingException;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -79,7 +80,7 @@ import static com.linkedin.kafka.cruisecontrol.analyzer.goals.GoalUtils.replicaS
 public class BrokerSetAwareGoal extends AbstractGoal {
   private static final Logger LOG = LoggerFactory.getLogger(BrokerSetAwareGoal.class);
   private BrokerSetResolutionHelper _brokerSetResolutionHelper;
-  private Map<String, Set<Broker>> _brokersByBrokerSet;
+  private Map<String, Set<Integer>> _brokersByBrokerSet;
   private ReplicaToBrokerSetMappingPolicy _replicaToBrokerSetMappingPolicy;
   private Set<String> _excludedTopics;
   private Set<String> _mustHaveTopicLeadersPerBroker;
@@ -188,10 +189,11 @@ public class BrokerSetAwareGoal extends AbstractGoal {
       String topicName = partitionsByTopic.getKey();
       if (!_excludedTopics.contains(topicName)) {
         List<Partition> partitions = partitionsByTopic.getValue();
-        Set<Broker> allBrokersForTopic = partitions.stream()
-                                                   .map(partition -> partition.partitionBrokers())
-                                                   .flatMap(brokers -> brokers.stream())
-                                                   .collect(Collectors.toSet());
+        Set<Integer> allBrokersForTopic = partitions.stream()
+                                                    .map(Partition::partitionBrokers)
+                                                    .flatMap(Collection::stream)
+                                                    .map(Broker::id)
+                                                    .collect(Collectors.toSet());
         // Checks if a topic's brokers do not all live in a single brokerSet
         if (_brokersByBrokerSet.values().stream().noneMatch(brokerSetBrokers -> brokerSetBrokers.containsAll(allBrokersForTopic))) {
           throw new OptimizationFailureException(
@@ -214,7 +216,7 @@ public class BrokerSetAwareGoal extends AbstractGoal {
                                     OptimizationOptions optimizationOptions) throws OptimizationFailureException {
     LOG.debug("balancing broker {}, optimized goals = {}", broker, optimizedGoals);
 
-    String currentBrokerSetId = _brokerSetResolutionHelper.brokerSetId(broker);
+    String currentBrokerSetId = _brokerSetResolutionHelper.brokerSetId(broker.id());
 
     for (Replica replica : broker.trackedSortedReplicas(replicaSortName(this, false, false)).sortedReplicas(true)) {
       String eligibleBrokerSetIdForReplica =
@@ -226,10 +228,10 @@ public class BrokerSetAwareGoal extends AbstractGoal {
         continue;
       }
       // If the brokerSet awareness condition is violated. Move replica to an eligible broker
-      Set<Broker> eligibleBrokers = _brokersByBrokerSet.get(eligibleBrokerSetIdForReplica)
-                                                       .stream()
-                                                       .filter(b -> clusterModel.aliveBrokers().contains(b))
-                                                       .collect(Collectors.toSet());
+      Set<Broker> eligibleBrokers = clusterModel.aliveBrokers()
+                                                .stream()
+                                                .filter(b -> _brokersByBrokerSet.get(eligibleBrokerSetIdForReplica).contains(b.id()))
+                                                .collect(Collectors.toSet());
       if (maybeApplyBalancingAction(clusterModel, replica, eligibleBrokers, ActionType.INTER_BROKER_REPLICA_MOVEMENT, optimizedGoals,
                                     optimizationOptions) == null) {
         // If balancing action can not be applied, provide recommendation to add new Brokers.
@@ -296,7 +298,7 @@ public class BrokerSetAwareGoal extends AbstractGoal {
     try {
       String expectedBrokerSetId =
           _replicaToBrokerSetMappingPolicy.brokerSetIdForReplica(sourceReplica, clusterModel, _brokerSetResolutionHelper);
-      String destinationBrokerSetId = _brokerSetResolutionHelper.brokerSetId(destinationBroker);
+      String destinationBrokerSetId = _brokerSetResolutionHelper.brokerSetId(destinationBroker.id());
 
       return !expectedBrokerSetId.equals(destinationBrokerSetId);
     } catch (BrokerSetResolutionException | ReplicaToBrokerSetMappingException e) {
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
@@ -59,10 +59,23 @@ public class BrokerSetFileResolver implements BrokerSetResolver {
     try {
       brokerIdsByBrokerSetId = loadBrokerSetData();
     } catch (IOException e) {
-      throw new IllegalArgumentException(e);
+      throw new BrokerSetResolutionException(e.getMessage());
     }
 
     return _brokerSetAssignmentPolicy.assignBrokerSetsForUnresolvedBrokers(clusterModel, brokerIdsByBrokerSetId);
+  }
+
+  @Override
+  public Map<String, Set<Integer>> brokerIdsByBrokerSetId(Map<Integer, String> rackIdByBrokerId)
+      throws BrokerSetResolutionException {
+    Map<String, Set<Integer>> brokerIdsByBrokerSetId;
+    try {
+      brokerIdsByBrokerSetId = loadBrokerSetData();
+    } catch (IOException e) {
+      throw new BrokerSetResolutionException(e.getMessage());
+    }
+
+    return _brokerSetAssignmentPolicy.assignBrokerSetsForUnresolvedBrokers(rackIdByBrokerId, brokerIdsByBrokerSetId);
   }
 
   private Map<String, Set<Integer>> loadBrokerSetData() throws IOException {
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetResolutionHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetResolutionHelper.java
@@ -5,73 +5,96 @@
 package com.linkedin.kafka.cruisecontrol.config;
 
 import com.linkedin.kafka.cruisecontrol.exception.BrokerSetResolutionException;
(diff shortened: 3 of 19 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
```diff
@@ -59,12 +59,25 @@ public Map<String, Set<Integer>> brokerIdsByBrokerSetId(ClusterModel clusterMode
...
   }
 
+  @Override
+  public Map<String, Set<Integer>> brokerIdsByBrokerSetId(Map<Integer, String> rackIdByBrokerId)
+      throws BrokerSetResolutionException {
+    Map<String, Set<Integer>> brokerIdsByBrokerSetId;
+    try {
+      brokerIdsByBrokerSetId = loadBrokerSetData();
+    } catch (IOException e) {
+      throw new BrokerSetResolutionException(e.getMessage());
```
wyuka, line 75: Why not handle it gracefully if the file is missing / malformed? I think this method should be robust, and rely on the `assignBrokerSetsForUnresolvedBrokers` to handle the error scenario.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
```diff
@@ -59,12 +59,25 @@ public Map<String, Set<Integer>> brokerIdsByBrokerSetId(ClusterModel clusterMode
...
+    }
+
+    return _brokerSetAssignmentPolicy.assignBrokerSetsForUnresolvedBrokers(rackIdByBrokerId, brokerIdsByBrokerSetId);
```
wyuka, line 78: Why are we not persisting this data back to the file? That would make sure that the unassigned brokers don't change colors on broker restart.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetResolutionHelper.java
```diff
@@ -5,73 +5,96 @@
...
-    throw new BrokerSetResolutionException(String.format("Failed to resolve BrokerSet for Broker %s", broker));
+    throw new BrokerSetResolutionException(String.format("Failed to resolve BrokerSet for Broker %s", brokerId));
+  }
+
+  /**
+   * Get the broker to rack mapping from cluster
+   * @param cluster the kafka cluster
+   * @return the map of broker id to rack id
+   */
+  public static Map<Integer, String> getRackIdByBrokerIdMapping(Cluster cluster) {
```
wyuka, line 84: Does this need to be public? Same goes for the other method with the same name.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
```diff
@@ -79,7 +80,7 @@
 public class BrokerSetAwareGoal extends AbstractGoal {
   private static final Logger LOG = LoggerFactory.getLogger(BrokerSetAwareGoal.class);
   private BrokerSetResolutionHelper _brokerSetResolutionHelper;
```
wyuka, line 82: I see that the `_brokerSetResolutionHelper` is initialized and stored as a field. What is the lifetime of the BrokerSetAwareGoal class? Is it created each time the goal calculation is done? Or is the same class re-used?

If it is the latter, then having a `BrokerSetResolutionHelper` as a member will have the side effect of re-using stale broker to color mappings even as they change.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
CCisGG: @wyuka I think your proposal is reasonable, but there might be some other considerations when @mohitpali implemented this. @mohitpali do you think it would work better if we write back to the file?

In anyways, that part should belong to a new issue/PR. We want to focus this PR on the exact issue it linked.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetResolutionHelper.java
CCisGG: Actually, they are also used in unit tests. So I'd keep them as public static.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
CCisGG: Those goals are indeed initialized multiple times from different places, for different purposes. We don't need to worry about that as in our case, the broker assignment is not called during goal init time.

However, you're right that the goal will re-calculate brokerSet assignment on each goal optimization. We think that is the expected behavior. The user of this goal should take care of the broker set cache.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
CCisGG: I think we intentionally want to remind the user that their config file is missing. E.g. they may put the file in wrong place, and in that case we don't want the user to mis-think that BrokerSetResolver is using their json file.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetResolver.java
```diff
@@ -23,6 +23,16 @@ public interface BrokerSetResolver extends CruiseControlConfigurable {
    *
    * @param clusterModel cluster model object
    * @return A map of broker Ids by their broker set Id
+   * @deprecated This method will be replaced by {@link #brokerIdsByBrokerSetId(Map)}
    */
+  @Deprecated
```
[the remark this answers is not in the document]
CCisGG: Sure let me add a TODO there.

CCisGG:
@mohitpali Could you take a look at this PR to see if the changes makes sense? This may not affect your current functionality, but once we remove the deprecated method in next few weeks, there should be some code changes needed from your side.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/NoOpBrokerSetAssignmentPolicy.java
```diff
@@ -15,26 +16,25 @@
...
 public class NoOpBrokerSetAssignmentPolicy implements BrokerSetAssignmentPolicy {
-  private static final String UNMAPPED_BROKER_SET_ID = "unmapped";
+  public static final String UNMAPPED_BROKER_SET_ID = "unmapped";
 
   /**
    * Does not assign any broker set to non-mapped brokers.
    *
    * @return A map of broker Ids by their broker set Id
    */
   @Override
```
mohitpali, line 30: should this be @Deprecated ?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolverTest.java
```diff
@@ -72,17 +73,28 @@ public void testNonExistentFile() {
...
+   * Tests if the cluster model has extra brokers that are unresolved from broker set mapping then throws broker set resolution exception
    */
   @Test
-  public void testBrokerSetResolutionExceptionWithNoOpDefaultAssignmentPolicy() {
+  public void testBrokerSetResolutionWithNoOpDefaultAssignmentPolicy() throws BrokerSetResolutionException {
     BrokerSetResolver brokerSetResolver = getBrokerSetResolver("testBrokerSets.json", this.getClass());
-    assertThrows(BrokerSetResolutionException.class,
-                 () -> brokerSetResolver.brokerIdsByBrokerSetId(DeterministicCluster.brokerSetUnSatisfiable3()));
+    final Map<String, Set<Integer>> brokerSets = brokerSetResolver.brokerIdsByBrokerSetId(
+        BrokerSetResolutionHelper.getRackIdByBrokerIdMapping(DeterministicCluster.brokerSetUnSatisfiable3()));
```
mohitpali, line 88: I dont remember why it was throwing Exception earlier.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolver.java
mohitpali: This was done to address a case where a broker is added to the cluster. Ideally the brokerSet to Broker mapping should be in-sync with the Cluster. But since the broker set mapping, even if dynamic, is not guaranteed to be updated atomically. In case of file based implementation, user should be updating the mapping json or implementing a resolution policy that is deterministic. I don't mind overwriting the file if user has saved the file at a non-ephemeral disk. But since this json is part of the code base, i would assume a new deployment would overwrite it.

Keeping the mapping read only makes more sense to me. For unmapped brokers, user could define a policy that is deterministic.

mohitpali (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
```diff
@@ -214,7 +216,7 @@ protected void rebalanceForBroker(Broker broker, ClusterModel clusterModel, Set<
                                     OptimizationOptions optimizationOptions) throws OptimizationFailureException {
     LOG.debug("balancing broker {}, optimized goals = {}", broker, optimizedGoals);
 
-    String currentBrokerSetId = _brokerSetResolutionHelper.brokerSetId(broker);
+    String currentBrokerSetId = _brokerSetResolutionHelper.brokerSetId(broker.id());
```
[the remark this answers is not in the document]
CCisGG: Hi Mohit, sorry I didn't get it. With the change in this PR, we change from using broker object to broker id, not the opposite way.

As for the case that broker is null, we never seen such case before. I think in that case the goal will fail. I'm fine with that because it should success in the next run once the broker object go back to non-null.

(I checked some other places in cruise-control, and didn't see any places make the assumption that broker object can be null. So I think it should be fine?)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/NoOpBrokerSetAssignmentPolicy.java
CCisGG: Now it is still the default BrokerSetAssignmentPolicy. Why do you think we should deprecate it?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolverTest.java
CCisGG: It throws exception if there is any broker that is not assigned with broker set. Now the NoOp policy will make sure that all brokers will be assigned with at least "unmapped" broker id. So now there shouldn't be any exceptions thrown with NoOp policy.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/BrokerSetAwareGoal.java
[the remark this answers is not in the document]
mohitpali: Please Ignore my comment for broker.id()

Broker object becomes null when the broker is provisioned and is being set-up. While populating rackIds to brokers the metadata sometimes returned a broker information without rack id and id rack id is not available then broker object comes as null. I think this is an edge case and unless reproduced again, you can continue with this PR.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/NoOpBrokerSetAssignmentPolicy.java
mohitpali: assignBrokerSetsForUnresolvedBrokers method is using ClusterModel argument here. I assumed you want the new rackIdToBrokerId mapping to be used as an argument going further. Ignore if the ClusterModel argument is still going to be used.

wyuka (approved):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/NoOpBrokerSetAssignmentPolicy.java
CCisGG: Ah I see your point. The method was already deprecated at the interface level. I think it should be good enough?

PR #1895, opened 2022-08-31
Rename the getGoalsByPriority to getDefaultGoalsByPriority

CCisGG:
This PR renames the getGoalsByPriority to getDefaultGoalsByPriority

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -473,7 +473,7 @@ public final class KafkaCruiseControlUtils {
    */
   public static List<Goal> goalsByPriority(List<String> goals, KafkaCruiseControlConfig config) {
     if (goals == null || goals.isEmpty()) {
-      return AnalyzerUtils.getGoalsByPriority(config);
+      return AnalyzerUtils.getDefaultGoalsByPriority(config);
     }
     Map<String, Goal> allGoals = AnalyzerUtils.getCaseInsensitiveGoalsByName(config);
     sanityCheckNonExistingGoal(goals, allGoals);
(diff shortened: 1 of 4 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
```diff
@@ -63,12 +61,7 @@ public static Consumer<String, MaintenancePlan> createMaintenanceEventConsumer(M
    * @return A list of names for goals {@link AnomalyDetectorConfig#SELF_HEALING_GOALS_CONFIG} in the order of priority.
    */
   public static List<String> getSelfHealingGoalNames(KafkaCruiseControlConfig config) {
-    List<Goal> goals = config.getConfiguredInstances(AnomalyDetectorConfig.SELF_HEALING_GOALS_CONFIG, Goal.class);
-    List<String> selfHealingGoalNames = new ArrayList<>(goals.size());
-    for (Goal goal : goals) {
-      selfHealingGoalNames.add(goal.name());
-    }
-    return selfHealingGoalNames;
```
efeg, line 71: Is the proposed behavior change intentional?

The existing code returns the names implemented by each individual goal. It happens to be the "simple name" of the class for the existing goals -- e.g. if the goal is `com.linkedin.kafka.cruisecontrol.analyzer.goals.CpuCapacityGoal`, the response will be `CpuCapacityGoal`. That being said, a goal implementation may choose to return any other name -- e.g. `Foo`.

The proposed code returns the full name of goals. For example, if the goal is `com.linkedin.kafka.cruisecontrol.analyzer.goals.CpuCapacityGoal`, the response will be `com.linkedin.kafka.cruisecontrol.analyzer.goals.CpuCapacityGoal`.

efeg (commented):
@CCisGG  Thanks for the PR -- left some comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorUtils.java
CCisGG: This is intentional, as we don't want to init a goal unless needed.

But you are raising a good point. Let me double check if it would break functionality.

CCisGG: I think you're right. Let me remove this change

efeg (approved):
Thanks for the update! LGTM.

PR #1901, opened 2022-09-09
Fix failedBrokers.txt permissions

### LINKED ISSUE #1900: Fix permissions of the failedBrokers.txt
Cruise Control has a file called failedBrokers.txt by default. "The file path to store the failed broker list. This is to persist the broker failure time in case Cruise Control failed and restarted when some brokers are down." The file's permission will be 644 which can be restricted to 600.

egyedt:
This PR resolves #1900.


egyedt:
@viktorsomogyi, @efeg 
Could you please review this improvement?

CCisGG:
Hi @egyedt, could you further explain what's the benefit of changing the permission to 600? Or what's the risk of keeping it as 644?

egyedt:
Hi @CCisGG, thank you for your question. This is a simple security leakage since there is no reason to others than the CC be able to see the file's content. This could be vulnerable. I suggest that we should restrict it because 600 permission is enough for CC.

CCisGG (approved):

PR #1908, opened 2022-09-19
Add endpoint to move all replicas from a disk to other disks of the same broker

### LINKED ISSUE #1907: Ability to move all replicas from a disk to other disks of the same broker
Right now it is possible to move leader replicas from a list of disks but there is no way to move all replicas from a disk to other disks of the same broker. 
This would be useful when needing to remove a disk from a broker without the loss of data.

ilievladiulian:
This PR resolves #1907 .

This PR introduces a new endpoint that moves replicas from a specified disk to other disks of the same broker. This will allow the removal of the disk without the loss of data.

The replicas are moved in a round-robin manner to the remaining disks, from the largest to the smallest, while checking the following constraint:
```
1 - (remainingUsageAfterRemoval / remainingCapacity) > errorMargin
```
where
```
remainingUsageAfterRemoval = current usage for remaining disks + additional usage from removed disks
remainingCapacity = sum of capacities of the remaining disks
errorMargin = configurable property (default 0.1); it makes sure that a disk percentage is always free when moving replicas
```

## What is new
The new endpoint is available as follows:
```
POST /remove_disks?brokerid_and_logdirs=[brokerid1-logdir1,brokerid2-logdir2...]
```

## What is missing
After all replicas are moved from the specified disk, the disk may still be used by CC during rebalances and Kafka can still use it when creating topics.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
@@ -264,11 +264,10 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
    * @return {@code true} if utilization is over the limit, {@code false} otherwise.
    */
   private boolean isUtilizationOverLimit(Disk disk) {
-    boolean diskUtilizationValid = true;
-    if (_shouldEmptyZeroCapacityDisks) {
-      diskUtilizationValid = disk.utilization() > 0;
+    if (_shouldEmptyZeroCapacityDisks && disk.capacity() == 0) {
+      return disk.utilization() > 0 || disk.replicas().size() > 0;
     }
-    return diskUtilizationValid && disk.utilization() > disk.capacity() * _balancingConstraint.capacityThreshold(RESOURCE);
+    return disk.utilization() > disk.capacity() * _balancingConstraint.capacityThreshold(RESOURCE);
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
@@ -43,7 +43,7 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
   private static final Logger LOG = LoggerFactory.getLogger(IntraBrokerDiskCapacityGoal.class);
   private static final int MIN_NUM_VALID_WINDOWS = 1;
   private static final Resource RESOURCE = Resource.DISK;
-  private boolean _shouldMoveEmptyReplicas = false;
+  private boolean _shouldEmptyZeroCapacityDisks = false;
 
   /**
    * Constructor for Capacity Goal.
@@ -55,10 +55,10 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
   /**
    * Constructor for Intra Broker Disk Capacity Goal.
    *
-   * @param shouldMoveEmptyReplicas specifies if the goal should move replicas with 0 expected disk utilization
+   * @param shouldEmptyZeroCapacityDisks specifies if the goal should move all replicas from disks with 0 capacity
    */
-  public IntraBrokerDiskCapacityGoal(boolean shouldMoveEmptyReplicas) {
-      _shouldMoveEmptyReplicas = shouldMoveEmptyReplicas;
+  public IntraBrokerDiskCapacityGoal(boolean shouldEmptyZeroCapacityDisks) {
+      _shouldEmptyZeroCapacityDisks = shouldEmptyZeroCapacityDisks;
   }
 
   /**
@@ -160,7 +160,8 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
     Replica sourceReplica = clusterModel.broker(action.sourceBrokerId()).replica(action.topicPartition());
     Disk destinationDisk = clusterModel.broker(action.destinationBrokerId()).disk(action.destinationBrokerLogdir());
     boolean shouldMoveReplica = sourceReplica.load().expectedUtilizationFor(RESOURCE) > 0;
-    if (_shouldMoveEmptyReplicas) {
+    if (_shouldEmptyZeroCapacityDisks) {
+      // should also move replicas with 0 expected disk utilization
       shouldMoveReplica = sourceReplica.load().expectedUtilizationFor(RESOURCE) >= 0;
     }
     return shouldMoveReplica && isMovementAcceptableForCapacity(sourceReplica, destinationDisk);
@@ -207,7 +208,7 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
         if (d == null) {
           LOG.debug("Failed to move replica {} to any disk {} in broker {}", replica, candidateDisks, replica.broker());
         }
-        if (!isUtilizationOverLimit(disk) && !_shouldMoveEmptyReplicas) {
+        if (!isUtilizationOverLimit(disk) && !_shouldEmptyZeroCapacityDisks) {
           break;
         }
       }
@@ -234,6 +235,10 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
           throw new OptimizationFailureException(String.format("[%s] Utilization (%.2f) for disk %s on broker %d is above capacity limit.",
                                                                name(), disk.utilization(), disk, broker.id()), recommendation);
         }
+        if (_shouldEmptyZeroCapacityDisks && disk.capacity() == 0 && disk.replicas().size() > 0) {
+          throw new OptimizationFailureException(String.format("[%s] Could not move all replicas from disk %s on broker %d",
+                                                               name(), disk.logDir(), broker.id()));
+        }
       }
     }
     finish();
@@ -252,14 +257,14 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
 
   /**
    * Check whether the combined replica utilization is above the given disk capacity limits.
-   * If _shouldMoveEmptyReplicas is true, the disk utilization is over limit only if it greater than 0.
+   * If _shouldEmptyZeroCapacityDisks is true, the disk utilization is over limit only if it greater than 0.
    *
    * @param disk Disk to be checked for capacity limit violation.
    * @return {@code true} if utilization is over the limit, {@code false} otherwise.
    */
   private boolean isUtilizationOverLimit(Disk disk) {
     boolean diskUtilizationValid = true;
-    if (_shouldMoveEmptyReplicas) {
+    if (_shouldEmptyZeroCapacityDisks) {
       diskUtilizationValid = disk.utilization() > 0;
     }
     return diskUtilizationValid && disk.utilization() > disk.capacity() * _balancingConstraint.capacityThreshold(RESOURCE);
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
@@ -43,13 +43,13 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
   private static final Logger LOG = LoggerFactory.getLogger(IntraBrokerDiskCapacityGoal.class);
   private static final int MIN_NUM_VALID_WINDOWS = 1;
   private static final Resource RESOURCE = Resource.DISK;
-  private boolean _shouldEmptyZeroCapacityDisks = false;
+  private final boolean _shouldEmptyZeroCapacityDisks;
 
   /**
    * Constructor for Capacity Goal.
    */
   public IntraBrokerDiskCapacityGoal() {
-
+    _shouldEmptyZeroCapacityDisks = false;
   }
 
   /**
@@ -66,6 +66,7 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
    */
   IntraBrokerDiskCapacityGoal(BalancingConstraint constraint) {
     _balancingConstraint = constraint;
+    _shouldEmptyZeroCapacityDisks = false;
   }
 
   @Override
@@ -257,7 +258,7 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
 
   /**
    * Check whether the combined replica utilization is above the given disk capacity limits.
-   * If _shouldEmptyZeroCapacityDisks is true, the disk utilization is over limit only if it greater than 0.
+   * If _shouldEmptyZeroCapacityDisks is true, the disk utilization is over limit only if it is greater than 0.
    *
    * @param disk Disk to be checked for capacity limit violation.
    * @return {@code true} if utilization is over the limit, {@code false} otherwise.
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
@@ -43,6 +43,7 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
   private static final Logger LOG = LoggerFactory.getLogger(IntraBrokerDiskCapacityGoal.class);
   private static final int MIN_NUM_VALID_WINDOWS = 1;
   private static final Resource RESOURCE = Resource.DISK;
+  // This is only used for the remove disks endpoint
   private final boolean _shouldEmptyZeroCapacityDisks;
 
   /**
@@ -160,10 +161,13 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
   protected boolean selfSatisfied(ClusterModel clusterModel, BalancingAction action) {
     Replica sourceReplica = clusterModel.broker(action.sourceBrokerId()).replica(action.topicPartition());
     Disk destinationDisk = clusterModel.broker(action.destinationBrokerId()).disk(action.destinationBrokerLogdir());
-    boolean shouldMoveReplica = sourceReplica.load().expectedUtilizationFor(RESOURCE) > 0;
+    boolean shouldMoveReplica;
     if (_shouldEmptyZeroCapacityDisks) {
       // should also move replicas with 0 expected disk utilization
       shouldMoveReplica = sourceReplica.load().expectedUtilizationFor(RESOURCE) >= 0;
+    } else {
+      // In general CC tries to reduce number of replica move. Since empty replicas does not affect balance, CC tries to avoid moving them.
+      shouldMoveReplica = sourceReplica.load().expectedUtilizationFor(RESOURCE) > 0;
     }
     return shouldMoveReplica && isMovementAcceptableForCapacity(sourceReplica, destinationDisk);
   }
@@ -209,7 +213,7 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
         if (d == null) {
           LOG.debug("Failed to move replica {} to any disk {} in broker {}", replica, candidateDisks, replica.broker());
         }
-        if (!isUtilizationOverLimit(disk) && !_shouldEmptyZeroCapacityDisks) {
+        if (!isUtilizationOverLimit(disk)) {
           break;
         }
       }
@@ -236,7 +240,7 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
           throw new OptimizationFailureException(String.format("[%s] Utilization (%.2f) for disk %s on broker %d is above capacity limit.",
                                                                name(), disk.utilization(), disk, broker.id()), recommendation);
         }
-        if (_shouldEmptyZeroCapacityDisks && disk.capacity() == 0 && disk.replicas().size() > 0) {
+        if (_shouldEmptyZeroCapacityDisks && disk.capacity() == 0 && !disk.replicas().isEmpty()) {
           throw new OptimizationFailureException(String.format("[%s] Could not move all replicas from disk %s on broker %d",
                                                                name(), disk.logDir(), broker.id()));
         }
@@ -265,7 +269,7 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
    */
   private boolean isUtilizationOverLimit(Disk disk) {
     if (_shouldEmptyZeroCapacityDisks && disk.capacity() == 0) {
-      return disk.utilization() > 0 || disk.replicas().size() > 0;
+      return disk.utilization() > 0 || !disk.replicas().isEmpty();
     }
     return disk.utilization() > disk.capacity() * _balancingConstraint.capacityThreshold(RESOURCE);
   }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
@@ -43,12 +43,22 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
   private static final Logger LOG = LoggerFactory.getLogger(IntraBrokerDiskCapacityGoal.class);
   private static final int MIN_NUM_VALID_WINDOWS = 1;
   private static final Resource RESOURCE = Resource.DISK;
+  private boolean _shouldMoveEmptyReplicas = false;
 
   /**
    * Constructor for Capacity Goal.
    */
   public IntraBrokerDiskCapacityGoal() {
 
+  }
+
+  /**
+   * Constructor for Intra Broker Disk Capacity Goal.
+   *
+   * @param shouldMoveEmptyReplicas specifies if the goal should move replicas with 0 expected disk utilization
+   */
+  public IntraBrokerDiskCapacityGoal(boolean shouldMoveEmptyReplicas) {
+      _shouldMoveEmptyReplicas = shouldMoveEmptyReplicas;
   }
 
   /**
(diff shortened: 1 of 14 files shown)


cruise-control/src/yaml/endpoints/removeDisks.yaml
```diff
@@ -0,0 +1,69 @@
...
+        in: query
+        description: List of broker id and logdir pair to be demoted in the cluster.
+        schema:
+          type: object
+          additionalProperties:
+            type: array
+            items:
+              type: string
+        required: true
+        example: 101-/tmp/kafka-logs-1,101-/tmp/kafka-logs-2
```
CCisGG, line 34: How did you handle if the broker/logDir doesn't exist?

ilievladiulian: Hi @CCisGG!
1. The remove broker endpoint marks the broker as dead, however we do not need to mark the whole broker as dead, only some of its disks.<br/>
I saw that there is an enum for disk states which contains the DEAD value, however when generating the cluster model at the start of the runnable (`clusterModel = _kafkaCruiseControl.clusterModel(...)`) the broker disk configuration is always read from the capacity configuration file. So even if the disk is marked as dead during the remove disks call, at a following rebalance call the cluster model will be read from the capacity file and the disk will have a healthy state. Please correct me if I am wrong.<br/>
I looked into this when trying to find a way to stop further rebalances to use the removed disks (until they are specifically removed from the Kafka broker configuration), but changing how the cluster model is read or how the disks are marked as dead seemed a bit too complex and could potentially affect other functionalities.<br/><br/>
2. Indeed the broker/logDir combination is validated in `RemoveDisksRunnable.checkCanRemoveDisks`. This goal should not be executed outside of a remove disks call. To prevent being called from the rebalance endpoint at the start of the`DiskRemovalGoal.optimize` method a call to `DiskRemovalGoal.sanityCheckOptimizationOptions` is made and throws an exception if the optimization was triggered by a goal violation. If this is not the way to prevent the call from the rebalance endpoint please tell.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
```diff
@@ -0,0 +1,159 @@
...
+    private void checkCanRemoveDisks(Map<Integer, Set<String>> brokerIdAndLogdirs, ClusterModel clusterModel) {
+        for (Map.Entry<Integer, Set<String>> entry : brokerIdAndLogdirs.entrySet()) {
+            Integer brokerId = entry.getKey();
+            Set<String> logDirs = entry.getValue();
+            Broker broker = clusterModel.broker(brokerId);
+            Set<String> brokerLogDirs = broker.disks().stream().map(Disk::logDir).collect(Collectors.toSet());
+            if (!brokerLogDirs.containsAll(logDirs)) {
+                throw new IllegalArgumentException(String.format("Invalid log dirs provided for broker %d.", brokerId));
+            }
+            if (broker.disks().size() < logDirs.size()) {
```
CCisGG, line 126: Can this be possible to true given the check on line 123?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
```diff
@@ -0,0 +1,159 @@
...
+            );
+        }
+
+        return result;
+    }
+
+    private void checkCanRemoveDisks(Map<Integer, Set<String>> brokerIdAndLogdirs, ClusterModel clusterModel) {
+        for (Map.Entry<Integer, Set<String>> entry : brokerIdAndLogdirs.entrySet()) {
+            Integer brokerId = entry.getKey();
+            Set<String> logDirs = entry.getValue();
```
CCisGG, line 120: Maybe change the name to something like "logDirsToRemove"

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
```diff
@@ -0,0 +1,159 @@
...
+            double currentUsage = 0.0;
+            for (Disk disk : broker.disks()) {
+                if (logDirs.contains(disk.logDir())) {
+                    removedUsage += disk.utilization();
+                } else {
+                    remainingCapacity += disk.capacity();
+                    currentUsage += disk.utilization();
+                }
+            }
+            double futureUsage = removedUsage + currentUsage;
```
CCisGG, line 143: Looks like you don't need to record removedUsage and currentUsage separately? The only used var is futureUsage, or total usage.

CCisGG (changes requested):
Made a second pass.

I still feel it is weird to make it a goal. I'll think more about it. 

With this implementation, I think at least you need to add protection in KafkaCruiseControlConfig.sanityCheckGoalNames, to avoid this goal being triggered NOT by the endpoint. 

Also I think you need to rebase the latest 2.5 branch as it has some changes to servlet.

CCisGG:
> I looked into this when trying to find a way to stop further rebalances to use the removed disks (until they are specifically removed from the Kafka broker configuration), but changing how the cluster model is read or how the disks are marked as dead seemed a bit too complex and could potentially affect other functionalities.

This is a valid concern actually. Do you have a solution so far?

CCisGG:
> I looked into this when trying to find a way to stop further rebalances to use the removed disks (until they are specifically removed from the Kafka broker configuration), but changing how the cluster model is read or how the disks are marked as dead seemed a bit too complex and could potentially affect other functionalities.
I discussed this with our team and we have 2 suggestions:
1. You can use something similar to "recentExcludedBrokers", to avoid replicas moving back to recent excluded disks.
2. We should be very careful that the add-on implementation should not affect any users who don't use this goal/feature. E.g. we don't want to degrade performance or increase lots of memory usage for adding this feature.

kyguy:
Hey @ilievladiulian eager to see this enhancement move forward, is this still being worked on?

ZVilusinsky:
We need to move kafka data to a pvc with different class (higher IOPS) and this would really help us with that and not commit seppuku :)

ilievladiulian:
Hi @CCisGG! I pushed a new implementation for the remove disks endpoint. Instead of creating a new goal for disk removal, the endpoint reuses the IntraBrokerDiskCapacityGoal that uses intra-broker replica movement so the disk usage is kept under a certain threshold.
The remove disks endpoint sets 0 capacity on the disks specified for removal and then lets IntraBrokerDiskCapacityGoal move the replicas from those disks to the other disks of the same broker.

ilievladiulian:
> > I looked into this when trying to find a way to stop further rebalances to use the removed disks (until they are specifically removed from the Kafka broker configuration), but changing how the cluster model is read or how the disks are marked as dead seemed a bit too complex and could potentially affect other functionalities.
> 
> This is a valid concern actually. Do you have a solution so far?

I did not look into more detail into preventing CC from using the deleted disks from further rebalances, as there is still a possibility for those disks to be used by Kafka when creating new topics and this cannot be prevented.

> Reading the src code, it looks like the inter broker replica movement will move the replica to the logDir with same name of the dst broker. E.g. brokerA.logDir1 -> brokerB.logDir1. It doesn't care about whether the logDir is removed or not:
> 
> https://github.com/linkedin/cruise-control/blob/29f9185a419fffc6d443fa6d4e19c99f8e4a1f42/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java#L365
> 
> I think you probably need to come up with some solutions to address this issue.

The endpoint only uses intra-broker replica movement to move replicas from the deleted disks. It is not influenced by inter-broker replica movements.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/RemoveDisksParameters.java
```diff
@@ -0,0 +1,78 @@
...
+    static {
+        SortedSet<String> validParameterNames = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        validParameterNames.add(BROKER_ID_AND_LOGDIRS_PARAM);
+        validParameterNames.add(DRY_RUN_PARAM);
+        validParameterNames.add(REASON_PARAM);
+        validParameterNames.add(SKIP_HARD_GOAL_CHECK_PARAM);
+        validParameterNames.add(STOP_ONGOING_EXECUTION_PARAM);
+        CASE_INSENSITIVE_PARAMETER_NAMES = Collections.unmodifiableSortedSet(validParameterNames);
+    }
+    protected boolean _dryRun;
```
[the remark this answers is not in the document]
ilievladiulian: No, just followed how other parameters classes were implemented. Changed to private.

cruise-control/src/yaml/endpoints/removeDisks.yaml
```diff
@@ -0,0 +1,69 @@
+RemoveDisksEndpoint:
+  post:
+    operationId: removeDisks
+    summary: Move all replicas from the specified disk to other disks of the same broker.
+    parameters:
+      - name: dryrun
```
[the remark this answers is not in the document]
ilievladiulian: Removed unnecessary parameter.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.Set;
+import java.util.Map;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.ArrayList;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+
+import static com.linkedin.kafka.cruisecontrol.config.constants.AnalyzerConfig.REMOVE_DISKS_REMAINING_SIZE_ERROR_MARGIN;
+import static com.linkedin.kafka.cruisecontrol.servlet.handler.async.runnable.RunnableUtils.*;
```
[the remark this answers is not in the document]
ilievladiulian: 👍

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
```diff
@@ -0,0 +1,159 @@
...
+    protected final Map<Integer, Set<String>> _brokerIdAndLogdirs;
+
+    protected final double _errorMargin;
+
+    public RemoveDisksRunnable(KafkaCruiseControl kafkaCruiseControl,
+                               OperationFuture future,
+                               RemoveDisksParameters parameters,
+                               String uuid) {
+        super(kafkaCruiseControl, future, parameters, parameters.dryRun(), parameters.stopOngoingExecution(), parameters.skipHardGoalCheck(),
+                uuid, parameters::reason);
```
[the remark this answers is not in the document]
ilievladiulian: The super constructor expects a Supplier<String> there. This seems to be present in all other runnables.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
```diff
@@ -0,0 +1,159 @@
...
+    @Override
+    protected OptimizationResult getResult() throws Exception {
+        return new OptimizationResult(computeResult(), _kafkaCruiseControl.config());
+    }
+
+    @Override
+    protected void init() {
+        _kafkaCruiseControl.sanityCheckDryRun(_dryRun, _stopOngoingExecution);
+        _goalsByPriority = new ArrayList<>(1);
+        _goalsByPriority.add(new DiskRemovalGoal(_brokerIdAndLogdirs, _errorMargin));
```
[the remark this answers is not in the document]
ilievladiulian: It is the only goal added, so it also has the highest priority.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
```diff
@@ -0,0 +1,159 @@
...
+                _fastMode
+        );
+
+        OptimizerResult result = _kafkaCruiseControl.optimizations(clusterModel, _goalsByPriority, _operationProgress, null, optimizationOptions);
+        if (!_dryRun) {
+            _kafkaCruiseControl.executeProposals(
+                    result.goalProposals(),
+                    Collections.emptySet(),
+                    isKafkaAssignerMode(_goals),
+                    0,
```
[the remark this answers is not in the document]
ilievladiulian: 👍

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
ilievladiulian: Good catch. Removed if branch.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
ilievladiulian: 👍

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
ilievladiulian: Good catch. Used only futureUsage.

ilievladiulian:
> > I looked into this when trying to find a way to stop further rebalances to use the removed disks (until they are specifically removed from the Kafka broker configuration), but changing how the cluster model is read or how the disks are marked as dead seemed a bit too complex and could potentially affect other functionalities.
> > I discussed this with our team and we have 2 suggestions:
> 
> 1. You can use something similar to "recentExcludedBrokers", to avoid replicas moving back to recent excluded disks.
> 2. We should be very careful that the add-on implementation should not affect any users who don't use this goal/feature. E.g. we don't want to degrade performance or increase lots of memory usage for adding this feature.

As stated in a previous comment, even if we prevent CC from using those disks, Kafka can still use them and this can't be prevented. If you think it still has value to prevent CC from using them, let me know and I will look into it with your suggestions in mind.

ilievladiulian:
@alex-necula Added exception if goal could not move all replicas from deleted disks, so that the task ends with the CompletedWithError status.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/DiskRemovalGoal.java
```diff
@@ -0,0 +1,200 @@
...
+        int diskIndex = (usedDiskIdx + 1) % remainingDisksNumber;
+        for (int i = 0; i < remainingDisksNumber; i++) {
+            Disk destinationDisk = remainingDisks.get(diskIndex);
+            if (isEnoughSpace(destinationDisk, replica)) {
+                clusterModel.relocateReplica(replica.topicPartition(), brokerId, destinationDisk.logDir());
+                return diskIndex;
+            }
+            diskIndex = (diskIndex + 1) % remainingDisksNumber;
+        }
+        LOG.info("Could not move replica {} to any of the remaining disks.", replica);
```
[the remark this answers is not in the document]
ilievladiulian: Addressed in commit ca18da1.

ilievladiulian:
Hi @CCisGG! Did you get a chance to look at the latest changes?

CCisGG:
Hi folks, I'll take a pass next week.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
```diff
@@ -43,19 +43,30 @@ public class IntraBrokerDiskCapacityGoal extends AbstractGoal {
   private static final Logger LOG = LoggerFactory.getLogger(IntraBrokerDiskCapacityGoal.class);
   private static final int MIN_NUM_VALID_WINDOWS = 1;
   private static final Resource RESOURCE = Resource.DISK;
+  private final boolean _shouldEmptyZeroCapacityDisks;
```
CCisGG, line 46: Just curious. Even if we don't add this boolean, shouldn't cruise control move replicas out if you set disk capacity to 0?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
```diff
@@ -239,12 +258,17 @@ public ModelCompletenessRequirements clusterModelCompletenessRequirements() {
...
   /**
    * Check whether the combined replica utilization is above the given disk capacity limits.
+   * If _shouldEmptyZeroCapacityDisks is true, the disk utilization is over limit only if it is greater than 0.
    *
    * @param disk Disk to be checked for capacity limit violation.
    * @return {@code true} if utilization is over the limit, {@code false} otherwise.
    */
   private boolean isUtilizationOverLimit(Disk disk) {
-    return disk.utilization() > disk.capacity() * _balancingConstraint.capacityThreshold(RESOURCE);
+    boolean diskUtilizationValid = true;
```
CCisGG, line 267: I see, so you do not want to trigger the movement if the disk utilization is already 0. I'm curious why do we add the logic here? Say if a disk utilization is already 0, what happens if we return true by this method?

CCisGG (changes requested):
Overall I think this new implementation is smart and should work. Left some minor comment. Also please add necessary unit test as part of this PR. Thanks for the contribution!

Aaron-ML:
Is this still being worked on? @ilievladiulian 

This is increasingly common for us with disk limitations in cloud providers causing us to have to rotate disks for faster storage tiers. This would be a big win for us!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
ilievladiulian: If the disk utilisation is 0 and the capacity is 0, it would not be considered over utilised, although there may be empty replicas. This broke the loop at line 213. I removed the 0 utilisation case and did an extra check on the flag in the if statement at line 212. This was under the assumption that if there is a replica on that disk, the utilisation will be greater than 0.

However, while writing the tests I saw that this assumption was wrong. There could be a 0 utilisation on the disk for empty replicas, at least in the tests environment. I fixed the formula in the latest commit.

ilievladiulian:
Hi @CCisGG! Thank you for the review. I added unit tests in the latest commit and answered the comments.

ilievladiulian:
Hi @CCisGG! Did you get a chance to look at the latest changes?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
```diff
@@ -149,8 +160,12 @@ public ActionAcceptance actionAcceptance(BalancingAction action, ClusterModel cl
   protected boolean selfSatisfied(ClusterModel clusterModel, BalancingAction action) {
     Replica sourceReplica = clusterModel.broker(action.sourceBrokerId()).replica(action.topicPartition());
     Disk destinationDisk = clusterModel.broker(action.destinationBrokerId()).disk(action.destinationBrokerLogdir());
-    return sourceReplica.load().expectedUtilizationFor(RESOURCE) > 0
-           && isMovementAcceptableForCapacity(sourceReplica, destinationDisk);
+    boolean shouldMoveReplica = sourceReplica.load().expectedUtilizationFor(RESOURCE) > 0;
```
CCisGG, line 163: I'd recommend to change this to:

```
    boolean shouldMoveReplica;
    if (_shouldEmptyZeroCapacityDisks) {
      // should also move replicas with 0 expected disk utilization
      shouldMoveReplica = sourceReplica.load().expectedUtilizationFor(RESOURCE) >= 0;
    } else {
     // In general CC tries to reduce number of replica move. Since empty replicas does not affect balance, CC tries to avoid moving them. 
      shouldMoveReplica = sourceReplica.load().expectedUtilizationFor(RESOURCE) > 0;
   }

```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
```diff
@@ -194,7 +209,7 @@ public int compare(Disk disk1, Disk disk2) {
         if (d == null) {
           LOG.debug("Failed to move replica {} to any disk {} in broker {}", replica, candidateDisks, replica.broker());
         }
-        if (!isUtilizationOverLimit(disk)) {
+        if (!isUtilizationOverLimit(disk) && !_shouldEmptyZeroCapacityDisks) {
```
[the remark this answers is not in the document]
CCisGG: Ok after rethink about it, it seems to just match the change in L163..

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
```diff
@@ -0,0 +1,168 @@
...
+        return new OptimizationResult(computeResult(), _kafkaCruiseControl.config());
+    }
+
+    @Override
+    protected void init() {
+        _kafkaCruiseControl.sanityCheckDryRun(_dryRun, _stopOngoingExecution);
+        _goalsByPriority = new ArrayList<>(1);
+        Goal intraBrokerDiskCapacityGoal = new IntraBrokerDiskCapacityGoal(true);
+        intraBrokerDiskCapacityGoal.configure(_kafkaCruiseControl.config().mergedConfigValues());
+        _goalsByPriority.add(intraBrokerDiskCapacityGoal);
```
CCisGG, line 67: I think we can do something like this instead: 

`_goalsByPriority = Collections.singletonList(intraBrokerDiskCapacityGoal);`

CCisGG (changes requested):
Left more comments.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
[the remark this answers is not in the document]
ilievladiulian: Good note. Added the comment in the latest commit.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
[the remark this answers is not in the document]
ilievladiulian: Good catch. After fixing the `isUtilizationOverLimit` method, this is no longer needed. Removed it in the latest commit.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/IntraBrokerDiskCapacityGoal.java
ilievladiulian: Good note. It looks more clear this way. Refactored in latest commit.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
ilievladiulian: Good note. One observation, in `GoalBasedOperationRunnable` after the goal is completed, the `_goalsByPriority` list is cleared. This does not work with `Collections.singletonList` as it provides an immutable list. In the latest commit, I used `Collections.singletonList` but wrapped in an `ArrayList` constructor call so the list would be mutable.

ilievladiulian:
Hi @CCisGG! Addressed your comments in the latest commit. I also added some logging in the `RemoveDisksRunnable` class. Can you take another look?

CCisGG (approved):
Looks good. Thanks for the great work! Could you please run some ad-hoc tests on state/cluster_state and on the new endpoint that you just added?

Feel free to let me know if the tests run well and I can merge the PR.

ilievladiulian:
Hi @CCisGG! Thank you for the time to review this PR.
I setup a cluster with 3 brokers, the first one having 2 disks. I ran a disk removal request for the first disk and every replica on the first disk was moved to the second disk.

Before the cluster state was:
image

The disk removal request returned:
image
image

After operation ended the cluster state is:
image

Running the `kafka-log-dirs.sh` tool for the first broker shows that there are no more replicas on the first disk.

sappusaketh:
We are all waiting for this feature is there a workaround  which could help us move data between disks and remove it safely?

CCisGG:
@sappusaketh Releasing 2.5.127 with this change.

CCisGG:
@ilievladiulian The CI workflow fails. Could you take a look at the issue? https://app.circleci.com/pipelines/github/linkedin/cruise-control/2259/workflows/ce2fdd75-615f-4784-9cb9-7b6d47fdb1d2/jobs/6294

```
[Error] /home/circleci/.gradle/workers/Picked up _JAVA_OPTIONS:-Xms512m -Xmx1g
/home/circleci/workspace/**************/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java:109:  error: method executeProposals in class KafkaCruiseControl cannot be applied to given types;
javac exited with exit code 1

> Task :**************:compileScala FAILED
```

Seems like there is a compile issue.

CCisGG:
Hi @ilievladiulian, any updates on this one? It's blocking the PR/release pipeline.

ilievladiulian:
Hi @CCisGG! Sorry for the delay. From what I can see, there was a change in the API for `KafkaCruiseControl.executeProposals`. The fix should be straightforward, will take a look at it tomorrow and create a new PR with the fix.

ilievladiulian:
Hi @CCisGG! Created [PR-2066](https://github.com/linkedin/cruise-control/pull/2066) that fixes the build.

ShubhamRwt:
Hi @ilievladiulian, I was curious and wanted to know, is there some reason we removed the `verbose` parameter from this endpoint? I mean not having the `verbose` would stop us from having the information about the `loadBeforeOptimization` and then we will not be able to compare it with `loadAfterOptimization`

PR #1910, opened 2022-09-21
Future-proof for Kafka 3.3 for KafkaYammerMetrics class name changes

### LINKED ISSUE #1903: Future-proofing Cruise Control for Kafka 3.3 for KafkaYammerMetrics class name changes
With Kafka 3.3, the KafkaYammerMetrics class name changes:

- KafkaYammerMetrics class in Kafka 3.2-: `kafka.metrics.KafkaYammerMetrics`
- KafkaYammerMetrics class in Kafka 3.3+: `org.apache.kafka.server.metrics.KafkaYammerMetrics` 

To future proof against `ClassNotFoundException` errors and/or stale metric registry usage, we should add some logic here [1]  to determine which class name to use based on what Kafka client version is in use! 

Let me know what you think! If it sounds like a reasonable request, I would be happy to contribute this fix!

[1] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java#L212-L221

kyguy:
This PR resolves #1903.

### DIFF
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -223,11 +223,11 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
    */
   private static MetricsRegistry metricsRegistry() {
     Object metricsRegistry;
-    Class<?> metrics;
+    Class<?> metricsClass;
 
     try {
       // First we try to get the KafkaYammerMetrics class for Kafka 3.3+
-      metrics = Class.forName(YAMMER_METRICS_IN_KAFKA_3_3_AND_LATER);
+      metricsClass = Class.forName(YAMMER_METRICS_IN_KAFKA_3_3_AND_LATER);
       LOG.info("Found class {} for Kafka 3.3 and newer.", YAMMER_METRICS_IN_KAFKA_3_3_AND_LATER);
     } catch (ClassNotFoundException e) {
       LOG.info("Class {} not found. We are probably on Kafka 3.2 or older.", YAMMER_METRICS_IN_KAFKA_3_3_AND_LATER);
@@ -235,7 +235,7 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
       // We did not find the KafkaYammerMetrics class from Kafka 3.3+. So we are probably on older Kafka version
       //     => we will try the older class for Kafka 2.6+.
       try {
-        metrics = Class.forName(YAMMER_METRICS_IN_KAFKA_2_6_AND_LATER);
+        metricsClass = Class.forName(YAMMER_METRICS_IN_KAFKA_2_6_AND_LATER);
         LOG.info("Found class {} for Kafka 2.6 and newer.", YAMMER_METRICS_IN_KAFKA_2_6_AND_LATER);
       } catch (ClassNotFoundException ee) {
         LOG.info("Class {} not found. We are probably on Kafka 2.5 or older.", YAMMER_METRICS_IN_KAFKA_2_6_AND_LATER);
@@ -243,7 +243,7 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
         // We did not find the KafkaYammerMetrics class from Kafka 2.6+. So we are probably on older Kafka version
         //     => we will try the older class for Kafka 2.5-.
         try {
-          metrics = Class.forName(YAMMER_METRICS_IN_KAFKA_2_5_AND_EARLIER);
+          metricsClass = Class.forName(YAMMER_METRICS_IN_KAFKA_2_5_AND_EARLIER);
           LOG.info("Found class {} for Kafka 2.5 and earlier.", YAMMER_METRICS_IN_KAFKA_2_5_AND_EARLIER);
         } catch (ClassNotFoundException eee) {
           // No class was found for any Kafka version => we should fail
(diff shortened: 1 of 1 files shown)


kyguy:
@CCisGG any thoughts on this?

CCisGG:
@kyguy thanks for creating the issue and this PR!

Upon reading through the issue/code, I feel it might be better to make the change when we bump Kafka version.

The reason I'm not in favor of this PR, is it introduces extra logic and make the code less readable, and more error-prone. 

I can understand that one risk is, when someone bump the kafka version to kafka 3.3, they may forget to make this change. Maybe we can simply add a test, and to make the test fail if someone forgot to update the class name. What do you think?

kyguy:
Hey @CCisGG thanks for the speedy response, I understand the concern

A test to verify the correct metric registry class name is being used sounds reasonable to me! I could adjust the PR to do just that.

 This would mean we would need a `migrate_to_kafka_3_3` when Cruise Control supports Kafka 3.3 correct?

CCisGG:
Hi @kyguy , I think our goal is to have a backwards-compatible CC from now on. It basically means we will always make sure the new cruise control releases should support all backward kafka versions. That said, ultimately we will probably just have a main branch and get rid of those naming conventions. 

For forward compatibility of Kafka, we'd also like to have it, and cruise control is probably already having it. The reasons we don't bump the kafka version for now, are:

- We don't see any compatibility issues for current CC version and kafka 3.3. (Please let me know if you found any such issues)
- Internally we are still using Kafka 3.0.* version.

We will try to keep Kafka version sync with LinkedIn internal kafka version.

scholzj:
> We don't see any compatibility issues for current CC version and kafka 3.3. (Please let me know if you found any such issues)

I was trying it with 2.5.100 and it did not seem compatible with Kafka 3.3.0 clusters. The YammerMetrics class was moved from `kafka.metrics.KafkaYammerMetrics` to `org.apache.kafka.server.metrics.KafkaYammerMetrics` in Kafka 3.3.0. The metrics reporter which runs inside Kafka broker does not seem to find the right metrics with Kafka 3.3.0 brokers and does not work properly. I haven't tried the code in this PR, but I think it should add compatibility with Kafka 3.3.0 but keep it also backwards compatible.

kyguy:
>  I think our goal is to have a backwards-compatible CC from now on. It basically means we will always make sure the new cruise control releases should support all backward kafka versions.

From what I understand, if we are planning on having CC be backwards-compatible and converge on a `main` branch we would need a change like this to address the issues that arise from the YammerMetrics class name changes that come with the Kafka 3.3 release

Otherwise we would need a `migrate_to_kafka_3_3` branch

CCisGG:
OK I see your points. That said I think it makes sense to have this change in the main branch. Let me review the change here.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -201,22 +207,62 @@ public void configure(Map<String, ?> configs) {
...
+   * The older default registries do not work with the newer versions of Kafka. Therefore, if the new class exists, we use it and if
+   * it doesn't exist we will fall back on the older ones.
    *
    * Once CC supports only 2.6.0 and newer, we can clean this up and use only KafkaYammerMetrics all the time.
    *
    * @return  MetricsRegistry with Kafka metrics
    */
   private static MetricsRegistry metricsRegistry() {
+    Object metricsRegistry;
+    Class<?> metrics;
```
CCisGG, line 226: nit: I think "metricsClass" would be a better name here.

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -201,22 +207,62 @@ public void configure(Map<String, ?> configs) {
...
+        LOG.info("Class {} not found. We are probably on Kafka 2.5 or older.", YAMMER_METRICS_IN_KAFKA_2_6_AND_LATER);
+
+        // We did not find the KafkaYammerMetrics class from Kafka 2.6+. So we are probably on older Kafka version
+        //     => we will try the older class for Kafka 2.5-.
+        try {
+          metrics = Class.forName(YAMMER_METRICS_IN_KAFKA_2_5_AND_EARLIER);
+          LOG.info("Found class {} for Kafka 2.5 and earlier.", YAMMER_METRICS_IN_KAFKA_2_5_AND_EARLIER);
+        } catch (ClassNotFoundException eee) {
+          // No class was found for any Kafka version => we should fail
+          throw new RuntimeException("Failed to find Yammer Metrics class", eee);
```
CCisGG, line 250: Is there a reason that we throws exception instead of returning Metrics.defaultRegistry()？ I'm asking because I see the code was returning Metrics.defaultRegistry() before.

kyguy: We do return Metrics.defaultRegistry() by the end of the method if the class and method exists, this makes it consistent with how we would check the defaultRegistry() methods of the other classes

CCisGG: Ah I see. Thanks for the explanation.

CCisGG (approved):

PR #1914, opened 2022-09-27
Fix leader replica cpu util when leader egress is zero

### LINKED ISSUE #1913: CPU Usage shown incorrectly 0.00% - No self healing triggered
## Problem

We did a test where we induced high CPU Usage on a broker by doing Produce only traffic. There was no consume load.

#### Actual CPU Usage metrics :

Screen Shot 2022-09-26 at 1 59 22 PM

#### What Prometheus reports :
```
 {
    "metric": {
      "instance": "b-18.XXXXX.us-east-1.amazonaws.com:11002"
    },
    "value": [
      1663883747.106,
      "0.7534427083333867"
    ]
  }
  ```
#### Additional Logs (number of sample windows is 5)

```
INFO reource : cpu, broker : 18, utilization : 0.0 (com.linkedin.kafka.cruisecontrol.model.ClusterModelStats)

INFO reource : cpu, broker : 18, load : Load[metricValues={metricId:"0", values:"{avg:0.0, max:0.0, {0:0.0, 1:0.0, 2:0.0, 3:0.0, 4:0.0}"
```

Screen Shot 2022-09-26 at 4 53 40 PM

This means that CPU Usage on the stats screen is being shown as 0% and there are no replica movements.

### Proposal

We need to change this code here - https://github.com/linkedin/cruise-control/blob/b4e44ec004e6f5e22bd1c4e203d92341ed9e1659/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ModelUtils.java#L104 to use an && instead of an || and calculate leader CPU load contribution even if there is no leader out traffic.

### PR DESCRIPTION
**Pull request by mohitpali**
This PR resolves #1913.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ModelUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ModelUtils.java
@@ -101,7 +101,7 @@ public final class ModelUtils {
     if (useLinearRegressionModel) {
       return estimateLeaderCpuUtilUsingLinearRegressionModel(partitionBytesInRate, partitionBytesOutRate);
     } else {
-      if (brokerLeaderBytesInRate == 0 || brokerLeaderBytesOutRate == 0) {
+      if (brokerLeaderBytesInRate == 0 && brokerLeaderBytesOutRate == 0) {
         return 0.0;
       } else if (brokerLeaderBytesInRate * ALLOWED_METRIC_ERROR_FACTOR < partitionBytesInRate
                  && brokerLeaderBytesInRate > UNSTABLE_METRIC_THROUGHPUT_THRESHOLD) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by CCisGG — COMMENTED** [review]
Changes in this PR look good to me. Just curious is there any existing test on this? And if not, could you help to add some tests for it? Thanks!

File: cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ModelUtils.java [hunk 1]
```diff
@@ -101,7 +101,7 @@ public static Double estimateLeaderCpuUtilPerCore(double brokerCpuUtil,
     if (useLinearRegressionModel) {
       return estimateLeaderCpuUtilUsingLinearRegressionModel(partitionBytesInRate, partitionBytesOutRate);
     } else {
-      if (brokerLeaderBytesInRate == 0 || brokerLeaderBytesOutRate == 0) {
+      if (brokerLeaderBytesInRate == 0 && brokerLeaderBytesOutRate == 0) {
```
CCisGG, line 104: I think we can also add comments in the javadoc of this method, to clarify this bytesOut does not include replication traffic.

mohitpali:
> Changes in this PR look good to me. Just curious is there any existing test on this? And if not, could you help to add some tests for it? Thanks!

Yes, let me add some tests to this.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/ModelUtils.java
mohitpali: Will do, thanks

CCisGG (approved):

PR #1919, opened 2022-10-04
Add fixability metrics to GoalOptimizer

### LINKED ISSUE #1918: Metric AnomalyDetector.UNDER_PROVISIONED not accurate with BrokerSetAwareGoal
### Problem

``BrokerSetAwareGoal`` restricts movements within the boundary of a broker set. This means that if a resource Capacity Goal e.g. ``DiskCapacityGoal`` is violated for brokers within a broker set in a manner where there is no disk capacity on the brokers in a broker set but there is capacity available in another broker set.

Lets say the disk capacity threshold is 70%
```
BS1 = B1(75% disk used) + B2(75% disk used)
BS2 = B3(10% disk used) + B4(10% disk used)
```

B1 and B2 both have ``DiskCapacityGoal`` violated. ``GoalViolationDetector`` will detect ``DiskCapacityGoal`` as violated. However, to ``GoalViolationDetector``, the ``DiskCapacityGoal`` is a **fixable** violation. Since for ``DiskCapacityGoal``, the capacity is available on other Brokers within the cluster.

However, for ``GoalOptimizer``, the optimization will fail since the ``BrokerSetAwareGoal`` along with ``DiskCapacityGoal`` will not be able to generate a proposal. This Optimizer behavior is correct and expected.

Since the ``ProvisionStatus`` metrics is published from ``GoalViolationDetector``, the ``UNDER_PROVISIONED`` status is not reported by AnomalyDetector sensor.

Screen Shot 2022-10-04 at 2 42 10 PM
Screen Shot 2022-10-04 at 2 43 03 PM

```
INFO Failed to generate proposals for execution (UUID: eccbe114-5be4-48b0-9a82-2961354a04f0 reason: Self healing for GOAL_VIOLATION: {Unfixable goal violations: {}, Fixable goal violations: {DiskCapacityGoal}, Exclude brokers recently (removed: true demoted: true), Provision: UNDECIDED}). (com.linkedin.kafka.cruisecontrol.executor.Executor)

com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException: [DiskCapacityGoal] Utilization (782404.13) of broker 7 violated capacity limit (614400.00) for resource disk. Add at least 1 broker for disk. || Tips: [1] There are 5 topics excluded from replica move. Potential mitigation: Remove selected topics from exclusion using excluded_topics parameter. Then, re-run your original request. Add at least 1 broker for disk. (operationLogger)

WARN Encountered optimization failure when trying to fix the anomaly {Unfixable goal violations: {}, Fixable goal violations: {DiskCapacityGoal}, Exclude brokers recently (removed: true demoted: true), Provision: UNDECIDED}. (com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorManager)

```

## Why is it relevant ?

The UNDER_PROVISIONED metrics is crucial to understand if a BrokerSet is out of capacity and requires broker addition/provisioning in the BrokerSet. (Since the movements can not happen outside the BrokerSet with BrokerSetAwareGoal.)

## Proposal

We should publish the provision status metrics from Goal Optimizer as well. I would like to add the ProvisionStatus metrics to be published from GoalOptimizer class as well under the existing sensor dimension - ``GoalOptimizer``.

These will be additional metrics and will not impact the existing ProvisionStatus metrics under ``AnomalyDetector`` sensor.

Please let me know what do you think ! @efeg @CCisGG 

Thanks, Mohit

mohitpali:
This PR resolves #1918.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
@@ -4,6 +4,7 @@
 
 package com.linkedin.kafka.cruisecontrol.analyzer;
 
+import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.linkedin.kafka.cruisecontrol.common.Utils;
@@ -88,6 +89,7 @@ public class GoalOptimizer implements Runnable {
   private final double _priorityWeight;
   private final double _strictnessWeight;
   private final OptimizationOptionsGenerator _optimizationOptionsGenerator;
+  private volatile ProvisionResponse _provisionResponse;
 
   /**
    * Constructor for Goal Optimizer takes the goals as input. The order of the list determines the priority of goals
@@ -124,6 +126,17 @@ public class GoalOptimizer implements Runnable {
     _proposalGenerationException = new AtomicReference<>();
     _proposalPrecomputingProgress = new OperationProgress();
     _proposalComputationTimer = dropwizardMetricRegistry.timer(MetricRegistry.name(GOAL_OPTIMIZER_SENSOR, "proposal-computation-timer"));
+
+    // The cluster is identified as under-provisioned, over-provisioned, or right-sized (undecided not reported).
+    dropwizardMetricRegistry.register(MetricRegistry.name(GOAL_OPTIMIZER_SENSOR, "under-provisioned"),
+                                      (Gauge<Integer>) () -> (provisionStatus() == ProvisionStatus.UNDER_PROVISIONED)
+                                                             ? 1 : 0);
+    dropwizardMetricRegistry.register(MetricRegistry.name(GOAL_OPTIMIZER_SENSOR, "over-provisioned"),
+                                      (Gauge<Integer>) () -> (provisionStatus() == ProvisionStatus.OVER_PROVISIONED)
+                                                             ? 1 : 0);
+    dropwizardMetricRegistry.register(MetricRegistry.name(GOAL_OPTIMIZER_SENSOR, "right-sized"),
+                                      (Gauge<Integer>) () -> (provisionStatus() == ProvisionStatus.RIGHT_SIZED)
+                                                             ? 1 : 0);
     _executor = executor;
     _hasOngoingExplicitPrecomputation = false;
     _priorityWeight = config.getDouble(AnalyzerConfig.GOAL_BALANCEDNESS_PRIORITY_WEIGHT_CONFIG);
@@ -133,6 +146,13 @@ public class GoalOptimizer implements Runnable {
     _optimizationOptionsGenerator = config.getConfiguredInstance(AnalyzerConfig.OPTIMIZATION_OPTIONS_GENERATOR_CLASS_CONFIG,
                                                                  OptimizationOptionsGenerator.class,
                                                                  overrideConfigs);
+  }
+
+  /**
+   * @return Provision status of the cluster based on the latest goal violation check.
+   */
+  private ProvisionStatus provisionStatus() {
+    return _provisionResponse.status();
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -448,7 +468,13 @@ public OptimizerResult optimizations(ClusterModel clusterModel,
       operationProgress.addStep(step);
       LOG.debug("Optimizing goal {}", goal.name());
       long startTimeMs = _time.milliseconds();
-      boolean succeeded = goal.optimize(clusterModel, optimizedGoals, optimizationOptions);
+      boolean succeeded;
+      try {
+        succeeded = goal.optimize(clusterModel, optimizedGoals, optimizationOptions);
+      } finally {
+        provisionResponse.aggregate(goal.provisionResponse());
```
CCisGG, line 475: Could you briefly explain why this change is needed, and what's the provisionResponse when the try fails?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -124,6 +126,17 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
     _proposalGenerationException = new AtomicReference<>();
     _proposalPrecomputingProgress = new OperationProgress();
     _proposalComputationTimer = dropwizardMetricRegistry.timer(MetricRegistry.name(GOAL_OPTIMIZER_SENSOR, "proposal-computation-timer"));
+
+    // The cluster is identified as under-provisioned, over-provisioned, or right-sized (undecided not reported).
+    dropwizardMetricRegistry.register(MetricRegistry.name(GOAL_OPTIMIZER_SENSOR, "under-provisioned"),
```
CCisGG, line 131: It looks like you're creating new metrics rather than share the same metric with the ones in AnomalyDetectorManager. I think we should use the same metric and emit values from both places.

CCisGG (changes requested):

mohitpali:
Screen Shot 2022-11-07 at 10 36 54 AM


I validated with replica capacity exceeding and disk capacity exceeding both and after topics are removed, the mnetrics gets cleared automatically.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
mohitpali: I have modified this in the next commit. Let me know if you have any more questions.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
mohitpali: Name is modified in new commit.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -135,6 +142,13 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
                                                                  overrideConfigs);
   }
 
+  /**
+   * @return if we have unfixable proposal optimization
+   */
+  private boolean hasUnfixableProposalOptimization() {
+    return _hasUnfixableProposalOptimization;
```
CCisGG, line 149: why do we need such private accessor for the _hasUnfixableProposalOptimization?

CCisGG (approved):

CCisGG:
> One more question: I'm not sure it would help much for your cases, unless you set the self-healing goals exactly same as goals.
> 
> So the problem here is: CC requires self-healing goals to be a subset of goals. Thus when self-healing for GoalViolations starts, it only adds self-healing goals to the RebalanceRunnable. Since self-healing goals is usually a proper subset of goals, it will not trigger the metric change. Did I miss something here?

Nvm now I can read it. So you only set the value when the goals is a super set of default goals. That makes sense.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -135,6 +142,13 @@ public GoalOptimizer(KafkaCruiseControlConfig config,
                                                                  overrideConfigs);
   }
 
+  /**
+   * @return if we have unfixable proposal optimization
```
CCisGG, line 146: nit: return true if we have unfixable proposal optimzation.

But I think we probably don't need this method anyways.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
mohitpali: i have seen sensors access the variables through a function in Cruise Control package. Hence i followed the same standard. Also in the past i have tried referencing the variable and it did not work well. I dont know the exact reason why that is the case though.

mohitpali:
Can this be merge ?

CCisGG:
Hi @mohitpali , I re-triggered the failed test. Will merge once the test passes.

Issue #1905, opened 2022-09-15
Compare exact member of ISR with RS when deciding whether the partition is  URP


CCisGG:
Today, the way we determine a partition is an URP (Under Replicated Partition) based on the size of ISR and RS. E.g. if ISR = [A, B], and RS = [A, B, C], then the partition is treated as URP by CC.

However, in certain cases (usually because kafka is in a bad/inconsistent state), the ISR can be diverged from RS. E.g. ISR = [A, B, B], while RS = [D, E, F]. In this case, we still want CC to report this replica as URP, since the partition is not in a good state.

The proposal is, instead of comparing the size of ISR and RS, we should comparing the members in ISR and RS. If ISR != RS, CC should report it as URP.

suyashtava:
Hi, @CCisGG , may I know if someone is working on it, if not? Can I pick this up

suyashtava:
Btw, I am interested to know, when ISR = [A, B, B], while RS = [D, E, F] This condition will happen?

CCisGG:
@suyashtava Sure! Please feel free to take it!

It happened recently in our kafka cluster, where there was a weird bug in ZooKeeper that different ZK host has different Data. Since Kafka relies on ZooKeeper, it messed up the replica set when it reads data from zookeeper. This should rarely happens, but just in case it happens, we should make this change.

AnvarKhatik:
Hey @suyashtava You're Still Working On This Issue

suyashtava:
Yup, yup. I was on Call last week, didn’t get a chance. Will fix it this
week.

Though I am trying to replicate the issue, but not able to replicate @CCisGG

CCisGG:
@suyashtava this is a rare case and it's hard to reproduce. I think it's ok to just change the logic and add some unit test for it.

suyashtava:
Raised a PR: https://github.com/linkedin/cruise-control/pull/1923

PR #1923, opened 2022-10-09
Compare exact member of ISR with RS when deciding whether the partition is URP

suyashtava:
This PR resolves [#1905 ](https://github.com/linkedin/cruise-control/issues/1905)

Today, the way we determine a partition is an URP (Under Replicated Partition) based on the size of ISR and RS. E.g. if ISR = [A, B], and RS = [A, B, C], then the partition is treated as URP by CC.

However, in certain cases (usually because kafka is in a bad/inconsistent state), the ISR can be diverged from RS. E.g. ISR = [A, B, B], while RS = [D, E, F]. In this case, we still want CC to report this replica as URP, since the partition is not in a good state.

The proposal is, instead of comparing the size of ISR and RS, we should comparing the members in ISR and RS. If ISR != RS, CC should report it as URP.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
@@ -15,10 +15,12 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
+import java.util.Arrays;
 import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.config.TopicConfig;
 import org.apache.kafka.common.Node;
+
 
 
 @JsonResponseClass
@@ -118,7 +120,7 @@ public class ClusterPartitionState {
           int numReplicas = partitionInfo.replicas().length;
           Node[] replicas = Arrays.copyOf(partitionInfo.replicas(), numReplicas);
           Arrays.sort(replicas);
-          boolean isURP = !(replicas == inSyncReplicas);
+          boolean isURP = !(Arrays.equals(replicas, inSyncReplicas));
           if (numInsyncReplicas < minInsyncReplicas) {
             underMinIsrPartitions.add(partitionInfo);
           }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
@@ -112,7 +112,7 @@ public class ClusterPartitionState {
         }
         for (PartitionInfo partitionInfo : _kafkaCluster.partitionsForTopic(topic)) {
           int numInsyncReplicas = partitionInfo.inSyncReplicas().length;
-          boolean isURP = numInsyncReplicas != partitionInfo.replicas().length;
+          boolean isURP = !(partitionInfo.inSyncReplicas() == partitionInfo.replicas());
           if (numInsyncReplicas < minInsyncReplicas) {
             underMinIsrPartitions.add(partitionInfo);
           }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
@@ -86,6 +86,24 @@ public class ClusterPartitionState {
     }
     return jsonMap;
   }
+  
+  /**
+   * Check the Kafka Partition State if it is Under Replicated or not.
+   * 
+   * @param partitionInfo the {@link PartitionInfo} object that contains partition info.
+   * @return the effective boolean under replicated status of a partition.
+   */
+  protected boolean isURP(PartitionInfo partitionInfo)
+  {
+    int numInsyncReplicas = partitionInfo.inSyncReplicas().length;
+    Node[] inSyncReplicas = Arrays.copyOf(partitionInfo.inSyncReplicas(), numInsyncReplicas);
+    Arrays.sort(inSyncReplicas);
+    int numReplicas = partitionInfo.replicas().length;
+    Node[] replicas = Arrays.copyOf(partitionInfo.replicas(), numReplicas);
+    Arrays.sort(replicas);
+    boolean isURP = !(Arrays.equals(replicas, inSyncReplicas));
+    return isURP;
+  }
 
   /**
    * Gather the Kafka partition state within the given under replicated, offline, under minIsr,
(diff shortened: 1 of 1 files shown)


suyashtava:
CC: @CCisGG

CCisGG:
@suyashtava that is a flaky test. I can re-trigger it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
```diff
@@ -112,7 +112,7 @@ protected void populateKafkaPartitionState(Set<PartitionInfo> underReplicatedPar
         }
         for (PartitionInfo partitionInfo : _kafkaCluster.partitionsForTopic(topic)) {
           int numInsyncReplicas = partitionInfo.inSyncReplicas().length;
-          boolean isURP = numInsyncReplicas != partitionInfo.replicas().length;
+          boolean isURP = !(partitionInfo.inSyncReplicas() == partitionInfo.replicas());
```
CCisGG, line 115: I don't think this is safe enough. E.g. if the ISR set is same as replica set, but they are not the same order, then isURP will be true. But I think in that case, isURP should be false.

I feel we should compare the two Node[] in a sorted manner

suyashtava:
Not very much aligned because as far as I have seen. Replicas always come
in this order. {Leader/isr1, Isr2/follower1, isr3/follower2}

While ISR comes in {isr1,isr2,isr3} considering all 3 are isr.

So they will always maintain the order.

If there’s a discrepancy in order then this is not a normal behaviour, and
should be called underrep.

@CCisGG What are your thoughts?

But if we are ignoring this abnormal behaviour than we can go for sorting.

CCisGG:
@suyashtava I just checked our cluster, and the order in ISR set is not what you listed. The leader is not always the first one. We should compare the two only after sorting.

CCisGG:
Besides, could you make the PR title more explicit and concrete? The PR title will be automatically convert to the commit message by default once merged. We should make sure the commit message is clear and informative.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
```diff
@@ -112,7 +113,12 @@ protected void populateKafkaPartitionState(Set<PartitionInfo> underReplicatedPar
         }
         for (PartitionInfo partitionInfo : _kafkaCluster.partitionsForTopic(topic)) {
           int numInsyncReplicas = partitionInfo.inSyncReplicas().length;
-          boolean isURP = !(partitionInfo.inSyncReplicas() == partitionInfo.replicas());
+          Node[] inSyncReplicas = Arrays.copyOf(partitionInfo.inSyncReplicas(), numInsyncReplicas);
+          Arrays.sort(inSyncReplicas);
+          int numReplicas = partitionInfo.replicas().length;
+          Node[] replicas = Arrays.copyOf(partitionInfo.replicas(), numReplicas);
+          Arrays.sort(replicas);
+          boolean isURP = !(replicas == inSyncReplicas);
```
CCisGG, line 121: "==" looks not right here, as it compares the reference, not the array content. I think Arrays.equal should be used here. That said, I think we can extract a method to compare between ISR and replica set, and unit test it. That way we can feel safer before ship it.

suyashtava: Yup, Array Equals makes more sense And Unit test aswell.

suyashtava:
Hi @CCisGG , I checked the Test file: There no any such Unit test here for any helper function([example](https://github.com/linkedin/cruise-control/blob/b68431fbe303558ca3868ba802ce2663690344c0/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java#L145)) in [Response.test](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/response/ResponseTest.java)

I would suggest let's go ahead with this, as it is not such a big change and I can create a new PR/Issue, where we can create all the missing unit tests, in this or any file. That would go a long way.

Your thoughts?

suyashtava:
Further was thinking will it make sense to have a response until file and we can move all our helpers functions there.

And then have unit test for all helpers functions in that file. That will be more systematic

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
```diff
@@ -83,6 +86,23 @@ public Map<String, Object> getJsonStructure() {
...
+   * @return the effective boolean under replicated status of a partition.
+   */
+  protected boolean isURP(PartitionInfo partitionInfo) {
+    int numInsyncReplicas = partitionInfo.inSyncReplicas().length;
+    Node[] inSyncReplicas = Arrays.copyOf(partitionInfo.inSyncReplicas(), numInsyncReplicas);
+    Arrays.sort(inSyncReplicas);
+    int numReplicas = partitionInfo.replicas().length;
+    Node[] replicas = Arrays.copyOf(partitionInfo.replicas(), numReplicas);
+    Arrays.sort(replicas);
+    boolean isURP = !(Arrays.equals(replicas, inSyncReplicas));
```
CCisGG, line 103: Redundant brackets here. "!Arrays.equals(replicas, inSyncReplicas);" should be good enough.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
```diff
@@ -112,7 +132,7 @@ protected void populateKafkaPartitionState(Set<PartitionInfo> underReplicatedPar
         }
         for (PartitionInfo partitionInfo : _kafkaCluster.partitionsForTopic(topic)) {
           int numInsyncReplicas = partitionInfo.inSyncReplicas().length;
```
CCisGG, line 134: Do we still need this var?

CCisGG (changes requested):
Changes look good to me. I'm sort of worrying about the perf impact when there are many partitions. We can merge it and we will see how does it work on our large clusters.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
suyashtava: Yes, we still need it to check if th3e current Partition is underMinIsrPartitions

suyashtava: Resolving this as of Now.

suyashtava:
Updated the redundant brackets.

Btw do you want me to work on creating a utils file and move all  helper functions there and unit tests for each of them?

CCisGG:
@suyashtava That would be great and helpful. Would you like to do it in this same PR or you want a new one?

suyashtava:
I would prefer we should do it seprately as a new WI.

That way we can track things down better which helpers functions need to be
moved and we will easier to track down mistake as it won’t create functions
in case of things break.


Should I create a new issue for the same, or directly create a placeholder
PR and start working there.

CCisGG:
@suyashtava a new issue would always be great.

CCisGG (approved):

suyashtava:
Sure will create new issue and start working on it.
 Will do it in Day time of IST in morning, tomorrow.

We are good to close this one.

CCisGG:
Hi @suyashtava , I have to revert the PR as it fails the integration test. 
https://app.circleci.com/pipelines/github/linkedin/cruise-control/1999/workflows/27fdca20-5672-4237-9c0c-a4a4f1c6cdcf/jobs/5725

Root cause is Node class cannot be cast to Comparable class. 

This is why we should add unit test for any changes. I'd say this time let's add the test before merge the PR.

PR #1935, opened 2022-10-31
Exclude guava compile dependencies from runtime

### LINKED ISSUE #1934: Guava compile dependencies on runtime classpath
Hi, we build/run cruise-control and noticed that some guava compile dependencies are on the runtime path. As a user we'd like the minimum set of dependencies on the classpath. We currently exclude them in our build and wondered if it was worth doing the same upstream.

They are:
```
com.google.code.findbugs:jsr305
com.google.errorprone:error_prone_annotations
com.google.j2objc:j2objc-annotations
org.checkerframework:checker-qual
```

https://github.com/google/guava/wiki/UseGuavaInYourBuild#what-about-guavas-own-dependencies

Thank you

robobario:
This PR resolves #1934

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -264,6 +264,12 @@ project(':cruise-control') {
       exclude group: 'org.slf4j', module: 'slf4j-log4j12'
       exclude group: 'log4j', module: 'log4j'
     }
+    configurations.runtimeClasspath {
+      exclude group: 'com.google.code.findbugs', module: 'jsr305'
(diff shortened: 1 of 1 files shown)


CCisGG:
Hi @robobario, I left some comments/questions in the source issue. Would you mind taking a look if there is a chance? Thanks!

robobario:
Hi @CCisGG, will do, sorry about the wait :bow:

PR #1939, opened 2022-11-24
Address issues with Vertx based Swagger UI

### LINKED ISSUE #1662: Add Vertx.io based API with Swagger UI
### Motivation
Currently Cruise Control uses the classic servlet technology today. This works but provides a limited extension points and it is generally a very basic framework. Extending Cruise Control in this framework with things like Swagger which is almost a must have these days is cumbersome.
Swagger standardizes API, provides documentation support and an easy to use API UI that makes the Cruise Control API much easier to use.

### About this task
This task attempts to tackle the task of introducing an alternative and modern Vertex.io based API parallel to the existing servlet one and based on the existing codebase. The new API will be fully compatible with the currently existing one to ease migration. It is basically just the same API running on Vertx and has a Swagger UI and the existing API documented with Swagger annotations.

### Testing
The main testing we carry out are unit tests for white box testing and integration level validation (black box) that compares the outputs of the APIs with various parameters to verify behavior and backward compatibility.

kismob:
This PR resolves #1662 .
Hi @CCisGG, I completely removed the VertxAvailability Integration test, because calling every endpoint did not rally added much to the test coverage, because in those tests I only checked if we get response or not. And because of the many many requests sent almost in the same time the tests could fail. But, I updated the existing integrations tests, so now they call the endpoints correctly when using vertx. So the already existing tests are now 100% testing the Vertx use-cases as well.
I updated the .gitignore as you suggested.
If there is anything else please let me know.


.gitignore
```diff
@@ -20,3 +20,6 @@ access.log
 .vertx
 **openapi.json
 fileStore/failedBrokers.txt
+cruise-control/src/main/resources/webroot/.openapi-generator-ignore
```
CCisGG, line 23: Just curious do we want to add all contents in webroot or we shouldn't do that? I'm not familiar with that part.

CCisGG (commented):
Thanks for the quick change! Left a minor question.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
```diff
@@ -934,4 +935,13 @@ public static List<CompletionType> cachedValues() {
       return Collections.unmodifiableList(CACHED_VALUES);
     }
   }
+
+  private static Integer getNewPort() {
+    try {
+      return new ServerSocket(0).getLocalPort();
```
viktorsomogyi, line 941: Do you need to close the socket?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlVertxApp.java
```diff
@@ -25,7 +25,7 @@ public KafkaCruiseControlVertxApp(KafkaCruiseControlConfig config, Integer port,
 
     @Override
     public String serverUrl() {
-        return null;
+        return "http://" + _hostname + ":" + _port + "/";
```
viktorsomogyi, line 28: Please use the URL class, it's a bit nicer.

.gitignore
kismob: The other contents are not generated, they are needed for the UI. So we don't want to add them here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
kismob: We only use this in tests, so the socket is closing when the test finishes.

CCisGG (approved):

CCisGG:
Hi @kismob , looks like there are other integration tests failures related to vertx io. https://app.circleci.com/pipelines/github/linkedin/cruise-control/2040/workflows/4578f8ca-c087-44e4-bd91-5d98f2480c32/jobs/5808

Looks like when vertx is enabled, the requests are slower that it wasn't able to meet the conditions before the timeout. I have tried to make the timeout longer, but it still doesn't help. Could you please take a look at those failures? Thank you!

kismob:
Hi @CCisGG, I was looking at it but I don't know what could cause the problem. I don't think it is a performance issue. All the integration tests are passing locally, it is really hard to guess what could cause the problem. Is there a way where I could use Circle CI to run the integration tests without commiting the changes so I could debug this?

CCisGG:
Hi @kismob, from the test failure output, it's clear that the failure reason is when vertx is enabled, some condition are not met within the given time. The reason that the tests pass locally, is probably your local dev machine has better performance than circle ci machines. I think now your question is how to identify the issue without reproducing it.

I discussed with the team and we don't have a perfect solution for that. I would recommend you gradually reduce the test timeout locally, until it fails. It's likely that at one point only the test with vertx enabled is failed, but the servlet one succeeded. Then you can start the investigation from there. Please let me know if this makes sense.

From our side, we will probably take actions on enforce integration test as a PR check in the future.

PR #1967, opened 2023-02-13
Add metrics to reflect the partition movement speed

### LINKED ISSUE #1966: Add metrics to reflect the partition movement performance.
Today in Cruise Control, there is no existing metric that directly shows the rebalance speed. 

As a recent effort of speeding up cruise control rebalance, we need to first have some metrics to reflect the partition movement performance. The basic idea is emit the (numTotalPartitions / time) and (numTotalMBs / time) values when the partition movement finishes.

CCisGG:
This PR resolves #1966 .

This PR adds 4 metrics:
1. Task execution rate
2. Partition data movement rate
3. Partition movement per sec
4. Partition data movement per sec

Those metrics would be a good indicator of the rebalance speed, and is a pre-requisite of optimizing the rebalance performance.

Tested:
Screen Shot 2023-02-14 at 2 42 42 PM

As there is no rebalance ongoing, the metrics value is 0.


mitchhh22 (approved):
LGTM

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1367,6 +1377,7 @@ private void interBrokerMoveReplicas() throws InterruptedException, ExecutionExc
       }
       // At this point it is guaranteed that there are no in execution tasks to wait -- i.e. all tasks are completed or dead.
       if (_stopSignal.get() == NO_STOP_EXECUTION) {
+        updatePartitionMovementMetrics(numTotalPartitionMovements, totalDataToMoveInMB, System.currentTimeMillis() - startTime);
```
sutambe, line 1380: I think both metrics should be updated in the loop and not just at the end when all is complete. There are already `numFinishedPartitionMovements` and `finishedDataMovementInMB` which can be used to update the gauge. This will give us better insight in the variability of the partition movement speed at a minute granularity. What do you think?

sutambe (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: I have considered that path. The problem of that is the value of first several loops can be misleading.

E.g. it can take 5 seconds to finish the first 100 partition movement, the the value will be 100/5 = 20. However for the total rebalance process, the average partition movement per sec can be only 0.01.

I believe the average speed of the whole rebalance time is more representative.

sutambe: This chart updated at the end of the rebalance is gonna look like a step function that changes only once a hour or something---when the next rebalance finishes, it will suddenly drop or increase. It's misleading. The correct average rebalance should be some sort of trailing X minutes rebalance speed. will work especially well for the bandwidth chart.   Ingraphs has support for that: https://iwww.corp.linkedin.com/wiki/cf/display/SOP/Ingraphs+Examples#IngraphsExamples-Smoothingbyaveragingthelast30minutesofdata

sutambe: If I understand you right, the value of first several loops is immaterial because the sensor framework will snapshot the value only once a minute. All intermediate values between two snapshots are overwritten. In that sense, the sensor should record only the cumulative time of the finished movements count and data in MB. It will smooth out towards the end of the rebalance towards the average value automatically. May be we don't need trailing x minutes then.

CCisGG: Ok I can see either way has some misleading points, but I still think we should only use the overall rebalance speed.

What we need to is a reflection of overall rebalance performance, rather than the rebalance throughput at a certain time. Think about this case: at some cycles, the partition movement can be much faster/slower than other cycles, but we don't care about them. 

In general, I think we should rule out the throughput change during the rebalance, and only emit the value at the end of the rebalance.

sutambe: Consider using 1-, 5-, and 15- minute average when reporting. It will smooth out the bandwidth signal over the last x minutes. For long running rebalances it will be a meaningful signal. For very short running rebalances not so much. but then short running rebalances are not a problem. 

See dropwizard: https://metrics.dropwizard.io/4.2.0/getting-started.html#gauges

lmr3796:
By the way, if you use "fixes #1966" instead of "resolves", it will close it on merge.

See https://github.blog/2013-01-22-closing-issues-via-commit-messages/

lmr3796:
Implementation-wise LGTM.  However I have the same concern as Sumant.

@sutambe 's idea of "Consider using 1-, 5-, and 15- minute average when reporting" sounds fairly reasonable to me, and you can co-exist that with your current "total average".

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
```diff
@@ -46,6 +49,8 @@ public class ExecutionTaskTracker {
   public static final String COMPLETED = "completed";
   public static final String GAUGE_ONGOING_EXECUTION_IN_KAFKA_ASSIGNER_MODE = "ongoing-execution-kafka_assigner";
   public static final String GAUGE_ONGOING_EXECUTION_IN_NON_KAFKA_ASSIGNER_MODE = "ongoing-execution-non_kafka_assigner";
+  public static final String METER_TASK_EXECUTION_RATE = "task-execution-rate";
+  public static final String METER_PARTITION_DATA_MOVEMENT_RATE = "partition-data-movement-rate";
```
sutambe: Can you please add the unit here? `dataToMove` below is in bytes? Please document briefly how to interpret the 5 constituent metrics partition-data-movement-rate.[Count, [1|5|15]MinuteRate, Mean]. 

I see count going into millions and Mean is about 1000 and continuously climbing. The rate should plateau at some point right?

I'm probably missing  the mapping of the Meter to ingraphs chart type.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
```diff
@@ -121,6 +135,7 @@ public void markTaskState(ExecutionTask task, ExecutionTaskState newState) {
         updateDataMovement(task);
         break;
       case COMPLETED:
+        _taskExecutionRateMeter.mark();
```
sutambe, line 138: I'm unclear if `executionRateMeter` will mark  completion regardless of the type of the task is being completed. In other words, leadership movements and partition movements captured by the same meter here? If yes, how can we distinguish?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
```diff
@@ -149,6 +164,7 @@ private void updateDataMovement(ExecutionTask task) {
     } else if (task.state() == ExecutionTaskState.ABORTED
                || task.state() == ExecutionTaskState.DEAD
                || task.state() == ExecutionTaskState.COMPLETED) {
+      _partitionDataMovementRateMeter.mark(dataToMove);
```
sutambe, line 167: Unit will be useful.

CCisGG: Unit was added on the metric name.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
CCisGG: Yes it's not distinguish the task type. I think it's fine to have it for all task types.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
CCisGG: Not sure what do you mean, but these 2 metrics are just meters not gauge. I have 2 other gauges added

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskTracker.java
sutambe: Can you please explain why distinguishing them is not necessary? If I recall correctly, Cruise-Control proposals contain a mix of partition movements, leadership movements, etc. The speed of execution of these different tasks is vastly different. 

For now, we could have just one cumulative meter. However, I think it would be useful to distinguish between the two so that they can be optimized separately, if needed.

CCisGG: I believe all task types in a "rebalance" operation is INTER_BROKER_REPLICA_ACTION, including tasks like [a, b, c] -> [b, c, a].

sutambe:
@CCisGG It looks like there are only [3 types of tasks](https://github.com/linkedin/cruise-control/blob/b4e44ec004e6f5e22bd1c4e203d92341ed9e1659/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTask.java#L274). Can you please add separate sensors to measure task-execution-rate for each kind of task?

PR #1968, opened 2023-02-21
Implement concurrency adjuster for each individual broker

### LINKED ISSUE #1965: Replace global execution concurrency to individual broker execution concurrency
### Background

Today during execution of inter-broker partition movement, Cruise Control is using a global broker-execution-concurrency to throttle the execution:

https://github.com/linkedin/cruise-control/blob/0832eb5e6ae9f230f6c4ee8a90fdd771a305ce09/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java#L157

This global broker-concurrency is not the most reasonable design, as each broker can perform differently. E.g. for slow brokers, it makes sense to set a low execution concurrency, but for fast brokers we should allow a higher concurrency.

What makes it worse, Cruise Control has a concurrency adjuster, which would dynamically change this concurrency based on the bad performers among all brokers. If one broker is very slow, Cruise Control would reduce the execution concurrency for all brokers. This significantly limit the overall performance for the rebalance.

### Solution

To speed up the rebalance, we should make changes to the concurrency throttle: instead of having a global value for broker execution concurrency, we will have a map to store the concurrency for each broker (by making change to the line below):

https://github.com/linkedin/cruise-control/blob/0832eb5e6ae9f230f6c4ee8a90fdd771a305ce09/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskManager.java#L44

With this change, the concurrency adjuster should only adjust the concurrency for the broker based on its own performance, not the worst performer in the cluster.

CCisGG:
This PR resolves #1965 .

### Description
Today Cruise Control use identical allowed execution concurrency for each broker, which can cause unnecessary slowdown of rebalance. E.g. if a single broker performs bad, the concurrency of all broker goes down together, and the rebalance can become very slow in that case.

This PR makes replace the global identical concurrency to individual broker concurrency. Each broker will have its own execution currency during the rebalance, and the concurrency will be adjusted only based on its own metric.

As a side effort, this PR change the way of reports the concurrency. Now that each broker can have a different concurrency, the concurrency report now displays the min/max/avg value of the concurrency of all brokers. Similarly, this PR update the metrics to emit min/max/avg value instead of a global value.

### Deprecation
The metrics GAUGE_EXECUTION_INTER_BROKER_PARTITION_MOVEMENTS_PER_BROKER_CAP and GAUGE_EXECUTION_INTRA_BROKER_PARTITION_MOVEMENTS_PER_BROKER_CAP are deprecated since now we don't have a common cap for each broker. We added several new metrics to reflect the broker concurrency, including max/min/avg concurrency value for each concurrency type.

### Test
Tested on dev cluster and verified it can adjust concurrency separately for each broker. Logs:
```

2023/02/21 17:50:38.253 INFO [Executor] [ConcurrencyAdjuster-0] [kafka-cruise-control] [] Concurrency adjuster recommended an increase in INTER_BROKER_REPLICA movement concurrency to 9 for broker 101598
2023/02/21 17:50:38.253 INFO [Executor] [ConcurrencyAdjuster-0] [kafka-cruise-control] [] Concurrency adjuster recommended an increase in INTER_BROKER_REPLICA movement concurrency to 9 for broker 102114
2023/02/21 17:50:38.253 INFO [Executor] [ConcurrencyAdjuster-0] [kafka-cruise-control] [] Concurrency adjuster recommended a decrease in INTER_BROKER_REPLICA movement concurrency to 4 for broker 84197
2023/02/21 17:50:38.253 INFO [Executor] [ConcurrencyAdjuster-0] [kafka-cruise-control] [] Concurrency adjuster recommended an increase in INTER_BROKER_REPLICA movement concurrency to 3 for broker 77284
2023/02/21 17:50:38.253 INFO [Executor] [ConcurrencyAdjuster-0] [kafka-cruise-control] [] Concurrency adjuster recommended an increase in INTER_BROKER_REPLICA movement concurrency to 5 for broker 101604
2023/02/21 17:50:38.253 INFO [Executor] [ConcurrencyAdjuster-0] [kafka-cruise-control] [] Concurrency adjuster recommended a decrease in INTER_BROKER_REPLICA movement concurrency to 1 for broker 81639
2023/02/21 17:50:38.253 INFO [Executor] [ConcurrencyAdjuster-0] [kafka-cruise-control] [] Concurrency adjuster recommended an increase in INTER_BROKER_REPLICA movement concurrency to 9 for broker 78311
```

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -26,7 +26,7 @@ import com.linkedin.kafka.cruisecontrol.detector.Provisioner;
 import com.linkedin.kafka.cruisecontrol.exception.BrokerCapacityResolutionException;
 import com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException;
 import com.linkedin.kafka.cruisecontrol.exception.OngoingExecutionException;
-import com.linkedin.kafka.cruisecontrol.executor.concurrency.ConcurrencyType;
+import com.linkedin.kafka.cruisecontrol.executor.ConcurrencyType;
 import com.linkedin.kafka.cruisecontrol.executor.ExecutionProposal;
 import com.linkedin.kafka.cruisecontrol.executor.Executor;
 import com.linkedin.kafka.cruisecontrol.executor.ExecutorState;
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -26,7 +26,7 @@ import com.linkedin.kafka.cruisecontrol.detector.Provisioner;
 import com.linkedin.kafka.cruisecontrol.exception.BrokerCapacityResolutionException;
 import com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException;
 import com.linkedin.kafka.cruisecontrol.exception.OngoingExecutionException;
-import com.linkedin.kafka.cruisecontrol.executor.ConcurrencyType;
+import com.linkedin.kafka.cruisecontrol.executor.concurrency.ConcurrencyType;
 import com.linkedin.kafka.cruisecontrol.executor.ExecutionProposal;
 import com.linkedin.kafka.cruisecontrol.executor.Executor;
 import com.linkedin.kafka.cruisecontrol.executor.ExecutorState;
@@ -497,9 +497,10 @@ public class KafkaCruiseControl {
 
   /**
    * Dynamically set the execution concurrency per broker.
+   * Notice that if concurrency type is LEADERSHIP, this value will also be used as max allowed concurrency at cluster level.
    *
-   * @param requestedExecutionConcurrency The maximum number of concurrent execution per broker.
-   * @param concurrencyType The type of the execution concurrency to change
+   * @param requestedExecutionConcurrency The maximum number of allowed concurrent execution per broker.
+   * @param concurrencyType The type of the execution concurrency to set value for.
    */
   public void setExecutionConcurrencyForAllBrokers(Integer requestedExecutionConcurrency, ConcurrencyType concurrencyType) {
     _executor.setExecutionConcurrencyForAllBrokers(requestedExecutionConcurrency, concurrencyType);
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -496,31 +496,13 @@ public class KafkaCruiseControl {
   }
 
   /**
-   * Dynamically set the inter-broker partition movement concurrency per broker.
+   * Dynamically set the execution concurrency per broker.
    *
-   * @param requestedInterBrokerPartitionMovementConcurrency The maximum number of concurrent inter-broker partition movements
-   *                                                         per broker.
+   * @param requestedExecutionConcurrency The maximum number of concurrent execution per broker.
+   * @param concurrencyType The type of the execution concurrency to change
    */
-  public void setRequestedInterBrokerPartitionMovementConcurrency(Integer requestedInterBrokerPartitionMovementConcurrency) {
-    _executor.setRequestedInterBrokerPartitionMovementConcurrency(requestedInterBrokerPartitionMovementConcurrency);
-  }
-
-  /**
-   * Dynamically set the intra-broker partition movement concurrency.
-   *
-   * @param requestedIntraBrokerPartitionMovementConcurrency The maximum number of concurrent intra-broker partition movements.
-   */
-  public void setRequestedIntraBrokerPartitionMovementConcurrency(Integer requestedIntraBrokerPartitionMovementConcurrency) {
-    _executor.setRequestedIntraBrokerPartitionMovementConcurrency(requestedIntraBrokerPartitionMovementConcurrency);
-  }
-
-  /**
-   * Dynamically set the leadership movement concurrency.
-   *
-   * @param requestedLeadershipMovementConcurrency The maximum number of concurrent leader movements.
-   */
-  public void setRequestedLeadershipMovementConcurrency(Integer requestedLeadershipMovementConcurrency) {
-    _executor.setRequestedLeadershipMovementConcurrency(requestedLeadershipMovementConcurrency);
+  public void setExecutionConcurrencyForAllBrokers(Integer requestedExecutionConcurrency, ConcurrencyType concurrencyType) {
+    _executor.setExecutionConcurrencyForAllBrokers(requestedExecutionConcurrency, concurrencyType);
   }
 
   /**
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ExecutionConcurrencyManager.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ExecutionConcurrencyManager.java
@@ -24,31 +24,46 @@ import org.slf4j.LoggerFactory;
  */
 public class ExecutionConcurrencyManager {
   private static final Logger LOG = LoggerFactory.getLogger(ExecutionConcurrencyManager.class);
-  private final int _maxNumClusterMovementConcurrency;
+
+  private static final int MIN_CONCURRENCY_PER_BROKER = 1;
+
+  // The total allowed movement concurrency for all types in the cluster. This value cannot be overridden at runtime.
+  private final int _clusterMovementConcurrency;
+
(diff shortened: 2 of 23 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -496,31 +496,13 @@ public void setRequestedExecutionProgressCheckIntervalMs(Long requestedExecution
...
   /**
-   * Dynamically set the inter-broker partition movement concurrency per broker.
+   * Dynamically set the execution concurrency per broker.
    *
-   * @param requestedInterBrokerPartitionMovementConcurrency The maximum number of concurrent inter-broker partition movements
-   *                                                         per broker.
+   * @param requestedExecutionConcurrency The maximum number of concurrent execution per broker.
+   * @param concurrencyType The type of the execution concurrency to change
    */
-  public void setRequestedInterBrokerPartitionMovementConcurrency(Integer requestedInterBrokerPartitionMovementConcurrency) {
```
groelofs, line 504: Given the lack of class-level javadocs, I can't tell: are these added and removed APIs public in the sense of being usable by external software, or are they used purely within CC itself?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ExecutionConcurrencyManager.java
```diff
@@ -0,0 +1,271 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * ExecutionConcurrencyManager: manages the allowed concurrency for all execution types: inter-broker/intra-broker/leadership movement.
+ * For each execution type, the default allowed concurrency is defined by configuration, and the requested concurrency is an optional
+ *  parameter announced on triggering the rebalance. At the beginning of the execution, each broker starts with the same requested
+ *  concurrency (or the default concurrency if requested concurrency not provided). During the execution, each broker adjust its
+ *  allowed concurrency based on its own metric.
```
groelofs, line 22: This is useful, but it feels like there's enough to the concurrency adjustment that it deserves its own doc (or section of a doc) describing how it works and how it fits together with the other pieces. For example, how are recommendations acted upon, is concurrency ever changed by more than 1 in a single step, how quickly do recommendations get updated, is there hysteresis to prevent ping-ponging, etc. (I suspect you could hit the necessary points in one page or less.)

Such a document would be useful for onboarding new devs to CC and for reminding older or infrequent contributors of changes they may not recall.

groelofs (commented):
I'm mainly looking at higher-level things; this isn't a "real" review, nor is it finished, and I won't be approving anything either way. Just FYI.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ExecutionConcurrencyManager.java
```diff
@@ -0,0 +1,271 @@
...
+   * @param requestedMaxInterBrokerPartitionMovements The maximum number of concurrent inter-broker partition movements per broker
+   *                                                  (if null, use {@link #_defaultInterBrokerPartitionMovementConcurrency}).
+   */
+  public synchronized void setRequestedMaxInterBrokerPartitionMovements(Integer requestedMaxInterBrokerPartitionMovements) {
+    if (requestedMaxInterBrokerPartitionMovements != null
+        && requestedMaxInterBrokerPartitionMovements > _maxNumClusterMovementConcurrency) {
+      throw new IllegalArgumentException("Attempt to set max inter-broker partition movements [" + requestedMaxInterBrokerPartitionMovements
+                                         + "] to greater than the maximum" + " number of allowed movements in cluster ["
+                                         + _maxNumClusterMovementConcurrency + "].");
+    }
```
groelofs, line 190: FWIW, I tend to hate this kind of logic since it ties together two different configs (which in itself is reasonable) whose relationship may not be well-documented in the source of truth (cfg2), and the result of a mistake in the latter is (AFAICT) a completely broken Cruise Control instance.  Is there a reason not to correct the too-large setting (requestedMaxInterBrokerPartitionMovements) to the sanity-check cap (_maxNumClusterMovementConcurrency) and log both this fact and what config name to change if the former was the correct value and the latter the mistake?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ExecutionConcurrencyManager.java
```diff
@@ -0,0 +1,271 @@
...
+      throw new IllegalArgumentException("The execution manager has not been initialized.");
+    }
+    return _executionConcurrencySummary;
+  }
+
+  private void sanityCheckRequestedConcurrency(Integer concurrency, ConcurrencyType concurrencyType) {
+    if (concurrency != null && concurrency >= _maxNumClusterMovementConcurrency) {
+      throw new IllegalArgumentException("Attempt to set " + concurrencyType + " concurrency ["
+                                         + concurrency + "] to greater than or equal to the maximum"
+                                         + " number of allowed movements in cluster [" + _maxNumClusterMovementConcurrency + "].");
```
groelofs, line 235: Same thing here. From what I've seen, even our SREs (who use CC every single day) struggle to understand the complexity and how various configs and goals interact with each other. There should be a conscious effort to simplify the operation of CC itself wherever possible, and that means not just logging sanity-check failures but actively correcting them where possible and simultaneously instructing users on how to fix things if the automatic correction is not what they wanted. (One might go so far as to include a new metric indicating the number of sanity-check auto-adjustments since startup so the potential user error isn't just buried in a log file. Or it could be included in the cluster-state command-line report or a new ccdo "things you should be aware of" option. Or both.)

groelofs (commented):
I'm mainly looking at higher-level things; this isn't a "real" review, nor is it finished, and I won't be approving anything either way. Just FYI.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
CCisGG: It's only used within CC itself.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ExecutionConcurrencyManager.java
CCisGG: Your concern totally makes sense, and the way we map cc config to li-cc config can totally cause a pain when missing sufficient description in li-cc...

>  Is there a reason not to correct the too-large setting (requestedMaxInterBrokerPartitionMovements) to the sanity-check cap (_maxNumClusterMovementConcurrency) and log both this fact and what config name to change if the former was the correct value and the latter the mistake?
I think it's because we don't want to silently use a value that not specified by user. E.g. if we just cap the requested value without telling the user, the user may think CC is using their requested value while it actually not. 

I think the best way to do it is to fail the execution, while output the failure reason as part of the command output. That may require another non-trivial refactor though.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
```diff
@@ -119,19 +120,70 @@ public void testGetLeaderMovementTasks() {
...
-    assertEquals("2 of the leader movements should return in one batch", 2, leaderMovementTasks.size());
-    assertEquals(4, leaderMovementTasks.get(0).executionId());
+    List<ExecutionTask> leaderMovementTasks = planner.getLeadershipMovementTasks(leadershipMovementConcurrencyPerBroker, 4);
+    assertEquals("4 of the leader movements should return in one batch", 4, leaderMovementTasks.size());
+    assertEquals(leaderMovementTasks.get(0).proposal(), _leaderMovement1);
+    assertEquals(leaderMovementTasks.get(1).proposal(), _leaderMovement2);
+    assertEquals(leaderMovementTasks.get(2).proposal(), _leaderMovement3);
+    assertEquals(leaderMovementTasks.get(3).proposal(), _leaderMovement4);
+
+    // 2: task is throttled by cluster concurrency:
```
lmr3796, line 143: Optional nit: Better practice is splitting these to different `@Test` method.  Since you're explaining the cases, I think they probably deserve a test case with good naming explaining

lmr3796 (approved):
+1

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlannerTest.java
CCisGG: Yeah I think that makes sense most of the cases. In cruise-control executor related tests, I see the tests classes tends to group many scenarios together in a single test. I feel the reason is there are quite a lot different branches needs to be tested for each method, which would potential cause the test number to easily 4X to 5X if separate each test branch. Another potential reason is setting up each test takes sometime, and we don't want the unit tests to spend much more time, as it is already very slow now.

mitchhh22 (approved):

PR #1980, opened 2023-03-09
Dynamically adjust the executionProgressCheckIntervalMs based on execution status

### LINKED ISSUE #1979: Dynamically adjust execution check interval
Today, the executionProgressCheckIntervalMs is a fixed value. This can slow down rebalance because of the long tail effect.

E.g. there can be one slow broker that has concurrency 1, but 2000 pending tasks. Each batch will only have 1 task, but the executor will wait for the full executionProgressCheckIntervalMs cycle to check the progress. That will significantly slow down the rebalance.

We should use some smart way to dynamically adjust execution check interval at runtime.

CCisGG:
This PR resolves #1979 .

Basically executor will reduce this executionProgressCheckIntervalMs if all tasks in a batch are completed within one check interval. Eventually it will hit minExecutionProgressCheckIntervalMs.

### DIFF
--- a/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
+++ b/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
@@ -367,11 +367,10 @@ public class ExecutorTest extends CCKafkaClientsIntegrationTestHarness {
   }
 
   @Test
-  public void testSetExecutionProgressCheckIntervalMs() {
+  public void testSetRequestedExecutionProgressCheckIntervalMs() {
     KafkaCruiseControlConfig config = new KafkaCruiseControlConfig(getExecutorProperties());
     Executor executor = new Executor(config, null, new MetricRegistry(), EasyMock.mock(MetadataClient.class),
                                      null, EasyMock.mock(AnomalyDetectorManager.class));
-    long defaultExecutionProgressCheckIntervalMs = config.getLong(ExecutorConfig.EXECUTION_PROGRESS_CHECK_INTERVAL_MS_CONFIG);
     long minExecutionProgressCheckIntervalMs = config.getLong(ExecutorConfig.MIN_EXECUTION_PROGRESS_CHECK_INTERVAL_MS_CONFIG);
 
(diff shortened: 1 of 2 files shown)


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -367,22 +367,55 @@ private static boolean verifyFutureError(Future<?> future, Class<? extends Throw
...
-    assertEquals(defaultExecutionProgressCheckIntervalMs, executor.executionProgressCheckIntervalMs());
     long minExecutionProgressCheckIntervalMs = config.getLong(ExecutorConfig.MIN_EXECUTION_PROGRESS_CHECK_INTERVAL_MS_CONFIG);
+
+    // RequestedExecutionProgressCheckIntervalMs has to be larger than minExecutionProgressCheckIntervalMs
     executor.setRequestedExecutionProgressCheckIntervalMs(minExecutionProgressCheckIntervalMs);
     assertEquals(minExecutionProgressCheckIntervalMs, executor.executionProgressCheckIntervalMs());
     assertThrows(IllegalArgumentException.class,
                  () -> executor.setRequestedExecutionProgressCheckIntervalMs(minExecutionProgressCheckIntervalMs - 1));
 
+    // With requested ≈, the value should be capped at requestedExecutionProgressCheckIntervalMs, and
```
lmr3796, line 383: nit: non-ascii char used

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -367,22 +367,55 @@ private static boolean verifyFutureError(Future<?> future, Class<? extends Throw
...
+    //  set to at least minExecutionProgressCheckIntervalMs
+    long requestedExecutionProgressCheckIntervalMs = 2 * defaultExecutionProgressCheckIntervalMs;
```
lmr3796, line 385: if underlying the executor is not too slow to spin up (e.g. initiates admin client), split the parts to multiple test methods.

But ultimately in unit tests, they should be mocked if those slow part exists.

lmr3796 (changes requested):

PR #1983, opened 2023-03-10
Add initialized state to concurrency adjuster and avoid null for the metric value

### LINKED ISSUE #1982: Upstream app throws error for concurrency metrics when no rebalance happens
In  #1968 , we implemented individual concurrency. As a side effort, we add new gauges to emit max/min/avg concurrency values.

As part of the implemention, it throws exception for max/min/avg concurrency if no rebalance is happening. This will make the gauges to be empty and upstream app can have exceptions in their sensors given such empty gauge.

CCisGG:
This PR resolves #1982.

This PR includes:
1. Avoid returning null for max/min/avg concurrencies if no ongoing execution. It will simply returns 0.
2. Add running state for ConcurrencyAdjuster and ConcurrencyManager. Only set state to running during execution.
3. Reseting the concurrency adjuster, manager and all concurrency values after execution. 
4. Add unit tests for the above 3 changes.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -402,6 +402,7 @@ public class Executor {
     private LoadMonitor _loadMonitor;
     private int _numChecks;
     private final ExecutionConcurrencyManager _executionConcurrencyManager;
+    private volatile boolean _started;
 
     public ConcurrencyAdjuster(int numMinIsrCheck) {
       _numMinIsrCheck = numMinIsrCheck;
@@ -430,6 +431,23 @@ public class Executor {
                                               requestedInterBrokerPartitionMovementConcurrency,
                                               requestedIntraBrokerPartitionMovementConcurrency,
                                               requestedLeadershipMovementConcurrency);
+      _started = true;
+    }
+
+    /**
+     * Concurrency should be reset after each execution is done.
+     */
+    public synchronized void clearAdjustment() {
+      _executionConcurrencyManager.reset();
+      _started = false;
+    }
+
+    /**
+     * Check if concurrency adjuster is started
+     * @return true if concurrency adjuster is started.
+     */
+    public boolean isStarted() {
+      return _started;
     }
 
     private boolean canRefreshConcurrency(ConcurrencyType concurrencyType) {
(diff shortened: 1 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -430,6 +431,23 @@ public synchronized void initAdjustment(LoadMonitor loadMonitor,
...
                                               requestedLeadershipMovementConcurrency);
+      _started = true;
+    }
+
+    /**
+     * Concurrency should be reset after each execution is done.
+     */
+    public synchronized void clearAdjustment() {
+      _executionConcurrencyManager.reset();
+      _started = false;
```
lmr3796, line 442: Could somebody be using `_executionConcurrencyManager` in a way that
1. check `_started` first
2. do something with the manager
3. it got reset in the middle
4. started mark as false

If that's so, either swap the 2 statements, or lock on the _started flag (by changing it to some other type suitable for such synchronization)

lmr3796 (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
lmr3796: Asking this because the code path that references `_started` here does not seem to be synced.

lmr3796: Also the other way around, if we swap that: is it possible some code path would assume if it is false, then the `_executionConcurrencyManager` has been reset?

Thinking more about it, I'm leaning forward that the code path on this `_started` flag should be revisited.  Maybe an RW lock is needed.

lmr3796 (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: Good point! I think your are right that we should swap the 2 lines.

CCisGG: > Also the other way around, if we swap that: is it possible some code path would assume if it is false, then the _executionConcurrencyManager has been reset?

Revisited the logic. Since ConcurrencyAdjuster and ConcurrencyManager are really different entities, one shouldn't make any assumptions that one is dependent on another. Thus RW lock is not needed here.

lmr3796 (approved):

PR #1992, opened 2023-03-22
Add a mapper hook to deal with encoded rack ID for `RackAwareGoal`s

### LINKED ISSUE #1993: Support transformed rack ID for replica placement
In LI's OSS Kafka linkedin/kafka#438, it adds a new capability to let the the broker have a "mapper class" to transform the rack ID for rack-aware replica placement, i.e. the "transformed" rack ID is used instead of the raw one for rack aware placement.

For a Kafka cluster enabling this, the corresponding cruise-control need to have the same mapper logic so that the rack-aware replica placement behavior is consistent during rebalances.

lmr3796:
This PR resolves #1993, which is to incorporate the feature in LI's OSS Kafka to add a mapper to translate rack IDs for replica placements (linkedin/kafka#438).

For a Kafka cluster enabling this, the corresponding cruise-control need to have the same mapper logic so that the rack-aware replica placement behavior is consistent during rebalances.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/kafkaassigner/KafkaAssignerDiskUsageDistributionGoal.java
```diff
@@ -463,8 +481,11 @@ Replica findReplicaToSwapWith(Replica replica,
   private boolean possibleToMove(Replica replica, Broker destinationBroker, ClusterModel clusterModel) {
     TopicPartition tp = replica.topicPartition();
 
-    boolean case1 = !clusterModel.partition(tp).partitionRacks().contains(destinationBroker.rack());
-    boolean case2 = replica.broker().rack() == destinationBroker.rack() && destinationBroker.replica(tp) == null;
+    boolean case1 = clusterModel.partition(tp).partitionRacks().stream()
+                                .map(this::mappedRackIdOf)
+                                .noneMatch(mappedRackIdOf(destinationBroker)::equals);
```
CCisGG, line 486: I think it's making redundant calls to mappedRackIdOf(destinationBroker) for each stream element. If someone creates complicated logic in mappedRackIdOf, then it would impact the performance.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/kafkaassigner/KafkaAssignerEvenRackAwareGoal.java
```diff
@@ -301,7 +319,7 @@ private boolean shouldExclude(Partition partition, int position, Set<String> exc
   private void ensureRackAwareSatisfiable(ClusterModel clusterModel, Set<String> excludedTopics)
       throws OptimizationFailureException {
     // Sanity Check: not enough racks to satisfy rack awareness.
-    int numAliveRacks = clusterModel.numAliveRacks();
+    int numAliveRacks = clusterModel.aliveRacks().stream().map(this::mappedRackIdOf).collect(Collectors.toSet()).size();
```
CCisGG, line 322: distinct().count()

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/AnalyzerConfig.java
```diff
@@ -642,6 +653,11 @@ public static ConfigDef define(ConfigDef configDef) {
                             DEFAULT_OPTIMIZATION_OPTIONS_GENERATOR_CLASS,
                             ConfigDef.Importance.LOW,
                             OPTIMIZATION_OPTIONS_GENERATOR_CLASS_DOC)
+                    .define(RACK_AWARE_GOAL_RACK_ID_MAPPER_CLASS_CONFIG,
```
CCisGG, line 656: Also add the configure in Configurations.md

CCisGG (commented):
Overall looks good to me. Left a few nits. I have some extra comments here:
1. I'd suggest to have create a CC issue and link it here. In the issue we can clarify the problem and tell open source users about how they can leverage this change.
2. Do a thorough search on rack.id() and make sure all the usage are covered in this PR.

lmr3796:
> 1. I'd suggest to have create a CC issue and link it here. In the issue we can clarify the problem and tell open source users about how they can leverage this change.
Added #1993  to the description.

> 2. Do a thorough search on rack.id() and make sure all the usage are covered in this PR.
Yes, actually the surface is done by a search starting from the `com.linkedin.kafka.cruisecontrol.model.Rack` class, which led to 
1. `com.linkedin.kafka.cruisecontrol.model.ClusterModel#_racksById`
2. `com.linkedin.kafka.cruisecontrol.model.Host#rack()`, which is only used by `com.linkedin.kafka.cruisecontrol.model.Broker#rack()`

`com.linkedin.kafka.cruisecontrol.model.Broker#rack()` then led to the `KafkaAssigner*Goal` series & `RackAwareGoals`

Finally, I did a `grep '[Rr]acks?()' goals/`
And found these 2
- numAliveRacks
- numRacks

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/kafkaassigner/KafkaAssignerEvenRackAwareGoal.java
lmr3796: Fixed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/kafkaassigner/KafkaAssignerDiskUsageDistributionGoal.java
lmr3796: fixed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/goals/GoalUtils.java
```diff
@@ -45,11 +52,41 @@ public final class GoalUtils {
...
+  }
+
+  /**
+   * Extract {@link KafkaCruiseControl} instance from configs.
+   *
+   * @param configs The configs used to configure {@link Goal} instance.
+   * @param errMsgSupplier supplier of the message to be used in the event that a IllegalArgumentException is thrown
+   * @return The extracted {@link KafkaCruiseControl} instance.
+   */
+  private static KafkaCruiseControl extractKafkaCruiseControlObjectFromConfig(Map<String, ?> configs, Supplier<String> errMsgSupplier) {
```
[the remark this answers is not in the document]
lmr3796: removed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/AnalyzerConfig.java
lmr3796: Added doc

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/kafkaassigner/KafkaAssignerDiskUsageDistributionGoal.java
lmr3796: For this
```java
    .noneMatch(mappedRackIdOf(destinationBroker)::equals);
```

It is only one `mappedRackIdOf(destinationBroker)` call. `mappedRackIdOf(destinationBroker)` is evaluated to an String first, and then the method reference `evaluatedStr::equals` is passed in.

CCisGG:
@lmr3796 Changes look good to me. 2 extra comments:
1. It would be helpful to have some unit tests
2. In the future if anyone is using Rack class to get the rack id, how can they realize this rack id mapper. This is a potential problem that we don't make the mapper as part of the rack.id() method. However I don't see any good ways of integrate the mapper with Rack class, so I think your solution is fine. Just want to hear your thoughts about this one.

lmr3796:
> 2\. In the future if anyone is using Rack class to get the rack id, how can they realize this rack id mapper. This is a potential problem that we don't make the mapper as part of the rack.id() method. However I don't see any good ways of integrate the mapper with Rack class, so I think your solution is fine. Just want to hear your thoughts about this one.

So this is intentional.  Because in the ultimate case we need this separation, as we will implement a goal that "extracts" the encoded power fault domain on the rack ID; and if it is tightly coupled to the `Rack#id` method, we would have no access to the "raw" rack ID to extract the part we need.

lmr3796:
> 1. It would be helpful to have some unit tests

Actually the unit test caught one bug :)

I originally forgot this `super.configure(configs)`, causing the `balancingConstraint` not initialized in the goals
https://github.com/linkedin/cruise-control/pull/1992/files#diff-0c353ddc2913093a1686a3605b6a7012fad492e8b5d1fb2bf497ab99d15f9f0cR56

CCisGG (approved):

CCisGG:
Looks good to me. Thanks @lmr3796 ! It would be helpful if we can test it on real cluster before roll out.

PR #1994, opened 2023-03-27
Slightly refactor RackAwareDistributionGoal

lmr3796:
The patch is a pure refactor without behavior change.

This opens the room for sub-classing it to customize the eligible broker priority. With this change, in the Linkedin internal build, we can achieve soft fault-zone-aware replica placement by
1. extend this goal
2. override the `SortedSet<Broker> rackAwareEligibleBrokers(Replica replica, ClusterModel clusterModel)`
3. provide number of replicas in fault zone as `extraSoftBrokerPriorityComparators`


CCisGG:
Could you further explain how do we use extraSoftBrokerPriorityComparators?

PR #1995, opened 2023-03-27
Add option to add preamble text to all slack notifications

tomncooper:
We use a single slack channel to receive notifications from multiple Cruise Control instances. The current slack notifier does not provided any uniquely identifiable information in its text so it is not possible to know from what instance a particular alert came from.

The notifier does let you customise the username and icon, which could allow us distinguish the alerts. However, as per that Slack webhook [docs](https://api.slack.com/messaging/webhooks#advanced_message_formatting):

> You cannot override the default channel (chosen by the user who installed your app), username, or icon when you're using Incoming Webhooks to post messages. Instead, these values will always inherit from the associated Slack app configuration.

This PR provides a simple solution by allowing a preamble string to added to every slack alert. This is set per instance and so users can easily provide unique identifiers.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/notifier/SlackSelfHealingNotifier.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/notifier/SlackSelfHealingNotifier.java
@@ -21,14 +21,17 @@ public class SlackSelfHealingNotifier extends SelfHealingNotifier {
     public static final String SLACK_SELF_HEALING_NOTIFIER_ICON = "slack.self.healing.notifier.icon";
     public static final String SLACK_SELF_HEALING_NOTIFIER_USER = "slack.self.healing.notifier.user";
     public static final String SLACK_SELF_HEALING_NOTIFIER_CHANNEL = "slack.self.healing.notifier.channel";
+    public static final String SLACK_SELF_HEALING_NOTIFIER_PREAMBLE = "slack.self.healing.notifier.preamble";
 
     public static final String DEFAULT_SLACK_SELF_HEALING_NOTIFIER_ICON = ":information_source:";
     public static final String DEFAULT_SLACK_SELF_HEALING_NOTIFIER_USER = "Cruise Control";
+    public static final String DEFAULT_SLACK_SELF_HEALING_NOTIFIER_PREAMBLE = "";
 
     protected String _slackWebhook;
     protected String _slackIcon;
     protected String _slackChannel;
     protected String _slackUser;
+    protected String _slackPreamble;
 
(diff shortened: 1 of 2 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/notifier/SlackSelfHealingNotifier.java
```diff
@@ -44,8 +47,10 @@ public void configure(Map<String, ?> config) {
         _slackIcon = (String) config.get(SLACK_SELF_HEALING_NOTIFIER_ICON);
         _slackChannel = (String) config.get(SLACK_SELF_HEALING_NOTIFIER_CHANNEL);
         _slackUser = (String) config.get(SLACK_SELF_HEALING_NOTIFIER_USER);
+        _slackPreamble = (String) config.get(SLACK_SELF_HEALING_NOTIFIER_PREAMBLE);
         _slackIcon = _slackIcon == null ? DEFAULT_SLACK_SELF_HEALING_NOTIFIER_ICON : _slackIcon;
         _slackUser = _slackUser == null ? DEFAULT_SLACK_SELF_HEALING_NOTIFIER_USER : _slackUser;
+        _slackPreamble = _slackPreamble == null ? DEFAULT_SLACK_SELF_HEALING_NOTIFIER_PREAMBLE : _slackPreamble;
```
CCisGG, line 53: You can use getOrDefault() method.
config.getOrDefault(SLACK_SELF_HEALING_NOTIFIER_PREAMBLE, DEFAULT_SLACK_SELF_HEALING_NOTIFIER_PREAMBLE)

CCisGG (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/notifier/SlackSelfHealingNotifier.java
tomncooper: That won't work as the default will have to be a string and what comes out of the config map is an Object. It has to be a two step process, somewhere the object has to be cast/converted to a string. So I just followed what had be done with other optional configs.

tomncooper: I could make it one line of course but I am not sure that buys us much?

CCisGG: That's fine then. Will approve. Thanks!

CCisGG (approved):

Issue #1828, opened 2022-05-10
Cruise Control sometimes is not able to detect offline partitions


CCisGG:
Today the way Cruise Control determine whether a partition is offline, is to check whether ISR set is empty:

https://github.com/linkedin/cruise-control/blob/a8a190f7c4662e1fc742994792c75b108fd9064d/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java#L105

While this works correctly for most of the time, it's actually not consistent with definition inside Kafka. In Kafka, an offline partition is a partition whose leader is not alive.

Recently we've have detect some issues that for a partition that is in a bad state:

{"controller_epoch":10000,"leader":-1,"version":1,"leader_epoch":50,"isr":[12345]}

And Cruise Control is not able to detect it as a offline partition while the kafka metric does show it as offline partition. Cruise Control only treat is as an URP.

I think cruise control should use the same criteria "whether leader is alive" to determine whether the partition is offline.

ayana-s:
👋 We've also experienced this behaviour, where Cruise Control doesn't accurately detect and display offline partitions. 

**Before creating offline partitions:**

<img width="1130" alt="Screenshot 2023-03-22 at 1 20 01 PM" src="https://user-images.githubusercontent.com/41079854/227235470-716b5030-9d4e-4220-be47-b65b45a21ab9.png">

**After creating offline partitions,** when describing our topic's partitions with kcat:
```
topic "test-ayana-3" with 1 partitions:
   partition 0, leader -1, replicas: 1, isrs: 1, Broker: Leader not available
```

However, Cruise Control still displays "No offline partitions":

<img width="1486" alt="Screenshot 2023-03-22 at 1 12 37 PM" src="https://user-images.githubusercontent.com/41079854/227235413-20f9b86c-febd-455c-a78e-656c16e417f9.png">

[...]

PR #2000, opened 2023-03-28
Set offline partitions as true if leader is null

ayana-s:
**🌊 This PR resolves https://github.com/linkedin/cruise-control/issues/1828.** 

## What does this PR accomplish?
Currently, cruise-control detects offline partitions if the number of in-sync replicas is zero, and only performs this check if the partition is under-replicated. This definition of offline partitions is not comprehensive nor completely correct.

This PR corrects the detection of offline partitions by using the **status of a partition's leader** to determine whether a partition is offline. It also cleans up the conditional logic that determines whether a partition's status is checked to begin with.

Screenshot 2023-04-18 at 12 07 55 PM

## How was it tested?
1. Removed brokers in a test cluster to create offline partitions. Confirmed that offline partitions are listed on the cruise-control UI (screenshot above). Confirmed that values close to my change (eg. URPs, offline replicas) do not seem affected. 

Checked http://localhost:9090/kafkacruisecontrol/kafka_cluster_state endpoint to confirm values matched UI:
Screenshot 2023-04-18 at 3 13 20 PM

2. Added a test for updated functionality
3. Deployed to Shopify's production environment

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
@@ -18,6 +18,8 @@ import java.util.stream.Stream;
 import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.config.TopicConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 @JsonResponseClass
@@ -35,6 +37,7 @@ public class ClusterPartitionState {
   public static final String MIN_INSYNC_REPLICAS = "min.insync.replicas";
   public static final int DEFAULT_MIN_INSYNC_REPLICAS = 1;
   public static final boolean DEFAULT_REMOTE_STORAGE_ENABLED = false;
+  private static final Logger LOG = LoggerFactory.getLogger(ClusterPartitionState.class);
 
   protected final Set<PartitionInfo> _underReplicatedPartitions;
   protected final Set<PartitionInfo> _offlinePartitions;
@@ -112,25 +115,25 @@ public class ClusterPartitionState {
         }
         for (PartitionInfo partitionInfo : _kafkaCluster.partitionsForTopic(topic)) {
           int numInsyncReplicas = partitionInfo.inSyncReplicas().length;
+          boolean hasOfflineReplicas = partitionInfo.offlineReplicas().length != 0;
           boolean isURP = numInsyncReplicas != partitionInfo.replicas().length;
+          boolean isOffline = partitionInfo.leader() == null;
+
           if (numInsyncReplicas < minInsyncReplicas) {
             underMinIsrPartitions.add(partitionInfo);
           }
-          if (isURP || verbose) {
-            boolean hasOfflineReplica = partitionInfo.offlineReplicas().length != 0;
-            if (hasOfflineReplica) {
-              partitionsWithOfflineReplicas.add(partitionInfo);
-            }
-            boolean isOffline = partitionInfo.inSyncReplicas().length == 0;
-            if (isOffline) {
-              offlinePartitions.add(partitionInfo);
-            } else if (isURP) {
-              underReplicatedPartitions.add(partitionInfo);
-            } else {
-              // verbose -- other
-              otherPartitions.add(partitionInfo);
-            }
+          if (hasOfflineReplicas) {
+            partitionsWithOfflineReplicas.add(partitionInfo);
           }
+
+          if (isOffline) {
+            offlinePartitions.add(partitionInfo);
+          } else if (isURP) {
+            underReplicatedPartitions.add(partitionInfo);
+          } else if (verbose) {
+            otherPartitions.add(partitionInfo);
+          }
+          LOG.trace("partitionInfo: {}", partitionInfo);
         }
       }
     }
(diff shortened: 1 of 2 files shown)


sanaasy (approved):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
```diff
@@ -112,25 +115,24 @@ protected void populateKafkaPartitionState(Set<PartitionInfo> underReplicatedPar
...
+            partitionsWithOfflineReplicas.add(partitionInfo);
+          }
+          if (isURP) {
+            underReplicatedPartitions.add(partitionInfo);
+          } else if (isOffline) {
+            offlinePartitions.add(partitionInfo);
+          } else if (verbose) {
+            otherPartitions.add(partitionInfo);
           }
+          LOG.warn("PI: {} -- LEADER: {}", partitionInfo, partitionInfo.leader());
```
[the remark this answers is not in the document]
vvuibert: could this be useful as `debug` if running verbose?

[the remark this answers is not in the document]
ayana-s: good suggestion - updated ✅

ayana-s:
Hi @CCisGG ! Could we please get a review on this PR, whenever you have the chance to look it over?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
```diff
@@ -112,25 +115,26 @@ protected void populateKafkaPartitionState(Set<PartitionInfo> underReplicatedPar
...
-            } else if (isURP) {
-              underReplicatedPartitions.add(partitionInfo);
-            } else {
-              // verbose -- other
-              otherPartitions.add(partitionInfo);
-            }
+          if (hasOfflineReplicas) {
+            partitionsWithOfflineReplicas.add(partitionInfo);
+          }
+          if (isURP) {
```
CCisGG, line 128: If it is offline, I think we don't want to add it into URP. This is to avoid verbose output.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
```diff
@@ -112,25 +115,26 @@ protected void populateKafkaPartitionState(Set<PartitionInfo> underReplicatedPar
         }
         for (PartitionInfo partitionInfo : _kafkaCluster.partitionsForTopic(topic)) {
           int numInsyncReplicas = partitionInfo.inSyncReplicas().length;
+          boolean hasOfflineReplicas = partitionInfo.offlineReplicas().length != 0;
           boolean isURP = numInsyncReplicas != partitionInfo.replicas().length;
+          boolean isOffline = partitionInfo.leader() == null;
```
CCisGG, line 120: Does this correctly reflect the leader is not alive? E.g. in kafka_cluster_state, it shows the leader is [-1]. Could you help to check why that is -1 while here it is null?

CCisGG (changes requested):
Added some comments

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
ayana-s: Good question - I believe it does. In `PartitionState.java` [(here)](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/PartitionState.java#L54), if the partition leader is null, `_leader` is set to `-1`. It looks like we use `PartitionState` to populate the kafkaClusterState [here](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/resources/yaml/responses/kafkaClusterState.yaml#L121-L148).

The null leader status is used to determine offline partitions elsewhere in the code as well [(example)](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java#L451-L452).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
ayana-s: Updated ✅

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/ClusterPartitionState.java
CCisGG: Perfect. Thank you!

CCisGG (approved):
Looks good to me. Thank you @ayana-s for the great work!

ayana-s:
Thank you @CCisGG ! Wanted to quickly confirm that we don't need to take any further action here, and that a maintainer will merge this sometime?

CCisGG:
@ayana-s I will merge it and create a release soon. No actions from your side. Thank you!

PR #2014, opened 2023-05-12
Initialize leadershipConcurrency for scaled brokers to avoid NPE

### LINKED ISSUE #1975: Exception thrown after starting leadership movements for `add_broker` request
When requesting Cruise Control to balance partitions to new brokers, I have been running into the following exception:
```
2023-03-03 16:27:31 INFO  Executor:1522 - Starting 69 leadership movements.
2023-03-03 16:27:31 ERROR Executor:1288 - Executor got exception during execution
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "java.util.Map.get(Object)" is null
        at com.linkedin.kafka.cruisecontrol.executor.ExecutionTaskPlanner.getLeadershipMovementTasks(ExecutionTaskPlanner.java:322) ~[cruise-control-2.5.113.jar:?]
        at com.linkedin.kafka.cruisecontrol.executor.ExecutionTaskManager.getLeadershipMovementTasks(ExecutionTaskManager.java:120) ~[cruise-control-2.5.113.jar:?]
        at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.moveLeadershipInBatch(Executor.java:1547) ~[cruise-control-2.5.113.jar:?]
        at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.moveLeaderships(Executor.java:1526) ~[cruise-control-2.5.113.jar:?]
        at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.execute(Executor.java:1284) ~[cruise-control-2.5.113.jar:?]
        at com.linkedin.kafka.cruisecontrol.executor.Executor$ProposalExecutionRunnable.run(Executor.java:1179) ~[cruise-control-2.5.113.jar:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) ~[?:?]
        at java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
        at java.lang.Thread.run(Thread.java:833) ~[?:?]
2023-03-03 16:27:31 WARN  operationLogger:1323 - Task [6062ba95-fd9f-469a-805a-54f5729c0567] userPOST /kafkacruisecontrol/add_broker execution is interrupted with exception Cannot invoke "java.lang.Integer.intValue()" because the return value of "java.util.Map.get(Object)" is null.
2023-03-03 16:27:31 INFO  Executor:1180 - Execution finished.
2023-03-03 16:27:34 INFO  UserTaskManager:349 - UserTask 0844f106-3038-4e71-a1e6-e76a2b0a2f37 is completed and removed from active tasks list
```
From what I understand, we need to add a key check like we have here [2] to the following code we have here [1]:
```
for (int broker: replicas) {
  if (leadershipConcurrency.containsKey(broker) && leadershipConcurrency.get(broker) > 0) {   // <--- Added line
    leadershipConcurrency.put(broker, leadershipConcurrency.get(broker) - 1);
  }
}
```
To avoid situations where `leadershipConcurrency.get(broker)` returns `null`.

Could someone confirm this is an issue? If so, I am happy to submit a patch for this if that makes things easier!

**Environment**
Cruise Control version: 2.5.113

**Additional context**
[1] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java#L322
[2] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java#L314

kyguy:
This PR resolves #1975.

Since we don't initialize the movement concurrency configuration for brokers that don't have any replicas, we run into an NPE when trying to move partitions to new brokers and away from stale brokers.

This PR ensures that the leadership concurrency configurations are set for all brokers in the cluster

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
@@ -7,6 +7,7 @@ package com.linkedin.kafka.cruisecontrol.executor;
 import com.linkedin.cruisecontrol.common.utils.Utils;
 import com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils;
 import com.linkedin.kafka.cruisecontrol.config.KafkaCruiseControlConfig;
+import com.linkedin.kafka.cruisecontrol.executor.concurrency.ExecutionConcurrencyManager;
 import com.linkedin.kafka.cruisecontrol.executor.strategy.BaseReplicaMovementStrategy;
 import com.linkedin.kafka.cruisecontrol.executor.strategy.ReplicaMovementStrategy;
 import com.linkedin.kafka.cruisecontrol.executor.strategy.StrategyOptions;
@@ -297,10 +298,12 @@ public class ExecutionTaskPlanner {
    *
    * @param leadershipConcurrencyByBrokerId the leadership movement concurrency of each broker
    * @param clusterLeadershipMovementConcurrency the allowed movement concurrency of the whole cluster
+   * @param executionConcurrencyManager the execution concurrency manager
    * @return The leadership movement tasks.
    */
   public List<ExecutionTask> getLeadershipMovementTasks(Map<Integer, Integer> leadershipConcurrencyByBrokerId,
-                                                        int clusterLeadershipMovementConcurrency) {
+                                                        int clusterLeadershipMovementConcurrency,
+                                                        ExecutionConcurrencyManager executionConcurrencyManager) {
     Map<Integer, Integer> leadershipConcurrency = new HashMap<>(leadershipConcurrencyByBrokerId);
     List<ExecutionTask> leadershipMovementsList = new ArrayList<>();
     Iterator<ExecutionTask> leadershipMovementIter = _remainingLeadershipMovements.values().iterator();
@@ -311,9 +314,13 @@ public class ExecutionTaskPlanner {
           Collectors.toSet());
       boolean canSchedule = true;
       for (int broker: replicas) {
-        if (leadershipConcurrency.containsKey(broker) && leadershipConcurrency.get(broker) <= 0) {
-          canSchedule = false;
-          break;
+        if (leadershipConcurrency.containsKey(broker)) {
+          if (leadershipConcurrency.get(broker) <= 0) {
+            canSchedule = false;
+            break;
+          }
+        } else {
+          leadershipConcurrency.put(broker, executionConcurrencyManager.getExecutionConcurrency(broker, ConcurrencyType.LEADERSHIP));
         }
       }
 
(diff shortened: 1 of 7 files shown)


kyguy:
@CCisGG  could you take a look at this when you get a chance?

CCisGG:
Hi @kyguy , this is definitely a good catch. Thanks for pointing that out! 

When we implemented it, we intentionally handle the missing broker case in the concurrency manager itself. E.g. if there is no concurrency for the broker, it should return the default value. 

Given that, could you try to pass the concurrency manager this method getLeadershipMovementTasks()? Let me know if that makes more sense.

kyguy:
> Given that, could you try to pass the concurrency manager this method getLeadershipMovementTasks()? Let me know if that makes more sense.

I think I understand, how does the next commit look? That is the general idea, right?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
```diff
@@ -297,10 +298,12 @@ public Collection<ExecutionTask> remainingLeadershipMovements() {
    *
    * @param leadershipConcurrencyByBrokerId the leadership movement concurrency of each broker
    * @param clusterLeadershipMovementConcurrency the allowed movement concurrency of the whole cluster
+   * @param executionConcurrencyManager the execution concurrency manager
    * @return The leadership movement tasks.
    */
   public List<ExecutionTask> getLeadershipMovementTasks(Map<Integer, Integer> leadershipConcurrencyByBrokerId,
-                                                        int clusterLeadershipMovementConcurrency) {
+                                                        int clusterLeadershipMovementConcurrency,
```
CCisGG, line 305: Since the executionConcurrencyManager has all the information needed, let's change the method signature to getLeadershipMovementTasks(ExecutionConcurrencyManager executionConcurrencyManager).

CCisGG (changes requested):
Left one more minor change.

CCisGG (changes requested):
Left one more minor change.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
kyguy: Thanks @CCisGG, you are right, that makes a lot more sense, just updated the PR to address this

CCisGG (approved):
LGTM. Thanks @kyguy

kyguy:
Argh I broke the unit tests, I'll fix them and ping you once it is done

kyguy:
Made adjustments to the unit tests to get them passing, if the changes look good to you then we should be good to go!

PR #2019, opened 2023-05-25
Future-proof against Kafka 3.5 breaking changes

### LINKED ISSUE #2018: Future-proof against breaking Kafka 3.5 changes
The upcoming Kafka 3.5 release comes with a change that will break Cruise Control. 

Most notable change is the package name change of:
 
`org.apache.kafka.server.log.internals` -> `org.apache.kafka.storage.internals.log` [1]

This change will cause issues with the metric reporter when run it with Kafka 3.5 brokers:
```
2023-05-23 17:17:15,315 ERROR Exiting Kafka due to fatal exception during startup. (kafka.Kafka$) [main]
java.lang.NoClassDefFoundError: kafka/log/LogConfig
        at com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter.createMetricsTopicFromReporterConfig(CruiseControlMetricsReporter.java:287)
        at com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter.configure(CruiseControlMetricsReporter.java:196)
        at org.apache.kafka.common.config.AbstractConfig.getConfiguredInstance(AbstractConfig.java:401)
        at org.apache.kafka.common.config.AbstractConfig.getConfiguredInstances(AbstractConfig.java:482)
        at kafka.server.DynamicMetricReporterState.createReporters(DynamicBrokerConfig.scala:871)
        at kafka.server.DynamicMetricReporterState.<init>(DynamicBrokerConfig.scala:857)
        at kafka.server.DynamicMetricsReporters.<init>(DynamicBrokerConfig.scala:797)
        at kafka.server.DynamicBrokerConfig.addReconfigurables(DynamicBrokerConfig.scala:265)
        at kafka.server.KafkaServer.startup(KafkaServer.scala:548)
        at kafka.Kafka$.main(Kafka.scala:113)
        at kafka.Kafka.main(Kafka.scala)
Caused by: java.lang.ClassNotFoundException: kafka.log.LogConfig
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
        ... 11 more
```
Since the metric reporter cannot find the the LogConfig class, because the location of the LogConfig class changes with Kafka 3.5.

[1] https://github.com/apache/kafka/pull/13213/files#diff-9d7d92cafcbea44968abada4825979d4eefa9d07bf49c444ce11c0c24b327127

kyguy:
This PR resolves #2018.

This PR future proofs Cruise Control against breaking changes coming with Kafka 3.5. Tested against a Kafka 3.5 cluster and a Kafka 3.4 cluster 

Let me know if it is preferable to submit this against a new  `migrate_to_kafka_3_5` branch. 

Let me know what you think!

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
@@ -424,40 +424,36 @@ class ReplicationThrottleHelper {
       // First we try to get the LogConfig class for Kafka 3.5+
       logConfigClass = Class.forName(LOG_CONFIG_IN_KAFKA_3_5_AND_LATER);
       LOG.info("Found class {} for Kafka 3.5 and newer.", LOG_CONFIG_IN_KAFKA_3_5_AND_LATER);
+      try {
+        Field field = logConfigClass.getDeclaredField(config.toString());
+        return field.get(null).toString();
+      } catch (NoSuchFieldException | IllegalAccessException e) {
+        LOG.info("Field {} not found. We are probably on Kafka 3.4 or older.", config);
+      }
     } catch (ClassNotFoundException e) {
       LOG.info("Class {} not found. We are probably on Kafka 3.4 or older.", LOG_CONFIG_IN_KAFKA_3_5_AND_LATER);
+    }
 
-      // We did not find the LogConfig class from Kafka 3.5+.
-      // So we are probably on older Kafka version => we will try the older class for Kafka 3.4-.
+    // We did not find the LogConfig class or field from Kafka 3.5+.
+    // So we are probably on older Kafka version => we will try the older class for Kafka 3.4-.
+    try {
+      logConfigClass = Class.forName(LOG_CONFIG_IN_KAFKA_3_4_AND_EARLIER);
+      LOG.info("Found class {} for Kafka 3.4 and earlier.", LOG_CONFIG_IN_KAFKA_3_4_AND_EARLIER);
       try {
-        logConfigClass = Class.forName(LOG_CONFIG_IN_KAFKA_3_4_AND_EARLIER);
-        LOG.info("Found class {} for Kafka 3.4 and earlier.", LOG_CONFIG_IN_KAFKA_3_4_AND_EARLIER);
-      } catch (ClassNotFoundException ee) {
-          // No class was found for any Kafka version => we should fail
-          throw new RuntimeException("Failed to find LogConfig class", ee);
+        String nameOfMethod = "";
+        if (config == LogConfig.LEADER_REPLICATION_THROTTLED_REPLICAS_CONFIG) {
+          nameOfMethod = "LeaderReplicationThrottledReplicasProp";
+        } else if (config == LogConfig.FOLLOWER_REPLICATION_THROTTLED_REPLICAS_CONFIG) {
+          nameOfMethod = "FollowerReplicationThrottledReplicasProp";
         }
-    }
-
-    try {
-      Field field = logConfigClass.getDeclaredField(config.toString());
-      return field.get(null).toString();
-    } catch (NoSuchFieldException | IllegalAccessException e) {
-      LOG.info("Field {} not found. We are probably on Kafka 3.4 or older.", config);
-
-      // We did not find the LogConfig class field from Kafka 3.5+ LogConfig class.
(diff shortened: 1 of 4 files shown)


kyguy:
@CCisGG Could you take a look when you get a chance?

CCisGG:
Hi @kyguy, this is a great catch and thanks for making the PR! I'd suggest to follow the pattern in this PR: https://github.com/linkedin/cruise-control/pull/1910.

kyguy:
> I'd suggest to follow the pattern in this PR

Thanks for the review, yes that sounds good to me! I have updated the PR to use reflection in the `ReplicationThrottleHelper` class to detect and use the correct `LogConfig` fields/methods depending on which `LogConfig` class is used - the older or the latest.

I have left the `TopicConfig` changes as they are since they are compatible with Kafka 2.5 and beyond

Let me know what you think!

kyguy:
Hi @CCisGG when you get a chance could you have another pass at this PR?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -390,4 +396,68 @@ static boolean configsEqual(Config configs, Map<String, String> expectedValues)
...
+      try {
+        logConfigClass = Class.forName(LOG_CONFIG_IN_KAFKA_3_4_AND_EARLIER);
+        LOG.info("Found class {} for Kafka 3.4 and earlier.", LOG_CONFIG_IN_KAFKA_3_4_AND_EARLIER);
+      } catch (ClassNotFoundException ee) {
+          // No class was found for any Kafka version => we should fail
+          throw new RuntimeException("Failed to find LogConfig class", ee);
+        }
+    }
+
+    try {
```
CCisGG, line 441: To me, it seems to better put move L442~L443 to L426, and move L451~L458 to L434. E.g. in general, there are only 2 cases: 3.4 or before, and 3.5 or after. We should merge the logic into only these 2 branches.

CCisGG (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
kyguy: @CCisGG  Yeah you are right, that does sound better.  Just refactored code based on the feedback, let me now what you think!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -390,4 +396,64 @@ static boolean configsEqual(Config configs, Map<String, String> expectedValues)
...
+    Class<?> logConfigClass;
+
+    try {
+      // First we try to get the LogConfig class for Kafka 3.5+
+      logConfigClass = Class.forName(LOG_CONFIG_IN_KAFKA_3_5_AND_LATER);
+      LOG.info("Found class {} for Kafka 3.5 and newer.", LOG_CONFIG_IN_KAFKA_3_5_AND_LATER);
+      try {
+        Field field = logConfigClass.getDeclaredField(config.toString());
+        return field.get(null).toString();
+      } catch (NoSuchFieldException | IllegalAccessException e) {
```
CCisGG, line 430: Thanks for the change! I think the error in L431 and error in L434 are likely always atomic. If we are confident enough about it, I think we can just catch one Exception, and log info with something like "Failed to read XXX config from LogConfig class since we are probably on kafka 3.4 or older: + e (the exception string)". That will further simplify the code. Same with L452 and L456. What do you think?

CCisGG (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
kyguy: @CCisGG yes that makes sense, just updated the code!

CCisGG (approved):

PR #2026, opened 2023-06-19
Add metrics for dead brokers

### LINKED ISSUE #1930: Publish new num-dead-broker-or-disk metrics
Currently we publish ``has-dead-brokers-or-disks``. With a replication factor and minISR setting, this would not impact the performance. However, it would be useful to publish the count of such dead brokers so that if the number is higher, we can use that to take any manual actions and investigate.

mohitpali:
This PR resolves #1930.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
@@ -62,6 +62,7 @@ import org.slf4j.LoggerFactory;
 
 import static com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUtils.LOAD_MONITOR_SENSOR;
 import static com.linkedin.kafka.cruisecontrol.config.constants.MonitorConfig.SKIP_LOADING_SAMPLES_CONFIG;
+import static com.linkedin.kafka.cruisecontrol.detector.AnomalyDetectorUtils.MAX_METADATA_WAIT_MS;
 import static com.linkedin.kafka.cruisecontrol.monitor.MonitorUtils.getRackHandleNull;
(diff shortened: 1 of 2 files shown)


mohitpali:
@CCisGG

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/LoadMonitor.java
```diff
@@ -196,6 +197,12 @@ public LoadMonitor(KafkaCruiseControlConfig config, Time time, MetricRegistry dr
     // The cluster has partitions with ISR > replicas (0: No such partitions, 1: Has such partitions)
     dropwizardMetricRegistry.register(MetricRegistry.name(LOAD_MONITOR_SENSOR, "has-partitions-with-isr-greater-than-replicas"),
                                       (Gauge<Integer>) () -> MonitorUtils.hasPartitionsWithIsrGreaterThanReplicas(kafkaCluster()) ? 1 : 0);
+    dropwizardMetricRegistry.register(MetricRegistry.name(LOAD_MONITOR_SENSOR, "dead-brokers-with-replicas"),
+                                      (Gauge<Integer>) () -> deadBrokersWithReplicas(MAX_METADATA_WAIT_MS).size());
```
CCisGG, line 201: Hi Mohit, could you double check whether deadBrokersWithReplicas(MAX_METADATA_WAIT_MS)  will request a full metadata copy from Kafka?

If that's the case, I think make it as part of metric emitting might be problematic especially for large clusters.

mohitpali: It would request metadata from the cluster only if the time is beyond metadataTTL - the default value for this is set to 55 seconds. I believe metadataTTL is the generic setting to prevent excessive calls from the cluster.
A user can always configure this setting to be larger to reduce metadata refreshes.

Even if this sensor did not exist, there is no way to deterministically prevent metadata calls being made every 55 seconds.

Issue #2037, opened 2023-07-14
Support basic auth for the PrometheusMetricSampler


leizor:
The PrometheusMetricSampler does not support querying Prometheus instances that are protected by [basic auth](https://prometheus.io/docs/guides/basic-auth/).

PR #2038, opened 2023-07-14
Add basic auth for Prometheus adapter

leizor:
This PR resolves https://github.com/linkedin/cruise-control/issues/2037.

This PR adds the ability to handle basic auth to the Prometheus adapter.

To use it, set the `prometheus.server.endpoint.auth.basic` configuration value in `cruisecontrol.properties` to your `user:password` (similar to how the `-u` flag of `curl` expects it). It is probably a good idea to inject the environment variable into the config instead of directly setting like detailed in https://github.com/linkedin/cruise-control/issues/1197.

This PR also removes the check for negative ports in `PrometheusMetricSampler.java`. If you're routing requests to your Prometheus instance based on incoming host, you're not always able to specify the port.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/prometheus/PrometheusMetricSampler.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/prometheus/PrometheusMetricSampler.java
@@ -40,6 +40,8 @@ import static com.linkedin.kafka.cruisecontrol.monitor.sampling.SamplingUtils.re
  * <ul>
  *   <li>{@link #PROMETHEUS_SERVER_ENDPOINT_CONFIG}: The config for the HTTP endpoint of the Prometheus server
  *   which is to be used as a source for sampling metrics.</li>
+ *   <li>{@link #PROMETHEUS_SERVER_ENDPOINT_BASIC_AUTH_CONFIG}: The config for the basic auth credentials of the
+ *   Prometheus server which is to be used as a source for sampling metrics in the form of {@code user:password}.</li>
  *   <li>{@link #PROMETHEUS_QUERY_RESOLUTION_STEP_MS_CONFIG}: The config for the resolution of the Prometheus
  *   query made to the server (default: {@link #DEFAULT_PROMETHEUS_QUERY_RESOLUTION_STEP_MS}).
  *   If this is set to 30 seconds for a 2 minutes query interval, the query returns with 4 values, which are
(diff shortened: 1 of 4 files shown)


CCisGG:
Hi @mohitpali , would you mind taking a look at this one? Thanks!

CCisGG (commented):
Looks good to me in general. Tagging @mohitpali to taking another look.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/prometheus/PrometheusMetricSampler.java
```diff
@@ -104,18 +107,16 @@ private void configurePrometheusAdapter(Map<String, ?> configs) {
             throw new ConfigException(String.format(
                 "%s config is required by Prometheus metric sampler", PROMETHEUS_SERVER_ENDPOINT_CONFIG));
         }
+        final String basicAuth = (String) configs.get(PROMETHEUS_SERVER_ENDPOINT_BASIC_AUTH_CONFIG);
 
         try {
             HttpHost host = HttpHost.create(endpoint);
-            if (host.getPort() < 0) {
```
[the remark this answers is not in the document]
leizor: I think it's valid to not specify a port (and therefore get a negative result from `host.getPort()`).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/prometheus/PrometheusMetricSampler.java
```diff
@@ -104,18 +107,16 @@ private void configurePrometheusAdapter(Map<String, ?> configs) {
             throw new ConfigException(String.format(
                 "%s config is required by Prometheus metric sampler", PROMETHEUS_SERVER_ENDPOINT_CONFIG));
         }
+        final String basicAuth = (String) configs.get(PROMETHEUS_SERVER_ENDPOINT_BASIC_AUTH_CONFIG);
```
mohitpali, line 110: Thanks for submitting this PR.

* While basic auth is a good start, I would suggest we introduce an interface called something like ``PrometheusAuthenticator`` with an ``addAuth(httpRequest)`` method. The Sampler can then accept the interface type as an argument instead of taking a String. You can implement a PrometheusBasicAuthenticator and use that implementation. 

* I would also recommend a PrometheusNoOpAuthenticator as a default implementation.
* Add a configuration that can be changed by user to provide the Authenticator class.


Since a lot of AWS MSK customers use the Open Monitoring, they will continue to use the NoOpAuthenticator which would be default. Amazon Managed Prometheus has a sigv4 signing authenticator and if someone is using AMP, then they could write or add a AMP based implementation that signs the http request. See this - https://docs.aws.amazon.com/prometheus/latest/userguide/AMP-onboard-query.html

I am not trying to look at it from only AWS perspective but a holistic Prometheus integration with CC. 

More on Prometheus auth can be found here - https://www.robustperception.io/prometheus-security-authentication-authorization-and-encryption/

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/sampling/prometheus/PrometheusMetricSampler.java
```diff
@@ -52,6 +54,7 @@
 public class PrometheusMetricSampler extends AbstractMetricSampler {
     // Config name visible to tests
     static final String PROMETHEUS_SERVER_ENDPOINT_CONFIG = "prometheus.server.endpoint";
+    static final String PROMETHEUS_SERVER_ENDPOINT_BASIC_AUTH_CONFIG = "prometheus.server.endpoint.auth.basic";
```
mohitpali, line 57: Whenever you are adding any configuration, please also add it to the documentation like configurations.md

mohitpali (changes requested):

PR #2043, opened 2023-08-03
Kerberos auth to local rules support

### LINKED ISSUE #2042: Kerberos auth to local rules support
ATL (auth to local rules) support does not exist yet, short principals will be used in authentication / authorization.
This impacts custom Kerberos principals and any use case relying on Kerberos auth to local rules.

--- o ---
**Details:**
In production environments, there is a need for customizing Kerberos principals. Furthermore, when somebody working with a Kafka cluster, then usually there are additional services like Ranger, Cruise Control and so on. Currently, Cruise Control does not support ATL (auth to local) rules.

By default, the Hadoop and other auth-to-local rules map a principal of the form <username>/<hostname>@<REALM> to <username>. This means if there are multiple clusters in the same realm, then principals associated with hosts of one cluster would map to the same user in all other clusters. Of course, Cruise Control will be able to handle other type of ATL rules as well with this feature (not only the default: "<username>/<hostname>@<REALM> to <username>").

For example, if you have two clusters, cluster1-host-[1..4].example.com and cluster2-host- [1..4].example.com, that are part of the same Kerberos realm, EXAMPLE.COM, then the cluster2 principal, kafka/cluster2-host1.example.com@EXAMPLE.COM, will map to the kafka user even on cluster1 hosts.

To prevent this, use auth-to-local rules as follows to ensure only principals containing hostnames of cluster1 are mapped to legitimate users. So, this means for Cruise Control that it will be able to handle only the properly authenticated and authorized user principals and there will be no circumstances when a long user principal will be mapped to a valid short principal form.
  

--- o ---
**The task is the following:**
- extend  SPNEGO login service to be able to handle ATLs
- extend TrustedProxy login service to be able to handle ATLs
- (fallback to SPNEGO will be using the extended SPNEGO login service)
- add `spnego.principal.to.local.rules` to WebServerConfig -> "A list of rules for mapping from principal names to short names (typically operating system usernames). The rules are evaluated in order and the first rule that matches a principal name is used to map it to a short name. Any later rules in the list are ignored. By default, principal names of the form {username}/{hostname}@{REALM} are mapped to {username}. When not specified, the short name will be used."

egyedt:
This PR resolves #2042.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/WebServerConfig.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/WebServerConfig.java
@@ -315,6 +315,18 @@ public final class WebServerConfig {
       + "in the service/host@REALM format (service is usually HTTP).";
 
   /**
+   * <code>spnego.principal.to.local.rules</code>
+   */
+  public static final String SPNEGO_PRINCIPAL_TO_LOCAL_RULES_CONFIG =
+          "spnego.principal.to.local.rules";
+  public static final String DEFAULT_SPNEGO_PRINCIPAL_TO_LOCAL_RULES = null;
+  public static final String SPNEGO_PRINCIPAL_TO_LOCAL_RULES_DOC = "A list of rules for mapping from principal "
+          + "names to short names (typically operating system usernames). The rules are evaluated in order and the "
+          + "first rule that matches a principal name is used to map it to a short name. Any later rules in the list are "
+          + "ignored. By default, principal names of the form <code>{username}/{hostname}@{REALM}</code> are mapped "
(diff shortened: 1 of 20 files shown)


egyedt:
@CCisGG could you please review this PR?

mhratson:
@egyedt thanks for the contribution. To help me review the PR could you please add some documentation, to introduce the problem and the solution, as well as end user documentation. 

Thanks!

egyedt:
@mhratson 
I extended the issue description here: 
https://github.com/linkedin/cruise-control/issues/2042
and added new part to the docs:
https://github.com/linkedin/cruise-control/blob/4c0c66144288077d0c45b2eca8b97f3c7480b022/docs/wiki/User%20Guide/Security.md

efeg (approved):

PR #2051, opened 2023-08-22
Support auto adjusting both broker and cluster leader move concurrency

hshi2022:
This PR resolves [#<2056>](https://github.com/linkedin/cruise-control/issues/2056).
Concurrency Adjuster currently does not take effect for leadership concurrency. This issue is because cc currently adjusts per broker concurrency (the max number of partitions to elect leader in a broker). However, for leadership movement, the cluster concurrency (the max total number of partitions to elect leader in a cluster) equals broker concurrency initially, of which the value is set by num.concurrent.leader.movements. That results in increasing per broker concurrency would not increase partitions in one request because it is still bounded by the cluster concurrency. 

This PR is to fix the above issue.

In this PR:
(1) support configuring broker and cluster leadership movement concurrency separately (num.concurrent.leader.movements.per.broker and num.concurrent.leader.movements)
(2) support auto adjusting both broker and cluster leader move concurrency. 
(3) support parameter broker_concurrent_leader_movements in Rest API for setting the upper bound of per broker ongoing leadership movements 

The leadership movement task is more bottleneck on controller, so in general, the cluster allowed concurrency is smaller than the broker allowed concurrency multiplied by the number of brokers in the cluster. It would be safer and more efficient to have both cluster and broker concurrency adjusters for leadership movements. If with broker concurrency adjustor alone, the execution of leadership movements might be unnecessarily slow due to low cluster concurrency; if with cluster concurrency adjuster alone, we do not have the ability to limit the traffic to slow brokers.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -235,6 +248,12 @@ static ConcurrencyAdjustingRecommendation recommendedConcurrency(Map<BrokerEntit
       }
     }
 
+    if (allBrokersWithinAdjusterLimit) {
```
CCisGG, line 251: Would you explain why do we want to increase cluster concurrency instead of broker concurrency when all brokers within adjuster limit?

Is that because concurrency adjuster tries its best to raise the broker concurrency, but still not able to hit the adjuster limit, meaning the cluster concurrency is too low currently?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -563,28 +582,60 @@ private ConcurrencyAdjustingRecommendation minIsrBasedConcurrency() {
...
-      if (currentMovementConcurrency > minMovementsConcurrency) {
-        int recommendedConcurrency = Math.max(minMovementsConcurrency, currentMovementConcurrency / MULTIPLICATIVE_DECREASE.get(concurrencyType));
-        _executionConcurrencyManager.setExecutionConcurrencyForBroker(brokerId, recommendedConcurrency, concurrencyType);
-        LOG.info("Concurrency adjuster decreased the {} movement concurrency to {} for broker {}",
-                 concurrencyType, recommendedConcurrency, brokerId);
+      if (currentConcurrency > minMovementsConcurrency) {
+        // Multiplicative-decrease reassignment concurrency (MIN: minMovementsConcurrency).
+        return Math.max(minMovementsConcurrency, currentConcurrency / MULTIPLICATIVE_DECREASE.get(concurrencyType));
       }
+      return -1;
```
CCisGG, line 610: Why not just return minMovementConcurrency here so that you don't need to check if "(decreasedConcurrency > 0) " in L598?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -563,28 +582,60 @@ private ConcurrencyAdjustingRecommendation minIsrBasedConcurrency() {
...
-      if (currentMovementConcurrency < maxMovementsConcurrency) {
-        int recommendedConcurrency = Math.min(maxMovementsConcurrency, currentMovementConcurrency + ADDITIVE_INCREASE.get(concurrencyType));
-        _executionConcurrencyManager.setExecutionConcurrencyForBroker(brokerId, recommendedConcurrency, concurrencyType);
-        LOG.info("Concurrency adjuster increased the {} movement concurrency to {} for broker {}",
-                 concurrencyType, recommendedConcurrency, brokerId);
+      if (currentConcurrency < maxMovementsConcurrency) {
+        // Additive-increase reassignment concurrency (MAX: maxMovementsConcurrency).
+        return Math.min(maxMovementsConcurrency, currentConcurrency + ADDITIVE_INCREASE.get(concurrencyType));
       }
+      return -1;
```
CCisGG, line 638: Ditto

CCisGG (commented):
Half way done. Currently completed review for Executor.java.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ConcurrencyAdjustingRecommendation.java
```diff
@@ -24,22 +24,32 @@ private enum ConcurrencyAdjustingRecommendationType {
...
+    DECREASE
+  }
+
   public static final ConcurrencyAdjustingRecommendation NO_CHANGE_RECOMMENDED
       = new ConcurrencyAdjustingRecommendation(ConcurrencyAdjustingRecommendationType.NO_CHANGE_RECOMMENDED);
   public static final ConcurrencyAdjustingRecommendation STOP_EXECUTION =
       new ConcurrencyAdjustingRecommendation(ConcurrencyAdjustingRecommendationType.STOP_EXECUTION);
   private final Set<Integer> _brokersToIncreaseConcurrency;
   private final Set<Integer> _brokersToDecreaseConcurrency;
+  private ClusterConcurrencyAdjustType _clusterConcurrencyAdjustRecommendation;
```
CCisGG, line 39: nit: Group all the final vars together, and non-final vars together.

CCisGG: Actually why is this value not final?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ExecutionConcurrencyManager.java
```diff
@@ -131,39 +150,53 @@ public boolean isInitialized() {
...
         throw new IllegalArgumentException("Unsupported concurrency type " + concurrencyType + " is provided.");
     }
   }
 
+  /**
+   * Retrieve the movement concurrency of the given cluster concurrency type.
+   * @param concurrencyType The type of concurrency for which the allowed movement concurrency is requested.
+   * @return The movement concurrency of the given concurrency type.
+   */
+  public synchronized int getExecutionClusterConcurrency(ConcurrencyType concurrencyType) {
```
CCisGG, line 177: You don't need to pass in the concurrencyType param here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/sync/AdminRequest.java
```diff
@@ -129,12 +129,20 @@ protected String processChangeExecutionConcurrencyRequest() {
       LOG.info("Intra-broker partition movement concurrency is set to: {} by user.", concurrentIntraBrokerPartitionMovements);
     }
     // 3. Change leadership concurrency for all brokers.
```
CCisGG, line 131: Since now we have the cluster concurrency, we should change this to "Change cluster leadership concurrency". And rename setExecutionConcurrencyForAllBrokers on L134 to something like setClusterExecutionConcurrency

CCisGG (commented):
Completed review half-way. Now completed adminRequest

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/AddedOrRemovedBrokerParameters.java
```diff
@@ -71,9 +75,10 @@ protected void initParameters() throws UnsupportedEncodingException {
     super.initParameters();
     _brokerIds = ParameterUtils.brokerIds(_requestContext, false);
     _dryRun = ParameterUtils.getDryRun(_requestContext);
-    _concurrentInterBrokerPartitionMovements = ParameterUtils.concurrentMovements(_requestContext, true, false);
+    _concurrentInterBrokerPartitionMovements = ParameterUtils.concurrentMovements(_requestContext, ConcurrencyType.INTER_BROKER_REPLICA);
```
CCisGG, line 78: Thumb up on this change.

CCisGG (commented):
Finished one pass.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
hshi2022: Thought more on this, actually, if the concurrency is explicitly set to a small value through rest api (< minMovementsConcurrency), we should let it as. So change it return currentConcurrency if no need change.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
hshi2022: updated as above

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ExecutionConcurrencyManager.java
CCisGG: I think we can make it this way only when we have the need for adding other cluster currency. I would prefer to make it getExecutionClusterLeadershipConcurrency

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -905,13 +905,13 @@ public synchronized void executeDemoteProposals(Collection<ExecutionProposal> pr
...
-   * Dynamically set the movement concurrency of the given type for all brokers.
+   * Dynamically set the per broker movement concurrency of the given type for all brokers or set the cluster concurrency.
    *
    * @param concurrency The maximum number of concurrent movements.
    * @param concurrencyType The type of concurrency for which the requested movement concurrency will be set.
    */
-  public synchronized void setExecutionConcurrencyForAllBrokers(Integer concurrency, ConcurrencyType concurrencyType) {
-    _executionTaskManager.getExecutionConcurrencyManager().setExecutionConcurrencyForAllBrokers(concurrency, concurrencyType);
+  public synchronized void setExecutionConcurrencyForAllBrokersOrCluster(Integer concurrency, ConcurrencyType concurrencyType) {
+    _executionTaskManager.getExecutionConcurrencyManager().setExecutionConcurrencyForAllBrokersOrCluster(concurrency, concurrencyType);
```
CCisGG, line 914: Just curious, why not make 2 functions?
1. setExecutionConcurrencyForAllBrokers
2. setExecutionConcurrencyForCluster
That seems to be more clear and makes more sense to me

CCisGG (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: btw since leadership is the only cluster level concurrency type, we can name it to setExecutionClusterLeadershipConcurrency

hshi2022: I do not think there is big difference in the functionality between setting concurrency for all brokers and for cluster. And one function makes the code more concise. Otherwise, there would be repeated code in both functions like "sanityCheckRequestedConcurrency(concurrency, concurrencyType);" and " default:
        throw new IllegalArgumentException("Unsupported concurrency type " + concurrencyType + " is provided.");
    }
    refreshExecutionConcurrencySummary();"

CCisGG: > I do not think there is big difference in the functionality between setting concurrency for all brokers and for cluster
Just curious, for "all brokers", we are setting a same concurrency value for each single broker, while for cluster concurrency, we are setting a single value for cluster level concurrency. Is that correct? In that case, I think they are still fundamentally different. E.g. In the concurrencyManager, we have separate functions for set broker concurrency and set cluster concurrency. Let me know if that makes sense.

hshi2022: In ExecutionConcurrencyManager, there are two methods setExecutionConcurrencyForAllBrokersOrCluster and setExecutionConcurrencyForBroker, so for allBrokers and Cluster there is also one method. setExecutionConcurrencyForBroker is a separate method because the method take an additional parameter (brokerId).

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/concurrency/ExecutionConcurrencyManager.java
hshi2022: Make sense. Also updated its caller method's name to be more specific for leadership.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: ack

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -225,7 +230,7 @@ static ConcurrencyAdjustingRecommendation recommendedConcurrency(Map<BrokerEntit
       overLimitDetailsByMetricName.put(metricName, new StringBuilder());
     }
 
-    boolean allBrokersWithinAdjusterLimit = true;
+    int numBrokerAboveMetricLimit = 0;
```
CCisGG, line 233: nit: numBrokersAboveMetricLimit

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
```diff
@@ -135,6 +138,8 @@ static void init(KafkaCruiseControlConfig config) {
                         config.getInt(ExecutorConfig.CONCURRENCY_ADJUSTER_MIN_LEADERSHIP_MOVEMENTS_PER_BROKER_CONFIG));
     listPartitionReassignmentsTimeoutMs = config.getLong(ExecutorConfig.LIST_PARTITION_REASSIGNMENTS_TIMEOUT_MS_CONFIG);
     listPartitionReassignmentsMaxAttempts = config.getInt(ExecutorConfig.LIST_PARTITION_REASSIGNMENTS_MAX_ATTEMPTS_CONFIG);
+    maxNumBrokersAboveMetricLimitNoDecreaseClusterConcurrency =
```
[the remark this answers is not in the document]
CCisGG: And if this is not caught by test, likely we are missing test coverage to the new code added in this commit, or the test did not work properly.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
```diff
@@ -524,6 +524,15 @@ public final class ExecutorConfig {
       + "per broker concurrency cap on leadership movements will be divided by the concurrency adjuster (if enabled) when any "
       + "considered metric exceeds the concurrency adjuster limit.";
 
+  /**
+   * <code>max.num.brokers.above.metric.limit.no.decrease.cluster.concurrency</code>
+   */
+  public static final String MAX_NUM_BROKERS_ABOVE_METRIC_LIMIT_NO_DECREASE_CLUSTER_CONCURRENCY_CONFIG
+      = "max.num.brokers.above.metric.limit.no.decrease.cluster.concurrency";
```
CCisGG, line 531: The name looks a bit weird to me, but I don't have a good idea on this config either. How about "min.num.brokers.violate.metric.limit.to.decrease.cluster.concurrency". Notice I intentionally change the name to "min"

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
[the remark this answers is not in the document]
hshi2022: I think it should be Math.max. If the config is <0, e.g., -4, Math.max(0, -4) = 0; and if the config is normal above 0, e..g, 2, Math.max(0,2) = 2. And this is expected.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
hshi2022: changed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
hshi2022: changed

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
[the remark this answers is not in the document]
CCisGG: You are absolutely right.

PR #2053, opened 2023-08-25
Use ExecutorService.execute where return value is unnecessary to make uncaught exception handler works fixes #2052

### LINKED ISSUE #2052: Uncaught exceptions on the executor tasks are not logged
- Despite CC's default thread factory [sets uncaught exception handler](https://github.com/linkedin/cruise-control/blob/28395928a2fed7dd01ff48efedd27ab351ca912a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/KafkaCruiseControlThreadFactory.java#L34) to log exceptions, it's not working as expected now because in most places, `ExecutorService#submit` is used instead of `ExecutorService#execute`
  * refs: https://stackoverflow.com/questions/1838923/why-is-uncaughtexceptionhandler-not-called-by-executorservice/29050729#29050729

ocadaruma:
This PR resolves #2052 

- We should use `ExecutorService#execute` to make uncaught exception handler works
  * In most places, we don't need return value of `submit`

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
@@ -222,7 +222,7 @@ public class KafkaCruiseControl {
     LOG.info("Starting Kafka Cruise Control...");
     _loadMonitor.startUp();
     _anomalyDetectorManager.startDetection();
-    _goalOptimizerExecutor.submit(_goalOptimizer);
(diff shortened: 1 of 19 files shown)


CCisGG (commented):
This is a very good catch and fix! 

Could you verify that rebalance still works smoothly after the fix?

ocadaruma:
@CCisGG Yes. On our environment, rebalances are working with patched version

CCisGG (approved):

CCisGG:
Merged. Thanks @ocadaruma !

PR #2054, opened 2023-08-29
Enable info level log after each goal optimization

CCisGG:
This PR is to enable info level log after each goal optimization.

Previously the log is a debug level log. Since we believe this log expose critical information of Cruise Control at runtime, we enabled the log in this PR. In addition, we add provision status to the log to show what of the provision status of each goal.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
@@ -481,13 +481,19 @@ public class GoalOptimizer implements Runnable {
         violatedGoalNamesAfterOptimization.add(goal.name());
       }
 
-      LOG.debug("[{}/{}] Generated {} proposals for {}{}.", optimizedGoals.size(), _goalsByPriority.size(), hasDiff ? "some" : "no",
-                isSelfHealing ? "self-healing " : "", goal.name());
       step.done();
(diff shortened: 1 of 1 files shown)


hshi2022 (approved):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java
```diff
@@ -481,13 +481,19 @@ public OptimizerResult optimizations(ClusterModel clusterModel,
         violatedGoalNamesAfterOptimization.add(goal.name());
       }
 
-      LOG.debug("[{}/{}] Generated {} proposals for {}{}.", optimizedGoals.size(), _goalsByPriority.size(), hasDiff ? "some" : "no",
-                isSelfHealing ? "self-healing " : "", goal.name());
       step.done();
       if (LOG.isDebugEnabled()) {
         LOG.debug("Broker level stats after optimization: {}", clusterModel.brokerStats(null));
```
hshi2022, line 486: Is it necessary to also add this to Info level

CCisGG: Good question!

I believe this is going to be very helpful for identifying production issues.

Also this will not make the logs look verbose, because the code path is only hit on start of each rebalance. It will provide critical meta-level information of the goal optimzation.

PR #2067, opened 2023-10-05
Add timers for proposal execution

hshi2022:
Add timers to measure the duration and count of of proposal executions. Separate timers are added for proposal executions involving broker removal and demotion.


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1316,12 +1335,31 @@ private class ProposalExecutionRunnable implements Runnable {
...
+      } else {
+        _executionTimerInvolveBrokerRemovalOrDemotion = null;
+      }
     }
 
     public void run() {
       LOG.info("Starting executing balancing proposals.");
-      UserTaskManager.UserTaskInfo userTaskInfo = initExecution();
-      execute(userTaskInfo);
+      final Timer.Context ctxExecutionBrokerRemovalOrDemotion = _executionTimerInvolveBrokerRemovalOrDemotion != null
```
CCisGG, line 1349: Do you actually want a timer or just a gauge? 

I think a timer helps to get statistical results like avg, p99, etc, especially when there are many concurrent operations. In this case, since there is only one execution at a time, I think a gauge might be more helpful. What do you think

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -133,6 +134,17 @@ public class Executor {
...
   private final AtomicDouble _partitionMovementCountPerSec;
   private final AtomicDouble _partitionMovementMbPerSec;
+  // The timers _proposalExecutionTimerInvolveBrokerRemoval and _proposalExecutionTimerInvolveBrokerDemotionOnly
+  // would be useful in estimating the cluster recover time during broker failures.
+  // Proposal execution for broker demotion would be faster compared to that for broker removal in general,
+  // so we have two metrics:
+  // (1) timer for broker removal (does not matter whether there is also broker demotion or not)
+  // (2) timer for broker demotion only
+  // Timer of executing proposal involved with broker removal (may or may not with broker demotion).
+  private final Timer _proposalExecutionTimerInvolveBrokerRemoval;
```
CCisGG, line 144: > (may or may not with broker demotion).

To my understand, the removal process does not include any demotion. In that case, we can probably just name the var below as _proposalExecutionTimerInvolveBrokerDemotion;

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1316,12 +1335,31 @@ private class ProposalExecutionRunnable implements Runnable {
...
+          ? _executionTimerInvolveBrokerRemovalOrDemotion.time() : null;
+      final Timer.Context ctxExecution = _proposalExecutionTimer.time();
+      try {
+        UserTaskManager.UserTaskInfo userTaskInfo = initExecution();
+        execute(userTaskInfo);
+      } catch (Exception e) {
+        LOG.error("ProposalExecutionRunnable got exception during run", e);
+      } finally {
+        if (ctxExecutionBrokerRemovalOrDemotion != null) {
+          ctxExecutionBrokerRemovalOrDemotion.stop();
```
CCisGG, line 1359: I'm thinking about another way of implementing it. E.g. have a startTime and an endTime, and then mark each timer with endTime-startTime. 

One potential issue of the current implementation is you need to take care of each individual timer logic here. If you have lots of sensors to mark conditionally, it might be a bit complicated. Eventually we only have a single value to mark, and you can implement the mark timer logic separately regardless how many timers you have. That would decouple the calculate time logic and mark timer logic.

CCisGG (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
groelofs: Yes, rebalances for broker removal tend to be pretty rare events from what I've seen, so a sticky gauge is probably fine.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
hshi2022: Changed to update the timer directly instead of using Context.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: Ok that a fair point.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1316,12 +1335,30 @@ private class ProposalExecutionRunnable implements Runnable {
...
+      } else {
+        _executionTimerInvolveBrokerRemovalOrDemotion = null;
+      }
     }
 
     public void run() {
       LOG.info("Starting executing balancing proposals.");
-      UserTaskManager.UserTaskInfo userTaskInfo = initExecution();
-      execute(userTaskInfo);
+      final long start = System.nanoTime();
```
CCisGG, line 1349: You probably don't want the nanoseconds as the unit. We usually use milliseconds.

Change it to something like` final long startTimeMs = System.miliseconds();`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: Sounds good.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1316,12 +1335,30 @@ private class ProposalExecutionRunnable implements Runnable {
       _replicationThrottle = replicationThrottle;
       _isTriggeredByUserRequest = isTriggeredByUserRequest;
       _lastSlowTaskReportingTimeMs = -1L;
+      if (removedBrokers != null && !removedBrokers.isEmpty()) {
```
CCisGG, line 1338: nit: use Collections.isEmpty() instead

CCisGG (changes requested):

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: nvm.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
hshi2022: done

CCisGG (approved):

PR #2072, opened 2023-10-27
Add anomaly detect-to-fix-complete-timer metrics

hshi2022:
Add anomaly detect-to-fix-complete-timer metrics which measures the time from anomaly being detected to self-healing completed.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
@@ -7,6 +7,7 @@ package com.linkedin.kafka.cruisecontrol.detector;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
 import com.linkedin.cruisecontrol.detector.Anomaly;
 import com.linkedin.cruisecontrol.detector.AnomalyType;
 import com.linkedin.kafka.cruisecontrol.detector.notifier.AnomalyNotifier;
@@ -18,6 +19,7 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
@@ -78,6 +80,7 @@ public class AnomalyDetectorState {
   private final Map<AnomalyType, Meter> _anomalyRateByType;
   private double _balancednessScore;
   private boolean _hasUnfixableGoals;
+  private final Map<AnomalyType, Timer> _anomalyDetectToFixCompleteTimer;
 
   public AnomalyDetectorState(Time time,
                               AnomalyNotifier anomalyNotifier,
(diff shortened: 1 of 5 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -287,12 +295,17 @@ private SelfHealingEnabledRatio selfHealingEnabledRatio() {
...
+   * @param completeWithError Whether the task execution finished with error.
    */
-  public synchronized void markSelfHealingFinished(String anomalyId) {
+  public synchronized void markSelfHealingFinished(String anomalyId, boolean completeWithError) {
     if (_ongoingSelfHealingAnomaly == null || !_ongoingSelfHealingAnomaly.anomalyId().equals(anomalyId)) {
       throw new IllegalStateException(String.format("Anomaly %s is not marked as %s state in AnomalyDetector.",
                                                     anomalyId, AnomalyState.Status.FIX_STARTED));
     }
+    long totalTime = _time.milliseconds() - _ongoingSelfHealingAnomaly.detectionTimeMs();
+    LOG.info("Self-healing for anomalyId {} completed{}. Total time: {} ms", anomalyId, completeWithError ? " with error" : "", totalTime);
```
groelofs, line 306: nit: it would be slightly easier to grep for successful ones if the string were "... completed successfully." for them (instead of just "... completed.")

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
```diff
@@ -287,12 +295,17 @@ private SelfHealingEnabledRatio selfHealingEnabledRatio() {
...
+    // Time the duration no matter whether the self-healing is completed successfully or not.
+    _anomalyDetectToFixCompleteTimer.get(_ongoingSelfHealingAnomaly.anomalyType()).update(totalTime, TimeUnit.MILLISECONDS);
```
groelofs, line 308: I don't feel like this is a useful metric--how can we determine (from graphs) which ones succeeded and which failed? Separating those into two metrics would be much better since then we'd have both "time to fix the anomaly" and "time wasted attempting to heal but failing at it."

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
hshi2022: updated

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/AnomalyDetectorState.java
hshi2022: Changed to only time the successful fix.  do not have a separate timer to track the failed fixes because the duration for failed fix is not very meaningful and there are other metrics/logs for tracking failures. There are two types of failures: (1) interrupted (receiving stop signal). There is already a counter to track the number of stopped execution. (2) encountering exception. This is unexpected and rare. We can use log if needed.

groelofs: Sounds good, thanks.

groelofs (approved):
Looks great, thanks!

PR #2075, opened 2023-10-31
Add operation log for execution time.

CCisGG:
This PR adds operation logs for the task execution time.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1391,7 +1391,7 @@ public class Executor {
         if (_executionException != null) {
           LOG.info("Execution failed: {}. Exception: {}", executionStatusString, _executionException.getMessage());
         } else {
-          LOG.info("Execution finished: {}. ", executionStatusString);
+          LOG.info("Execution succeeded: {}. ", executionStatusString);
         }
       }
     }
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1381,8 +1381,8 @@ public class Executor {
           _executionTimerInvolveBrokerRemovalOrDemotion.update(duration, TimeUnit.MILLISECONDS);
         }
 
-        OPERATION_LOG.info("Execution Finished: task Id: {}; removed brokers: {}; demoted brokers: {}; total time used: {}ms.",
-                           _userTaskInfo != null ? _userTaskInfo.userTaskId() : "N/A (triggered by self-healing)",
+        OPERATION_LOG.info("Execution finished: task Id: {}; removed brokers: {}; demoted brokers: {}; total time used: {}ms.",
+                           _uuid,
                            _removedBrokers,
                            _demotedBrokers,
                            duration);
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1382,7 +1382,7 @@ public class Executor {
         }
 
         OPERATION_LOG.info("Execution Finished: task Id: {}; removed brokers: {}; demoted brokers: {}; total time used (in milliseconds): {}.",
-                           _userTaskInfo != null ? _userTaskInfo.userTaskId(): "N/A (triggered by self-healing)",
+                           _userTaskInfo != null ? _userTaskInfo.userTaskId() : "N/A (triggered by self-healing)",
                            _removedBrokers,
                            _demotedBrokers,
                            duration);
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1381,7 +1381,7 @@ public class Executor {
           _executionTimerInvolveBrokerRemovalOrDemotion.update(duration, TimeUnit.MILLISECONDS);
         }
 
-        String executionStatusString = String.format(" task Id: %s; removed brokers: %s; demoted brokers: %s; total time used: %dms.",
+        String executionStatusString = String.format("task Id: %s; removed brokers: %s; demoted brokers: %s; total time used: %dms.",
                                                _uuid,
                                                _removedBrokers,
                                                _demotedBrokers,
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1374,6 +1380,12 @@ public void run() {
         if (_executionTimerInvolveBrokerRemovalOrDemotion != null) {
           _executionTimerInvolveBrokerRemovalOrDemotion.update(duration, TimeUnit.MILLISECONDS);
         }
+
+        OPERATION_LOG.info("Execution Finished: task Id: {}; removed brokers: {}; demoted brokers: {}; total time used (in milliseconds): {}.",
```
mhratson, line 1384:
```suggestion
        OPERATION_LOG.info("Execution Finished: task Id: {}; removed brokers: {}; demoted brokers: {}; total time used: {}ms.",
```

mhratson (approved):
LGTM

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1374,6 +1380,12 @@ public void run() {
         if (_executionTimerInvolveBrokerRemovalOrDemotion != null) {
           _executionTimerInvolveBrokerRemovalOrDemotion.update(duration, TimeUnit.MILLISECONDS);
         }
+
+        OPERATION_LOG.info("Execution Finished: task Id: {}; removed brokers: {}; demoted brokers: {}; total time used: {}ms.",
```
groelofs, line 1384: nit: s/Finished/finished/  and please insert "successfully" as well

Where do interrupted or failed executions get logged? We should also track the time spent on those since, for example, a human-interrupted self-healing could mean we miss our Nimbus SLA.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1374,6 +1380,12 @@ public void run() {
...
+                           _userTaskInfo != null ? _userTaskInfo.userTaskId() : "N/A (triggered by self-healing)",
```
groelofs, line 1385: I thought there was always a task ID, even for self-healing? I'm pretty sure the emails include one, and the existing logs certainly do.

groelofs (changes requested):
I'm completely unfamiliar with CC's code base, but adding a new log seems like the wrong approach when there's already an existing log that just needs timing info:

CCisGG:
> I'm completely unfamiliar with CC's code base, but adding a new log seems like the wrong approach when there's already an existing log that just needs timing info:

@groelofs Thanks for the review! Yeah I was aware of it and hesitated between which place is better to add the log. I finally decided to put a new one there because 1. Huilin's new timer was updated there, so I want to keep the time to be consistent to avoid any confusions. 2. This place has access to removed/demoted brokers info that also can be added to logs.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: Interesting. Let me double check.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: Fixed the nit but didn't add successfully. (It could also be failure or interruptions)

I believe this place will always be reached regardless of failure or interruption.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: You are completely right. Looks like uuid always exists while userTaskInfo does not. Thanks for the catch!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: @groelofs do you have any other concerns about this one?

groelofs: Do you have a way to distinguish success/failure? That's absolutely critical info, so I'd like it included. (The other logging location does indicate this.)

Also, since you're adding it here rather than the other location, can you include the UUID?

CCisGG: > That's absolutely critical info, so I'd like it included. 

Good point. I have included that piece of information in the latest commit. 

The latest commit will show exceptions if execution failed, but it doesn't show interruptions. I think if we want to also include interruptions in the log, I'll prefer to just using the existing log lines + execution duration. The drawback is it will miss the information of whether it's demote/remove brokers.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1374,8 +1380,20 @@ public void run() {
         if (_executionTimerInvolveBrokerRemovalOrDemotion != null) {
           _executionTimerInvolveBrokerRemovalOrDemotion.update(duration, TimeUnit.MILLISECONDS);
         }
+
+        String executionStatusString = String.format(" task Id: %s; removed brokers: %s; demoted brokers: %s; total time used: %dms.",
```
groelofs, line 1384: nit: you don't need the leading space since both log messages have one (alternatively, nuke those and keep this one)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1374,8 +1380,20 @@ public void run() {
...
+                                               _uuid,
+                                               _removedBrokers,
+                                               _demotedBrokers,
+                                               duration
+                                               );
+
+        if (_executionException != null) {
+          LOG.info("Execution failed: {}. Exception: {}", executionStatusString, _executionException.getMessage());
+        } else {
+          LOG.info("Execution finished: {}. ", executionStatusString);
```
groelofs, line 1394: For consistency (or better consistency, anyway) with the other logging, could you s/finished/succeeded/  ?  "Finished" is a little bit ambiguous.

groelofs (approved):
Minor nits (sorry!) but looks good, thanks.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: Good catch! Updated.

Issue #2078, opened 2023-11-13
Cruise control rebalance hangs when setting static broker configurations along with `default.replication.throttle`


rosakng:
Today the way Cruise Control removes broker throttle rate configurations during a rebalance is to [delete](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java#L342) the current configurations for leader and follower replication throttled rate, then expect those configurations to be `null` after removing them and before continuing on with the rebalance. Though this is true in the case of removing dynamic broker configurations, however, if there is a static broker configuration set for `leader.replication.throttled.rate` or `follower.replication.throttled.rate`, the configurations will never equal in the [check](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java#L376), leaving the rebalance to hang and exhaust all of its retries without continuing. 

The state of cruise control then appears to have paused in either moving more data or fully completing the process.
Cruise control should be able to run rebalances even if a static configuration is set for `leader.replication.throttled.rate` or `follower.replication.throttled.rate` along with [`default.replication.throttle`](https://github.com/linkedin/cruise-control/wiki/Configurations).

rosakng:
A proposed solution when clearing throttles is to remove the throttle rate configuration if the current throttle rate is of a dynamic broker config source [type](https://kafka.apache.org/25/javadoc/org/apache/kafka/clients/admin/ConfigEntry.ConfigSource.html#DYNAMIC_BROKER_CONFIG), such that only dynamic broker configurations are being removed. In order to support static configurations being returned after a previous dynamic configuration is removed in [waitForConfigs](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java#L370) is to set the expected configuration value to the current static configuration value, only if the current configuration is set static-ly. 

👋 @CCisGG , would you happen to know if this is a known issue or something I can contribute for a fix?

rosakng:
HI @CCisGG  Thanks for your reply! 
My concern with setting `default.replication.throttle` to 0 is that based off of the config specifications
[...]

PR #2080, opened 2023-11-17
Handle setting and removing static throttle rate configs during rebalances

rosakng:
This PR resolves https://github.com/linkedin/cruise-control/issues/2078

Today, CC uses the `default.replication.throttle` config to set the `leader.replication.throttled.rate` and `follower.replication.throttled.rate` only [if unset](https://github.com/linkedin/cruise-control/blob/2cdc79db1840d9e2951eeb43a33f236be03e8689/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java#L186). After the throttle is applied, CC attempts to [remove](https://github.com/linkedin/cruise-control/blob/2cdc79db1840d9e2951eeb43a33f236be03e8689/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java#L342C32-L342C32) the configs and expect them to be `null` in the [check](https://github.com/linkedin/cruise-control/blob/2cdc79db1840d9e2951eeb43a33f236be03e8689/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java#L376) before it continues the rebalance. In the case that a static broker config exists for either throttle rate, this check will fail and retry, causing the rebalance to hang. 

The solution composes of cruise control applying the `default.replication.throttle` to the leader and follower throttle configs only if they are not set to the desired throttled rate already. Since static broker configurations cannot be removed, they are skipped in the removal as well as in the comparison for checking that the configs are equal.

### Testing Steps:
- Used Shopify's sandbox environment to test different combinations of configurations
ex: 
```
follower.replication.throttled.rate=157286400 (static)
cc - default.replication.throttle=52430000
```
Relevant logs:
```
{"source":"cruise-control_log","level":"INFO","time":"2023-11-16T23:11:10,288","msg":"Setting a rebalance throttle of 52430000 bytes/sec "}
{"source":"cruise-control_log","level":"DEBUG","time":"2023-11-16T23:11:10,293","msg":"Setting leader.replication.throttled.rate to 52430000 bytes/second for broker 0 "}
{"source":"cruise-control_log","level":"DEBUG","time":"2023-11-16T23:11:10,294","msg":"Setting follower.replication.throttled.rate to 52430000 bytes/second for broker 0 "}
...
{"source":"cruise-control_log","level":"DEBUG","time":"2023-11-16T23:18:37,535","msg":"Removing leader throttle rate: ConfigEntry(name=leader.replication.throttled.rate, value=52430000, source=DYNAMIC_BROKER_CONFIG, isSensitive=false, isReadOnly=true, synonyms=[], type=LONG, documentation=null) on broker 0 "}
{"source":"cruise-control_log","level":"DEBUG","time":"2023-11-16T23:18:37,535","msg":"Removing follower throttle rate: ConfigEntry(name=follower.replication.throttled.rate, value=52430000, source=DYNAMIC_BROKER_CONFIG, isSensitive=false, isReadOnly=true, synonyms=[], type=LONG, documentation=null) on broker 0 "}
{"source":"cruise-control_log","level":"DEBUG","time":"2023-11-16T23:19:37,796","msg":"Found static broker config: ConfigEntry(name=follower.replication.throttled.rate, value=157286400, source=STATIC_BROKER_CONFIG, isSensitive=false, isReadOnly=true, synonyms=[], type=LONG, documentation=null), skipping comparison "}
...
{"level":"INFO","time":"2023-11-16T23:29:06,779","msg":"Inter-broker partition movements finished."}
```
- added mocks for static broker configurations in unit tests, updated/added test coverage


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -346,18 +343,18 @@ private void removeThrottledRateFromBroker(Integer brokerId)
     ConfigEntry currFollowerThrottle = brokerConfigs.get(FOLLOWER_THROTTLED_RATE);
     List<AlterConfigOp> ops = new ArrayList<>();
     if (currLeaderThrottle != null) {
-      if (currLeaderThrottle.value().equals(WILDCARD_ASTERISK)) {
```
rosakng, line 349: The [leader.replication.throttled.rate](https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#leader-replication-throttled-rate) does not seem to have an asterisk as one of it's valid values as part of the specificication. The same applies to [follower.replication.throttled.rate](https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#follower-replication-throttled-rate), so the checks are removed.

rosakng:
Hi @CCisGG ! Could you please review whenever you have the chance? Thank you!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -183,20 +183,17 @@ private Map<String, Set<String>> getThrottledReplicasByTopic(List<ExecutionPropo
...
+  private void setThrottledRateIfNecessary(int brokerId) throws ExecutionException, InterruptedException, TimeoutException {
     if (_throttleRate == null) {
       throw new IllegalStateException("Throttle rate cannot be null");
     }
     Config brokerConfigs = getBrokerConfigs(brokerId);
     List<AlterConfigOp> ops = new ArrayList<>();
     for (String replicaThrottleRateConfigKey : Arrays.asList(LEADER_THROTTLED_RATE, FOLLOWER_THROTTLED_RATE)) {
       ConfigEntry currThrottleRate = brokerConfigs.get(replicaThrottleRateConfigKey);
-      if (currThrottleRate == null) {
+      if (currThrottleRate == null || !Objects.equals(currThrottleRate.value(), String.valueOf(_throttleRate))) {
```
CCisGG, line 194: I think Objects.equals essentially compare the references of 2 objects. I feel Objects.equals(currThrottleRate.value(), String.valueOf(_throttleRate)) will always return false since they are different objects. Please double confirm. 

For string compare, I would recommend just use currThrottleRate.value().equals(String.valueOf(_throttleRate))

CCisGG (commented):
Thanks for the clear PR description. This is really nice. Left one minor comment.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
rosakng: 👋 The string compare is tested in example 2 in the PR description where the `leader.replication.throttled.rate` gets updated as well as the [unit test](https://github.com/linkedin/cruise-control/pull/2080/files#diff-f54134039747ea091a5603a0a8f3bd23f63efe4a8a31873ba6beb8ad0b568897R314) where the throttle rate is overwritten now. 
I also like the recommendation though so I can change that.

CCisGG (approved):
LGTM. Thanks!

PR #2089, opened 2023-12-07
Fix SASL/OAUTHBEARER configuration

### LINKED ISSUE #2088: SASL/OAUTHBEARER doesn't work because 'sasl.oauthbearer.token.endpoint.url' config is not allowed
For SASL/OAUTHBEARER to work properly users must be able to set token endpoint url through the `sasl.oauthbearer.token.endpoint.url` but this config is ignored even if it is specified.

bachmanity1:
This PR resolves #2088.


bachmanity1:
@CCisGG @hshi2022 can you have a look, please?

CCisGG (approved):

PR #2096, opened 2023-12-21
Update _sortedReplicas correctly

### LINKED ISSUE #2094: Bug? Update _sortedReplicas correctly
Hi, team. 
[updateSortedReplicas(replica)](https://github.com/linkedin/cruise-control/blob/fa21067d968eff7aba06358642c7aeb85e7a9836/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java#L455) is called when make `makeFollower` or `makeLeader` for a tp. 
Replica load can get changed before it by
https://github.com/linkedin/cruise-control/blob/fa21067d968eff7aba06358642c7aeb85e7a9836/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java#L451
https://github.com/linkedin/cruise-control/blob/fa21067d968eff7aba06358642c7aeb85e7a9836/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Replica.java#L248

So the ordering of TreeSet `_sortedReplicas` would get destroyed especially when we use load related `_scoreFunc`. 
https://github.com/linkedin/cruise-control/blob/fa21067d968eff7aba06358642c7aeb85e7a9836/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/SortedReplicas.java#L83
As a result, [sr.remove(replica)](https://github.com/linkedin/cruise-control/blob/fa21067d968eff7aba06358642c7aeb85e7a9836/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java#L434) would not work as expected.

To correctly update `_sortedReplicas` we need do `sr.remove(replica);` before updating the load of the replica, WDYT?

jiao-zhangS:
This PR resolves #2094

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/model/Broker.java
@@ -429,9 +429,14 @@ public class Broker implements Serializable, Comparable<Broker> {
     }
   }
 
-  private void updateSortedReplicas(Replica replica) {
+  private void removeSortedReplicas(Replica replica) {
     _sortedReplicas.values().forEach(sr -> {
       sr.remove(replica);
+    });
+  }
+
+  private void addSortedReplicas(Replica replica) {
+    _sortedReplicas.values().forEach(sr -> {
       sr.add(replica);
     });
   }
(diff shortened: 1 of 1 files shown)


CCisGG (approved):
Good catch! Thanks for the change!

PR #2097, opened 2023-12-21
Fix slow task calculation bug

### LINKED ISSUE #2095: Slow task calculation bug?
`executionAlertingThresholdMs` is calculated as following
https://github.com/linkedin/cruise-control/blob/fa21067d968eff7aba06358642c7aeb85e7a9836/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java#L195

the default value for `inter.broker.replica.movement.rate.alerting.threshold` is 0.1 MB/s
https://github.com/linkedin/cruise-control/blob/fa21067d968eff7aba06358642c7aeb85e7a9836/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java#L273

It was used as 0.1 MB/ms. So calculation should be updated to
```
long executionAlertingThresholdMs = Math.max(Math.round(proposal.dataToMoveInMB() / _interBrokerReplicaMovementRateAlertingThreshold * 1000),
```

jiao-zhangS:
This PR resolves #2095

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
@@ -192,7 +192,7 @@ public class ExecutionTaskPlanner {
       }
       if (!proposal.isInterBrokerMovementCompleted(partitionInfo)) {
         long replicaActionExecutionId = _executionId++;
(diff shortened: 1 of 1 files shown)


CCisGG (approved):
LGTM. Thanks for the catch!

CCisGG (approved):

PR #2098, opened 2024-01-03
Add min.insync.replicas config for sample store and maintanence topics

### LINKED ISSUE #2093: RFE:  Add default `min.insync.replicas` config for sample store topics
The default replication factor for sample store topics (set by `sample.store.topic.replication.factor` in the Cruise Control configuration) is `2`. Cruise Control also relies on the `min.insync.replicas` setting of the Kafka cluster it is managing for these sample topics. Therefore, when a Kafka cluster's `min.insync.replicas` field is set to a value greater than `1` the sample topics could easily become under replicated. To avoid issues such as these, would it make sense to provide a `min.insync.replicas` config for sample store topics? e.g.
```
sample.store.topic.min.insync.replicas
```
in the same manner we do for the metrics reporter topic, `cruise.control.metrics.topic.min.insync.replicas`

If it sounds like a reasonable request, I am happy to contribute it myself!

kyguy:
This PR resolves #2093.


CCisGG:
Hi @mhratson, could you take a look at this one? Thanks!

PR #2101, opened 2024-01-05
Update zookeeper to 3.8.3 in gradle.properties

### LINKED ISSUE #2100: Update zookeeper to 3.8.x
Could we Upgrade zookeeper as dependency because old version of zookeeper (3.5.7) still uses log4j library. Starting from zookeeper 3.8.x they removed log4j and replaced to logback lib. It will be great if we can bump the version and get rid of log4j at all in dependency tree

basically I want to get rid of https://www.cvedetails.com/cve/CVE-2019-17571  and any other issues brought by log4j

proggga:
This PR resolves #2100


CCisGG:
Running our internal check pipeline to see if the version is secure.

CCisGG (approved):

PR #2103, opened 2024-01-09
Update build.gradle - Support AWS IAM auth

### LINKED ISSUE #2102: Feature Request: Support AWS IAM Authentication
This library https://github.com/aws/aws-msk-iam-auth enables developers to use AWS Identity and Access Management (IAM) to connect to their Amazon Managed Streaming for Apache Kafka (Amazon MSK). 

**Problem**
Cruise-control does not support AWS IAM Authentication by default, returning this error "org.apache.kafka.common.config.ConfigException: Invalid value software.amazon.msk.auth.iam.IAMClientCallbackHandler for configuration sasl.client.callback.handler.class: Class software.amazon.msk.auth.iam.IAMClientCallbackHandler could not be found.", forcing developers to re-compile the current code to include it.

**Solution**
Adding the aws library (https://github.com/aws/aws-msk-iam-auth) in the gradle imports should be enough

```java
implementation "software.amazon.msk:aws-msk-iam-auth:2.0.2"
```

matias-gm:
This PR resolves #2102

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -276,6 +276,7 @@ project(':cruise-control') {
     implementation "org.apache.zookeeper:zookeeper:${zookeeperVersion}"
     implementation "io.netty:netty-handler:${nettyVersion}"
     implementation "io.netty:netty-transport-native-epoll:${nettyVersion}"
+    implementation "software.amazon.msk:aws-msk-iam-auth:2.0.2"
     api "org.apache.kafka:kafka_$scalaBinaryVersion:$kafkaVersion"
     api "org.apache.kafka:kafka-clients:$kafkaVersion"
     // Add following dependency when upgrading to Kafka 3.5


build.gradle
```diff
@@ -276,6 +276,7 @@ project(':cruise-control') {
     implementation "org.apache.zookeeper:zookeeper:${zookeeperVersion}"
     implementation "io.netty:netty-handler:${nettyVersion}"
     implementation "io.netty:netty-transport-native-epoll:${nettyVersion}"
+    implementation "software.amazon.msk:aws-msk-iam-auth:2.0.2"
```
matiasgarciamarset, line 279: I think I need to use ' ' instead

marcelloromani: Why do you think you need single quotes? All other `implementation` lines are with double quotes.
I'm just curious about the rationale of your comment.
Also to follow the convention you should probably add something like "iamVersion" variable.

darend, line 279: Bump to latest https://github.com/aws/aws-msk-iam-auth/releases

```suggestion
    implementation "software.amazon.msk:aws-msk-iam-auth:2.3.0"
```

danielgospodinow: Even maybe `2.3.5` now since that's the latest version atm.

gecube:
Really looking forward to this landing. We need to run Cruise Control against Amazon MSK clusters with IAM auth enabled, and the missing `aws-msk-iam-auth` callback handler on the classpath is currently a hard blocker for us — no way to connect without rebuilding the image with the jar added manually. Merging this would be a big help. Thanks @matias-gm!

PR #2106, opened 2024-01-11
Added `verbose` and `skip_hard_goal_check` parameter to remove-disks endpoint valid parameter list

### LINKED ISSUE #2105: `Remove-disks` endpoint doesn't work when used with `skip_hard_goal_check` and `verbose` parameter
When we try to use the `remove-disks` endpoint and try to use a `curl` command with `skip_hard_goal_check` or verbose parameter
```
curl -vv -X POST "http://localhost:9090/kafkacruisecontrol/remove_disks?json=true&dryrun=true&verbose=false&skip_hard_goal_check=false&brokerid_and_logdirs=0-%2Fvar%2Flib%2Fkafka%2Fdata-0%2Fkafka-log0"
```
 the endpoints doesn't work and return this error 
```
{"version":1,"errorMessage":"Unrecognized endpoint parameters in REMOVE_DISKS POST request: [skip_hard_goal_check, verbose].","stackTrace":""}
```

ShubhamRwt:
Description

This PR resolves #2105. This PR adds `verbose` and `skip_hard_goal_check` parameter to the valid parameters list of `remove-disks` endpoint.

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/RemoveDisksRunnable.java
@@ -49,7 +49,7 @@ public class RemoveDisksRunnable extends GoalBasedOperationRunnable {
                                OperationFuture future,
                                RemoveDisksParameters parameters,
                                String uuid) {
-        super(kafkaCruiseControl, future, parameters, false, parameters.stopOngoingExecution(), parameters.skipHardGoalCheck(),
+        super(kafkaCruiseControl, future, parameters, parameters.dryRun(), parameters.stopOngoingExecution(), parameters.skipHardGoalCheck(),
                 uuid, parameters::reason);
         _brokerIdAndLogdirs = parameters.brokerIdAndLogdirs();
     }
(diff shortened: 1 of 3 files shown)


mhratson (changes requested):
Those parameters aren't being used for anything, what's the purpose of adding them?
So [the request doesn't error](https://github.com/linkedin/cruise-control/issues/2105)?

ShubhamRwt:
Yes @mhratson, so that it dosn't fail

ShubhamRwt:
Yes @kyguy. Also When we have `verbose`  as a parameter in our requests through  we are able to get the `loadBeforeOptimization` which is a very beneficial thing so that we can we can compare it `loadAfterOptimization`. But without `verbose` I see the `loadBeforeOptimization` field doesn't show up.

ShubhamRwt:
Hi @mhratson, just wanted to follow up on this issue, Does the above reason sounds suffice to get these changes merged. I mean when we move the data it can cause the broker load to change so I guess it would be good to have both `loadBeforeOptimization` and `loadAfterOptimization` field?

ShubhamRwt:
@mhratson Hi, we can probably skip the `skip_hard_goal_check` parameter but I think `verbose` is still something we are interested in. I am not 100% sure but wouldn't the `loadbeforeOptimization` be different then the `loadAfterOptimization` whe will move data between the disks? In such cases I do think that `verbose` parameter is required since it provided you with the `loadBeforeOptimization`

ShubhamRwt:
Hi @mhratson, based on the above suggestion I went through all the `POST` endpoints and I guess we already have ` verbose` enabled for the ones where we require it and disabled at the places where it is not required. Testing the `verbose` parameter for the `remove-disks` endpoint, I do see it can help us when we want to compare the `before/after` load. As you can see here -> https://gist.github.com/ShubhamRwt/c12fa47d36d53a0cba3473667020319f, I can see that the ` DISK_CAP(MB)` and the `FOLLOWER_NW_IN(KB/s)` has changed. I think it would be good to add ` verbose`. Regarding the `skip_hard_goal_check` I don't think its needed

PR #2114, opened 2024-01-19
Upgrade Gradle to 8.5

### LINKED ISSUE #2113: Upgrade Gradle to 8.5
There are some new features in Gradle 7.6+ and 8+ which can be required for CVE fixing, so it would be useful to upgrade the current Gradle version.

egyedt:
This PR resolves #2113.

> There are some new features in Gradle 7.6+ and 8+ which can be required for CVE fixing, so it would be useful to upgrade the current Gradle version.


egyedt:
@CCisGG, @viktorsomogyi 
Could you please review it?

build.gradle
```diff
@@ -8,7 +8,7 @@ plugins {
   id "idea"
   id "jacoco" // Java Code Coverage plugin
   id "com.github.ben-manes.versions" version "0.42.0"
-  id "com.github.spotbugs" version "5.0.9" apply false
+  id "com.github.spotbugs" version "5.1.5" apply false
```
viktorsomogyi, line 11: seems to be that the latest is 6.0.6. Would you please use that?
https://plugins.gradle.org/plugin/com.github.spotbugs

viktorsomogyi (changes requested):

viktorsomogyi (approved):

build.gradle
```diff
@@ -8,7 +8,7 @@ plugins {
   id "idea"
   id "jacoco" // Java Code Coverage plugin
   id "com.github.ben-manes.versions" version "0.42.0"
-  id "com.github.spotbugs" version "5.0.9" apply false
```
mhratson, line 11: Is this related change?

egyedt: Yes, there was a Spotbugs related failure with this version and Gradle 8.5.
This was the error with Spotbugs:5.0.9 
```
Some problems were found with the configuration of task ':cruise-control:spotbugsIntegrationTest' (type 'SpotBugsTask').
  - In plugin 'com.github.spotbugs.snom.SpotBugsBasePlugin' type 'com.github.spotbugs.snom.SpotBugsTask' property 'firstEnabledReport.destination' is missing an input or output annotation.
```

 This is gone with Spotbugs:5.1.5, but the latest version is Spotbugs:6.0.6, so we should use it, if we already bump the version!

egyedt:
@efeg Could you please check this PR?

Issue #2109, opened 2024-01-12
Persist Demoted and Removed brokers


morgangalpin:
### Overview
Changes to the sets of demoted and removed brokers are retained in memory only. So when Cruise Control restarts, the information is lost and upon startup, load is again shifted back onto the demoted/removed brokers. The original intent was to add persistence for this data, as indicated by this old [TODO in Executor](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L119).

The main goal of this change is to add a persistence mechanism that will serve to persist the sets of demoted/removed brokers, while also being general enough that other runtime data can be persisted as well in the future. A secondary goal of this change is to refactor the logic around manipulating demoted/removed brokers to reuse the common functionality, e.g. removing expired brokers from the sets.

#### PersistedData package
Introduce a package to contain the layers of persisted data classes introduced below.

#### PersistedMap
An implementation of the standard `Map` interface that maintains a cached version of the map data in memory while also persisting any changes to the map. Upon creation, the map reads and initializes to the existing persisted state.

#### KafkaPersistedMap
A specialization of the `PersistedMap` class where the data is published to a kafka topic. Data in the map is eventually consistent as updates to the map data are read from the topic and used to update the cached data. This would also allow multiple instances of cruise control to share access to the persisted data.

#### ExecutorPersistedData
The Executor has two pieces of data that need persisting: the maps of demoted and removed brokers. Each map contains a mapping of broker id to the time they were demoted/removed. The data is read many times, but updated infrequently. This class would maintain both maps and ensure they're persisted and available for reading.

#### Executor
Needs to be updated to use an instance of `ExecutorPersistedData` as the store for both demoted and removed brokers. Also some refactoring to consolidate common functionality.

#### PersistedDataConfig
Introduces configuration for `PersistedData`. Each configuration will be prefixed with `"persisted.data."`.

##### Configuration Values
- `persisted.data.backing.method`: The method to use to store persisted data. This is the first "persisted.data" config to set which will determine which other configs of the series should be configured. At present, the only option is `"kafka"` which is also the default.
[...]

PR #2115, opened 2024-01-23
Persist demoted and removed brokers

morgangalpin:
This PR implements #2109.


mhratson:
Hey @morgangalpin thanks for this!

Do you mind sharing your testing approach/details aside from code tests?

morgangalpin:
Sure thing. We've been using it in production for several months now. I've also tried it out on a local dockerized kafka cluster. My manual testing only involved restarting CC to ensure the previous removed/demoted broker state was restored.

mhratson:
Got it, let me try reproduce this in my test environment and i'll get back to you…

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/persisteddata/BackingMethod.java
```diff
@@ -0,0 +1,38 @@
...
+package com.linkedin.kafka.cruisecontrol.persisteddata;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Defines the backing method types of persistent data.
+ */
+public enum BackingMethod {
```
mhratson, line 13: How about `PersistMethod`?

morgangalpin: I like it. I'll update it.

morgangalpin:
@mhratson any progress with trying it out?

morgangalpin:
@mhratson just following up, is there anything else needed before we can merge this change?

morgangalpin:
@CCisGG or @viktorsomogyi, I haven't heard back from @mhratson . Are you able to review and merge this PR?

morgangalpin:
@mhratson @CCisGG Is there anything left to do before merging this PR? Is this even the right medium to communicate about this?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -38,9 +41,18 @@ public class KafkaCruiseControlConfig extends AbstractConfig {
   private static final ConfigDef CONFIG;
 
   static {
-    CONFIG = CruiseControlRequestConfig.define(CruiseControlParametersConfig.define(AnomalyDetectorConfig.define(
-        AnalyzerConfig.define(ExecutorConfig.define(MonitorConfig.define(WebServerConfig.define(
-            UserTaskManagerConfig.define(new ConfigDef())))))))).withClientSslSupport().withClientSaslSupport();
+    ConfigDef configDef = new ConfigDef();
```
mhratson, line 44: Changes in this file look like a refactoring rather than part of the feature, can this be extracted in a separate PR?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReader.java
```diff
@@ -333,11 +331,7 @@ protected void ensureTopicCreated(Map<String, ?> config) {
    * @param maintenanceEventTopic Maintenance event topic.
    */
   protected void maybeCreateOrUpdateTopic(AdminClient adminClient, NewTopic maintenanceEventTopic) {
-    if (!createTopic(adminClient, maintenanceEventTopic)) {
-      // Update topic config and partition count to ensure desired properties.
-      maybeUpdateTopicConfig(adminClient, maintenanceEventTopic);
-      maybeIncreasePartitionCount(adminClient, maintenanceEventTopic);
-    }
+    KafkaCruiseControlUtils.maybeCreateOrUpdateTopic(adminClient, maintenanceEventTopic);
```
mhratson, line 334: same here, an unrelated (though maybe tempting) refactoring…

viktorsomogyi (commented):
@morgangalpin sorry for this, I missed your ping. I started reviewing it, will likely do multiple rounds on it in the following days.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/persisteddata/namespace/ExecutorPersistedData.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * Executor data that needs persisting is stored in the given map. It assumes that the map's
+     * data is persisted independently. This class is concerned with providing a clean view of that
+     * data and not with how it is stored.
+     *
+     * @param persistedMap The map to store {@link Executor} data in.
+     */
+    public ExecutorPersistedData(Map<String, String> persistedMap) {
+        this._demotedOrRemovedBrokers = Namespace.EXECUTOR.embed(
```
viktorsomogyi, line 59: If I understand correctly, you do this because this way we can serialize demoted or removed brokers with their timestamp the same way in both cases (in-memory vs Kafka).
To be completely honest, I find this a bit less intuitive to understand at first. For this I'd like to propose that what if we move the type of serialization to the respective storage method? For instance we could have a class like this:
```
BrokerEvent implements Serializable {
  private String brokerId;
  private long timestamp;
  private EventType eventType; // demote or remove
  private Namespace namespace; // currently "executor" only
}
```
For memory serialization we could use the standard ByteArrayInput/OutputStream serialization while for Kafka topics we could use json that is serialized into Strings. That would make the Kafka topics easily debuggable too.
When deserialized, we can represent them in a Map<String, BrokerEvent> structure to map broker ID to the event to make it easier to query.
Let me know your thoughts.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReader.java
morgangalpin: This code is needed by the new code, but isn't available when written as a protected method. The protected method looks like it can be removed because it is only used in one place, but I left it since removing it could break any subclasses that override it. I could mark it as `@deprecated` instead.

config/cruisecontrol.properties
```diff
@@ -198,6 +198,52 @@ num.concurrent.leader.movements=1000
...
+#persisted.data.kafka.topic.partition.count=2
+
+# The replication factor to use for the kafka topic. Only applies when
+# "persisted.data.persist.method" is set to "kafka". Multiple partition replicas are desirable to
+# ensure the topic is reasonably available.
+#persisted.data.kafka.topic.replication.factor=2
+
+# The configs to apply to the kafka topic used to persist Cruise Control data. Only applies if
+# "persisted.data.persist.method" is set to "kafka". This "list" should be a semicolon separated
+# string of 'key=value' pairs. The keys and values need to be valid Kafka Topic configs.
```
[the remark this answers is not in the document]
morgangalpin: Ok. I can switch the semicolons to commas and provide an example that demonstrates tidy formatting. In case I've misunderstood your request, the reason for the separator character is that I want to provide a mechanism for users to add any number of key=value pairs to the list rather than mandating which configs are permitted.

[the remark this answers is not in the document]
morgangalpin: Sorry I can't switch the semicolons to commas because individual values can have commas. I can make sure the docs make it clear that readable formatting is easy and desirable.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/persisteddata/kafka/KafkaPersistedMap.java
```diff
@@ -0,0 +1,348 @@
...
+     * Reads and writes data in a kafka topic and keeps an eventually consistent view of the
+     * persisted data.
+     */
+    public KafkaPersistedMap(KafkaCruiseControlConfig config, AdminClient adminClient) {
+        this(config.getString(PersistedDataConfig.KAFKA_TOPIC_NAME_CONFIG),
+                config.getInt(PersistedDataConfig.KAFKA_TOPIC_PARTITION_COUNT_CONFIG),
+                config.getShort(PersistedDataConfig.KAFKA_TOPIC_REPLICATION_FACTOR_CONFIG),
+                config.getMap(PersistedDataConfig.KAFKA_TOPIC_ADDITIONAL_CONFIGS_MAP_CONFIG),
+                config.getMap(PersistedDataConfig.KAFKA_PRODUCER_ADDITIONAL_CONFIGS_MAP_CONFIG),
+                config.getMap(PersistedDataConfig.KAFKA_CONSUMER_ADDITIONAL_CONFIGS_MAP_CONFIG),
```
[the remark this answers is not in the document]
morgangalpin: Ok, makes sense. Thanks for the examples.

Issue #170, opened 2018-03-27
Expose the leadership movement progress.


efeg:
Enable users to track the progress of ongoing leadership movement.

efeg:
Added in patch https://github.com/linkedin/cruise-control/pull/196.

PR #2122, opened 2024-02-15
Add loggings for partition load aggregation

### LINKED ISSUE #2123: Partition load result shows discrepancy with the actual kafka partition
In one of our largest clusters, there are only ~20% partitions are listed in the partition load. This issue was super hard to debug given the current logs and heapdump, and was gone after a restart of the cruise control instance.

In order to better debug it next time, we need to add more useful logs.

CCisGG:
This PR resolves #2123 .

### DIFF
--- a/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
+++ b/cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
@@ -217,9 +217,11 @@ public class MetricSampleAggregator<G, E extends Entity<G>> extends LongGenerati
       MetricSampleAggregationResult<G, E> result = new MetricSampleAggregationResult<>(generation(), completeness);
       Set<E> entitiesToInclude =
           interpretedOptions.includeInvalidEntities() ? interpretedOptions.interestedEntities() : completeness.validEntities();
+      LOG.debug("Including {} entities during metric aggregation.", entitiesToInclude.size());
       for (E entity : entitiesToInclude) {
         RawMetricValues rawValues = _rawMetrics.get(entity);
         if (rawValues == null) {
+          LOG.debug("Failed to find entity {} from _rawMetrics.", entity);
           ValuesAndExtrapolations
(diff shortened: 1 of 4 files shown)


cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
```diff
@@ -217,9 +217,11 @@ public MetricSampleAggregationResult<G, E> aggregate(long from, long to, Aggrega
       MetricSampleAggregationResult<G, E> result = new MetricSampleAggregationResult<>(generation(), completeness);
       Set<E> entitiesToInclude =
           interpretedOptions.includeInvalidEntities() ? interpretedOptions.interestedEntities() : completeness.validEntities();
+      LOG.debug("Including {} entities during metric aggregation.", entitiesToInclude.size());
       for (E entity : entitiesToInclude) {
         RawMetricValues rawValues = _rawMetrics.get(entity);
         if (rawValues == null) {
+          LOG.debug("Failed to find entity {} from _rawMetrics.", entity);
```
bsandeep23, line 224: IMO, would be better to log it as a WARN

bsandeep23 (approved):
LGTM, With one non blocking comment

mhratson:
> This PR resolves #170 .

Does it resolve the issue? Or adds more related logging?

cruise-control-core/src/main/java/com/linkedin/cruisecontrol/monitor/sampling/aggregator/MetricSampleAggregator.java
CCisGG: Hi @bsandeep23 , I originally added a WARN level log here, but then after thinking it deeper I think it may be safer to make it debug for now.

The reason is we currently don't clearly know whether this is a legit path or not, and how frequently this happens. I would worry about this log line will be hit as hot path for some cases which could cause logging pollution. So for now I think it's probably safer to make it debug level.

CCisGG:
> Does it resolve the issue? Or adds more related logging? Or you meant https://github.com/linkedin/cruise-control/issues/2123?

Good catch! It's 2123. Updated

mhratson (approved):
🚢

PR #2143, opened 2024-04-10
Create dependency-submission.yml

stephengroat:
Currently seeing a lot of dependency vulnerabilities being triggered by `cruise-control`, wondering if changes could be made to gradle to address these issues

I've made this PR (to enable dependabot) so that there can be some visibility into the dependencies that are currently being marked as vulnerable

Can share access to the dependency vulnerabilties in https://github.com/stephengroat/cruise-control/security/dependabot if anyone is interested

Screenshot 2024-04-11 at 08 58 48


mhratson:
Thanks for suggestion, we're discussing Dependabot integration internally. Closing for now.

PR #2145, opened 2024-04-21
Support intra-broker throttling (replica.alter.log.dirs.io.max.bytes.per.second)

### LINKED ISSUE #1851: disk balance
Hi,
Running disk balance is very bad via CC.
I'm running JBOD cluster with 10 brokers, 4 disks on each machine.
even when setting  num.concurrent.intra.broker.partition.movements=1 my cluster IO goes to 100% on couple of brokers, leading to offline partitions and service degradation...
Now, i would like to ask:
Why CC doesn't support logdir movement throttle?
Why i can't run disk balance to only 1 broker at a time ?

dongjinleekr:
This PR resolves #1851.


dongjinleekr:
This PR is an WIP; I will test this feature in our in-house fork (@naver) this week. Stay tuned!! :upside_down_face:

dongjinleekr:
@aswinayyolath @mhratson @jiao-zhangS Could you kindly have a look? We adopted this patch to our in-house distribution and confirmed it works correctly. :pray:

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1707,6 +1729,11 @@ private void intraBrokerMoveReplicas() {
         waitForIntraBrokerReplicaTasksToFinish();
         inExecutionTasks = inExecutionTasks();
       }
+
+      if (_logDirThrottle != null) {
```
cpaika, line 1733: for my understanding, why the underscore in the variable names here?

cpaika (approved):
Will make things a lot more stable for JBOD clusters, good change

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
dongjinleekr: In consistency with `_replicationThrottle`.

adkafka (approved):

dongjinleekr:
@cpaika @adkafka

Excuse me. Is there any issue with this PR? It initially worked on 2.5.138, and 2.5.141 was already released last month, but this PR has yet to be merged or released.

If any modifications are needed, don't hesitate to leave me a mention.

adkafka:
> @cpaika @adkafka
> 
> Excuse me. Is there any issue with this PR? It initially worked on 2.5.138, and 2.5.141 was already released last month, but this PR has yet to be merged or released.
> 
> If any modifications are needed, don't hesitate to leave me a mention.

I don't see any issues with this PR, but I'm not a maintainer of this project. We need a maintainer to review this and add their approval.

aswinayyolath:
@CCisGG could you please review this PR?

aswinayyolath (approved):

dongjinleekr:
@CCisGG Hello. Could you please have a look when you are free? :pray:

CCisGG:
Hi @mhratson, would you mind taking a look at this one? Thanks!

imans777:
Anyone caring to review this?
It's really a must-have feature.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
```diff
@@ -206,6 +206,14 @@ public final class ExecutorConfig {
   public static final String DEFAULT_REPLICATION_THROTTLE_DOC = "The replication throttle applied to replicas being "
       + "moved, in bytes per second.";
 
+  /**
+   * <code>default.log.dir.throttle</code>
+   */
+  public static final String DEFAULT_LOG_DIR_THROTTLE_CONFIG = "default.log.dir.throttle";
+  public static final Long DEFAULT_DEFAULT_LOG_DIR_THROTTLE = null;
```
CCisGG, line 213: Would that be more reasonable to have it Long.MAX_VALUE rather than null? I'm a bit concerned to have a null default value for a LONG type. I'm not entirely sure what will happen here by default

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -830,7 +834,7 @@ public synchronized void executeProposals(Collection<ExecutionProposal> proposal
                             requestedIntraBrokerPartitionMovementConcurrency, requestedClusterLeadershipMovementConcurrency,
                             requestedBrokerLeadershipMovementConcurrency, requestedExecutionProgressCheckIntervalMs, replicaMovementStrategy,
                             isTriggeredByUserRequest, loadMonitor);
-      startExecution(loadMonitor, null, removedBrokers, replicationThrottle, isTriggeredByUserRequest);
+      startExecution(loadMonitor, null, removedBrokers, replicationThrottle, logDirThrottle, isTriggeredByUserRequest);
```
CCisGG, line 837: Just curious, does that make more sense to have a separate method for intra-broker throttling since this parameter only applies for inter-broker movement?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -918,7 +922,7 @@ public synchronized void executeDemoteProposals(Collection<ExecutionProposal> pr
       initProposalExecution(proposals, demotedBrokers, concurrentSwaps, null, 0,
                             requestedClusterLeadershipMovementConcurrency, requestedBrokerLeadershipMovementConcurrency,
                             requestedExecutionProgressCheckIntervalMs, replicaMovementStrategy, isTriggeredByUserRequest, loadMonitor);
-      startExecution(loadMonitor, demotedBrokers, null, replicationThrottle, isTriggeredByUserRequest);
+      startExecution(loadMonitor, demotedBrokers, null, replicationThrottle, replicationThrottle, isTriggeredByUserRequest);
```
CCisGG, line 925: What is the implication here to have log dir throttling set to replication throttle here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -51,48 +52,54 @@ class ReplicationThrottleHelper {
...
-      LOG.info("Setting a rebalance throttle of {} bytes/sec", _throttleRate);
-      Set<Integer> participatingBrokers = getParticipatingBrokers(replicaMovementProposals);
-      Map<String, Set<String>> throttledReplicas = getThrottledReplicasByTopic(replicaMovementProposals);
-      for (int broker : participatingBrokers) {
-        setThrottledRateIfNecessary(broker);
-      }
-      for (Map.Entry<String, Set<String>> entry : throttledReplicas.entrySet()) {
-        setThrottledReplicas(entry.getKey(), entry.getValue());
-      }
+    LOG.info("Setting a rebalance throttle of {} bytes/sec", throttleRate);
```
CCisGG, line 81: Why do we remove the "if (throttlingEnabled())" check here?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -336,7 +355,7 @@ private void removeThrottledReplicasFromTopic(String topic, Set<String> replicas
     }
   }
 
-  private void removeThrottledRateFromBroker(Integer brokerId)
+  private void removeThrottledRatesFromBroker(Integer brokerId)
```
CCisGG, line 358: Is this method now become "remove both throttle rates"? In that case we should add a comment to specify that.

CCisGG (commented):
Added some comments/questions.

Some other questions I have:
1. Did you test that with your change, when no replication/intra broker throttling was set (which I think is by default), does this work properly?
2. Is their any case that this intra-broker throttling can interfere with inter-broker thorttling? E.g. what happens if they are enabled together?
3. Please update Configuration.md for this new config.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
dongjinleekr: I agree with you. But, since the `startExecution` method is invoked for all rebalance tasks yet (i.e., both inter-broker and intra-broker with demote, remove broker options), I thought the splitting task is better to be done with independent PR.

It is why I chose to add a new parameter here.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
dongjinleekr: Mistake; Since it is demote task, we don't need `logDirThrottle` parameter so it should be `null`. fixed.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
dongjinleekr: Oh no, it is just a simple mistake. There are no changes in the method body. fixed.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
dongjinleekr: Before this PR, `clearThrottles` (now `clearInterBrokerThrottles`) had `if (throttlingEnabled())` check, which dependes on `private final Long _throttleRate;`, right?

```
private boolean throttlingEnabled() {
  return _throttleRate != null;
}
```

Since `ReplicationThrottleHelper` does not have `_throttleRate` anymore, the nullity check is moved into `Executor#interBrokerMoveReplicas` method:

```
# before
throttleHelper.setThrottles(tasksToExecute.stream().map(ExecutionTask::proposal).collect(Collectors.toList()));

# after
if (_replicationThrottle != null) {
  throttleHelper.setReplicationThrottles(tasksToExecute.stream().map(ExecutionTask::proposal).collect(Collectors.toList()),
      _replicationThrottle);
}
```

```
# before
throttleHelper.clearThrottles(completedTasks, inProgressTasks);

# after
if (_replicationThrottle != null) {
  throttleHelper.clearInterBrokerThrottles(completedTasks, inProgressTasks);
}
```

In short, the `if (throttlingEnabled())` check is moved from the body of `clearThrottles` to the outer if statement of the caller side, `Executor`.

dongjinleekr:
@CCisGG 

Sorry for being late. Here are the updates, rebased onto the latest main, and the anwsers:

> Did you test that with your change, when no replication/intra broker throttling was set (which I think is by default), does this work properly?

YES. I tested it before I opened the PR, and we have been running it with our in-house distribution for more than 6 months and experienced no trouble. (As you already pointed out, `default.log.dir.throttle` is `null` by default.)

> Is their any case that this intra-broker throttling can interfere with inter-broker thorttling? E.g. what happens if they are enabled together?

As far as I know, there is no case where intra and inter-broker throttlings interfere with each other. The reason is simple: the intra-broker rebalancing is only executed when `rebalance_disk` [parameter](https://github.com/linkedin/cruise-control/blob/main/cruise-control/src/main/resources/yaml/endpoints/rebalance.yaml#L144) is set to `true`. If this parameter is set to `false` (default), only inter-broker rebalancing is executed.

In short, intra- and inter-broker rebalancings are executed exclusively for each other, so if `default.replication.throttle` and `default.log.dir.throttle` are enabled together, they will not impact each other.

luciano-sabenca:
Any news about this MR? 
It's a really useful feature for us! 
:D

docs/wiki/User Guide/Configurations.md
```diff
@@ -155,7 +155,8 @@ The following configurations are inherited from the open source Kafka client con
...
+| default.log.dir.throttle                                           | Long	| N         | null	                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 | The replication throttle applied to replicas being moved, in bytes per second.	                                                                                                                                                                                                                                                                                             |
```
CCisGG, line 159: Please add different descriptions for both config since one of them is inter broker and the other is intra broker.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -651,6 +651,8 @@ private static boolean hasProposalsToExecute(Collection<ExecutionProposal> propo
    *                                (if null, use default.replica.movement.strategies).
    * @param replicationThrottle The replication throttle (bytes/second) to apply to both leaders and followers
```
CCisGG, line 652: Let's clarify in the comment that replicationThrottle is for inter-broker, logDirThrottle is for intra-broker rebalance.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
CCisGG: Let's create a task to track the split. Thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
CCisGG: Got it. Thanks

CCisGG (changes requested):
Thanks for the change! I left some minor comments.

docs/wiki/User Guide/Configurations.md
dongjinleekr: OMG. What a silly mistake 😓 fixed.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
dongjinleekr: Fixed. I elaborated the Javadoc by explicitly stating inter-broker movements vs. intra-broker movements.

dongjinleekr:
@CCisGG Rebased onto the latest main, with some additional improvements.

-----

> [Let's create a task to track the split. Thanks!](https://github.com/linkedin/cruise-control/pull/2145#discussion_r2002098925)

You mean splitting the `Executor#startExecution` method call to inter-broker mode and intra-broker mode, applying `replicationThrottle` and `logDirThrottle` respectively, right? I tried to, but it seems impossible. Here is why:

**The `Executor#executeProposals` method, which calls `Executor#startExecution`, is agnostic to the inter-broker and intra-broker modes.**

(Note: The method call chain is `KafkaCruiseControl#executeProposals` → `Executor#executeProposals` → `Executor#startExecution`)

The `rebalance_disk` parameter (`ParameterUtils.REBALANCE_DISK_MODE_PARAM`) is only used to generate the proposal, and then the proposal is delivered to the `Executor#executeProposals` method to be executed. That is, the `Executor#executeProposals` method can not discriminate between the two modes.

Then, is splitting `KafkaCruiseControl#executeProposals` into inter- and intra-broker modes feasible? I also investigated the possibility, but it seems impossible. Here is why: it is called by the following 5 API calls/Runnables:

1. `add_broker` (`AddBrokersRunnable`)
2. `fix_offline_replicas` (`FixOfflineReplicasRunnable`)
3. `rebalance` (`RebalanceRunnable`)
4. `RemoveDisksRunnable`
5. `UpdateTopicConfigurationRunnable`

5 incurs no movements, 1 incurs inter-broker movements only, 3 incurs either inter-broker or intra-broker movements, and 2 and 4 incur BOTH. For this reason, splitting the two methods' calls seems unreasonable - we can not assert whether 2 or 4 invocations need intra-broker movements beforehand.

For the reasons above, I did nothing to the call chain. It is true that these methods seem too overloaded, but improving them with another PR may be much better IMHO. (I have an idea, and will open another PR on this issue after this PR is merged.)

dongjinleekr:
@CCisGG Instead, I did:

1. removed `logDirThrottle` from the `Executor#executeRemoval` method.
    Since its goal is to move all replicas in the specified broker, intra-broker replica movement does not occur at all. So, it is totally useless.
2. Remove `_logDirThrottle` from the AddBrokersRunnable, RemoveBrokersRunnable, and UpdateTopicConfigurationRunnable.
    DITTO. These Runnables do not incur any intra-broker replica movements. So, it is also useless.

With one commit for `Configurations.md`, one commit for `KafkaCruiseControl` Javadoc, and two commits for the improvements above. In total, I added four commits after your last review. :pray: Please have a look when you are free!! Thanks.

CCisGG:
@dongjinleekr Thanks for the explanations and the new commits. I think they are all reasonable. 

Since I probably won't have much time to do another round of review and audit, would you mind inviting some other person (maybe from your org) to review the PR? And Let's do a final round of testing to ensure this won't interfere with inter-broker rebalance. Once those are done, I can help to merge to PR. Let me know if this works. Thank you!

CCisGG (approved):
Looks good in general to me. Please ask for another review from a different person if possible. Then I can help to merge the PR.

dongjinleekr:
@CCisGG 

> Would you mind inviting some other person (maybe from your org) to review the PR?

Oh no, no problem. In fact, I am now working on supporting this feature in the [Strimzi Kafka operator](https://github.com/strimzi/strimzi-kafka-operator), and it will be completed by this weekend. As soon as I open the PR, there may be other people who can review it. (Sadly, I am the only person who is working on CC internals in my org :cry:)

Plus: We are running this feature in our in-house fork but, let me double-check if there is any problem 🙏 Thank you!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1676,20 +1690,28 @@ private void interBrokerMoveReplicas() throws InterruptedException, ExecutionExc
...
-    private void intraBrokerMoveReplicas() {
+    private void intraBrokerMoveReplicas() throws InterruptedException, ExecutionException, TimeoutException {
+      ReplicationThrottleHelper throttleHelper = new ReplicationThrottleHelper(_adminClient);
       int numTotalPartitionMovements = _executionTaskManager.numRemainingIntraBrokerPartitionMovements();
       long totalDataToMoveInMB = _executionTaskManager.remainingIntraBrokerDataToMoveInMB();
       long startTime = System.currentTimeMillis();
       LOG.info("User task {}: Starting {} intra-broker partition movements.", _uuid, numTotalPartitionMovements);
 
       int partitionsToMove = numTotalPartitionMovements;
+      Set<Integer> participatingBrokers = Sets.newHashSet();
```
kyguy, line 1701: Could we use `new HashSet<>()`  from java.util  instead? 

It appears that `Sets.newHashSet()` is deprecated and uses  `java.util.HashSet` under the hood anyway.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -39,54 +39,61 @@ class ReplicationThrottleHelper {
...
-  ReplicationThrottleHelper(AdminClient adminClient, Long throttleRate, int retries, Set<Integer> deadBrokers) {
+  ReplicationThrottleHelper(AdminClient adminClient, int retries, Set<Integer> deadBrokers) {
     this._adminClient = adminClient;
-    this._throttleRate = throttleRate;
     this._retries = retries;
     this._deadBrokers = deadBrokers;
   }
 
-  void setThrottles(List<ExecutionProposal> replicaMovementProposals)
+  void setReplicationThrottles(List<ExecutionProposal> replicaMovementProposals, long throttleRate)
```
kyguy, line 73: It looks like we refactor `setThrottles()` -> `setReplicationThrottles()` but then refactor `clearThrottles()` -> `clearInterBrokerThrottles()` instead of `clearReplicationThrottles()`. To improve readability, could we use a consistent naming scheme for the new and refactored method names in this class? For example, use:

(a) `setReplicationThrottles()` and `clearReplicationThrottles()`

or

(b) `setInterBrokerThrottles()` and `clearInterBrokerThrottles()`

Option (b) seems better aligned with Cruise Control terminology but either would be fine so long as it is consistent!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -39,54 +39,61 @@ class ReplicationThrottleHelper {
...
+    Map<String, Set<String>> throttledReplicas = getThrottledReplicasByTopic(replicaMovementProposals);
+    for (int broker : participatingBrokers) {
+      setReplicationThrottledRateIfNecessary(broker, throttleRate);
+    }
+    for (Map.Entry<String, Set<String>> entry : throttledReplicas.entrySet()) {
+      setThrottledReplicas(entry.getKey(), entry.getValue());
     }
   }
 
+  Set<Integer> setLogDirThrottles(List<ExecutionProposal> replicaMovementProposals, long throttleRate)
```
kyguy, line 86: Similar comment as above

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -177,24 +185,35 @@ private Map<String, Set<String>> getThrottledReplicasByTopic(List<ExecutionPropo
...
+    List<AlterConfigOp> ops = new ArrayList<>();
+    ConfigEntry currThrottleRate = brokerConfigs.get(LOG_DIR_THROTTLED_RATE);
+    if (currThrottleRate == null || !currThrottleRate.value().equals(String.valueOf(throttleRate))) {
+      LOG.debug("Setting {} to {} bytes/second for broker {}", LOG_DIR_THROTTLED_RATE, throttleRate, brokerId);
+      ops.add(new AlterConfigOp(new ConfigEntry(LOG_DIR_THROTTLED_RATE, String.valueOf(throttleRate)), AlterConfigOp.OpType.SET));
+    }
+    if (!ops.isEmpty()) {
+      changeBrokerConfigs(brokerId, ops);
+    }
+  }
```
kyguy, line 215: Since there will only ever be one "ops", we could refactor to something like this:
```suggestion
    ConfigEntry currThrottleRate = brokerConfigs.get(LOG_DIR_THROTTLED_RATE);
    if (currThrottleRate == null || !currThrottleRate.value().equals(String.valueOf(throttleRate))) {
      LOG.debug("Setting {} to {} bytes/second for broker {}", LOG_DIR_THROTTLED_RATE, throttleRate, brokerId);
      changeBrokerConfigs(brokerId, List.of(new AlterConfigOp(new ConfigEntry(LOG_DIR_THROTTLED_RATE, String.valueOf(throttleRate)), AlterConfigOp.OpType.SET)));
    }
  }
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -357,6 +376,28 @@ private void removeThrottledRateFromBroker(Integer brokerId)
...
+      if (currLogDirThrottle.source().equals(ConfigEntry.ConfigSource.STATIC_BROKER_CONFIG)) {
+        LOG.debug("Skipping removal for static log dir throttle rate: {}", currLogDirThrottle);
+      } else {
+        LOG.debug("Removing log dir throttle rate: {} on broker {}", currLogDirThrottle, brokerId);
+        ops.add(new AlterConfigOp(new ConfigEntry(LOG_DIR_THROTTLED_RATE, null), AlterConfigOp.OpType.DELETE));
+      }
+    }
+    if (!ops.isEmpty()) {
+      changeBrokerConfigs(brokerId, ops);
+    }
```
kyguy, line 398: Similar comment as above, since there is only ever one "ops", this could be refactored like the following:
```suggestion
    if (currLogDirThrottle != null) {
      if (currLogDirThrottle.source().equals(ConfigEntry.ConfigSource.STATIC_BROKER_CONFIG)) {
        LOG.debug("Skipping removal for static log dir throttle rate: {}", currLogDirThrottle);
      } else {
        LOG.debug("Removing log dir throttle rate: {} on broker {}", currLogDirThrottle, brokerId);
        changeBrokerConfigs(brokerId, List.of(new AlterConfigOp(new ConfigEntry(LOG_DIR_THROTTLED_RATE, String.valueOf(throttleRate)), AlterConfigOp.OpType.SET)));
      }
    }
```

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
```diff
@@ -4,6 +4,7 @@
 
 package com.linkedin.kafka.cruisecontrol.executor;
 
+import com.google.common.collect.Sets;
```
kyguy, line 7: Could we use new HashSet<>() from java.util here instead?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -7,6 +7,7 @@
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.google.common.collect.Sets;
```
kyguy, line 10: Could we use new HashSet<>() from java.util here instead?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java
```diff
@@ -667,14 +671,15 @@ public void executeProposals(Set<ExecutionProposal> proposals,
                                Long executionProgressCheckIntervalMs,
                                ReplicaMovementStrategy replicaMovementStrategy,
                                Long replicationThrottle,
+                               Long logDirThrottle,
```
kyguy, line 674: Why not add `logDirThrottle` to the `executeRemoval()` method as well? This way we could clear the throttles from the broker metadata when brokers are removed from the cluster.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/IdempotenceCacheTest.java
```diff
@@ -74,7 +74,7 @@ public void testIdempotence() {
     for (MaintenanceEventType eventType : MaintenanceEventType.cachedValues()) {
       parameterConfigOverrides.put(MAINTENANCE_EVENT_TYPE_CONFIG, eventType);
       // Expect mocks.
-      EasyMock.expect(_mockKafkaCruiseControl.config()).andReturn(_config).times(4);
+      EasyMock.expect(_mockKafkaCruiseControl.config()).andReturn(_config).anyTimes();
```
kyguy, line 77: Is there any specific reason we changed this to `anyTimes()`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/UpdateTopicConfigurationRunnable.java
```diff
@@ -209,8 +209,8 @@ protected OptimizerResult workWithClusterModel() throws KafkaCruiseControlExcept
       _kafkaCruiseControl.executeProposals(result.goalProposals(), Collections.emptySet(), false, _concurrentInterBrokerPartitionMovements,
           _maxInterBrokerPartitionMovements,
           0, _clusterLeaderMovementConcurrency, _brokerLeaderMovementConcurrency,
-          _executionProgressCheckIntervalMs, _replicaMovementStrategy, _replicationThrottle, _isTriggeredByUserRequest, _uuid,
-          SKIP_AUTO_REFRESHING_CONCURRENCY);
+          _executionProgressCheckIntervalMs, _replicaMovementStrategy, _replicationThrottle, null,
```
kyguy, line 212: The `logDirThrottle` is null here because partitions are never any intra-broker partition movements as part of topic configuration changes, is my understanding correct?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/handler/async/runnable/AddBrokersRunnable.java
```diff
@@ -127,6 +127,7 @@ protected OptimizerResult workWithClusterModel() throws KafkaCruiseControlExcept
                                            _executionProgressCheckIntervalMs,
                                            _replicaMovementStrategy,
                                            _replicationThrottle,
+                                           null,
```
kyguy, line 130: Why is the `logDirThrottle` argument set to null here? Wouldn't we want new brokers to have the configured logDirThrottle set?

kyguy (commented):
Hi @dongjinleekr, the PR looks great overall! I just left a few minor suggestions, mostly related to refactoring for consistency and readability.

palpandy:
Hello, 
Any news about this feature ? It is really a important feature for us.

imans777:
Is there a parameter like `replication_throttle` to be able to set on specific requests (all APIs)? I only saw the `default.log.dir.throttle` which is the static configurations. We should support `log_dir_throttle` parameters for all APIs.
Also it should be added to `/admin` API so we can modify the default value dynamically.
(I took a look at code but didn't found answer to these)

PR #2165, opened 2024-06-18
Adjust log4j configuration to use `SizeBasedTriggeringPolicy` and include RollverStrategy

### LINKED ISSUE #2163: Logfile rotation isn't removing old log files
Current log4j configuration is performing logfile rotation using `TimeBasedTriggeringPolicy` which rotates file each hour.
This configuration isn't removing old log files and this can led to usage of all disk space.

Cute-Monster:
This PR resolves #2163.

With the following settings `logs` directory content look is: 
image

### DIFF
--- a/config/log4j.properties
+++ b/config/log4j.properties
@@ -2,11 +2,20 @@
 # Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
 #
 
+# ===============
+# Properties
+# ===============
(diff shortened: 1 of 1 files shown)


CCisGG (changes requested):
Hi I think we intentionally use time based rolling strategy as it is useful during debugging issues. I think you can config the max retention time for your logs.

We would probably close this PR for the above reason, but please feel free to make the change to your own fork.

mhratson:
There's no single solution for everyone so  please add  wiki/documentation how to configure it instead. Can reference this PR as an example.

Issue #2168, opened 2024-06-23
java.lang.NoSuchMethodError: 'java.lang.String kafka.server.KafkaConfig.ListenersProp()' when running MetricsReporter for Kafka 3.8


HenryCaiHaiying:
When running embedded MetricsReporter with the upcoming Kafka 3.8, hit the following exception:

[2024-06-18 16:08:06,722] ERROR [KafkaServer id=1025] Fatal error during KafkaServer startup. Prepare to shutdown (kafka.server.KafkaServer)
java.lang.NoSuchMethodError: 'java.lang.String kafka.server.KafkaConfig.ListenersProp()'
        at com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter.getBootstrapServers(CruiseControlMetricsReporter.java:129)

HenryCaiHaiying:
The root cause is Kafka 3.8 (maybe also 3.7) refactored the core module for many configuration parameters, see this PR: https://github.com/apache/kafka/commit/1b301b30207ed8fca9f0aea5cf940b0353a1abca?diff=split&w=0#diff-8eb3e01716508551f203b0b37c2d5f951e93cce7ffed7c00c2b33633d7c8ed23

HenryCaiHaiying:
The easiest fix is just to use hard-coded literal string 'listeners' for this config parameter, the name of this parameter is unlikely to change (otherwise it will break million's customer's server.properties).  Although you can try to change the variable name reference from KafkaConfig.ListenersProp() to SocketServerConfigs.LISTENERS_CONFIG, but SocketServerConfigs is also a new class introduced in Kafka 3.8, to do that you would have to create a new CC branch migrate_to_3_8 to make the compilation work, but maintaining different branches for different kafka version is cumbersome and hard to maintain in the long run.

HenryCaiHaiying:
See fix: https://github.com/linkedin/cruise-control/pull/2169/files

mhratson:
#2169 is merged

PR #2169, opened 2024-06-23
Use literal config name for listeners and broker.id config

HenryCaiHaiying:
Fix issue: https://github.com/linkedin/cruise-control/issues/2168

When running embedded MetricsReporter with the upcoming Kafka 3.8, hit the following exception:

[2024-06-18 16:08:06,722] ERROR [KafkaServer id=1025] Fatal error during KafkaServer startup. Prepare to shutdown (kafka.server.KafkaServer) java.lang.NoSuchMethodError: 'java.lang.String kafka.server.KafkaConfig.ListenersProp()' at com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter.getBootstrapServers(CruiseControlMetricsReporter.java:129)

The root cause is Kafka 3.8 (maybe also 3.7) refactored the core module for many configuration parameters, see this PR: apache/kafka@1b301b3?diff=split&w=0#diff-8eb3e01716508551f203b0b37c2d5f951e93cce7ffed7c00c2b33633d7c8ed23

The easiest fix is just to use hard-coded literal string 'listeners' for this config parameter, the name of this parameter is unlikely to change (otherwise it will break million's customer's server.properties). Although you can try to change the variable name reference from KafkaConfig.ListenersProp() to SocketServerConfigs.LISTENERS_CONFIG, but SocketServerConfigs is also a new class introduced in Kafka 3.8, to do that you would have to create a new CC branch migrate_to_3_8 to make the compilation work, but maintaining different branches for different kafka version is cumbersome and hard to maintain in the long run.

This also applies on another property broker.id

This PR resolves #<Replace-Me-With-The-Issue-Number-Addressed-By-This-PR>.

### DIFF
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -25,7 +25,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import kafka.server.KafkaConfig;
 import org.apache.kafka.clients.ClientUtils;
 import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.clients.admin.AdminClient;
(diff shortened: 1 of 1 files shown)


zornhsu (commented):
+1
Suggest we merge this change to `main` instead of maintaining a new `migrate_to_3_8` branch

mimaison:
> The easiest fix is just to use hard-coded literal string 'listeners' for this config parameter, the name of this parameter is unlikely to change (otherwise it will break million's customer's server.properties). Although you can try to change the variable name reference from KafkaConfig.ListenersProp() to SocketServerConfigs.LISTENERS_CONFIG, but SocketServerConfigs is also a new class introduced in Kafka 3.8, to do that you would have to create a new CC branch migrate_to_3_8 to make the compilation work, but maintaining different branches for different kafka version is cumbersome and hard to maintain in the long run.

`KafkaConfig` is not part of the public API, this is why its structure changed without a KIP or deprecation notice. `SocketServerConfigs` is not part of the public API either. So using the string literals, which are part of the public API, is in my opinion the best option to make the Cruise Control metrics reporter work with Kafka 3.8.

Cruise Control still depends on many non-public APIs but at least this quick fix addresses this immediate issue, as 3.8 will release shortly and is already being tested by several organizations.

HenryCaiHaiying:
I don't see any objections, can we move forward to merge in this PR?

mimaison:
@CCisGG Can you take a look? Thanks

fvaleri:
+1

PR #2203, opened 2024-10-03
Fix the Circle CI error by removing the test-multi-arch job

allenxwang:
## Summary
1. Why: 

`test-multi-arch` job fails due to image ubuntu-2004:202107-02 is unavailable. See https://discuss.circleci.com/t/linux-image-deprecations-and-eol-for-2024/.

On the other hand, builds and integration tests on other platforms pass.

2. What: 

Removed the job `test-multi-arch` to unblock the PR workflow.

I also tried to fix this by using the `default` machine image as suggested in the above link but it [failed](https://app.circleci.com/pipelines/github/linkedin/cruise-control/2467/workflows/42acc46b-01da-44cc-8fff-d06800a6f1d7/jobs/6740) in findbugs due to class file version issue. 

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [x ] other

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,13 +75,6 @@ workflows:
   version: 2.1
   build-and-publish:
     jobs:
-      - test-multi-arch:
-          matrix:
-            parameters:
-              platform: ["s390x"]
-          filters:
-            tags:
-              only: /.*/
       - build:
           filters:
             tags:
@@ -89,7 +82,6 @@ workflows:
       - integration-test:
           requires:
             - build
-            - test-multi-arch
           filters:
             branches:
               ignore: /.*/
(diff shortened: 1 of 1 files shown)


yasiribmcon:
Hi @allenxwang default machine image is referring to ubuntu 22.04 image - `ubuntu-2204:2024.05.1` instead of ubuntu 20.04, which could be the reason CI build started to fail.
Could you please try to run jobs with `ubuntu-2004:current` image which is actually `ubuntu-2004:2024.05.1` which is not a part of the deprecated list.
It would be preferable if building s390x remains a part of CI.

CCisGG (approved):
I'm ok with merge this to unblock the pipeline for now, but I think we should RC the build error and add this back once we know how to fix the error. IMO other companies or users may still depend on this test-multi-arch-s390x and removing it may introduce uncertainties to those users.

PR #2204, opened 2024-10-03
Improve per task observability through additional logging

allenxwang:
## Summary
1. Why: 

We want to track and analyze task execution for each task in the log analysis tool. But currently there is not enough information logged and the log does not have task UUID so it is difficult to track on the per task base.

2. What: 

Added more INFO logging with UUID.

Also fixed a bug where `null` UUID is logged by moving the call to `clearCompletedExecution()` (where uuid is reset to `null`) to after the logging.

## Additional evidence
```
Execution succeeded: task Id: null; removed brokers: [103986]; demoted brokers: null; total time used: 874448ms.. 
```

## Categorization
- [ ] documentation
- [x ] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [ x] other

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -833,7 +833,9 @@ public class Executor {
       startExecution(loadMonitor, null, removedBrokers, replicationThrottle, isTriggeredByUserRequest);
     } catch (Exception e) {
       if (e instanceof OngoingExecutionException) {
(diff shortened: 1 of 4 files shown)


CCisGG (changes requested):
Looks like are confusing "task" and "operation" here. In  general, I think a task is a single partition reassignment, for example partition X: {broker A, B, C} -> {broker D, E, F}. An operation what CC executes to complete user request or self-healing. Broker removals or rebalances should be operations rather than task. So I think here you probably want to change them to "Operation id: {}"

CCisGG (approved):
Thanks Allen!

PR #2214, opened 2024-10-27
Optimize replication throttling in Executor to reduce Kafka admin requests

### LINKED ISSUE #1972: Redundant set/unset throttling during the rebalance

aswinayyolath:
## Summary

- **Why:**  
  The current implementation of CC sets and clears replication throttles for each execution task individually, resulting in a high volume of admin requests and slowing down the rebalancing process. This change optimizes throttling by only setting and clearing throttles once per broker during the start and end of a group of tasks, improving efficiency.

- **What:**  
  Updated `Executor.java` to introduce logic for setting throttles per broker at the beginning of task execution and clearing throttles only after all tasks for a broker are complete.

## Expected Behavior  
With this improvement, the replication throttles are set once per broker and cleared after all related tasks are finished, reducing admin request load and improving rebalance speed.

## Actual Behavior  
Currently, throttling is set and cleared for every individual task, which can cause unnecessary delays during rebalancing due to excessive Kafka admin requests.

## Additional Evidence  

I have ran the test to make sure all the tests pass.

image

----------------------------------------

image

## Additional notes
Made changes like

```java
partitionMovementTasksByState.getOrDefault(ExecutionTaskState.PENDING, 0)
```

Now the code uses `getOrDefault` for accessing task states, which avoids potential NullPointerExceptions if any state is not present in the map.

## Categorization  
- [ ] documentation
- [x] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [ ] other

## This PR resolves # 

This PR resolves #1972

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1613,6 +1613,8 @@ public class Executor {
       LOG.info("User task {}: Starting {} inter-broker partition movements.", _uuid, numTotalPartitionMovements);
 
       int partitionsToMove = numTotalPartitionMovements;
+      Set<Integer> throttledBrokers = new HashSet<>();
+
       // Exhaust all the pending partition movements.
       while ((partitionsToMove > 0 || !inExecutionTasks().isEmpty()) && _stopSignal.get() == NO_STOP_EXECUTION) {
         // Get tasks to execute.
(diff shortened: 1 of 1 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1644,7 +1656,11 @@ private void interBrokerMoveReplicas() throws InterruptedException, ExecutionExc
             .collect(Collectors.toList());
         inProgressTasks.addAll(inExecutionTasks());
 
-        throttleHelper.clearThrottles(completedTasks, inProgressTasks);
+          // Clear throttles only after all tasks are complete for the brokers involved
+          if (partitionsToMove == 0) {
```
CCisGG, line 1660: I think you may want to check all of these conditions: 

(partitionsToMove == 0 && inExecutionTasks().isEmpty()) ||  (_stopSignal.get() != NO_STOP_EXECUTION)

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1621,11 +1623,21 @@ private void interBrokerMoveReplicas() throws InterruptedException, ExecutionExc
 
         AlterPartitionReassignmentsResult result = null;
         if (!tasksToExecute.isEmpty()) {
-          throttleHelper.setThrottles(tasksToExecute.stream().map(ExecutionTask::proposal).collect(Collectors.toList()));
+            // Set throttles only if not already set for brokers involved
+            List<ExecutionProposal> proposals = tasksToExecute.stream().map(ExecutionTask::proposal).collect(Collectors.toList());
+            for (ExecutionProposal proposal : proposals) {
```
CCisGG, line 1628: Does it makes more sense to add a map inside throttle helper, so that you don't need to expose the logic in executor. I think the thorttle helper should be responsible to remember which broker is already throttled and just don't send redundant request to those brokers. Same with the clear throttle logic.

CCisGG (changes requested):
Added a few comments. Also please unit tests for it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
aswinayyolath: Okay 👍

HerveRiviere:
Hello @aswinayyolath 
Thanks for the contribution it can solve a pain point about rebalance execution slowness in my organisation ( CC can take up to 15 min to launch rebalance in case of a cluster with about 100 node and 50k partitions...)  

I see the review is on hold since Nov 2024, do you plan to work again on it ? 
If not I can take the handover to address comments, let me know :) 
Thanks !

aswinayyolath:
@HerveRiviere Thanks for reaching out! I'm currently tied up with other priority items, so please feel free to pick it up. I appreciate you offering to help!

PR #2221, opened 2024-11-08
Fix: intra.broker.goals cannot be configured as default.goals

k0b3rIT:
## Summary
1. Why: KafkaCruiseControlConfig sanity checks prevent to configure inter broker goals as default goals.
2. What: Modify the sanity check to allow adding intra broker goals to the `default.goals`

## Expected Behavior
I can add (enable) goals to the `default.goals` list from both the intra and inter broker goals
image

## Actual Behavior
Sanity checks ensure the following rules:
- **goals and intra broker goals should be disjoint** (an intra broker goal cant be in the `goals` and the `intra.broker.goals` at the same time)
- **`default.goals` should be a subset of `goals`** (As a result of this, we could not add intra broker goal to the `default.goals`)
- **`intra.broker.goals` cant be empty** (you could not move all the intra broker roles to the goals)
image

We could not add intra goal to the `default.goals` without a trick (see in the workaround section) 

## Steps to Reproduce
Try to enable an intra broker goal by adding it to the `default. goals`

## Known Workarounds
You can remove 1 intra broker goal from the `intra.broker.goals` config and add that to the `goals` (keep in mind that the other one should be in the `intra.broker.goals` as the list cant be empty) this way you can add that 1 goal to the `default.goals` and enable it

Note that this way we still cant enable both of the currently available intra broker goals as 1 have to be in the intra.broker.goals list to avoid hitting the mentioned sanity check. So only 1 intra goal at a time.

## Categorization
- [ ] documentation
- [x] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [ ] other

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
@@ -197,8 +197,8 @@ public class KafkaCruiseControlConfig extends AbstractConfig {
       throw new ConfigException("Attempt to configure default goals configuration with an empty list of goals.");
     }
 
-    // Ensure that default goals are supported inter-broker goals.
-    if (defaultGoalNames.stream().anyMatch(g -> !interBrokerGoalNames.contains(g))) {
+    // Ensure that default goals are supported inter-broker or intra-broker goals.
+    if (defaultGoalNames.stream().anyMatch(g -> !interBrokerGoalNames.contains(g) && !intraBrokerGoalNames.contains(g))) {
       throw new ConfigException(String.format("Attempt to configure default goals with unsupported goals (%s:%s and %s:%s).",
                                               AnalyzerConfig.DEFAULT_GOALS_CONFIG, defaultGoalNames,
                                               AnalyzerConfig.GOALS_CONFIG, interBrokerGoalNames));
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
@@ -199,9 +199,10 @@ public class KafkaCruiseControlConfig extends AbstractConfig {
 
     // Ensure that default goals are supported inter-broker or intra-broker goals.
     if (defaultGoalNames.stream().anyMatch(g -> !interBrokerGoalNames.contains(g) && !intraBrokerGoalNames.contains(g))) {
-      throw new ConfigException(String.format("Attempt to configure default goals with unsupported goals (%s:%s and %s:%s).",
+      throw new ConfigException(String.format("Attempt to configure default goals with unsupported goals (%s:%s and %s:%s and %s:%s).",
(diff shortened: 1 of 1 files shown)


nattilabalint (approved):

viktorsomogyi (approved):
@mhratson would you please review this?

viktorsomogyi:
@CCisGG would you review this please if you have some time?

imans777:
?

k0b3rIT:
> ?

Hi @imans777,
Do you have a concern regarding this change?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
```diff
@@ -197,8 +197,8 @@ private void sanityCheckGoalNames() {
       throw new ConfigException("Attempt to configure default goals configuration with an empty list of goals.");
     }
 
-    // Ensure that default goals are supported inter-broker goals.
-    if (defaultGoalNames.stream().anyMatch(g -> !interBrokerGoalNames.contains(g))) {
+    // Ensure that default goals are supported inter-broker or intra-broker goals.
+    if (defaultGoalNames.stream().anyMatch(g -> !interBrokerGoalNames.contains(g) && !intraBrokerGoalNames.contains(g))) {
       throw new ConfigException(String.format("Attempt to configure default goals with unsupported goals (%s:%s and %s:%s).",
                                               AnalyzerConfig.DEFAULT_GOALS_CONFIG, defaultGoalNames,
                                               AnalyzerConfig.GOALS_CONFIG, interBrokerGoalNames));
```
CCisGG, line 204: Looks like you also need to update the exception logs to make it more clear.

CCisGG (changes requested):

CCisGG:
To me, allowing intra broker goals to be added into default goals is a valid intention. I'm just trying to make sure this won't break users who don't use intra broker goals. From the code change, looks like it won't.

k0b3rIT:
> To me, allowing intra broker goals to be added into default goals is a valid intention. I'm just trying to make sure this won't break users who don't use intra broker goals. From the code change, looks like it won't.

@CCisGG thanks for your feedback.
My understanding is that intra.broker.goals is just a list of supported intra goals, just like the goals is the list of supported inter goals.
There is no other way to enable the intra broker roles, other than adding it to the default goals.

But now it is not possible, just with the mentioned workaround.
So I dont think it will break anyone.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/KafkaCruiseControlConfig.java
k0b3rIT: You are right, I have fixed it!

CCisGG (approved):

imans777:
> > ?
> 
> Hi @imans777, Do you have a concern regarding this change?

Hey there, I just wanted to pop up this change to be reviewed and merged sooner :)

k0b3rIT:
> > > ?
> > 
> > 
> > Hi @imans777, Do you have a concern regarding this change?
> 
> Hey there, I just wanted to pop up this change to be reviewed and merged sooner :)

Ahh, got it, THX!

k0b3rIT:
Thank you for the approval @CCisGG! Do you have write access to merge this PR?

CCisGG:
Merged. Thanks for the change!

PR #2244, opened 2025-01-29
Remove zk upstream

### LINKED ISSUE #2243: Remove ZooKeeper from Cruise Control since Kafka 4.0 will not support it
Remove ZooKeeper from Cruise Control since Kafka 4.0 will not support it.
Cruise Control use ZK in production codes and in test codes as well. The most of the production codes - where ZK is in use - are deprecated or not defaults and there is already another solution. Furthermore, Cruise Control is able to work without ZK in its current state.
The tests are heavily depends on ZK since the integration ones are using Kafka brokers with ZK as metadata store, additionally ZK client is also used, which can be replaced with Kafka AdminClient.
We should remove ZK from Cruise Control, since CC not requires it and ZK will be removed from Kafka as well.

egyedt:
## Summary
1. Why: Kafka 4.0 will not support ZooKeeper.
2. What: Remove ZooKeeper from Cruise Control.

## Expected Behavior
Cruise Control is able to work without ZooKeeper and no ZooKeeper dependency is needed.

## Actual Behavior
Cruise Control is able to work without ZooKeeper but ZooKeeper dependency is needed for some features and tests.

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [X] refactor
- [ ] security/CVE
- [ ] other

This PR resolves #2243.


egyedt:
Hi @bgrishinko, could you please review this change?
Thanks in advance!

danielgospodinow:
Great effort! Kudos! I hope your PR gets reviewed.

kyguy (commented):
Given that ZooKeeper support was deprecated in Kafka 3.5.0 and fully removed in Kafka 4.0.0 this seems like a reasonable step to me. 

In terms of implementation, we should aim to maintain forward and backward compatibility with as many Kafka versions as possible to serve the broadest range of users in the community. 

To that end, since the `main` Cruise Control branch supports Kafka 3.7.0+, we should ensure that this update allows users to run against Zookeeper-based Kafka clusters. This should be possible by replacing existing ZkClient usages with the Admin API where applicable. There are already some existing Cruise Control classes that do this to replacement so we can safely remove those old ZkClient-specific classes as part of this PR. This would enable us to support Kafka users from versions 3.7.0 through 4.0.0 and beyond regardless of whether they are using a Zookeeper-based or Kraft-based clusters.

@egyedt Once the CI tests passing I would be happy to take a closer look at the code and review the changes if you would like!

egyedt:
@kyguy the CI was green, but as @pnagy-cldr mentioned, in this current form (now also the Kafka 4.0 support is added to this PR) the #2308 is also required, that is why the CI is not green now. 

We have separate commits at this PR (ZK removal, ZK removal from tests and Kafka 4.0 support) and also the #2308 is needed (remove Java 11 support). Should we cherry-pick #2308 changes into this PR? Or should we separate Kafka 4.0 support into a new PR? (If we will do this, then somehow Kafka 4.0 support must depend on this ZK removal and Java 11 support removal PR)

Furthermore, as you mentioned Admin API is used by default every place, where the usage of ZK Client was also possible. So there will be no change in the default funcionality and there will be no feature loss with this change. 
The only place where ZK was a bit useful is the backup store to save failed brokers. (But this was also deprecated, since failed.brokers.file.path can be used instead as persistent file storage <- this is the default behaviour.)

kyguy:
> the CI was green, but as @pnagy-cldr mentioned, in this current form (now also the Kafka 4.0 support is added to this PR) the https://github.com/linkedin/cruise-control/pull/2308 is also required, that is why the CI is not green now.

Thanks @egyedt, that makes sense.

> We have separate commits at this PR (ZK removal, ZK removal from tests and Kafka 4.0 support) and also the https://github.com/linkedin/cruise-control/pull/2308 is needed (remove Java 11 support). Should we cherry-pick https://github.com/linkedin/cruise-control/pull/2308 changes into this PR? Or should we separate Kafka 4.0 support into a new PR? (If we will do this, then somehow Kafka 4.0 support must depend on this ZK removal and Java 11 support removal PR)

Personally, I would keep the ZK removal changes (including related test updates) in this PR and move the Kafka 4.0 support changes into a separate PR. From what I understand the ZK removal changes here don't dependent on [Java 11 support removal](https://github.com/linkedin/cruise-control/pull/2308), is that correct? 

Splitting the work this way could help unblock this PR and make both PRs easier to review!

> Furthermore, as you mentioned Admin API is used by default every place, where the usage of ZK Client was also possible. So there will be no change in the default funcionality and there will be no feature loss with this change.

That is awesome, thanks!

gradle.properties
```diff
@@ -2,8 +2,7 @@ org.gradle.daemon=false
 org.gradle.parallel=false
 org.gradle.jvmargs=-Xms512m -Xmx512m
 scalaVersion=2.13.13
-kafkaVersion=3.9.1
-zookeeperVersion=3.9.3
+kafkaVersion=4.0.0
```
danielgospodinow: Just for my understanding - do we bump to `4.0.0` just for the sake of using newer Kafka APIs in CC's testing?

So if we were to keep `3.9.1`, we could've had that too - just the tests would've looked differently.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterTest.java
```diff
@@ -207,19 +209,26 @@ public void testUpdatingMetricsTopicConfig() throws InterruptedException {
     brokerConfig.put(CRUISE_CONTROL_METRICS_TOPIC_AUTO_CREATE_CONFIG, "true");
     brokerConfig.put(CRUISE_CONTROL_METRICS_TOPIC_NUM_PARTITIONS_CONFIG, "2");
     brokerConfig.put(CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR_CONFIG, "1");
-    CCEmbeddedBroker broker = new CCEmbeddedBroker(brokerConfig);
-    // Restart broker
-    broker.startup();
-    // Wait for broker to boot up
-    Thread.sleep(5000);
+    try (CCEmbeddedBroker broker = new CCEmbeddedBroker(brokerConfig)) {
```
danielgospodinow, line 212: Nice catch here!

gradle.properties
danielgospodinow: Also, bumping to `4.0.0` doesn't automatically prevent us from talking to older Kafka clusters, right? Assuming `4.0.0` clients have compatibility with older Kafka broker versions and that CC doesn't use specific `4.0.0` features, of course.

pnagy-cldr:
@kyguy @danielgospodinow 
Now that Java 17 support has been merged, I rebased this PR and it should be green now.
You asked earlier to create a new PR for Bumping kafka version to 4.0.0 instead of having it here.
I can do it, but since Kafka 4.0.0 does not contain ZK anymore, It won't even compile. ZK removal is meaningless alone, it is needed for bumping Kafka. So I could either:
1. Leave it as is. Bumping Kafka, and removing ZK is a separated commit
2. Create a new PR for Kafka bump, but it will need to depend on ZK removal, otherwise it won't even compile. Until ZK removal is not merged, the new PR will still contain ZK removal commit to make sure it compiles, so the new PR will be the same as this one until ZK removal merge :)
What sounds better for you guys?

pnagy-cldr:
@danielgospodinow @CCisGG @kyguy I removed the **kafka bump to 4.0.0** commit from this PR.
Once this is merged, I will create a new PR based on main with Kafka version bump to 4.0.0.

README.md
```diff
@@ -91,9 +91,9 @@ Control). The metrics reporter periodically samples the Kafka raw metrics on the
     * If the default broker cleanup policy is `compact`, make sure that the topic to which Cruise Control metrics
     reporter should send messages is created with the `delete` cleanup policy -- the default metrics reporter topic
     is `__CruiseControlMetrics`.
-2. Start ZooKeeper and Kafka server ([See tutorial](https://kafka.apache.org/quickstart)).
+2. Start Kafka server ([See tutorial](https://kafka.apache.org/quickstart)).
```
kyguy, line 94:
```suggestion
2. Start Kafka server ([See tutorial](https://kafka.apache.org/quickstart)), and if you're using a ZooKeeper-based Kafka cluster also start a ZooKeeper server.
```

README.md
```diff
@@ -91,9 +91,9 @@ Control). The metrics reporter periodically samples the Kafka raw metrics on the
...
 3. Modify `config/cruisecontrol.properties` of Cruise Control:
-    * (Required) fill in `bootstrap.servers` and `zookeeper.connect` to the Kafka cluster to be monitored.
+    * (Required) fill in `bootstrap.servers` to the Kafka cluster to be monitored.
```
kyguy, line 96:
```suggestion
    * (Required) fill in `bootstrap.servers`, and if you're using a ZooKeeper-based Kafka cluster also `zookeeper.connect` for the Kafka cluster to be monitored.
```

build.gradle
```diff
@@ -481,9 +480,10 @@ project(':cruise-control-metrics-reporter') {
     testImplementation "org.apache.kafka:kafka-clients:$kafkaVersion:test"
     testImplementation "org.apache.kafka:kafka-server-common:$kafkaVersion"
     testImplementation "org.apache.kafka:kafka-group-coordinator:$kafkaVersion"
+    testImplementation "org.apache.kafka:kafka-metadata:$kafkaVersion"
+    testImplementation "org.apache.kafka:kafka-raft:$kafkaVersion"
```
kyguy, line 484: Although I think we should avoid adding more dependence on internal Kafka APIs [1], from what I understand these new dependencies are necessary for the new `CCEmbeddedKraftController` and related classes needed for the immediate move off of internal Kafka Zookeeper APIs.

We can remove these dependencies in a future PR, maybe as part of the migration work to use TestContainers in place of internal Kafka APIs

[1] https://github.com/linkedin/cruise-control/issues/2282

config/cruise_control_jaas.conf_template
```diff
@@ -1,16 +1,4 @@
...
-//Enter appropriate Client entry for secured zookeeper client connections
-Client {
-  com.sun.security.auth.module.Krb5LoginModule required
-  useKeyTab=true
-  keyTab="/path/to/zookeeper_client.keytab"
-  storeKey=true
-  useTicketCache=false
-  principal="zookeeper_client@<REALM>";
-};
-
```
kyguy, line 13: I assume that the `Admin` API we are migrating to as part of this PR allows Cruise Control to connect to secured Zookeeper clusters without this config, hence why this is being removed, is that correct?

config/cruisecontrol.properties
```diff
@@ -173,12 +173,6 @@ num.proposal.precompute.threads=1
 # Configurations for the executor
 # =======================================
 
-# The zookeeper connect of the Kafka cluster
-zookeeper.connect=localhost:2181/
```
kyguy, line 177: Similar comment to the one above, this configuration is unnecessary with the move to the `Admin` API, correct? We don't need this configuration to have Cruise Control to connect to a Zookeeper-based Kafka cluster?

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBroker.java
```diff
@@ -5,51 +5,49 @@
...
     _hosts = new HashMap<>();
+    _logDirs = new ArrayList<>();
 
     try {
       // Also validates the config
       KafkaConfig kafkaConfig = new KafkaConfig(config, true);
       parseConfigs(config);
 
-      _kafkaServer = createKafkaServer(kafkaConfig);
+      _kafkaServer = new CCKafkaRaftServer(kafkaConfig, config.get(CLUSTER_ID_CONFIG).toString(), Time.SYSTEM);
```
kyguy, line 45: I assume this maintains compatibility with the same Kafka versions that the original function did? Is that correct? I guess for the `main` Cruise Control branch we would want to maintain compilation support with at least Kafka 3.7 but I think the wider range the better!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedKRaftController.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+package com.linkedin.kafka.cruisecontrol.metricsreporter.utils;
+
+import kafka.server.KafkaConfig;
+import org.apache.commons.io.FileUtils;
+import org.apache.kafka.common.utils.Time;
+import org.apache.kafka.network.SocketServerConfigs;
+import org.apache.kafka.raft.QuorumConfig;
+import org.apache.kafka.server.config.KRaftConfigs;
+import org.apache.kafka.server.config.ServerLogConfigs;
```
kyguy, line 12: I know the internal Kafka API usages are necessary for classes like `CCKafkaRaftServer` but for the sake of making the future migration off of internal Kafka APIs easier we could isolate the internal APIs usages to that class and removing them here in this class by hard-coding the configuration keys.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUnitTestUtils.java
```diff
@@ -18,6 +18,7 @@
 import com.linkedin.kafka.cruisecontrol.detector.NoopTopicAnomalyFinder;
 import com.linkedin.kafka.cruisecontrol.monitor.metricdefinition.KafkaMetricDef;
 import com.linkedin.kafka.cruisecontrol.monitor.sampling.NoopSampler;
+import org.apache.kafka.server.config.ZkConfigs;
```
kyguy, line 21: To further reduce the dependence on the internal Kafka APIs we could remove this line here and hard-code the configuration key below instead!

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUnitTestUtils.java
```diff
@@ -51,7 +52,7 @@ public static Properties getKafkaCruiseControlProperties() {
             TestConstants.DEFAULT_CLUSTER_CONFIGS_FILE)).getFile();
     String brokerSetsDataFile = Objects.requireNonNull(KafkaCruiseControlUnitTestUtils.class.getClassLoader().getResource(
             TestConstants.DEFAULT_BROKER_SET_RESOLVER_FILE)).getFile();
-    props.setProperty(ExecutorConfig.ZOOKEEPER_CONNECT_CONFIG, "localhost:2121");
+    props.setProperty(ZkConfigs.ZK_CONNECT_CONFIG, "localhost:2121");
```
kyguy, line 55: We could hard-code the configuration key here instead of using the internal Kafka API for the configuration variable.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizerTest.java
```diff
@@ -97,7 +97,7 @@ private GoalOptimizer createGoalOptimizer() {
   private GoalOptimizer createGoalOptimizer(Properties overrideProps) {
     Properties props = new Properties();
     props.setProperty(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG, "bootstrap.servers");
-    props.setProperty(ExecutorConfig.ZOOKEEPER_CONNECT_CONFIG, "connect:1234");
+    props.setProperty(ZkConfigs.ZK_CONNECT_CONFIG, "connect:1234");
```
kyguy, line 100: We could hard-code the configuration key here instead of using the internal Kafka API for the configuration variable.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolverTest.java
```diff
@@ -35,7 +35,7 @@ private static BrokerSetResolver getBrokerSetResolver(String configFileName, Cla
     BrokerSetResolver brokerSetResolver = new BrokerSetFileResolver();
     String fileName = Objects.requireNonNull(clazz.getClassLoader().getResource(configFileName)).getFile();
     Map<String, Object> configs =
-        Map.of(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG, "bootstrap.servers", ExecutorConfig.ZOOKEEPER_CONNECT_CONFIG, "connect:1234",
+        Map.of(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG, "bootstrap.servers", ZkConfigs.ZK_CONNECT_CONFIG, "connect:1234",
```
kyguy, line 38: We could hard-code the configuration key here instead of using the internal Kafka API for the configuration variable.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolverTest.java
```diff
@@ -69,7 +69,7 @@ public void testNonExistentFile() {
     BrokerSetResolver brokerSetResolver = new BrokerSetFileResolver();
     String fileName = "testBrokerSetz.json";
     Map<String, Object> configs =
-        Map.of(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG, "bootstrap.servers", ExecutorConfig.ZOOKEEPER_CONNECT_CONFIG, "connect:1234",
+        Map.of(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG, "bootstrap.servers", ZkConfigs.ZK_CONNECT_CONFIG, "connect:1234",
```
kyguy, line 72: We could hard-code the configuration key here instead of using the internal Kafka API for the configuration variable.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/CaseInsensitiveGoalConfigTest.java
```diff
@@ -45,7 +45,7 @@ public static Collection<Object[]> data() {
     Collection<Object[]> params = new ArrayList<>();
 
     Properties sharedProps = new Properties();
-    sharedProps.setProperty(ExecutorConfig.ZOOKEEPER_CONNECT_CONFIG, "localhost:2121");
+    sharedProps.setProperty(ZkConfigs.ZK_CONNECT_CONFIG, "localhost:2121");
```
kyguy, line 48: We could hard-code the configuration key here instead of using the internal Kafka API for the configuration variable.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ConcurrencyAdjusterTest.java
```diff
@@ -304,7 +305,7 @@ private static Properties getExecutorProperties() {
     Properties props = new Properties();
     props.setProperty(MonitorConfig.BOOTSTRAP_SERVERS_CONFIG, "bootstrap.servers");
     props.setProperty(MonitorConfig.METRIC_SAMPLER_CLASS_CONFIG, NoopSampler.class.getName());
-    props.setProperty(ExecutorConfig.ZOOKEEPER_CONNECT_CONFIG, "localhost:2121");
+    props.setProperty(ZkConfigs.ZK_CONNECT_CONFIG, "localhost:2121");
```
kyguy, line 308: We could hard-code the configuration key here instead of using the internal Kafka API for the configuration variable.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReaderTest.java
```diff
@@ -220,6 +229,10 @@ public void testMaintenanceEventTopicCreationUpdateAndRead()
                                                                                   parameterConfigOverrides);
 
     assertNotNull(maintenanceEventReader);
+    long start = System.currentTimeMillis();
+    while (System.currentTimeMillis() < start + 15000) {
```
kyguy, line 233: Busy loops like this and sleeps used for waiting for an action to complete eats resources and can lead to test flakes, I have seen them throughout the existing tests. It doesn't have to be part of this PR but we should consider adding some logic to check if the action is complete at some interval as opposed to busy waiting. Maybe it is something we can fix up in a future PR.

docs/wiki/User Guide/Secure-zookeeper-configuration.md
```diff
@@ -1,15 +0,0 @@
...
-* Set the config _zookeeper.security.enabled_ in "$base_dir/config/cruisecontrol.properties" to _true_.
-* Rename the file "$base_dir/config/cruise_control_jaas.conf_template" to "$base_dir/config/cruise_control_jaas.conf". 
-* In the file cruise_control_jaas.conf, enter the appropriate _Client{ .. }_ entry for the zookeeper client.
-
-Cruise Control will export the "$base_dir/config/cruise_control_jaas.conf" configuration file only if it is present.
-Please ensure that the jaas file contains the correct entry for successful authentication. The authentication failure/success
-messages will appear in the Cruise Control logs on startup.
-
-NOTE: If using the SASL protocol, you could enter the _KafkaClient{ .. }_ entry here as this configuration file will be
-exported.
```
kyguy, line 15: So none of these configurations are relevant any more with this move to the `Admin` API? We can connect and use a secured Zookeeper-based Kafka cluster without these configurations, correct?

egyedt: yes

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReaderTest.java
egyedt: I agree!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterTest.java
egyedt: thx

gradle.properties
egyedt: It will be compiled with Kafka 4.0, but it will work with different Kafka versions as well.
Anyway this will be done in a follow-up PR based on the final discussion.

egyedt: yes, you can work with other Kafka clusters with different versions as well.

README.md
egyedt: nice catch

README.md
egyedt: nice catch

config/cruise_control_jaas.conf_template
egyedt: Yes, there will be no direct communication between ZK and CC anymore.
CC communicates with Kafka and Kafka with ZK and finally Kafka will respond the answer for CC. So only "indirect" communication can happen between CC and ZK in the future, so we do not need this template anymore.

config/cruisecontrol.properties
egyedt: yes, your understanding is correct in my opinion.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBroker.java
egyedt: Yes, Kafka 3.7.0 was able to run in KRaft mode. Anyway this class is only about tests. CC is still able to run with a Kafka which use ZK as metadata store.

README.md
pnagy-cldr: @kyguy This is not needed:
 **and if you're using a ZooKeeper-based Kafka cluster also `zookeeper.connect**
This is what the zookeeper removal PR is about, CruiseControl will not connect to zookeeper directly anymore, it will rather use Kafka admin API for getting the same information. So setting a zookeeper.connect is useless.

build.gradle
pnagy-cldr: I agree with you about using as few kafka server dependencies as possible since in an ideal world, CC is only a Kafka client. But CCEmbeddedKraftController is needed for testing. As you also mentioned, reducing these kind of dependencies would be a bigger effort, I won't touch this part in the followup PR, https://github.com/linkedin/cruise-control/issues/2282 might be appropriate for this.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedKRaftController.java
pnagy-cldr: I will do this in the followup PR, thanks for the idea. The only disadvantage I can see is that if we hardcode property names, they are not following upstream changes. Since property names rarely/never change, this might be a low risk for the cost of independence.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUnitTestUtils.java
pnagy-cldr: ZkConfigs should not be used anywhere/anymore, since CC will not connect to ZK directly.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUnitTestUtils.java
pnagy-cldr: ZkConfigs should not be used anywhere/anymore, since CC will not connect to ZK directly. I think these kind of lines should have been removed in ZK removal, but I will remove it in the followup PR, since Kafka 4.0 does not even have ZkConfigs :)

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizerTest.java
pnagy-cldr: I removed it in Kafka version bump. Reason and details are the same as above.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolverTest.java
pnagy-cldr: I removed it in Kafka version bump. Reason and details are the same as above.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/BrokerSetFileResolverTest.java
pnagy-cldr: I removed it in Kafka version bump. Reason and details are the same as above.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/config/CaseInsensitiveGoalConfigTest.java
pnagy-cldr: I removed it in Kafka version bump. Reason and details are the same as above.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ConcurrencyAdjusterTest.java
pnagy-cldr: I removed it in Kafka version bump. Reason and details are the same as above.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/detector/MaintenanceEventTopicReaderTest.java
pnagy-cldr: I think this is not needed at all, it seems a leftover, so I removed it in followup PR.
The next line is a verify:
    verify(newPartitionCount, newRF, newRetentionMs, true);
Which has been improved in ZK removal PR to use:
waitUntilTrue(() -> {...}
Which periodically checks for actions to complete.

Sidenote: Awaitility is a good library for periodically checking for actions to be completed, and it is highly configurable. If you would like to remove such manually written loops for checking completed actions in the future, Awaitility could be a good choice.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBroker.java
kyguy: It's not just runtime compatibility but build compatibility as well, from what I understand it was one of the requirements for Kafka version/API changes from LinkedIn for some PRs/upgrades in the past.

Anyways, It looks like the `main` branch already lost build compatibility with Kafka 3.7.0 before this PR so it doesn't matter. It looks like we can easily get the Kafka 3.7.0 build support back when we remove some private Kafka API config classes we use in some of the tests. We can handle this in a future PR!

Thanks for the response, looks like it isn't an issue for this PR!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedKRaftController.java
kyguy: > Since property names rarely/never change, this might be a low risk for the cost of independence.

Was thinking along similar lines!

kyguy:
Thanks for the fixes/responses @pnagy-cldr @egyedt!

PR #2248, opened 2025-01-31
Java 17 support

akatona84:
## Summary
1. Why: To be able to build and test with java 17 too.
2. What: some spotbugs fixes, version bumps and adjusting jvm args for testing

## Expected Behavior
build goes smoothly with java 17

## Actual Behavior
build and tests fail with java 17

## Steps to Reproduce
1. set your java to 17
2. run `./gradlew build` and see the crashes

## Known Workarounds
-

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [x] other

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ plugins {
   id "idea"
   id "jacoco" // Java Code Coverage plugin
   id "com.github.ben-manes.versions" version "0.42.0"
-  id "com.github.spotbugs" version "6.0.6" apply false
+  id "com.github.spotbugs" version "6.0.25" apply false
   id "checkstyle"
   id "org.openapi.generator" version "5.3.0"
 }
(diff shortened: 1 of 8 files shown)


akatona84:
This change is not movin away from jdk11 yet, it's just making it possible to compile with java17 too.

build.gradle
```diff
@@ -45,6 +45,18 @@ project.ext {
   buildVersionFileName = "cruise-control-version.properties"
   commitId = project.hasProperty('commitId') ? commitId : null
   scalaBinaryVersion = getScalaBinaryVersion(scalaVersion)
+
+  defaultTestJvmArgs = []
+  // "JEP 403: Strongly Encapsulate JDK Internals" causes some tests to fail when they try
+  // to access internals (often via mocking libraries). We use `--add-opens` as a workaround
```
viktorsomogyi, line 51: often or always? As I see it is only used in the test task def, so perhaps we could rephrase this.

akatona84: I think it's more about the occurrences rather than temporal frequency. OFC those tests always fail without the add-opens.
BTW I copied it from Kafka :D. Most of the cases tests were failing because of easymock and powermock to be specific, but I have no clue why exactly `java.time` had to be opened, gradle wasn't kind enough to provide me more info about the failing test(s) before I added it.

akatona84: In Kafka easymock/powermock was replaced with mockito over time. It's still not done, but that's the way to get rid of these add-opens lateron.

viktorsomogyi: Yea I agree that we should rather use Mockito, however I think that comes with some test refactoring as well, so let's not do that now.

viktorsomogyi (approved):

viktorsomogyi:
@bgrishinko would you please review this?

build.gradle
Wadimz: Hi folks!
Just checking in — do we have an estimated timeline for merging this? It’s a long-awaited improvement that we’re really looking forward to, and we’d love to see it move forward. 😊
Thanks in advance!

CCisGG (approved):
Took a quick pass on this and looks ok. Also relying on the other reviewers' approval. Since many people are waiting this excited change, let's merge this

PR #2253, opened 2025-02-26
Update cruise control to support 3.9

jlisam:
## Summary
1. Why: To support kafka 3.9
2. What: This PR updates the kafka libraries to 3.9 and mostly replaces `new SystemTime()` with `Time.System` 

## Expected Behavior
Should compile for 3.9 server code

## Actual Behavior
Tested this with local broker and it works

## Steps to Reproduce
1. Compile and generate jar for metrics reporter
2. Drop metrics reporter jar in kafka installation
3. Start 3.9 server

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [ ] other


shaneikennedy:
It looks like https://github.com/linkedin/cruise-control/pull/2254 is a bit more active, will that PR cover 3.9 compatibility as well? It would be nice to merge this PR tbh i think it makes sense to get 3.9 compat before 4.0 since there seems to be quite a few complexities in making 4.0 compat happen

CCisGG:
Hi @jlisam thanks for the change. We have decided to have our main branch support kafka 3.8+. Can you help to verify that this change is compatible with kafka 3.8?

CCisGG:
> @CCisGG yeah, I can definitely test that. Is there a test suite I can use to do that or just doing some local testing with some 3.8 brokers would be sufficient?

@jlisam Test with 3.8 brokers and also build/run unit tests with kafka version 3.8. We need to ensure the APIs you use in this PR are compatible with kafka 3.8

jlisam:
@CCisGG ok let me try that out. Sorry for the delay

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaClientsIntegrationTestHarness.java
```diff
@@ -58,7 +58,7 @@ protected void setSecurityConfigs(Properties clientProps, String certAlias) {
       clientProps.setProperty(CommonClientConfigs.SECURITY_PROTOCOL_CONFIG, protocol.name);
       clientProps.setProperty(SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG, "");
       try {
-        clientProps.putAll(TestSslUtils.createSslConfig(true, true, Mode.CLIENT, trustStoreFile, certAlias));
+        clientProps.putAll(TestSslUtils.createSslConfig(true, true, ConnectionMode.CLIENT, trustStoreFile, certAlias));
```
kyguy, line 61: This change will break compilation with Kafka versions earlier than 3.9.0. However, it will still run with earlier versions since the changes are limited to the test code. We made a similar decision when we upgraded the Cruise Control code base to Kafka 3.8.0.

Unless we start migrating the Cruise Control code base off of non-public Kafka API usages, we will continue to encounter breaking changes every time we upgrade Kafka. This is due to the unstable and frequently changing nature of Kafka’s internal APIs.

As an alternative to this change here, we could implement our own `createSslConfig` method that avoids internal Kafka APIs and use that to generate the SSL configurations for tests. This would require a bit more effort but would move us closer to eliminating our dependence on internal Kafka APIs while preserving compatibility with Kafka 3.8.0.

kyguy:
Support for Kafka 3.9.1 has been added as part of the PR here [1] The main branch is now compatible with Kafka versions in the range [3.8.0, 4.0.0). This can be closed!

CC @jlisam 

[1] https://github.com/linkedin/cruise-control/pull/2286

PR #2254, opened 2025-02-27
Replace deprecated methods to support Kafka 4.0.0

ShubhamRwt:
## Summary
1. Why: Cruise Control metrics reporter package seems to be using some deprecated method for some time now which are completely removed in Kafka 4.0.0. These deprecated method makes CC incompatible with Kafka 4.0 and beyond.
```
2025-02-22T22:27:38,802 ERROR Uncaught exception in thread 'CruiseControlMetricsReporterRunner': (org.apache.kafka.common.utils.KafkaThread) [CruiseControlMetricsReporterRunner]
java.lang.NoSuchMethodError: 'java.util.Map org.apache.kafka.clients.admin.DescribeTopicsResult.values()'
        at com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter.maybeIncreaseTopicPartitionCount(CruiseControlMetricsReporter.java:376) ~[cruise-control-metrics-reporter-2.5.141.jar:?]
        at com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter.maybeUpdateCruiseControlMetricsTopic(CruiseControlMetricsReporter.java:348) ~[cruise-control-metrics-reporter-2.5.141.jar:?]
        at com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter.run(CruiseControlMetricsReporter.java:395) ~[cruise-control-metrics-reporter-2.5.141.jar:?]
        at java.base/java.lang.Thread.run(Thread.java:840) [?:?] 
```        

2. What: This PR replaces the deprecated method with the ones that should be used in place of them .

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [ ] other

This PR resolves #<Replace-Me-With-The-Issue-Number-Addressed-By-This-PR> if any.

### DIFF
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -374,6 +374,7 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
     String cruiseControlMetricsTopic = _metricsTopic.name();
 
     try {
+      // For compatibility with Kafka 4.0 and beyond we must use new API methods.
       Method topicDescriptionMethod = topicNameValuesMethod();
 
       DescribeTopicsResult describeTopicsResult = _adminClient.describeTopics(Collections.singletonList(cruiseControlMetricsTopic));
@@ -521,8 +522,8 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
 
   /**
    * Attempts to retrieve the method for mapping topic names to futures from the {@link org.apache.kafka.clients.admin.DescribeTopicsResult} class.
-   * This method first tries to get the {@code topicNameValues()} method, which is available in Kafka 4.0.0 or later.
-   * If the method is not found, it falls back to trying to retrieve the {@code values()} method, which is available in Kafka 3.0.0 or earlier.
+   * This method first tries to get the {@code topicNameValues()} method, which is available in Kafka 3.1.0 and later.
+   * If the method is not found, it falls back to trying to retrieve the {@code values()} method, which is available in Kafka 3.9.0 and earlier.
    *
    * If neither of these methods is found, a {@link RuntimeException} is thrown.
    *
@@ -546,7 +547,7 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
         // Second we try to get the values() method
         topicDescriptionMethod = DescribeTopicsResult.class.getMethod("values");
       } catch (NoSuchMethodException exception) {
-        LOG.info("Failed to get method values() from DescribeTopicsResult class since we are probably on kafka 3.0.0 or older: ", exception);
+        LOG.info("Failed to get method values() from DescribeTopicsResult class: ", exception);
       }
     }
 
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -33,6 +34,7 @@ import org.apache.kafka.clients.admin.AlterConfigsResult;
 import org.apache.kafka.clients.admin.Config;
 import org.apache.kafka.clients.admin.CreateTopicsResult;
 import org.apache.kafka.clients.admin.DescribeConfigsResult;
+import org.apache.kafka.clients.admin.DescribeTopicsResult;
 import org.apache.kafka.clients.admin.NewPartitions;
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.admin.TopicDescription;
@@ -42,6 +44,7 @@ import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.KafkaException;
+import org.apache.kafka.common.KafkaFuture;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.ConfigResource;
 import org.apache.kafka.common.config.TopicConfig;
@@ -375,23 +378,36 @@ public class CruiseControlMetricsReporter implements MetricsReporter, Runnable {
     // so we have to use the new method "topicNameValues".
     // To make sure that the internal build pass, this logic is introduced to choose the API based on the Kafka version.
     try {
-      Method topicNameValuesMethod = null;
+      Method topicDescriptionMethod = null;
       try {
         // First we try to get the topicNameValues() method
-        topicNameValuesMethod = Class.forName("org.apache.kafka.clients.admin.DescribeTopicsResult").getMethod("topicNameValues");
+       topicDescriptionMethod = Class.forName("org.apache.kafka.clients.admin.DescribeTopicsResult").getMethod("topicNameValues");
       } catch (ClassNotFoundException | NoSuchMethodException exception) {
         LOG.info("Failed to get method topicNameValues() from DescribeTopicsResult class since we are probably on kafka 3.0.0 or older: ", exception);
       }
 
-      TopicDescription topicDescription = topicNameValuesMethod != null? _adminClient.describeTopics(Collections.singletonList(cruiseControlMetricsTopic)).topicNameValues()
-              .get(cruiseControlMetricsTopic).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS) : _adminClient.describeTopics(Collections.singletonList(cruiseControlMetricsTopic)).values()
-              .get(cruiseControlMetricsTopic).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
-
-      if (topicDescription.partitions().size() < _metricsTopic.numPartitions()) {
-        _adminClient.createPartitions(Collections.singletonMap(cruiseControlMetricsTopic,
-                                                               NewPartitions.increaseTo(_metricsTopic.numPartitions())));
+      if (topicDescriptionMethod == null) {
+        try {
+          // Second we try to get the values() method
+          topicDescriptionMethod = Class.forName("org.apache.kafka.clients.admin.DescribeTopicsResult").getMethod("values");
+        } catch (ClassNotFoundException | NoSuchMethodException exception) {
+          LOG.info("Failed to get method values() from DescribeTopicsResult class since we are probably on kafka 3.0.0 or older: ", exception);
+        }
       }
-    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+
+      if (topicDescriptionMethod != null) {
+        Map<String, KafkaFuture<TopicDescription>> topicDescriptionMap = (Map<String, KafkaFuture<TopicDescription>>) topicDescriptionMethod.invoke(_adminClient.describeTopics(Collections.singletonList(cruiseControlMetricsTopic)));
+
+        TopicDescription topicDescription = topicDescriptionMap.get(cruiseControlMetricsTopic).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+
+        if (topicDescription.partitions().size() < _metricsTopic.numPartitions()) {
+          _adminClient.createPartitions(Collections.singletonMap(cruiseControlMetricsTopic,
+                  NewPartitions.increaseTo(_metricsTopic.numPartitions())));
+        }
+      } else {
+        throw new NoSuchElementException("Unable to find both values() and topicNameValues() method in the DescribeTopicsResult class ");
+      }
+    } catch (InterruptedException | ExecutionException | TimeoutException | InvocationTargetException | IllegalAccessException e) {
       LOG.warn("Partition count increase to {} for topic {} failed{}.", _metricsTopic.numPartitions(), cruiseControlMetricsTopic,
                (e.getCause() instanceof ReassignmentInProgressException) ? " due to ongoing reassignment" : "", e);
     }
--- a/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
+++ b/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
@@ -5,6 +5,7 @@
 package com.linkedin.kafka.cruisecontrol.metricsreporter;
 
 import com.linkedin.kafka.cruisecontrol.metricsreporter.exception.CruiseControlMetricsReporterException;
+import com.linkedin.kafka.cruisecontrol.metricsreporter.exception.KafkaTopicDescriptionException;
 import com.linkedin.kafka.cruisecontrol.metricsreporter.metric.CruiseControlMetric;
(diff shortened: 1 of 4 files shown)


kyguy:
Hey @CCisGG could you take a look at this one? It is a small change, it future-proofs the Cruise Control metric reporter so it can be run on brokers using Kafka 4.0.

CCisGG (approved):
Looks good. Thanks! One minor question: will people use older kafka version be bothered from this change?

ShubhamRwt:
The replaced method are deprecated since Kafka 3.1.0 and the new methods were also introduced in Kafka 3.1.0. Sof if brokers are using Kafka older then 3.1.0, only then it can be a problem

kyguy:
> The replaced method are deprecated since Kafka 3.1.0 and the new methods were also introduced in Kafka 3.1.0. Sof if brokers are using Kafka older then 3.1.0, only then it can be a problem

+1

Here is a reference to the deprecated methods in the Kafka 3.1 API docs [1]

[1] https://kafka.apache.org/31/javadoc/org/apache/kafka/clients/admin/DescribeTopicsResult.html#values()

CCisGG:
Oops. We are internally still using kafka 3.0 so this will break our internal build. Can we choose the API based on the kafka version?

CCisGG (changes requested):
We are internally still using kafka 3.0 so this will break our internal build. Can we choose the API based on the kafka version?

ShubhamRwt:
Thanks for the info @CCisGG, I will add the corresponding changes to the PR

CCisGG (changes requested):
Hi @ShubhamRwt thanks for the change!

However I think this won't work, because the compiler still see both methods at compile time and one of the method doesn't exist. I think it will cause build errors.

I think you need to use invoke to invoke the method calls, instead of directly calling these methods. Both the old method and new method might need to called by "invoke".

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterTest.java
```diff
@@ -190,7 +202,42 @@ public void testUpdatingMetricsTopicConfig() throws ExecutionException, Interrup
     setSecurityConfigs(props, "admin");
     props.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers());
     AdminClient adminClient = AdminClient.create(props);
-    TopicDescription topicDescription = adminClient.describeTopics(Collections.singleton(TOPIC)).values().get(TOPIC).get();
+
+    // Starting with Kafka 4.0.0, the deprecated method "values()" class is completely removed from "org.apache.kafka.clients.admin.DescribeTopicsResult"
+    // so we have to use the new method "topicNameValues".
+    // To make sure that the internal build pass, this logic is introduced to choose the API based on the Kafka version
+    Method topicDescriptionMethod = null;
```
kyguy, line 209: I would do the same for the tests, we can separate the DescribeTopicsResult method detection logic into its own method, `topicNameValuesMethod()` in the `CruiseControlMetricRepoter` class, then call `topicNameValuesMethod()` in our tests to save us from the extra lines.

```suggestion
    Method topicDescriptionMethod = topicNameValuesMethod();
```

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -370,16 +373,41 @@ protected void maybeUpdateTopicConfig() {
...
+    // To make sure that the internal build pass, this logic is introduced to choose the API based on the Kafka version.
     try {
-      // Retrieve topic partition count to check and update.
-      TopicDescription topicDescription =
-          _adminClient.describeTopics(Collections.singletonList(cruiseControlMetricsTopic)).values()
-                      .get(cruiseControlMetricsTopic).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
-      if (topicDescription.partitions().size() < _metricsTopic.numPartitions()) {
-        _adminClient.createPartitions(Collections.singletonMap(cruiseControlMetricsTopic,
-                                                               NewPartitions.increaseTo(_metricsTopic.numPartitions())));
+      Method topicDescriptionMethod = null;
```
kyguy, line 381: It may be worth separating out the DescribeTopicsResult method detection into a separate method. This would keep the focus of the maybeIncreaseTopicPartitionCount() method on the original logic.

We could separate the DescribeTopicsResult method detection logic out like this: 

```
/**
   * Attempts to retrieve the method for mapping topic names to futures from the {@link org.apache.kafka.clients.admin.DescribeTopicsResult} class.
   * This method first tries to get the {@code topicNameValues()} method, which is available in Kafka 4.x or later.
   * If the method is not found, it falls back to trying to retrieve the {@code values()} method, which is available in Kafka 3.x or earlier.
   *
   * If neither of these methods is found, a {@link RuntimeException} is thrown.
   *
   * <p>This method is useful for ensuring compatibility with both older and newer versions of Kafka clients.</p>
   *
   * @return the {@link Method} object representing the {@code topicNameValues()} or {@code values()} method.
   * @throws RuntimeException if neither the {@code values()} nor {@code topicNameValues()} methods are found.
   */
  /* test */ static Method topicNameValuesMethod() {
    //
    Method topicDescriptionMethod = null;
    try {
      // First we try to get the topicNameValues() method
      topicDescriptionMethod = Class.forName("org.apache.kafka.clients.admin.DescribeTopicsResult").getMethod("topicNameValues");
    } catch (ClassNotFoundException | NoSuchMethodException exception) {
      LOG.info("Failed to get method topicNameValues() from DescribeTopicsResult class since we are probably on kafka 3.0.0 or older: ", exception);
    }

    if (topicDescriptionMethod == null) {
      try {
        // Second we try to get the values() method
        topicDescriptionMethod = Class.forName("org.apache.kafka.clients.admin.DescribeTopicsResult").getMethod("values");
      } catch (ClassNotFoundException | NoSuchMethodException exception) {
        LOG.info("Failed to get method values() from DescribeTopicsResult class since we are probably on kafka 3.0.0 or older: ", exception);
      }
    }

    if (topicDescriptionMethod != null) {
      return topicDescriptionMethod;
    } else {
      throw new RuntimeException("Unable to find both values() and topicNameValues() method in the DescribeTopicsResult class ");
    }
  }
```
Then call it like this:

```suggestion
      Method topicDescriptionMethod = topicNameValuesMethod();
```

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
```diff
@@ -107,7 +116,40 @@ public void testAutoCreateMetricsTopic() throws ExecutionException, InterruptedE
         Properties props = new Properties();
         props.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers());
         AdminClient adminClient = AdminClient.create(props);
-        TopicDescription topicDescription = adminClient.describeTopics(Collections.singleton(TOPIC)).values().get(TOPIC).get();
+
+        // Starting with Kafka 4.0.0, the deprecated method "values()" class is completely removed from "org.apache.kafka.clients.admin.DescribeTopicsResult"
+        // so we have to use the new method "topicNameValues".
+        // To make sure that the internal build pass, this logic is introduced to choose the API based on the Kafka version
+        Method topicDescriptionMethod = null;
```
kyguy, line 123: I would do the same for the tests, we can separate the DescribeTopicsResult method detection logic into its own method, `topicNameValuesMethod()` in the `CruiseControlMetricRepoter` class, then call `topicNameValuesMethod()` in our tests to save us from the extra lines.

```suggestion
    Method topicDescriptionMethod = topicNameValuesMethod();
```

kyguy (commented):
Nice work @ShubhamRwt, just took a quick pass, was thinking that it might be easier/neater if we organize the DescribeTopicsResult method detection logic into its own method in the CruiseControlMetricReporter class, we could call it something like "topicNameValuesMethod()".

I left more specific details in the code comments, let me know what you think!

ShubhamRwt:
@CCisGG @kyguy is it good to merge now?

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -370,16 +372,22 @@ protected void maybeUpdateTopicConfig() {
 
   protected void maybeIncreaseTopicPartitionCount() {
     String cruiseControlMetricsTopic = _metricsTopic.name();
+
     try {
-      // Retrieve topic partition count to check and update.
-      TopicDescription topicDescription =
-          _adminClient.describeTopics(Collections.singletonList(cruiseControlMetricsTopic)).values()
-                      .get(cruiseControlMetricsTopic).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      Method topicDescriptionMethod = topicNameValuesMethod();
```
kyguy, line 377: It may be worth adding a comment right above this line and other places where we call `topicNameValuesMethod()`, e.g. "// For compatibility with Kafka 4.0 and beyond we must use new API methods."

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
```diff
@@ -370,16 +372,23 @@ protected void maybeUpdateTopicConfig() {
...
-                      .get(cruiseControlMetricsTopic).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      // For compatibility with Kafka 4.0 and beyond we must use new API methods.
+      Method topicDescriptionMethod = topicNameValuesMethod();
+
+      DescribeTopicsResult describeTopicsResult = _adminClient.describeTopics(Collections.singletonList(cruiseControlMetricsTopic));
+
+      Map<String, KafkaFuture<TopicDescription>> topicDescriptionMap = (Map<String, KafkaFuture<TopicDescription>>) topicDescriptionMethod
+              .invoke(describeTopicsResult);
+
+      TopicDescription topicDescription = topicDescriptionMap.get(cruiseControlMetricsTopic).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
kyguy, line 385: We could also abstract these the above lines into a `getTopicDescription()` method to simplify the code across here and the tests even further.  If you want to go that route you could separate the code like this:
```
  /**
   * Retrieves the {@link TopicDescription} for the specified Kafka topic, handling compatibility
   * with Kafka versions 4.0 and above. This method uses reflection to invoke the appropriate method
   * for retrieving topic description information, depending on the Kafka version.
   *
   * @param _adminClient The Kafka {@link AdminClient} used to interact with the Kafka cluster.
   * @param cruiseControlMetricsTopic The name of the Kafka topic for which the description is to be retrieved.
   *
   * @return The {@link TopicDescription} for the specified Kafka topic.
   *
   * @throws KafkaTopicDescriptionException If an error occurs while retrieving the topic description,
   *         or if the topic name retrieval method cannot be found or invoked properly. This includes
   *         exceptions related to reflection (e.g., {@link NoSuchMethodException}), invocation issues,
   *         execution exceptions, timeouts, and interruptions.
   */
  /* test */ static TopicDescription getTopicDescription(AdminClient _adminClient, String cruiseControlMetricsTopic) throws KafkaTopicDescriptionException {
    try {
      // For compatibility with Kafka 4.0 and beyond we must use new API methods.
      Method topicDescriptionMethod = topicNameValuesMethod();
      DescribeTopicsResult describeTopicsResult = _adminClient.describeTopics(Collections.singletonList(cruiseControlMetricsTopic));
      Map<String, KafkaFuture<TopicDescription>> topicDescriptionMap = (Map<String, KafkaFuture<TopicDescription>>) topicDescriptionMethod
              .invoke(describeTopicsResult);

      return topicDescriptionMap.get(cruiseControlMetricsTopic).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
    }
    catch (InvocationTargetException | IllegalAccessException | ExecutionException | InterruptedException | TimeoutException | NoSuchMethodException e) {
      throw new KafkaTopicDescriptionException(String.format("Unable to retrieve config of Cruise Cruise Control metrics topic {}.", cruiseControlMetricsTopic), e);
    }
  }
```
Note that if you did this, it would probably be a good idea to create a custom exception class in the `cruisecontrol.metricsrepoter/exception` folder so you don't have to add all the Exception classes to the method signatures where this method is called.
```
public class KafkaTopicDescriptionException extends Exception {
    public KafkaTopicDescriptionException(String message, Throwable cause) {
        super(message, cause);
    }
}
```
This way, you can just call `getTopicDescription()` whenever you need the topic description for a topic and not worry about calling and invoking the right method there as well, it would just be done in the method here in the `CruiseControlMetricsReporter` class.

kyguy (commented):
Thanks for the updates @ShubhamRwt, looking better, I have left a little more feedback if you are interested, let me know what you think!

cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java
ShubhamRwt: I was looking into applying this suggestion but there is one more thing -> The call made in `CruiseControlMetricsReporter` is  
```
      return topicDescriptionMap.get(cruiseControlMetricsTopic).get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
while in the test we don't use the `get` method with `CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS`  parameters
```
 `topicDescriptionMap.get(TOPIC).get()`
 ```

kyguy: Does the inclusion of `.get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS)` interfere with the tests? From what I understand, it still returns the `TopicDescription` object we need. 

Even if the inclusion of `.get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS)`  did cause any problems in the tests, couldn't we remove it from the `getTopicDescription()` method itself and append it to the end of the places where we need it?

kyguy (commented):
Looking even better @ShubhamRwt, I left one last comment about creating a `getTopicDescription()` method to further reduce lines, let me know what you think!

shaneikennedy:
Sorry for the drive by comment here but does this get us compatibility with 3.9 as well? I imagine there will be lots of teams upgrading older versions of kafka to 3.9 before making the switch to 4.0, so 3.9 compatibility feels quite important

ShubhamRwt:
This PR is mostly to make sure that the cruise control metrics reporter can run on broker using 4.0 since it has multiple deprecated methods which are removed in Kafka 4.0.0(The metrics reporter works correctly with 3.9.0 since the deprecated method are not deleted in 3.9.0).

shaneikennedy:
> @shaneikennedy This PR is mostly to make sure that the cruise control metrics reporter can run on broker using 4.0 since it has multiple deprecated methods which are removed in Kafka 4.0.0(The metrics reporter works correctly with 3.9.0 since the deprecated method are not deleted in 3.9.0).

Gotcha, thanks for the info 🙌

CCisGG (approved):

CCisGG:
LGTM. Thanks @kyguy

ShubhamRwt:
Hi @CCisGG, can you help me understand the release cadence of Cruise Control? Are we going to have a release coming out in the next few days?

We were planning to ship the new release of Cruise Control with Strimzi 0.46.0 release which is going to use Kafka 4.0.0 since it will contain the fix[1] for issues we are currently having with Kafka 4.0.0.

[1] https://github.com/linkedin/cruise-control/pull/2254

CCisGG:
@ShubhamRwt it's usually on demand. I'll make a new release soon.

ShubhamRwt:
Thanks a lot @CCisGG

PR #2294, opened 2025-07-24
CVE fix: commons-lang3 v3.12.0 -> 3.20.0

stecurran-est:
Why:
Address CVE-2025-48924, a known vulnerability in org.apache.commons:commons-lang3.

What:
Added a Gradle constraint to enforce the use of commons-lang3 version 3.20.0, replacing the previously used 3.12.0, which is affected by the CVE.

Expected Behavior:
Vulnerability scans should no longer report the affected version of commons-lang3.
cruise-control should build successfully with the updated dependency.

Actual Behavior:
Vulnerability scanning no longer reports the affected version.
cruise-control builds successfully after the update.

Categorization:
security/CVE


kyguy (approved):
Tested this change with a few builds, everything looks solid!

PR #2304, opened 2025-08-17
Add bulk replication throttle mode (set throttle once inter-broker)

### LINKED ISSUE #1972: Redundant set/unset throttling during the rebalance

il-kyun:
## Summary
1. Why: Per-batch replication throttle set/clear causes excessive AdminClient calls when many batches involve the same brokers, significantly slowing rebalances. This PR proposes a bulk mode that sets the replication bytes throttle once before inter-broker movements and clears it once after, reducing Admin overhead and improving execution time. 
2. What:
   - Add `bulk.replication.throttle.enabled` (default: `false`) to `ExecutorConfig`.
   - In `Executor`, when enabled:
     - Set throttles once before inter-broker partition movements begin (filtering to existing topics).
     - Clear throttles once after all inter-broker partition movements complete, with summary logging (completed/aborted/dead/in-progress/aborting).
   - Keep existing per-batch behavior when disabled.
   - Update tests to cover both bulk and non-bulk paths.
   - Follow-ups: Improvements to the Kafka calls inside ReplicationThrottleHelper are being addressed in a separate PR; this PR focuses solely on bulk set/clear semantics.

## Expected Behavior
- With `bulk.replication.throttle.enabled=true`:
  - Throttles are applied once at the start of inter-broker moves and cleared once at the end.
  - Fewer AdminClient set/clear calls → faster rebalances.
  - No functional change to movement semantics; only fewer config mutations.
- With `bulk.replication.throttle.enabled=false`:
  - Current per-batch set/clear behavior remains unchanged.

## Actual Behavior
- Throttles are currently set before each batch and cleared after each batch if no other tasks are active on the same broker.
- When many batches affect the same brokers, the same throttle is repeatedly set/cleared.
- Each Admin request can take seconds (~20s observed locally), causing noticeable slowdowns.

## Steps to Reproduce
1. Prepare a rebalance plan with many inter-broker replica movements that repeatedly involve the same brokers.
2. Run the rebalance with current behavior (bulk disabled) and observe repeated set/clear throttle Admin calls per batch.
3. Measure end-to-end inter-broker phase duration.
4. Enable `bulk.replication.throttle.enabled=true` and re-run to observe reduced Admin calls and improved completion time.

## Additional evidence
1. Environment (test settings):
   - `concurrency.adjuster.max.partition.movements.per.broker=12`
   - `default.replica.movement.strategies=com.linkedin.kafka.cruisecontrol.executor.strategy.PrioritizeMinIsrWithOfflineReplicasStrategy,com.linkedin.kafka.cruisecontrol.executor.strategy.PrioritizeOneAboveMinIsrWithOfflineReplicasStrategy,com.linkedin.kafka.cruisecontrol.executor.strategy.PrioritizeSmallReplicaMovementStrategy,com.linkedin.kafka.cruisecontrol.executor.strategy.BaseReplicaMovementStrategy`
2. Scenario and results:
   - 900 partitions to rebalance (800 small, 100 large).
   - Before (bulk disabled): moving the 800 small partitions took over 1 hour 30 minutes.
   - After enabling `bulk.replication.throttle.enabled`: the 800 small partitions completed within a few minutes.
3. Logs: added summary logs when clearing bulk throttles (Completed/Aborted/Dead/InProgress/Aborting counts).
4. Follow-ups:
   - We are improving the Kafka call patterns inside `ReplicationThrottleHelper` in a separate PR; this PR intentionally limits scope to the bulk set/clear behavior.

## Categorization
- [ ] documentation
- [ ] bugfix
- [x] new feature
- [ ] refactor
- [ ] security/CVE
- [ ] other

This PR resolves #1972

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
@@ -1606,13 +1606,33 @@ public class Executor {
 
     private void interBrokerMoveReplicas() throws InterruptedException, ExecutionException, TimeoutException {
       Set<Integer> currentDeadBrokersWithReplicas = _loadMonitor.deadBrokersWithReplicas(MAX_METADATA_WAIT_MS);
-      ReplicationThrottleHelper throttleHelper = new ReplicationThrottleHelper(_adminClient, _replicationThrottle,
-          currentDeadBrokersWithReplicas);
+      ReplicationThrottleHelper throttleHelper = new ReplicationThrottleHelper(_adminClient, _replicationThrottle, currentDeadBrokersWithReplicas);
+      final boolean bulkThrottleEnabled = _config.getBoolean(ExecutorConfig.BULK_REPLICATION_THROTTLE_ENABLED_CONFIG);
       int numTotalPartitionMovements = _executionTaskManager.numRemainingInterBrokerPartitionMovements();
       long totalDataToMoveInMB = _executionTaskManager.remainingInterBrokerDataToMoveInMB();
       long startTime = System.currentTimeMillis();
       LOG.info("User task {}: Starting {} inter-broker partition movements.", _uuid, numTotalPartitionMovements);
 
+      // If bulk throttle is enabled, set throttles once for all pending inter-broker tasks before entering the loop.
+      if (bulkThrottleEnabled && _replicationThrottle != null && numTotalPartitionMovements > 0) {
+        ExecutionTasksSummary summaryAtStart = _executionTaskManager.getExecutionTasksSummary(
+            Collections.singleton(INTER_BROKER_REPLICA_ACTION));
+        Map<ExecutionTaskState, Set<ExecutionTask>> interBrokerTasksByState = summaryAtStart.filteredTasksByState()
+            .get(INTER_BROKER_REPLICA_ACTION);
+        Set<ExecutionTask> pendingAtStart = interBrokerTasksByState.getOrDefault(ExecutionTaskState.PENDING,
+            Collections.emptySet());
+        // Filter out proposals for non-existent topics to avoid admin timeouts on config changes for non-existent topics.
+        Set<String> existingTopics = _metadataClient.refreshMetadata().cluster().topics();
+        List<ExecutionProposal> proposalsForExistingTopics = pendingAtStart.stream()
+            .map(ExecutionTask::proposal)
+            .filter(p -> existingTopics.contains(p.topic()))
+            .collect(Collectors.toList());
+        long numExistingTopicsToThrottle = proposalsForExistingTopics.stream().map(ExecutionProposal::topic).distinct().count();
+        LOG.info("User task {}: Applying bulk replication throttle of {} B/s to {} inter-broker movements "
+                + "({} pending before filtering) across {} topics.",
+          _uuid, _replicationThrottle, proposalsForExistingTopics.size(), pendingAtStart.size(), numExistingTopicsToThrottle);
+        throttleHelper.setThrottles(proposalsForExistingTopics);
+      }
       int partitionsToMove = numTotalPartitionMovements;
       // Exhaust all the pending partition movements.
       while ((partitionsToMove > 0 || !inExecutionTasks().isEmpty()) && _stopSignal.get() == NO_STOP_EXECUTION) {
@@ -1622,7 +1642,9 @@ public class Executor {
 
         AlterPartitionReassignmentsResult result = null;
         if (!tasksToExecute.isEmpty()) {
-          throttleHelper.setThrottles(tasksToExecute.stream().map(ExecutionTask::proposal).collect(Collectors.toList()));
+          if (!bulkThrottleEnabled) {
+            throttleHelper.setThrottles(tasksToExecute.stream().map(ExecutionTask::proposal).collect(Collectors.toList()));
+          }
           // Execute the tasks.
           _executionTaskManager.markTasksInProgress(tasksToExecute);
           result = ExecutionUtils.submitReplicaReassignmentTasks(_adminClient, tasksToExecute);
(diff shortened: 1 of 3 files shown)


cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1606,53 +1606,118 @@ private void updateOngoingExecutionState() {
...
+                : interBrokerTasksByState.getOrDefault(ExecutionTaskState.DEAD, Collections.emptySet()).size();
+            int inProgressCount = interBrokerTasksByState == null ? 0
+                : interBrokerTasksByState.getOrDefault(ExecutionTaskState.IN_PROGRESS, Collections.emptySet()).size();
+            int abortingCount = interBrokerTasksByState == null ? 0
+                : interBrokerTasksByState.getOrDefault(ExecutionTaskState.ABORTING, Collections.emptySet()).size();
+            LOG.info("User task {}: Clearing bulk replication throttles (configured rate: {} B/s). "
+                    + "Completed: {}, Aborted: {}, Dead: {}, InProgress: {}, Aborting: {}.",
+                _uuid, _replicationThrottle, completedCount, abortedCount, deadCount, inProgressCount, abortingCount);
+            throttleHelper.clearThrottles(completedTasks, inProgressTasks);
+          } catch (ExecutionException | InterruptedException | TimeoutException e) {
```
kyguy: Should we need to re-interrupt the thread here when we catch an InterruptException?

[1] https://michaelscharf.blogspot.com/2006/09/dont-swallow-interruptedexception-call.html

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
```diff
@@ -1606,53 +1606,118 @@ private void updateOngoingExecutionState() {
...
+            }
+            int completedCount = interBrokerTasksByState == null ? 0
+                : interBrokerTasksByState.getOrDefault(ExecutionTaskState.COMPLETED, Collections.emptySet()).size();
+            int abortedCount = interBrokerTasksByState == null ? 0
+                : interBrokerTasksByState.getOrDefault(ExecutionTaskState.ABORTED, Collections.emptySet()).size();
+            int deadCount = interBrokerTasksByState == null ? 0
+                : interBrokerTasksByState.getOrDefault(ExecutionTaskState.DEAD, Collections.emptySet()).size();
+            int inProgressCount = interBrokerTasksByState == null ? 0
+                : interBrokerTasksByState.getOrDefault(ExecutionTaskState.IN_PROGRESS, Collections.emptySet()).size();
+            int abortingCount = interBrokerTasksByState == null ? 0
```
kyguy: Should we separate the `interBrokerTasksByState` null check into a signal conditional before these statements or isolate the logic in a separate method?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
il-kyun: I revisited the cleanup block as you suggested and caught the interrupted thread, making sure we re-interrupt before logging so the calling code still sees the interruption. https://github.com/linkedin/cruise-control/pull/2304/commits/422ec93922cb920c18de2d9a2d4cdafd5eee4650

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java
il-kyun: I also simplified the repeated null-check logic by factoring it into a helper, and later pulled the logging and list-building into small methods so the finally block stays readable. https://github.com/linkedin/cruise-control/pull/2304/commits/422ec93922cb920c18de2d9a2d4cdafd5eee4650

il-kyun:
Thanks for starting the review on my PR, I really appreciate it!

> Is there any specific reason why the bulk replication throttle mode shouldn't be enabled by default? It seems like it would be helpful for most rebalances.

I initially set `bulk.replication.throttle.enabled` to false to avoid surprising changes in existing deployments and to preserve backward compatibility.
I agree — I also think switching the default to true is better.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -422,6 +426,205 @@ public void testSetExecutionProgressCheckIntervalMsWithNoRequestedValue() {
...
+      ExecutionProposal proposal0 =
+          new ExecutionProposal(TP0, PRODUCE_SIZE_IN_BYTES, new ReplicaPlacementInfo(initialLeader0),
+              Collections.singletonList(new ReplicaPlacementInfo(initialLeader0)),
+              Collections.singletonList(new ReplicaPlacementInfo(initialLeader0 == 0 ? 1 : 0)));
+      ExecutionProposal proposal1 =
+          new ExecutionProposal(TP1, 0, new ReplicaPlacementInfo(initialLeader1),
+              Arrays.asList(new ReplicaPlacementInfo(initialLeader1), new ReplicaPlacementInfo(initialLeader1 == 0 ? 1 : 0)),
+              Arrays.asList(new ReplicaPlacementInfo(initialLeader1 == 0 ? 1 : 0), new ReplicaPlacementInfo(initialLeader1)));
+
+      Collection<ExecutionProposal> proposalsToExecute = Arrays.asList(proposal0, proposal1);
```
kyguy: Could this logic be separated into a separate method and used in this test method and reused in `testInterBrokerMoveReplicasWhenBulkThrottleEnabledThenThrottleClearedAtEnd()`

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -422,6 +426,205 @@ public void testSetExecutionProgressCheckIntervalMsWithNoRequestedValue() {
...
+
+      MetricRegistry metricRegistry = new MetricRegistry();
+      Executor executor = new Executor(configs, Time.SYSTEM, metricRegistry, mockAnomalyDetectorManager);
+      executor.setUserTaskManager(mockUserTaskManager);
+
+      // Start execution with replication throttle
+      final long replicationThrottle = PRODUCE_SIZE_IN_BYTES;
+      executor.setGeneratingProposalsForExecution(RANDOM_UUID, ExecutorTest.class::getSimpleName, true);
+      executor.executeProposals(proposalsToExecute, Collections.emptySet(), null, mockLoadMonitor, null, null, null, null, null,
+          null, null, replicationThrottle, true, RANDOM_UUID, false, false);
```
kyguy: Would it be possible to do this same with this logic? Separate it into another method and then call it in both tests?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
il-kyun: I’ve just done that. https://github.com/linkedin/cruise-control/pull/2304/commits/7c9bfa880b453b6ad9eeec1487844b3e6d31174f
The common logic is now extracted into a helper method and reused in both this test.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
il-kyun: I’ve just done that. https://github.com/linkedin/cruise-control/pull/2304/commits/7c9bfa880b453b6ad9eeec1487844b3e6d31174f
The common logic is now extracted into a helper method and reused in both this test.

kyguy: Looks great, thanks!

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
kyguy: Looks great, thanks!

il-kyun:
> Thanks for the updates! Just left some minor comments.
> 
> Similar to the related PR here: #2305 I wonder if it would be better to simply update the existing non-batching logic to this batching implementation instead of having it be configurable to save us the code complexity. I can't think of a reason why users would not want to batch requests like this. Anyways, I'll defer the maintainers on that!

https://github.com/linkedin/cruise-control/pull/2305#discussion_r2426413296

Removes the flag and makes the bulk path the default behavior to simplify the codebase and reduce configuration complexity.
@kyguy Thank you.

il-kyun:
This is an improvement based on [the issue](https://github.com/linkedin/cruise-control/issues/1972) @CCisGG created. I’d like to hear your thoughts on it when you have some time. I know you’re busy, but please take a look when you get a chance.

CCisGG:
Hey @il-kyun sorry for the delay, and thanks for working on this long-lasting pain point. It looks like after your PR, CC triggers throttling before the whole rebalance starts, and remove throttling after the rebalance completely rebalance. My only concern is what if the rebalance takes very long (10+ hours), and we could potentially make unnecessary throttling before a broker starts to moving replicas. There was a previous contributor making a PR that seems to better address by this  on only set the throttling to brokers right before the task execution: https://github.com/linkedin/cruise-control/pull/2214/files

Please let me know what do you think

PR #2305, opened 2025-08-18
ReplicationThrottleHelper: Add support for AdminClient bulk operations

### LINKED ISSUE #1972: Redundant set/unset throttling during the rebalance

il-kyun:
### Summary
1. Why: Per-entity AdminClient calls (`incrementalAlterConfigs`/`describeConfigs`) scale poorly on large clusters and are fragile when topics are deleted mid-run.
2. What: Replaced the legacy per-entity configuration path with a new bulk-based implementation as the default behavior.
The new design applies and verifies broker and topic throttles in batches by default

### Expected Behavior
- Same throttling semantics as before with significantly fewer AdminClient round-trips and lower latency on large operations.
- Bulk verification reduces describe calls by grouping resources.
- Non-existent topics detected and skipped without failing the whole operation.
- Safe no-op on empty inputs.
- Continues to respect wildcard `*` and static broker configs (skips removal of static values).

### Actual Behavior (Before Change)
- Previous per-entity path triggers many calls per broker/topic, causing slowdowns and timeouts in large clusters.
- Config verification performed per resource, increasing cost.

### Steps to Reproduce
1. In a cluster with 100+ brokers and many topics/partitions, generate large inter-broker replica movements.
2. Compare old per-entity implementation (pre-change) vs. new default bulk implementation.
3. Record total execution time, AdminClient call counts, and any timeouts/failures.

### Additional evidence
1. Environment: Kafka version, cluster size (brokers/topics/partitions), Cruise Control version.
2. Logs: Presence of “Removing leader/follower throttle rate …”, and bulk verification messages.
3. Metrics: Before/after comparisons of call counts, execution time.

### Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [x] refactor
- [ ] security/CVE
- [ ] other

This PR resolves #1972

### DIFF
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
@@ -434,22 +434,6 @@ class ReplicationThrottleHelper {
   }
 
   // Retries until we can read the configs changes we just wrote
-  void waitForConfigs(ConfigResource cf, Collection<AlterConfigOp> ops) {
-    // Use HashMap::new instead of Collectors.toMap to allow inserting null values
-    Map<String, String> expectedConfigs = ops.stream()
-            .collect(HashMap::new, (m, o) -> m.put(o.configEntry().name(), o.configEntry().value()), HashMap::putAll);
-    boolean retryResponse = CruiseControlMetricsUtils.retry(() -> {
-      try {
-        return !configsEqual(getEntityConfigs(cf), expectedConfigs);
-      } catch (ExecutionException | InterruptedException | TimeoutException e) {
-        return false;
-      }
-    }, _retries);
-    if (!retryResponse) {
-      throw new IllegalStateException("The following configs " + ops + " were not applied to " + cf + " within the time limit");
-    }
-  }
-
   void waitForConfigs(Map<ConfigResource, Collection<AlterConfigOp>> opsByResource) {
     Map<ConfigResource, Map<String, String>> expectedByResource = new HashMap<>();
     for (Map.Entry<ConfigResource, Collection<AlterConfigOp>> entry : opsByResource.entrySet()) {
--- a/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
+++ b/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java
@@ -521,15 +521,15 @@ public class ReplicationThrottleHelperTest extends CCKafkaIntegrationTestHarness
     EasyMock.replay(mockAdminClient);
     ReplicationThrottleHelper throttleHelper = new ReplicationThrottleHelper(mockAdminClient, 100L, retries);
     ConfigResource cf = new ConfigResource(ConfigResource.Type.TOPIC, TOPIC0);
-    assertThrows(IllegalStateException.class, () -> throttleHelper.waitForConfigs(cf, Collections.singletonList(
-            new AlterConfigOp(new ConfigEntry("k", "v"), AlterConfigOp.OpType.SET)
-    )));
+    Map<ConfigResource, Collection<AlterConfigOp>> opsByResource = Collections.singletonMap(cf,
+            Collections.singletonList(new AlterConfigOp(new ConfigEntry("k", "v"), AlterConfigOp.OpType.SET)));
+    assertThrows(IllegalStateException.class, () -> throttleHelper.waitForConfigs(opsByResource));
 
     // Case 2: queue a single result and call checkConfigs with matching configs, so it succeeds
     EasyMock.reset(mockAdminClient);
     expectDescribeTopicConfigs(mockAdminClient, TOPIC0, EMPTY_CONFIG, true);
     EasyMock.replay(mockAdminClient);
-    throttleHelper.waitForConfigs(cf, Collections.emptyList());
+    throttleHelper.waitForConfigs(Collections.singletonMap(cf, Collections.emptyList()));
   }
 
   @Test


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -855,6 +855,7 @@ private Properties getExecutorProperties() {
     props.setProperty(ExecutorConfig.REMOVAL_HISTORY_RETENTION_TIME_MS_CONFIG, Long.toString(REMOVAL_HISTORY_RETENTION_TIME_MS));
     props.setProperty(ExecutorConfig.LIST_PARTITION_REASSIGNMENTS_TIMEOUT_MS_CONFIG, Long.toString(LIST_PARTITION_REASSIGNMENTS_TIMEOUT_MS));
     props.setProperty(ExecutorConfig.LIST_PARTITION_REASSIGNMENTS_MAX_ATTEMPTS_CONFIG, Long.toString(LIST_PARTITION_REASSIGNMENTS_MAX_ATTEMPTS));
+    props.setProperty(ExecutorConfig.BULK_REPLICATION_THROTTLE_BULK_OPS_ENABLED_CONFIG, "true");
```
il-kyun: The test “test with JDK=microsoft:11” failed, but it passed after applying BULK_REPLICATION_THROTTLE_BULK_OPS_ENABLED_CONFIG.

https://github.com/linkedin/cruise-control/actions/runs/17787726830/job/50558296433
```
com.linkedin.kafka.cruisecontrol.executor.ExecutorTest > testReplicaReassignmentProgressWithThrottle FAILED
    java.lang.AssertionError: Proposal execution did not finish within the time limit
        at org.junit.Assert.fail(Assert.java:89)
        at com.linkedin.kafka.cruisecontrol.KafkaCruiseControlUnitTestUtils.waitUntilTrue(KafkaCruiseControlUnitTestUtils.java:115)
        at com.linkedin.kafka.cruisecontrol.executor.ExecutorTest.executeAndVerifyProposals(ExecutorTest.java:802)
        at com.linkedin.kafka.cruisecontrol.executor.ExecutorTest.testReplicaReassignmentProgressWithThrottle(ExecutorTest.java:132)
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -357,6 +455,150 @@ private void removeThrottledRateFromBroker(Integer brokerId)
...
+            replicas);
+          if (newThrottledReplicas.isEmpty()) {
+            ops.add(new AlterConfigOp(new ConfigEntry(FOLLOWER_THROTTLED_REPLICAS, null),
+              AlterConfigOp.OpType.DELETE));
+          } else {
+            ops.add(new AlterConfigOp(new ConfigEntry(FOLLOWER_THROTTLED_REPLICAS, newThrottledReplicas),
+              AlterConfigOp.OpType.SET));
+          }
+        }
+      }
```
kyguy: Would it be neater if we grouped the common code here into a separate method and call it here for the leader replicas and then the follower replicas?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -357,6 +455,150 @@ private void removeThrottledRateFromBroker(Integer brokerId)
...
+      if (currFollowerThrottle != null) {
+        if (currFollowerThrottle.source().equals(ConfigEntry.ConfigSource.STATIC_BROKER_CONFIG)) {
+          LOG.debug("Skipping removal for static follower throttle rate: {}", currFollowerThrottle);
+        } else {
+          LOG.debug("Removing follower throttle rate: {} on broker {}", currFollowerThrottle, brokerId);
+          ops.add(new AlterConfigOp(new ConfigEntry(FOLLOWER_THROTTLED_RATE, null), AlterConfigOp.OpType.DELETE));
+        }
+      }
```
kyguy, line 380: Would it be neater to group the common code here into a separate method as well?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
il-kyun: I’ve just done that. https://github.com/linkedin/cruise-control/pull/2305/commits/99c4c0c32a5b72083fcf1f55a6ac8a64e5a568b4
The common logic is now extracted into a helper method and reused it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
il-kyun: I’ve just done that. https://github.com/linkedin/cruise-control/pull/2305/commits/99c4c0c32a5b72083fcf1f55a6ac8a64e5a568b4
The common logic is now extracted into a helper method and reused it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
kyguy: Looks great, thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
kyguy: Looks great, thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
kyguy: Looks great, thanks!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
```diff
@@ -610,6 +610,15 @@ public final class ExecutorConfig {
   public static final String AUTO_STOP_EXTERNAL_AGENT_DOC = "When starting a new proposal execution while external agent is reassigning partitions,"
       + " automatically stop the external agent and start the execution."
       + " Set to false to keep the external agent reassignment and skip starting the execution.";
+
+  /**
+   * <code>bulk.replication.throttle.bulk.ops.enabled</code>
+   */
+  public static final String BULK_REPLICATION_THROTTLE_BULK_OPS_ENABLED_CONFIG = "bulk.replication.throttle.bulk.ops.enabled";
```
mimaison: In which scenarios would users not want to bulk alter/describe configs operations? Are there any drawbacks of enabling this by default?

il-kyun: @mimaison Thanks for bringing that up!
We actually discussed this earlier in [here](https://github.com/linkedin/cruise-control/pull/2304#issuecomment-3311531334) — at the time, we kept _useBulkOps configurable mainly out of caution, since we hadn’t yet fully validated that replacing the per-entity logic wouldn’t introduce unexpected side effects.
Removes the _useBulkOps flag and makes the bulk path the default behavior to simplify the codebase and reduce configuration complexity.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
```diff
@@ -497,6 +497,7 @@ public final class ExecutorConfig {
   public static final String AUTO_STOP_EXTERNAL_AGENT_DOC = "When starting a new proposal execution while external agent is reassigning partitions,"
       + " automatically stop the external agent and start the execution."
       + " Set to false to keep the external agent reassignment and skip starting the execution.";
+
```
mimaison: Let's undo this change since we're not touching this file anymore

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -9,8 +9,11 @@
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AlterConfigOp;
 import org.apache.kafka.clients.admin.Config;
+import org.apache.kafka.clients.admin.AlterConfigsResult;
 import org.apache.kafka.clients.admin.ConfigEntry;
 import org.apache.kafka.common.config.ConfigResource;
+import org.apache.kafka.common.KafkaFuture;
+import org.apache.kafka.server.config.QuotaConfigs;
```
mimaison: This is not part of the Kafka public API, we should avoid using it.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
mimaison, line 1: There's something wrong with this file, it does not compile

il-kyun: I rebased onto main and resolved the conflicts — it’s all good now.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/ExecutorConfig.java
il-kyun: I’ve just done that.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -154,8 +152,10 @@ private boolean throttlingEnabled() {
   private Set<Integer> getParticipatingBrokers(List<ExecutionProposal> replicaMovementProposals) {
     Set<Integer> participatingBrokers = new TreeSet<>();
     for (ExecutionProposal proposal : replicaMovementProposals) {
-      participatingBrokers.addAll(proposal.oldReplicas().stream().map(ReplicaPlacementInfo::brokerId).collect(Collectors.toSet()));
-      participatingBrokers.addAll(proposal.newReplicas().stream().map(ReplicaPlacementInfo::brokerId).collect(Collectors.toSet()));
+      participatingBrokers.addAll(
+          proposal.oldReplicas().stream().map(ReplicaPlacementInfo::brokerId).collect(Collectors.toSet()));
+      participatingBrokers.addAll(
+          proposal.newReplicas().stream().map(ReplicaPlacementInfo::brokerId).collect(Collectors.toSet()));
```
kyguy: Are these changes necessary? It looks like this is just changing the formatting

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -280,83 +287,160 @@ static String removeReplicasFromConfig(String throttleConfig, Set<String> replic
...
 
+  private Map<ConfigResource, Config> getBrokerConfigs(Set<Integer> brokerIds)
+    throws ExecutionException, InterruptedException, TimeoutException {
+    if (brokerIds == null || brokerIds.isEmpty()) {
+      return Collections.emptyMap();
+    }
+    List<ConfigResource> resources = brokerIds.stream()
+      .map(id -> new ConfigResource(ConfigResource.Type.BROKER, String.valueOf(id)))
+      .collect(Collectors.toList());
+    return _adminClient.describeConfigs(resources).all().get(CLIENT_REQUEST_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
kyguy, line 433: For the cases where we are dealing with large Kafka clusters, would it make sense to batch the requests here? From what I understand it would help avoid memory and concurrency issues with the admin client. Alternatively, maybe it would make more sense to batch the set of brokerIds in the calling methods to reduce the memory impact of storing the broker configs as well. What do you think?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -38,8 +40,14 @@ class ReplicationThrottleHelper {
   static final String WILDCARD_ASTERISK = "*";
   static final String LEADER_REPLICATION_THROTTLED_RATE_CONFIG = "leader.replication.throttled.rate";
   static final String FOLLOWER_REPLICATION_THROTTLED_RATE_CONFIG = "follower.replication.throttled.rate";
+  private static final List<String> REPLICATION_THROTTLED_RATE_CONFIGS = Arrays.asList(
+      LEADER_REPLICATION_THROTTLED_RATE_CONFIG,
+      FOLLOWER_REPLICATION_THROTTLED_RATE_CONFIG);
```
kyguy: Is this necessary? It appears we only the value use it in one place

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -38,8 +40,14 @@ class ReplicationThrottleHelper {
...
   static final String LEADER_REPLICATION_THROTTLED_REPLICAS_CONFIG = "leader.replication.throttled.replicas";
   static final String FOLLOWER_REPLICATION_THROTTLED_REPLICAS_CONFIG = "follower.replication.throttled.replicas";
+  private static final List<String> REPLICATION_THROTTLED_REPLICAS_CONFIGS = Arrays.asList(
+      LEADER_REPLICATION_THROTTLED_REPLICAS_CONFIG,
+      FOLLOWER_REPLICATION_THROTTLED_REPLICAS_CONFIG);
```
kyguy: Is this necessary? It appears we only the value use it in one place

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -374,6 +458,33 @@ void waitForConfigs(ConfigResource cf, Collection<AlterConfigOp> ops) {
     }
   }
 
+  void waitForConfigs(Map<ConfigResource, Collection<AlterConfigOp>> opsByResource) {
```
kyguy, line 437: Is there any reason we are keeping the original `waitForConfigs()` method? Can we completely replace it with this implementation?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -280,83 +287,160 @@ static String removeReplicasFromConfig(String throttleConfig, Set<String> replic
...
+      if (!ops.isEmpty()) {
+        bulkOps.put(cf, ops);
+      }
     }
-    if (!ops.isEmpty()) {
-      changeTopicConfigs(topic, ops);
+    applyIncrementalAlterConfigsForBrokers(bulkOps);
+  }
+
+  private void applyIncrementalAlterConfigsForBrokers(Map<ConfigResource, Collection<AlterConfigOp>> bulkOps)
```
kyguy: This method is meant to replace `changeBrokerConfigs`, right? If so, would it make sense to keep the method name a `changeBrokerConfigs`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -280,83 +287,160 @@ static String removeReplicasFromConfig(String throttleConfig, Set<String> replic
...
-    ConfigEntry currFollowerThrottle = brokerConfigs.get(FOLLOWER_REPLICATION_THROTTLED_RATE_CONFIG);
-    List<AlterConfigOp> ops = new ArrayList<>();
-    if (currLeaderThrottle != null) {
-      if (currLeaderThrottle.source().equals(ConfigEntry.ConfigSource.STATIC_BROKER_CONFIG)) {
-        LOG.debug("Skipping removal for static leader throttle rate: {}", currFollowerThrottle);
-      } else {
-        LOG.debug("Removing leader throttle rate: {} on broker {}", currLeaderThrottle, brokerId);
-        ops.add(new AlterConfigOp(new ConfigEntry(LEADER_REPLICATION_THROTTLED_RATE_CONFIG, null), AlterConfigOp.OpType.DELETE));
-      }
+  private void applyIncrementalAlterConfigsForTopics(Map<ConfigResource, Collection<AlterConfigOp>> bulkOps)
```
kyguy: This method is meant to replace `changeTopicConfigs`, right? If so, would it make sense to keep the method name a `changeTopicConfigs`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
il-kyun: I've renamed it back to changeBrokerConfigs() to maintain consistency with the previous implementation and make the purpose clearer.  https://github.com/linkedin/cruise-control/pull/2305/commits/44251ab856acc69e219d0569cc1cf135153e3314

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
il-kyun: I've renamed it back to changeTopicConfigs() for consistency and clarity. https://github.com/linkedin/cruise-control/pull/2305/commits/44251ab856acc69e219d0569cc1cf135153e3314

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
il-kyun: I just changed it to use List.of() https://github.com/linkedin/cruise-control/pull/2305/commits/96b64714fed42d5f460dfa5b020278b72da2d749

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
il-kyun: I just changed it to use List.of() https://github.com/linkedin/cruise-control/pull/2305/commits/96b64714fed42d5f460dfa5b020278b72da2d749

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
il-kyun: I reverted it. https://github.com/linkedin/cruise-control/pull/2305/commits/44251ab856acc69e219d0569cc1cf135153e3314

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
il-kyun: You're right - we can remove the single-resource version. https://github.com/linkedin/cruise-control/pull/2305/commits/0fc26ab7b16b55a8a43b25b8081ca4e39387b853

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
il-kyun: Yes — splitting the broker set before calling describeConfigs makes sense to reduce payload size, heap pressure, and timeout risk in the AdminClient.

Recomputing with ~100 KB per broker:
	•	50 brokers ≈ ~5 MB per call
	•	40 brokers ≈ ~4 MB per call
	•	25 brokers ≈ ~2.5 MB per call

Given this, I think 25 is a reasonable default batch size — it keeps each request lightweight while maintaining good throughput. The value should remain configurable so it can be tuned based on cluster size and network performance.
Would 25 per batch as a default (configurable) work for you?

il-kyun:
@CCisGG 
Thanks for checking it so quickly.

I think I may have misunderstood earlier — this PR seems to be a more appropriate solution.
Please take a look when you have some time.
https://github.com/linkedin/cruise-control/pull/2304#issuecomment-3473331671

CCisGG:
> @CCisGG
> Thanks for checking it so quickly.
> 
> I think I may have misunderstood earlier — this PR seems to be a more appropriate solution.
> Please take a look when you have some time.
> https://github.com/linkedin/cruise-control/pull/2304#issuecomment-3473331671

Just to confirm, are you saying #2304 is not a good path to go, and instead we use this PR 2305 to address the issue?

il-kyun:
> Just to confirm, are you saying https://github.com/linkedin/cruise-control/pull/2304 is not a good path to go, and instead we use this PR 2305 to address the issue?

@CCisGG 
Not exactly. #2304 optimizes when we apply throttling by setting the replication throttle once before the inter-broker phase and clearing it once after, which avoids per-batch set/clear churn. #2305 optimizes a different layer—how we apply it—by batching the AdminClient operations in ReplicationThrottleHelper (apply/verify) instead of issuing one call per entity. So I don’t see them as substitutes; they’re complementary options that can be used together: #2304 reduces throttle churn, and #2305 reduces AdminClient round-trips. Also, for #2304, I’d keep it configuration-driven as in the initial proposal.
I’d appreciate your thoughts on this approach—does this align with your view?

PR #2307, opened 2025-09-17
Upgrade jetty to 12.0.39

JvD-Ericsson:
## Summary
1. Why: To remove CVEs:
	- CVE-2024-6763
	- CVE-2024-13009

2. What: Upgrade jetty to 12.0.12 to remove CVE-2024-13009 and CVE-2024-6763

- [x] security/CVE

Jetty 12 only works with Java 17+ so, this change will not work with the Java 11

PR #2308, opened 2025-09-19
Remove support for Java 11

JvD-Ericsson:
## Summary
1. Why: Jetty 12 requires a minimum of Java 17, so Java 11 must be dropped to facilitate this upgrade #2307. The current version of Jetty in cruise-control is 9.4.56.v20240826. This version contains two CVEs 
- CVE-2024-13009 - patched in 9.4.57.v20241219
- CVE-2024-6763 - patched in 12.0.12

So, Jetty 12 is needed to patch CVE-2024-6763.

2. What: Remove support support for Java 11. Raise minimum JDK to 17 in builds, docs, and workflows

## Categorization
- [x] refactor

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
       _JAVA_OPTIONS: "-Xms512m -Xmx1g"
     working_directory: ~/workspace
     docker:
-      - image: cimg/openjdk:11.0
+      - image: cimg/openjdk:17.0
     steps:
       - checkout
       - restore_cache:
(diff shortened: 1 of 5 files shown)


danielgospodinow:
@kyguy, can you review this PR once you have some bandwidth? Because it's kind of blocking #2244, which is a fantastic initiative IMO.

kyguy (approved):
In hindsight, I wish we had added a deprecation notice for Java 11 when we added support for Java 17 some months back. That being said, dropping Java 11 now is a reasonable step especially given that it's required to address certain CVEs (some listed in the PR description) and to support newer Kafka versions (Kafka ≥ 4.0).

As far as backward compatibility goes, Java 17 has been supported in Kafka since version 3.2.0 [1]. Although Cruise Control’s main branch currently only supports Kafka 3.7.0+, it’s still important to maintain compatibility with as many Kafka versions as possible to serve the broadest range of users in the community. So this change is in line with that goal as well.

I have tested this change and have been running builds with Java 17 for some time now, this looks like a reasonable update to me. We just need a maintainer to sign off and merge.

[1] https://kafka.apache.org/32/documentation.html#java

Issue #2282, opened 2025-07-14
Migrate off internal Kafka APIs to Improve Compatibility and Maintainability


kyguy:
# Problem:

Cruise Control currently depends on several internal (non-public) Kafka APIs. While this may have been a pragmatic choice in the past, it’s now a growing liability. Internal APIs are not stable, lack compatibility guarantees, and are increasingly diverging across Kafka versions. This results in:

- **Upgrade pain**: Upgrading Kafka versions becomes a time-consuming, error-prone process, often requiring patching and ad hoc fixes in Cruise Control. [[1]](https://github.com/linkedin/cruise-control/pull/2254) [[2]](https://github.com/linkedin/cruise-control/pull/2019) [[3]](https://github.com/linkedin/cruise-control/pull/1910)
- **Compatibility constraints**: Users are forced to stick to specific Kafka versions, limiting flexibility in modern deployments and increasing vulnerability to CVEs.
- **Technical debt**: Each Kafka release amplifies the maintenance burden, as internal APIs evolve or disappear entirely.

Kafka is releasing faster and with more aggressive changes. Every Kafka release makes our current approach harder to sustain. Migrating off internal APIs is an opportunity to:

- **Future-proof Cruise Control** facilitating upgrades to newer Kafka versions;  enabling access to feature improvements, compatibility changes, and important CVE fixes.
- **Expand Kafka version compatibility** enabling Cruise Control to be run safely with a wider range of Kafka versions.
- **Reduce maintenance overhead**, freeing contributors to focus on feature development rather than API breakage fixes and less `migrate_to_kafka` branches to support.

# Proposed Solutions:

I’ve audited the Cruise Control codebase and identified several usages of internal Kafka APIs. Below is a proposed plan for migrating to Kafka’s public APIs:

## Remove `KafkaTopicConfigProvider`

The `KafkaTopicConfigProvider` relies on internal Kafka APIs, has been deprecated since 2021, is incompatible with Kafka versions 3.0 and above, and is no longer needed. Its responsibilities are fully handled by the `KafkaAdminTopicConfigProvider` class. Given its deprecated status and limited compatibility, we can safely remove the `KafkaTopicConfigProvider` class from the codebase. We can additionally remove the place in the codebase that checks its configuration as well. 

## Deprecate and Remove `ZkBrokerFailureDetector`

[...]

PR #2311, opened 2025-09-30
Upgrade Kafka to 4.0.0

### LINKED ISSUE #2309: Bump Kafka version to 4.x
We should bump the Kafka version to at least 4.0.0, to be able to compile the project with a Kafka version that is at least 4.0.0.
Removing Zookeeper is a prerequisite for this, because Kafka dropped everything related to Zookeeper from 4.0.0
https://github.com/linkedin/cruise-control/issues/2243

pnagy-cldr:
## Summary
1. Why: Kafka version bump is done periodically in the project. Now it's time to increase it to 4.0.0.
2. What: Kafka related API calls change over time. When bumping Kafka version, these CC code should adapt to it. 

## Categorization
- [x] refactor

This PR resolves #2309 .


pnagy-cldr:
@kyguy @CCisGG As I promised, here is the kafka bump 4.0.0 PR based on latest main branch.
It is a much lighter/smaller change than removing ZK from the project. Mostly it just includes using latest Kafka admin API calls instead of deprecated ones that were removed in 4.0.0.
I also added minor changes according to ZK removal review in a separate commit, and we also answered all questions in ZK removal PR with the help of @egyedt 
Please review when you will have time!

egyedt (approved):
LGTM!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBrokerBuilder.java
```diff
@@ -200,15 +199,6 @@ public CCEmbeddedBrokerBuilder enableControlledShutdown(boolean enableControlled
     return this;
   }
 
-  /**
-   * @param controlledShutdownRetryBackoff controlled shutdown retry backoff.
-   * @return This
-   */
-  public CCEmbeddedBrokerBuilder controlledShutdownRetryBackoff(long controlledShutdownRetryBackoff) {
```
danielgospodinow, line 207: This was just not used anywhere?

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedKRaftController.java
```diff
@@ -88,13 +84,13 @@ public String clusterId() {
...
     private Properties createControllerProperties() {
         Properties props = new Properties();
-        props.setProperty(KRaftConfigs.PROCESS_ROLES_CONFIG, "controller");
-        props.setProperty(KRaftConfigs.NODE_ID_CONFIG, String.valueOf(ID));
-        props.setProperty(KRaftConfigs.CONTROLLER_LISTENER_NAMES_CONFIG, "CONTROLLER");
-        props.setProperty(SocketServerConfigs.LISTENERS_CONFIG, "CONTROLLER://:" + _port);
-        props.setProperty(QuorumConfig.QUORUM_VOTERS_CONFIG, quorumVoters());
-        props.setProperty(ServerLogConfigs.LOG_DIR_CONFIG, _logDir.getAbsolutePath());
-        props.setProperty(KRaftConfigs.METADATA_LOG_DIR_CONFIG, _logDir.getAbsolutePath());
+        props.setProperty("process.roles", "controller");
```
danielgospodinow: This is really smart to prevent dependencies on the Kafka libs.

Although I do think it'd be better to have them defined somewhere as constants.

E.g.,
``` java
    private static final String PROCESS_ROLE_PROP = "process.roles";
```

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -37,10 +37,10 @@
 class ReplicationThrottleHelper {
```
danielgospodinow, line 36: There's a ZK-related comment in this file's test. 
https://github.com/linkedin/cruise-control/blob/27c0bf4fcae9583a4bd1efad140ad47d44ba21a6/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelperTest.java#L118

Should we modify/remove it?

danielgospodinow (approved):
Thanks for the quick follow-up PR! 🚀 LGTM with some small comments.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBrokerBuilder.java
pnagy-cldr: It was used and set in this class, but from 4.0.0, Kafka removed the possibility to set this property:
https://issues.apache.org/jira/browse/KAFKA-18417
Because of this, I removed the variable that holds the backoff value, and I also removed the line where we set it:
    props.put(ServerConfigs.CONTROLLED_SHUTDOWN_RETRY_BACKOFF_MS_CONFIG, Long.toString(_controlledShutdownRetryBackoff));

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedKRaftController.java
pnagy-cldr: It is actually a good idea, thanks! At this point, I do not want to reinvent where to store them, but I created constants for them inside the same class, with the same name that they had in kafka internal classes. From this point, it is easy to move those constants anywhere else.

danielgospodinow: Yup, that's totally fine by me.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
pnagy-cldr: Yes, this comment seems a leftover for me, so I changed it, thanks!
It seems to me that in the following commit (~3years ago):
Update ReplicationThrottleHelper to use AdminClient (#1781)
The test was changed to use admin client instead of zk client, but the comment remained zk related.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
```diff
@@ -102,9 +103,10 @@ public CCKafkaRaftServer(KafkaConfig config, String clusterId, Time time) {
                 metaPropsEnsemble,
                 time,
                 _metrics,
-                CompletableFuture.completedFuture(QuorumConfig.parseVoterConnections(config.quorumVoters())),
+                CompletableFuture.completedFuture(QuorumConfig.parseVoterConnections(config.quorumConfig().voters())),
                 new ArrayList<>(),
-                new StandardFaultHandlerFactory()
+                new StandardFaultHandlerFactory(),
+                new ServerSocketFactory.KafkaServerSocketFactory()
```
kyguy, line 109: It looks like the method signature of `SharedServer` was changed in Kafka 4.0.0 to require an additional argument (hence why we need to add `new ServerSocketFactory.KafkaServerSocketFactory()` as an argument here). However, once we add this argument to the code, Cruise Control loses build compatibility with all Kafka versions < v4.0.0 which use the original method signature.

Could we refactor this code to maintain build compatibility with Kafka versions < 4.0.0?

We have leveraged Java reflection to accomplish this in the past [1] I know using Java reflection is not good practice but it may be a necessary evil until we finish the migration off of internal Kafka APIs. 

[1] https://github.com/linkedin/cruise-control/pull/2254

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
```diff
@@ -207,8 +209,7 @@ private static MetaTuple initializeLogDirs(KafkaConfig config, String logPrefix)
         MetaPropertiesEnsemble metaPropsEnsemble = copier.copy();
 
         // Load the BootstrapMetadata
-        BootstrapDirectory bootstrapDirectory = new BootstrapDirectory(
-                config.metadataLogDir(), Optional.ofNullable(config.interBrokerProtocolVersionString()));
+        BootstrapDirectory bootstrapDirectory = new BootstrapDirectory(config.metadataLogDir());
```
kyguy, line 212: Same ask as above, could we refactor this code to maintain build compatibility with Kafka versions < 4.0.0

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
```diff
@@ -37,10 +37,10 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(ReplicationThrottleHelper.class);
   static final String WILDCARD_ASTERISK = "*";
-  static final String LEADER_THROTTLED_RATE = QuotaConfigs.LEADER_REPLICATION_THROTTLED_RATE_CONFIG;
-  static final String FOLLOWER_THROTTLED_RATE = QuotaConfigs.FOLLOWER_REPLICATION_THROTTLED_RATE_CONFIG;
-  static final String LEADER_THROTTLED_REPLICAS = QuotaConfigs.LEADER_REPLICATION_THROTTLED_REPLICAS_CONFIG;
-  static final String FOLLOWER_THROTTLED_REPLICAS = QuotaConfigs.FOLLOWER_REPLICATION_THROTTLED_REPLICAS_CONFIG;
+  static final String LEADER_THROTTLED_RATE = QuotaConfig.LEADER_REPLICATION_THROTTLED_RATE_CONFIG;
+  static final String FOLLOWER_THROTTLED_RATE = QuotaConfig.FOLLOWER_REPLICATION_THROTTLED_RATE_CONFIG;
+  static final String LEADER_THROTTLED_REPLICAS = QuotaConfig.LEADER_REPLICATION_THROTTLED_REPLICAS_CONFIG;
+  static final String FOLLOWER_THROTTLED_REPLICAS = QuotaConfig.FOLLOWER_REPLICATION_THROTTLED_REPLICAS_CONFIG;
```
kyguy: Since the `org.apache.kafka.server.config` package is not part of the public Kafka API these class names may change again in the near future. I would hard code these values as well instead of relying on `org.apache.kafka.server.config` package. The literal configurations strings should be more stable.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
pnagy-cldr: Thanks for the idea, I have checked the PR. This seems very hacky for me. I totally agree that internal kafka API dependencies should be avoided in the project, that is why there is your issue https://github.com/linkedin/cruise-control/issues/2282
But what is the benefit in being able to build this main line with kafka 3.7.0? We only talk about build/compile time kafka 4.0.0, the CC will be able to talk with ANY older Kafka cluster runtime.
This is just a test class, so in my opinion this is not worth the effort, rather your testcontainers suggested solution should be implemented.
WDYT?

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
pnagy-cldr: Same as above, I am waiting for your opinion.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
pnagy-cldr: Done, thanks for the idea! I have renamed the variables to the original one to be able to find them easier in the future if we need to check them in internal Kafka classes.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
danielgospodinow: > Cruise Control loses build compatibility with all Kafka versions < v4.0.0 ... Could we refactor this code to maintain build compatibility with Kafka versions < 4.0.0?

Hm, but what's the goal here? Even if we build for Kafka >= 4.0.0 only, that's fine, IMO, because I don't see a use case where in the future we decide to depend on an older Kafka lib - we're only going to move forward with this.

> CC will be able to talk with ANY older Kafka cluster runtime

I agree with @pnagy-cldr, I also think that this is the case.

kyguy: > But what is the benefit in being able to build this main line with kafka 3.7.0? We only talk about build/compile time kafka 4.0.0, the CC will be able to talk with ANY older Kafka cluster runtime.

From what I understand, it was a strict requirement coming from LinkedIn (the maintainers) in the past. I think they build on an older Kafka version internally and wanted to keep build support for those older Kafka versions. Hopefully, it is not a requirement anymore and we can move forward with this PR without the need of any hacky Java reflection. I am also keen on getting and keeping Cruise Control aligned with the latest Kafka versions!

Anyways, as you noted, this won't be an issue anyway once we finish the API migration https://github.com/linkedin/cruise-control/issues/2282, I'll keep driving on that end!

kyguy (approved):
Looks great @pnagy-cldr, I think the points made about avoiding the extra code complexity required to keep build compatibility with older Kafka versions are fair. Hopefully the maintainers are on board so we can move this PR forward!

egyedt:
@pnagy-cldr thanks for the PR. 
@CCisGG could you please take a look, since this PR seems to be approved from several community members.
Thanks in advance!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
CCisGG: > From what I understand, it was a strict requirement coming from LinkedIn (the maintainers) in the past. I think they build on an older Kafka version internally and wanted to keep build support for those older Kafka versions. Hopefully, it is not a requirement anymore and we can move forward with this PR without the need of any hacky Java reflection. I am also keen on getting and keeping Cruise Control aligned with the latest Kafka versions!

Thanks Kyle for provide the context here. Yeah we (LinkedIn) are not depending on the main branch anymore. So I'm good with it.

CCisGG (approved):

PR #2303, opened 2025-08-11
Migrate CruiseControlMetricsReporterTest to TestContainers

kyguy:
## Summary

1. Why: As detailed in issue #2282, the existing Cruise Control tests rely on the Kafka server wrapper class, `CCEmbeddedBroker`, which uses several non-public Kafka APIs. These internal APIs lack stability and compatibility guarantees, making it increasingly difficult to upgrade Kafka versions. As Kafka evolves, this reliance introduces upgrade friction, version lock-in, and a growing maintenance burden.

2. What: This PR starts the migration of the Cruise Control tests away from internal Kafka APIs by refactoring the `CruiseControlMetricsReporterTest` to replace its dependency on the `CCEmbeddedBroker` class with the [TestContainers]( https://java.testcontainers.org/modules/kafka/) Kafka module instead. This helps future-proof Cruise Control by easing upgrades to newer Kafka versions, expanding Kafka versions compatibility, and reducing the ongoing maintenance overhead. 

If an approach looks good, passes review, and is merged, we can continue migrating the remaining test classes incrementally, one PR at a time, until we completely remove the dependency on private Kafka APIs in the Cruise Control tests.

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [x] refactor
- [ ] security/CVE
- [ ] other

This PR partially resolves #2282

PR #2313, opened 2025-10-06
Migrate AutoCreateTopicTest to TestContainers

kyguy:
## Summary
1. Why: This PR is the next iteration of the TestContainers migration following the initial migration [PR](https://github.com/linkedin/cruise-control/pull/2303) in effort to remove the Cruise Control tests dependence on internal Kafka APIs. As detailed in issue Migrate off internal Kafka APIs to Improve Compatibility and Maintainability #2282, the existing Cruise Control tests rely on the Kafka server wrapper class, `CCEmbeddedBroker`, which uses several non-public Kafka APIs. These internal APIs lack stability and compatibility guarantees, making it increasingly difficult to upgrade Kafka versions. As Kafka evolves, this reliance introduces upgrade friction, version lock-in, and a growing maintenance burden.

2. What: This PR is the next step of the TestContainers migration following this [PR](https://github.com/linkedin/cruise-control/pull/2303). It refactors the `CruiseControlMetricsReporterAutoCreateTopicTest` to replace its dependency on the `CCEmbeddedBroker` class with the [TestContainers](https://java.testcontainers.org/modules/kafka/) Kafka module instead. This helps future-proof Cruise Control by easing upgrades to newer Kafka versions, expanding Kafka versions compatibility, and reducing the ongoing maintenance overhead.

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [x] refactor
- [ ] security/CVE
- [ ] other

This PR partially resolves [#2282](https://github.com/linkedin/cruise-control/issues/2282).

### DIFF
--- a/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
+++ b/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
@@ -7,6 +7,7 @@ package com.linkedin.kafka.cruisecontrol.metricsreporter;
 import com.linkedin.kafka.cruisecontrol.metricsreporter.exception.KafkaTopicDescriptionException;
 import com.linkedin.kafka.cruisecontrol.metricsreporter.utils.CCContainerizedKraftCluster;
 import com.linkedin.kafka.cruisecontrol.metricsreporter.utils.CCKafkaClientsIntegrationTestHarness;
+import com.linkedin.kafka.cruisecontrol.metricsreporter.utils.KafkaServerConfigs;
 import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.CreateTopicsResult;
@@ -27,8 +28,6 @@ import static com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetr
 import static org.junit.Assert.assertEquals;
 
 public class CruiseControlMetricsReporterAutoCreateTopicTest extends CCKafkaClientsIntegrationTestHarness {
-    private static final int NUM_OF_BROKERS = 2;
-    private static final String HOST = "127.0.0.1";
     protected static final String TOPIC = "CruiseControlMetricsReporterTest";
     protected static final String TEST_TOPIC = "TestTopic";
     private CCContainerizedKraftCluster _cluster;
@@ -41,7 +40,7 @@ public class CruiseControlMetricsReporterAutoCreateTopicTest extends CCKafkaClie
         Properties adminClientProps = new Properties();
         setSecurityConfigs(adminClientProps, "admin");
 
-        _cluster = new CCContainerizedKraftCluster(NUM_OF_BROKERS, buildBrokerConfigs(), adminClientProps);
+        _cluster = new CCContainerizedKraftCluster(2, buildBrokerConfigs(), adminClientProps);
         _cluster.start();
         _bootstrapUrl = _cluster.getExternalBootstrapAddress();
 
@@ -92,7 +91,7 @@ public class CruiseControlMetricsReporterAutoCreateTopicTest extends CCKafkaClie
         Properties props = new Properties();
         props.setProperty(CommonClientConfigs.METRIC_REPORTER_CLASSES_CONFIG, CruiseControlMetricsReporter.class.getName());
         props.setProperty(CruiseControlMetricsReporterConfig.config(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG),
-          HOST + ":" + CCContainerizedKraftCluster.CONTAINER_INTERNAL_LISTENER_PORT);
+          "127.0.0.1:" + CCContainerizedKraftCluster.CONTAINER_INTERNAL_LISTENER_PORT);
         props.put("listener.security.protocol.map", String.join(",",
           CCContainerizedKraftCluster.CONTROLLER_LISTENER_NAME + ":PLAINTEXT",
           CCContainerizedKraftCluster.INTERNAL_LISTENER_NAME + ":PLAINTEXT",
@@ -106,10 +105,10 @@ public class CruiseControlMetricsReporterAutoCreateTopicTest extends CCKafkaClie
         props.setProperty(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_NUM_PARTITIONS_CONFIG, "1");
         props.setProperty(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR_CONFIG, "1");
         // disable topic auto-creation to leave the metrics reporter to create the metrics topic
-        props.setProperty("auto.create.topics.enable", "false");
-        props.setProperty("offsets.topic.replication.factor", "1");
-        props.setProperty("default.replication.factor", "2");
-        props.setProperty("num.partitions", "2");
+        props.setProperty(KafkaServerConfigs.AUTO_CREATE_TOPICS_ENABLE_CONFIG, "false");
+        props.setProperty(KafkaServerConfigs.OFFSETS_TOPIC_REPLICATION_FACTOR_CONFIG, "1");
+        props.setProperty(KafkaServerConfigs.DEFAULT_REPLICATION_FACTOR_CONFIG, "2");
+        props.setProperty(KafkaServerConfigs.NUM_PARTITIONS_CONFIG, "2");
         return props;
     }
 
--- a/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/KafkaServerConfigs.java
+++ b/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/KafkaServerConfigs.java
@@ -1,0 +1,18 @@
+/*
+ * Copyright 2020 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License"). See License in the project root for license information.
+ */
+package com.linkedin.kafka.cruisecontrol.metricsreporter.utils;
+
+/**
+ * Defines Kafka broker configuration keys to avoid relying on internal Kafka classes.
+ * This reduces coupling to Kafka's internal APIs and improves long-term maintainability.
+ */
+public final class KafkaServerConfigs {
+    public static final String AUTO_CREATE_TOPICS_ENABLE_CONFIG = "auto.create.topics.enable";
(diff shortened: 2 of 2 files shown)


cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
```diff
@@ -95,11 +100,11 @@ public Properties overridingProps() {
         props.setProperty(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_NUM_PARTITIONS_CONFIG, "1");
         props.setProperty(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR_CONFIG, "1");
         // disable topic auto-creation to leave the metrics reporter to create the metrics topic
-        props.setProperty(ServerLogConfigs.AUTO_CREATE_TOPICS_ENABLE_CONFIG, "false");
-        props.setProperty(ServerLogConfigs.LOG_FLUSH_INTERVAL_MESSAGES_CONFIG, "1");
-        props.setProperty(GroupCoordinatorConfig.OFFSETS_TOPIC_REPLICATION_FACTOR_CONFIG, "1");
-        props.setProperty(ReplicationConfigs.DEFAULT_REPLICATION_FACTOR_CONFIG, "2");
-        props.setProperty(ServerLogConfigs.NUM_PARTITIONS_CONFIG, "2");
+        props.setProperty("auto.create.topics.enable", "false");
```
tomncooper: We shouldn't be hard coding configs, AFAIK these are available programatically via the relevant client config class (eg `org.apache.kafka.clients.consumer.ConsumerConfig`).

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
```diff
@@ -31,6 +27,8 @@
 import static org.junit.Assert.assertEquals;
 
 public class CruiseControlMetricsReporterAutoCreateTopicTest extends CCKafkaClientsIntegrationTestHarness {
+    private static final int NUM_OF_BROKERS = 2;
+    private static final String HOST = "127.0.0.1";
```
tomncooper: I would typically use `localhost` instead of the ip4 address as it is more generic. Unless you are getting weird network behaviour with test containers?

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
```diff
@@ -95,11 +100,11 @@ public Properties overridingProps() {
         props.setProperty(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_NUM_PARTITIONS_CONFIG, "1");
         props.setProperty(CruiseControlMetricsReporterConfig.CRUISE_CONTROL_METRICS_TOPIC_REPLICATION_FACTOR_CONFIG, "1");
         // disable topic auto-creation to leave the metrics reporter to create the metrics topic
-        props.setProperty(ServerLogConfigs.AUTO_CREATE_TOPICS_ENABLE_CONFIG, "false");
-        props.setProperty(ServerLogConfigs.LOG_FLUSH_INTERVAL_MESSAGES_CONFIG, "1");
-        props.setProperty(GroupCoordinatorConfig.OFFSETS_TOPIC_REPLICATION_FACTOR_CONFIG, "1");
-        props.setProperty(ReplicationConfigs.DEFAULT_REPLICATION_FACTOR_CONFIG, "2");
-        props.setProperty(ServerLogConfigs.NUM_PARTITIONS_CONFIG, "2");
+        props.setProperty("auto.create.topics.enable", "false");
+        props.setProperty("flush.messages", "1");
```
fvaleri: Why is this needed? This is usually not recommended with Kafka.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
fvaleri: Also, do we really need these two vars given that they are used in just one place?

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
kyguy: I accidentally translated this to the incorrect property, it is supposed to be `log.flush.interval.messages` as it is in the `CruiseControlMetricsReporterTest` and `CruiseControlMetricsReporterSslTest` classes! 

That being said, I am not sure how necessary its inclusion is, the tests seem to run fine without it. I left the property in since it was already part of these tests. This change here is just for removing dependence on the internal Kafka API, `org.apache.kafka.coordinator.group.GroupCoordinatorConfig`. 

If you are confident that the property is unnecessary, I will remove it (The tests pass without it anyway)

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
kyguy: I understand that it is bad practice but these server configs are not available via Kafka's public API. Luckily, the hard coded configs should be relatively stable, more stable than the server config package/variable names!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
kyguy: We could change this to `localhost` (it would work fine with TestContainers networking) and/or remove these variables but to be consistent with the other related test classes `CruiseControlMetricsReporterTest` and `CruiseControlMetricsReporterSslTest` I was thinking we could keep them for now. I plan on moving these variables to the super class anyway once I finish migrating all of the subclasses to TestContainers, maybe we could remove these variables from this class and change the HOST value to `localhost` then? 

It doesn't really matter to me either way, if either of you feel strongly about either change I am happy to make them. Let me know!

fvaleri: What's the benefit of moving to a super class? I usually prefer to have them in the same test class, as it makes it easier to reason about the setup.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
fvaleri: Yeah, I don't see how this can be related to this test, so let's remove it.

kyguy: Just removed!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
tomncooper: Fair point, I didn't realise the [ServerLogConfigs](https://github.com/apache/kafka/blob/4.1.0/server-common/src/main/java/org/apache/kafka/server/config/ServerLogConfigs.java) were an internal API. 

If these are going to be used in many places we should put them in a common class.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
kyguy: > What's the benefit of moving to a super class? I usually prefer to have them in the same test class, as it makes it easier to reason about the setup.

To reduce code duplication and highlight the configuration differences between the test subclasses.

fvaleri: I agree in principle and this helps with shared helper methods, but in practice a little duplication is worth the clarity and independence it provides. Having them in a super class does not allow to have the test requirements clear at a glance, you have to inspect another class and this just adds to the cognitive load.

kyguy: Maybe the benefit (or lack thereof) of moving these and related fields to the superclass will be more clear in the future PR where we plan on doing the refactoring. At that point we will have a diff where we can directly compare the changes between the classes. We can revisit the debate there!

As for this PR, it doesn't matter if we keep or the remove these fields, I'll remove them.

fvaleri (approved):
LGTM. Thanks for addressing my comments.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
kyguy: > If these are going to be used in many places we should put them in a common class.

Fair point, just added a utility class to hold the Kafka server configuration key constants, let me know what you think!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterAutoCreateTopicTest.java
kyguy: Just removed!

PR #2338, opened 2025-12-02
Ensure Executor tests use up-to-date cluster metadata

kyguy:
## Summary
1. Why: The clients used to fetch metadata in the Executor tests (the admin client of the `CCContainerizedKraftCluster` class and the `MetadataClient` of the `Executor` class) can fetch stale cluster information while the controller propagates metadata updates to the brokers. This can cause flakiness in tests that depend on that metadata.

4. What: This PR updates the Executor tests to have the clients refresh metadata more frequently and removes the Executor's dependence on the `MetadataClient` and private Kafka APIs to avoid testing against stale metadata.

## Categorization
- [ ] documentation
- [x] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [ ] other

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -339,7 +339,7 @@ project(':cruise-control') {
     testImplementation 'com.jayway.jsonpath:json-path:2.7.0'
     testImplementation 'org.powermock:powermock-module-junit4:2.0.9'
     testImplementation 'org.powermock:powermock-api-easymock:2.0.9'
-    testImplementation "org.testcontainers:kafka:$testcontainersVersion"
+    testImplementation "org.testcontainers:testcontainers-kafka:$testcontainersVersion"
   }
 
   publishing {
@@ -484,7 +484,7 @@ project(':cruise-control-metrics-reporter') {
     testImplementation "org.apache.kafka:kafka-raft:$kafkaVersion"
     testImplementation "org.apache.kafka:kafka-storage:$kafkaVersion"
     testImplementation 'commons-io:commons-io:2.11.0'
-    testImplementation "org.testcontainers:kafka:$testcontainersVersion"
+    testImplementation "org.testcontainers:testcontainers-kafka:$testcontainersVersion"
     testOutput sourceSets.test.output
   }
 
--- a/build.gradle
+++ b/build.gradle
@@ -339,7 +339,7 @@ project(':cruise-control') {
     testImplementation 'com.jayway.jsonpath:json-path:2.7.0'
     testImplementation 'org.powermock:powermock-module-junit4:2.0.9'
     testImplementation 'org.powermock:powermock-api-easymock:2.0.9'
-    testImplementation "org.testcontainers:testcontainers-kafka:$testcontainersVersion"
+    testImplementation "org.testcontainers:kafka:$testcontainersVersion"
   }
 
   publishing {
@@ -484,7 +484,7 @@ project(':cruise-control-metrics-reporter') {
     testImplementation "org.apache.kafka:kafka-raft:$kafkaVersion"
     testImplementation "org.apache.kafka:kafka-storage:$kafkaVersion"
     testImplementation 'commons-io:commons-io:2.11.0'
-    testImplementation "org.testcontainers:testcontainers-kafka:$testcontainersVersion"
+    testImplementation "org.testcontainers:kafka:$testcontainersVersion"
     testOutput sourceSets.test.output
   }
 
--- a/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
+++ b/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
@@ -13,6 +13,7 @@ import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.admin.TopicDescription;
 import org.apache.kafka.common.Uuid;
 import org.apache.kafka.common.config.types.Password;
+import org.apache.kafka.common.errors.RetriableException;
 import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
@@ -376,8 +377,8 @@ public class CCContainerizedKraftCluster implements Startable {
           Thread.currentThread().interrupt();
           throw new RuntimeException("Interrupted while waiting for broker to become ready", e);
         } catch (ExecutionException e) {
-          if (e.getCause() instanceof UnknownTopicOrPartitionException) {
-            // Topic doesn't exist yet, retry
+          if (e.getCause() instanceof UnknownTopicOrPartitionException || e.getCause() instanceof RetriableException) {
+            // Topic doesn't exist yet or transient error, retry
             return null;
           }
           throw new RuntimeException("Failed to describe topics: " + topicNames, e);
@@ -412,11 +413,10 @@ public class CCContainerizedKraftCluster implements Startable {
             Thread.currentThread().interrupt();
             throw new RuntimeException("Interrupted while waiting for broker removal", e);
         } catch (ExecutionException e) {
-          if (e.getCause() instanceof UnknownTopicOrPartitionException) {
-            // Topic doesn't exist yet, retry
+          if (e.getCause() instanceof RetriableException) {
             return false;
           }
-          throw new RuntimeException("Failed to describe cluster: " + e);
+          throw new RuntimeException("Failed to describe cluster: ", e);
         } catch (TimeoutException e) {
           // Timeout fetching metadata, treat as transient and retry
           return false;
--- a/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
+++ b/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
@@ -396,6 +396,25 @@ public class CCContainerizedKraftCluster implements Startable {
       timeout,
       String.format("Broker %s did not shutdown properly.", brokerId)
     );
+
+    // Wait until describeLogDirs fails
+    waitUntil(
+      () -> {
+        try {
+          _adminClient.describeLogDirs(Collections.singletonList(brokerId))
+            .allDescriptions()
+            .get(5, TimeUnit.SECONDS);
+          return false;
+        } catch (InterruptedException ie) {
+          throw new RuntimeException(ie);
+        } catch (Exception e) {
+          return true;
+        }
+      },
+      result -> result,
+      timeout,
+      String.format("Broker %s did not fully shut down (logDirs RPC still succeeds).", brokerId)
+    );
   }
 
   public static class BrokerWaitStrategy extends AbstractWaitStrategy {
--- a/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
+++ b/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
@@ -31,6 +31,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -274,6 +275,15 @@ public class CCContainerizedKraftCluster implements Startable {
   }
 
   /**
+   * Returns the {@link Admin} client used by this cluster instance.
+   *
+   * @return the {@link Admin} client associated with this cluster
+   */
+  public Admin adminClient() {
+    return _adminClient;
+  }
+
+  /**
    * Returns list of KafkaContainer broker objects within the TestContainer Kafka cluster.
    *
    * @return List of KafkaContainer broker objects within the TestContainer Kafka cluster.
@@ -338,36 +348,47 @@ public class CCContainerizedKraftCluster implements Startable {
   }
 
   /**
-   * Waits until the metadata for a Kafka topic meets the specified condition,
+   * Waits until the metadata for the listed Kafka topics meet the specified condition,
    * or until the given timeout period elapses.
    *
-   * @param topicName the name of the topic whose metadata should be monitored
-   * @param timeout   the maximum duration to wait for the condition to be satisfied
+   * @param topicNames the names of the topics whose metadata should be monitored
+   * @param fetchTimeout the timeout duration for the metadata retrieval call
+   * @param waitTimeout the maximum duration to wait for the condition to be satisfied
    * @param condition a {@link Predicate} that tests the {@link TopicDescription} for readiness
-   * @return the {@link TopicDescription} once the condition evaluates to {@code true}
+   * @return the {@link Map}{@code <String, TopicDescription>} once the condition evaluates to {@code true}
    */
-  public TopicDescription waitForTopicMetadata(String topicName, Duration timeout, Predicate<TopicDescription> condition) {
+  public Map<String, TopicDescription> waitForTopicMetadata(List<String> topicNames,
+                                               Duration fetchTimeout,
+                                               Duration waitTimeout,
+                                               Predicate<TopicDescription> condition) {
     return waitUntil(
       () -> {
         try {
-          return _adminClient.describeTopics(Collections.singleton(topicName))
-              .topicNameValues()
-              .get(topicName)
-              .get();
+          Map<String, TopicDescription> descriptions =
+            _adminClient.describeTopics(topicNames)
+              .allTopicNames()
+              .get(fetchTimeout.toSeconds(), TimeUnit.SECONDS);
+
+          boolean topicMetadataReady = descriptions.values().stream().allMatch(td -> td != null && condition.test(td));
+
+          return topicMetadataReady ? descriptions : null;
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+          throw new RuntimeException("Interrupted while waiting for broker to become ready", e);
         } catch (ExecutionException e) {
           if (e.getCause() instanceof UnknownTopicOrPartitionException) {
             // Topic doesn't exist yet, retry
             return null;
           }
-          throw new RuntimeException("Failed to describe topic: " + topicName, e);
-        } catch (InterruptedException e) {
-          Thread.currentThread().interrupt();
-          throw new RuntimeException("Interrupted while waiting for broker to become ready", e);
+          throw new RuntimeException("Failed to describe topics: " + topicNames, e);
+        } catch (TimeoutException e) {
+          // Timeout fetching metadata, treat as transient and retry
+          return null;
         }
       },
-      td -> td != null && condition.test(td),
-      timeout,
-      String.format("Timeout waiting for topic %s metadata to be ready.", topicName)
+      Objects::nonNull,
+      waitTimeout,
+      String.format("Timeout waiting for topics %s metadata to be ready.", topicNames)
     );
   }
 
@@ -385,35 +406,25 @@ public class CCContainerizedKraftCluster implements Startable {
     waitUntil(
       () -> {
         try {
-          return _adminClient.describeCluster().nodes().get().stream()
-            .map(node -> String.valueOf(node.id()))
-            .collect(Collectors.toSet());
-        } catch (InterruptedException | ExecutionException e) {
-          throw new RuntimeException(e);
+          return _adminClient.describeCluster().nodes().get(10, TimeUnit.SECONDS)
+            .stream().noneMatch(node -> node.id() == brokerId);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Interrupted while waiting for broker removal", e);
+        } catch (ExecutionException e) {
+          if (e.getCause() instanceof UnknownTopicOrPartitionException) {
+            // Topic doesn't exist yet, retry
+            return false;
+          }
+          throw new RuntimeException("Failed to describe cluster: " + e);
+        } catch (TimeoutException e) {
+          // Timeout fetching metadata, treat as transient and retry
+          return false;
         }
       },
-      brokerIds -> !brokerIds.contains(String.valueOf(brokerId)),
+      Boolean::booleanValue,
       timeout,
       String.format("Broker %s did not shutdown properly.", brokerId)
-    );
-
-    // Wait until describeLogDirs fails
-    waitUntil(
-      () -> {
-        try {
-          _adminClient.describeLogDirs(Collections.singletonList(brokerId))
-            .allDescriptions()
-            .get(5, TimeUnit.SECONDS);
-          return false;
-        } catch (InterruptedException ie) {
-          throw new RuntimeException(ie);
-        } catch (Exception e) {
-          return true;
-        }
-      },
-      result -> result,
-      timeout,
-      String.format("Broker %s did not fully shut down (logDirs RPC still succeeds).", brokerId)
     );
   }
 
--- a/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
+++ b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
@@ -17,6 +17,7 @@ import com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsUtil
 import com.linkedin.kafka.cruisecontrol.metricsreporter.config.EnvConfigProvider;
 import com.linkedin.kafka.cruisecontrol.monitor.ModelCompletenessRequirements;
 import com.linkedin.kafka.cruisecontrol.monitor.task.LoadMonitorTaskRunner;
+import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.admin.AlterConfigOp;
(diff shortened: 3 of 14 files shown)


danielgospodinow (approved):
LGTM! Seems like worst case we would wait for almost 1 min for a broker to terminate. But I guess that should be a rare case.

kyguy:
Hey @mimaison, been hunting down some flaky tests, could you have a pass at this as well when you get a chance?

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
```diff
@@ -396,6 +396,25 @@ public void shutDownBroker(int brokerId) {
       timeout,
       String.format("Broker %s did not shutdown properly.", brokerId)
     );
+
+    // Wait until describeLogDirs fails
+    waitUntil(
+      () -> {
+        try {
+          _adminClient.describeLogDirs(Collections.singletonList(brokerId))
```
mimaison: Can you detail the tests flakiness you're trying to fix with this?

I don't understand how this call could work when the container has been stopped and is not part of the cluster metadata anymore.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -869,8 +869,13 @@ private void executeAndVerifyProposals(AdminClient adminClient,
                   topic.partitions().get(tp.partition()).replicas().stream().anyMatch(rep -> r.brokerId().equals(rep.id())));
         }
       }
+
+      // Wait for leader movement
+      TopicDescription topicAfterLeaderMove = _cluster.waitForTopicMetadata(tp.topic(), Duration.ofSeconds(60),
```
mimaison: Do you have the logs of a failure? What are the partition replicas? Above we checked the replication factor matches the expected value, so I'm curious what the replica list is.

kyguy: > Do you have the logs of a failure?

Unfortunately not, the PRs that tripped the race condition have had the tests re-run, this was the assertion that failed [1] IIRC, it was the `testReplicaReassignmentProgressWithThrottle()` that was failing specifically, I'll try reproducing and getting logs.

[1] https://github.com/linkedin/cruise-control/blob/main/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java#L872-L873

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
kyguy: The `testBrokerDiesBeforeMovingPartition` test of the `ExecutorTest` class has been failing intermittently for CI in some PRs. [1] [2] 
```
com.linkedin.kafka.cruisecontrol.executor.ExecutorTest > testBrokerDiesBeforeMovingPartition FAILED
    java.lang.IllegalStateException: Failed to retrieve if there are already ongoing intra-broker replica reassignments.
        at com.linkedin.kafka.cruisecontrol.executor.Executor.sanityCheckOngoingMovement(Executor.java:1077)
        at com.linkedin.kafka.cruisecontrol.executor.Executor.startExecution(Executor.java:1016)
        at com.linkedin.kafka.cruisecontrol.executor.Executor.executeProposals(Executor.java:834)
        at com.linkedin.kafka.cruisecontrol.executor.ExecutorTest.executeAndVerifyProposals(ExecutorTest.java:828)
        at com.linkedin.kafka.cruisecontrol.executor.ExecutorTest.testBrokerDiesBeforeMovingPartition(ExecutorTest.java:198)

        Caused by:
        java.util.concurrent.TimeoutException
            at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1960)
            at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2095)
            at org.apache.kafka.common.internals.KafkaFutureImpl.get(KafkaFutureImpl.java:173)
            at com.linkedin.kafka.cruisecontrol.executor.ExecutorAdminUtils.hasOngoingIntraBrokerReplicaMovement(ExecutorAdminUtils.java:112)
            at com.linkedin.kafka.cruisecontrol.executor.Executor.sanityCheckOngoingMovement(Executor.java:1073)
            ... 4 more
```

> I don't understand how this call could work when the container has been stopped and is not part of the cluster metadata anymore.

From what I understand, even after the broker is removed from the cluster metadata, the controller may still be busy processing the broker shutdown. During this time, it can be delayed or fail to respond to `listPartitionReassignments()` requests [3] here leading to a timeout error here [4]

This extra wait was added to wait for the controller isn't overloaded any more and that the broker is completely shutdown.

[1] https://circleci.com/api/v1.1/project/github/linkedin/cruise-control/7014/output/106/0?file=true&allocation-id=692e17b7cb705e774ad6fd9c-0-build%2FABCDEFGH
[2] https://productionresultssa7.blob.core.windows.net/actions-results/a214b7ca-4dfe-4f26-a306-b5c629799e56/workflow-job-run-9dd3c006-698a-57a5-a8ca-be66dd4c4066/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-12-08T14%3A20%3A34Z&sig=VC46M8yrXq%2B17YI6MaP0SJTgc%2F9goso5ZEPN82kvl1s%3D&ske=2025-12-08T22%3A35%3A22Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-12-08T10%3A35%3A22Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2025-12-08T14%3A10%3A29Z&sv=2025-11-05
[3] https://github.com/linkedin/cruise-control/blob/main/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java#L371
[4] https://github.com/linkedin/cruise-control/blob/main/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L1059-L1065

CCisGG:
Hi @kyguy , just an fyi that me and @efeg will leave the company and we will probably lose the admin access to this repo. Please find @mhratson to merge/approve the PRs. Meanwhile I really appreciate you (and other active contributors) to help review the PRs and contribute a lot to make Cruise Control better. Thank you!

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
kyguy: I wasn't able to reproduce this issue locally or in the CI so I am going to strip this code out for now. If this assertion fails in the future I'll make sure to grab the logs and open a separate issue/fix for it.

kyguy:
@mimaison could you take another look at this when you get a chance?

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
```diff
@@ -396,6 +396,25 @@ public void shutDownBroker(int brokerId) {
       timeout,
       String.format("Broker %s did not shutdown properly.", brokerId)
     );
+
```
mimaison: This change does not make sense. It's trying to hide an issue specific to `hasOngoingIntraBrokerReplicaMovement()` that probably stems from the`MetadataClient` usage in `Executor`.

If we can't fix `Executor` now, I think it's acceptable to have this as an interim fix, but in that case it should be clearly documented so we remove this in the future. Have you tried updating `Executor` to not rely on internal Kafka APIs?

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
kyguy: Upon further investigation, it looks like metadataClient reports the killed broker in its live broker list in the call here `_metadataClient.cluster().nodes().stream().mapToInt(Node::id).boxed().collect(Collectors.toSet())` [1] even though the broker has been shutdown and removed from the cluster metadata here. [2] The test fails when the Admin client calls in the `hasOngoingIntraBrokerReplicaMovement()` method attempt to query log directory information from the dead broker. [3]

From what I understand, even after migrating away from the MetadataClient and onto the public Kafka API for metadata management, we will still need some sort of wait in the tests to wait for the dead broker to be removed from the cluster metadata and the topic metadata. 

I have been testing a new metadata client that exclusively uses public Kafka APIs (named `MetadataAdminClient`) and the `describeCluster().nodes.get()` Admin client calls (after the `shutdown()` method)  list the dead broker as alive sometimes when we don't explicitly wait for the dead broker to be removed from the topic metadata beforehand. My current suspicion is that the Admin client is seeing the id of the dead broker in topic metadata and thinks the broker is still alive.

Anyways, I have added an explicit wait for the topic metadata in the affected test in my latest commit so that the test passes consistently. That being said, I am still considering where the best place for this wait should be, either in the `shutdown()` method or the affected test (assuming we need the topic metadata wait). 

[1] https://github.com/linkedin/cruise-control/blob/main/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/Executor.java#L1073
[2] https://github.com/linkedin/cruise-control/blob/main/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java#L380-L399
[3] https://github.com/linkedin/cruise-control/blob/main/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorAdminUtils.java#L112

kyguy: > I have been testing a new metadata client that exclusively uses public Kafka APIs (named MetadataAdminClient) and the describeCluster().nodes.get() Admin client calls (after the shutdown() method) list the dead broker as alive sometimes when we don't explicitly wait for the dead broker to be removed from the topic metadata beforehand.

Wait ... I was short circuiting my cluster metadata waits with my exception handling in the `shutdown()` method. It looks like I shouldn't need this extra logic afterall. I'll strip it out

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
kyguy: > It's trying to hide an issue specific to hasOngoingIntraBrokerReplicaMovement() that probably stems from the `MetadataClient` usage in Executor.

After much debugging and testing it appears the issue was in fact surrounding the `MetadataClient` usage in the Executor. The `MetadataClient` and the Admin client used in the Executor class were not detecting cluster and topic metadata changes before the testing timeouts. 

I have updated the Executor to take an Admin client as an parameter so the tests can load an Admin client configured with a higher refresh rate to avoid using stale metadata.

kyguy:
Hi @mimaison, could you have another pass when you get a chance?

build.gradle
```diff
@@ -339,7 +339,7 @@ project(':cruise-control') {
     testImplementation 'com.jayway.jsonpath:json-path:2.7.0'
     testImplementation 'org.powermock:powermock-module-junit4:2.0.9'
     testImplementation 'org.powermock:powermock-api-easymock:2.0.9'
-    testImplementation "org.testcontainers:kafka:$testcontainersVersion"
+    testImplementation "org.testcontainers:testcontainers-kafka:$testcontainersVersion"
```
mimaison: How is this related to fixing the Executor tests?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java
mimaison, line 1: This a good cleanup but I'd rather do this in another PR as it's not related to the other changes

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ReplicationThrottleHelper.java
mimaison, line 1: his a good cleanup but I'd rather do this in another PR as it's not related to the other changes

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionUtils.java
mimaison, line 1: This a good cleanup but I'd rather do this in another PR as it's not related to the other changes

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/executor/ExecutionTaskPlanner.java
mimaison, line 1: This a good cleanup but I'd rather do this in another PR as it's not related to the other changes

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataAdminClient.java
```diff
@@ -0,0 +1,96 @@
...
+   * @param adminClient The AdminClient to use for fetching cluster metadata.
+   */
+  public MetadataAdminClient(Admin adminClient) {
+    _adminClient = adminClient;
+  }
+
+  /**
+   * Close adminClient
+   */
+  public void close() {
```
mimaison: It's a bit weird this is responsible for closing the `Admin` instance as this just received it and did not create it.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -103,12 +106,14 @@ public int clusterSize() {
   @Before
   public void setUp() {
     Properties adminClientProps = new Properties();
+    adminClientProps.setProperty(AdminClientConfig.METADATA_MAX_AGE_CONFIG, "0");
```
mimaison, line 109: Is this really needed?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -103,12 +106,14 @@ public int clusterSize() {
...
     setSecurityConfigs(adminClientProps, "admin");
 
     _cluster = new CCContainerizedKraftCluster(clusterSize(), buildBrokerConfigs(), adminClientProps);
     _cluster.start();
     _brokerAddressList = _cluster.getBrokerAddressList();
     _bootstrapUrl = _cluster.getExternalBootstrapAddress();
+    _adminClient = _cluster.adminClient();
```
mimaison, line 116: That client is never closed

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
```diff
@@ -659,45 +644,30 @@ private void populateProposals(List<ExecutionProposal> proposalToExecute,
...
+      List.of(TOPIC0, TOPIC1),
+      Duration.ofSeconds(15),
+      Duration.ofSeconds(60),
+      topicDescription -> {
+        return topicDescription.partitions() != null
+          && !topicDescription.partitions().isEmpty()
+          && topicDescription.partitions().stream().allMatch(p ->
+            p.leader() != null
+            && p.replicas() != null && !p.replicas().isEmpty()
+            && p.isr() != null && !p.isr().isEmpty());
```
mimaison: `replicas()` and `isr()` should not return null, see https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/TopicPartitionInfo.java#L58-L67

mimaison (commented):
Thanks for the PR. I left a few suggestions

build.gradle
kyguy: Due to breaking changes in recent Docker Engine updates we need to upgrade Testcontainers to a version that adds compatibility fixes, either [1.21.4](https://github.com/testcontainers/testcontainers-java/releases/tag/1.21.4) or [2.0.3](https://github.com/testcontainers/testcontainers-java/releases/tag/2.0.3).

That being said, we could avoid this specific artifact name change by sticking with the 1.x versions instead or jumping to 2.x versions as I have done as part of my commit.

Let me update the version to 1.21.4 in my next commit to avoid this artifact name change.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
kyguy: From what I understand, the client is closed at the end of every test by the `teardown()` method which calls `_cluster.close()` which calls `CCContainerizedKraftCluster`'s stop() method:

```
@Override
  public void stop() {
    this._adminClient.close();
    this._brokers.stream().parallel().forEach(GenericContainer::close);
  }
```

Does this not suffice?

Would it be better if I called `_cluster._stop()` or `._adminClient.close()` directly in the `teardown()` method instead?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataAdminClient.java
kyguy: Fair point, just removed in the latest commit.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
kyguy: I was having trouble getting the tests to pass consistently without it in the past, but the tests are seem to be working fine without it now. I think passing the Admin client to the Executor class was the main fix that was needed to address the flaky tests.

Just removed it in the latest commit.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
kyguy: Thanks for the information, I removed the null checks for these in the latest commit.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
kyguy: I noticed the flakiness reappearing in the CircleCI tests without this configuration so I added this configuration back in. 

From what I understand, without this configuration, the adminClient may get stale metadata when querying the cluster for the number of active brokers in the `hasOngoingIntraBrokerReplicaMovement()` method. This happens in the `testBrokerDiesBeforeMovingPartition()` test where a broker is shutdown and then metadata is requested from the active brokers (and the dead broker appears in the active brokers list)

From what I understand, we have a couple of options to solve this (1) Set AdminClientConfig.METADATA_MAX_AGE_CONFIG = "0" or (2) Update the `shutDownBroker()` method of the `CCContainerizedKraftCluster` class  to block until a few consecutive describeCluster() calls show that the shutdown broker is no longer apart of the cluster.

kyguy:
@mimaison Ready for another review (at your leisure of course)

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
mimaison: You're right `CCContainerizedKraftCluster` will close it when `tearDown()` runs.

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/executor/ExecutorTest.java
mimaison: Fair enough, it's probably something to revisit later but I guess we can have it for now.

mimaison (approved):
Thanks for the updates and tracking the underlying issues. Not all things are quite perfect yet but I think it's good enough so we can continue make progress on the other issues.

kyguy:
Hi @mhratson, I am not sure if you are still active here but if you are could you merge this when you get chance?

It will address the flaky tests that are causing the CI to fail on other PRs.

kyguy:
Hi @mimaison @viktorsomogyi I rebased these changes on top of `main` but want to make sure everything still looks ok from your point of view. I want to get this PR sorted and merged since it is causing CI failures for other PRs!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataAdminClient.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * Fetches the current metadata for the Kafka cluster.
+   *
+   * @return a {@link Cluster} containing the cluster ID, broker nodes, and partition information for all topics
+   */
+  public Cluster cluster() {
+    try {
+      Set<String> topicNames = _adminClient.listTopics().names().get();
```
viktorsomogyi: should we include internal topics as well?
`listTopics(new ListTopicsOptions().listInternal(true))`

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataAdminClient.java
```diff
@@ -0,0 +1,82 @@
...
+      List<PartitionInfo> partitionInfos = new ArrayList<>();
+
+      for (TopicDescription desc : topicDescriptions.values()) {
+        for (TopicPartitionInfo partInfo : desc.partitions()) {
+
+          Node leader = partInfo.leader();
+          Node[] replicas = partInfo.replicas().toArray(Node[]::new);
+          Node[] isr = partInfo.isr().toArray(Node[]::new);
+
+          partitionInfos.add(new PartitionInfo(desc.name(), partInfo.partition(), leader, replicas, isr));
```
viktorsomogyi: shouldn't we use the 6-arg constructor to include the offline replicas from `partInfo`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataAdminClient.java
```diff
@@ -0,0 +1,82 @@
...
+      return new Cluster(clusterId, nodes, partitionInfos, Collections.emptySet(), Collections.emptySet());
+    } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        LOG.error("Interrupted while fetching cluster metadata", e);
+        throw new RuntimeException("Interrupted while fetching cluster metadata", e);
+
+    } catch (ExecutionException e) {
+        LOG.error("ExecutionException while fetching cluster metadata", e);
+        throw new RuntimeException("Failed to fetch cluster metadata", e);
+    }
```
viktorsomogyi, line 96: Not sure if we should rethrow it as `RuntimeException` as we don't handle it on all call paths. What do you think about keeping the current (yet stale) metadata instead and keep retrying? Although we'd have a stale metadata, it wouldn't fail mid-rebalance or during some other operation.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataAdminClient.java
kyguy: Yes, added in latest commit.

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataAdminClient.java
kyguy: Yes, we ought to do this as well as it is information that is provided from the existing implementation. Unfortunately, `TopicPartitionInfo` doesn't expose offline replicas so I had to derive it from live broker set.

Let me know if that looks alright or if you think there is a better way to handle it!

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataAdminClient.java
kyguy: That’s a fair point. It looks like the MetadataClient class (which MetadataAdminClient is intended to replace) follows the behavior you suggested: it returns the cached (potentially stale) metadata and logs a warning if the calls fail.

I’ve updated the code to follow the same behavior. Now, it will only throw a `RuntimeException` if the repeated calls fail and no cached metadata already exists.

Let me know if that looks okay to you!

viktorsomogyi (commented):
I think it's all good from the production perspective, however, would you please add some unit tests to cover `MetadataAdminClient`?

cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/common/MetadataAdminClient.java
viktorsomogyi: Ok, I think it is a good approach here.

kyguy:
> I think it's all good from the production perspective, however, would you please add some unit tests to cover MetadataAdminClient?

Thanks for the review @viktorsomogyi, I have just added MetadataAdminClientTest class and Apache 2.0 headers to all the classes created or edited as part of this PR.

Let me know what you think!

viktorsomogyi (approved):
LGTM, if you can resolve the conflicts, I think we can merge it.

kyguy:
@mimaison There have been a couple changes to This PR since your last review/approval:

(1) Rebased on `main`
(2) We return cached cluster metadata on fetch failure.
(3) Added a  `MetadataAdminClientTest` class to test the methods in `MetadataAdminClient` class.
(4) Apache 2.0 headers were added to the new and edited classes.

Do these changes look ok to you?

mimaison:
> Do these changes look ok to you?

Yes, feel free to merge

kyguy:
Thanks Mickael!

PR #2286, opened 2025-07-16
Upgrade kafka to 3.9.1

JvD-Ericsson:
## Summary
1. Why: 
To remove CVEs:
	- CVE-2025-27817
	- CVE-2024-56128
	- CVE-2025-48734

2. What: 
Upgrade kafka to 3.9.1 to remove CVE-2025-27817 and CVE-2024-56128,
and add dependency constraint for commons-beanutils:1.11.0  to remove CVE-2025-48734

## Additional evidence
Partial output from security scanner Trivy
image

## Categorization
- [x] security/CVE

PR #2342, opened 2025-12-09
Upgrade Kafka to 4.3.1

### LINKED ISSUE #2341: Bump Kafka version to 4.1.1
Bump Kafka version to 4.1.1.

egyedt:
## Summary
Bump Kafka version to 4.3.0.

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [x] other

This PR resolves #2341 if any.


gradle.properties
```diff
@@ -2,7 +2,7 @@ org.gradle.daemon=false
 org.gradle.parallel=false
 org.gradle.jvmargs=-Xms512m -Xmx512m
 scalaVersion=2.13.13
-kafkaVersion=4.0.0
+kafkaVersion=4.1.1
```
dbahatSAP, line 5: Can you please also remove the now un-needed gradle override in https://github.com/linkedin/cruise-control/pull/2286/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R311?
See https://github.com/linkedin/cruise-control/issues/2354

egyedt: good point, thanks!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBroker.java
```diff
@@ -60,7 +61,7 @@ public CCEmbeddedBroker(Map<Object, Object> config) {
   private void parseConfigs(Map<Object, Object> config) {
     readLogDirs(config);
     _id = Integer.parseInt((String) config.get(KRaftConfigs.NODE_ID_CONFIG));
-    _metadataLogDir = new File((String) config.get(KRaftConfigs.METADATA_LOG_DIR_CONFIG));
+    _metadataLogDir = new File((String) config.get(MetadataLogConfig.METADATA_LOG_DIR_CONFIG));
```
kyguy: In effort to avoid dependence on internal/private Kafka APIs like `MetadataLogConfig` [1] could we create a field like the following:
```
public static final String METADATA_LOG_DIR_CONFIG = "metadata.log.dir";
```
in the `KafkaServerConfigs` class [2] instead?

[1] https://github.com/linkedin/cruise-control/issues/2282
[2] https://github.com/linkedin/cruise-control/blob/main/cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/KafkaServerConfigs.java

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBrokerBuilder.java
```diff
@@ -285,13 +274,12 @@ public Map<Object, Object> buildConfig() {
     props.put(SocketServerConfigs.LISTENER_SECURITY_PROTOCOL_MAP_CONFIG, "CONTROLLER:PLAINTEXT,PLAINTEXT:PLAINTEXT,SSL:SSL");
     props.put(SocketServerConfigs.LISTENERS_CONFIG, csvJoiner.toString());
     props.put(ServerLogConfigs.LOG_DIR_CONFIG, _logDirectory.getAbsolutePath());
-    props.put(KRaftConfigs.METADATA_LOG_DIR_CONFIG, _metadataLogDirectory.getAbsolutePath());
+    props.put(MetadataLogConfig.METADATA_LOG_DIR_CONFIG, _metadataLogDirectory.getAbsolutePath());
```
kyguy: Same comment as above.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
```diff
@@ -222,7 +222,7 @@ private static MetaTuple initializeLogDirs(KafkaConfig config, String logPrefix)
 
     private void initializeMetaData(KafkaConfig config) {
         Set<File> allLogDirs = new HashSet<>(readLogDirs(config));
-        allLogDirs.add(new File(config.getString(KRaftConfigs.METADATA_LOG_DIR_CONFIG)));
+        allLogDirs.add(new File(config.getString(MetadataLogConfig.METADATA_LOG_DIR_CONFIG)));
```
kyguy: Same comment as above.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterTest.java
```diff
@@ -201,7 +201,8 @@ public void testReportingMetrics() {
   }
 
   @Test
-  public void testUpdatingMetricsTopicConfig() throws InterruptedException, TimeoutException {
+  @Ignore("Container stop and start will destroy Kafka broker and create a new one so this test is not valid in this form.")
```
kyguy: Even if we are recreating a broker here, isn't the test still valid? From what I understand the cruise control metric reporter agent of the recreated broker will update the metrics topic config won't it?

[1] https://github.com/linkedin/cruise-control/blob/e30eaf352c31511241f4dfae457fbcb77022a9c6/cruise-control-metrics-reporter/src/main/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporter.java#L354-L371

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
```diff
@@ -42,9 +42,9 @@
 import org.apache.kafka.server.config.ServerTopicConfigSynonyms;
 import org.apache.kafka.server.metrics.KafkaYammerMetrics;
 import org.apache.kafka.storage.internals.log.LogConfig;
+import org.apache.kafka.storage.internals.log.UnifiedLog;
```
kyguy: Same comment for this as well, is there a way we can avoid using private Kafka APIs?

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBroker.java
egyedt: ack

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBrokerBuilder.java
egyedt: ack

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
egyedt: ack

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterTest.java
egyedt: I think TestContainers are not OK with it. I tested this multiple times and it seems like it is flaky and it can stuck in a bad state (The still available node tries to connect to the other node which is not existing, and the destroyed node cannot be created again due to Kafka errors).
It may need a deeper investigation, but I disabled this test now, since it was very flaky on my PRs.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCKafkaRaftServer.java
egyedt: ack

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/CruiseControlMetricsReporterTest.java
kyguy: I took a closer look at this today and was able to reproduce the issue on my side. It appears that this test was never fully migrated for KRaft clusters. Currently, it sets up a cluster with only two brokers, so when the test restarts one of them, the quorum is temporarily lost. As a result, the check for the topic metadata change times out while waiting for the quorum to recover.

From what I understand, Kafka 4.x enforces stricter Raft semantics compared to Kafka 3.9.1, which is why this test used to pass consistently before the upgrade.

To address this issue we can create a cluster with 3 brokers so that the quorum isn't lost when one of the brokers is restarted. We can do this by adding the following to the class:

```
  private static final int NUM_OF_BROKERS = 3;

  @Override
  protected int clusterSize() {
    return NUM_OF_BROKERS;
  }
```

With this change the test seems to pass consistently on my side, let me know what you think and/or if it works for you!

egyedt: Thanks for investigating it. I applied this fix in a new commit, since if it was more frequent with Kafka 4+, then it should be fixed in this PR.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCEmbeddedBroker.java
danielgospodinow: @kyguy, do you think there's some way we can clearly instruct people to avoid adding dependencies on the internal Kafka APIs? Because doing so would start nullifying your previous work to decouple CC from them.

Maybe a README heads up, maybe a contribution guide, maybe a PR template hint...

kyguy: > Maybe a README heads up, maybe a contribution guide, maybe a PR template hint...

Those all sound like good ideas to me. It'll definitely be a lot easier after we have finished the migration off of internal Kafka APIs and have the related dependencies removed from the gradle file. Then we could have some sort of check in the PR template to make sure no internal Kafka dependencies have been added to the gradle file as part of the PR.

kyguy (approved):
Thanks for the changes @egyedt, looks great!

egyedt:
@bgrishinko 
This PR has an approval and is available to be merged.
Thanks in advance!

kyguy:
@egyedt Since Kafka 4.3.0 has been released since this PR was opened, would it be worth updating the PR to bump the Kafka version to 4.3.0? What do you think?

egyedt:
> @egyedt Since Kafka 4.3.0 has been released since this PR was opened, would it be worth updating the PR to bump the Kafka version to 4.3.0? What do you think?

Good idea @kyguy, I updated the PR.

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
```diff
@@ -55,7 +55,7 @@
  * </ul>
  */
 public class CCContainerizedKraftCluster implements Startable {
-  private static final String KAFKA_IMAGE = System.getenv().getOrDefault("KAFKA_IMAGE", "apache/kafka:3.9.1");
+  private static final String KAFKA_IMAGE = System.getenv().getOrDefault("KAFKA_IMAGE", "apache/kafka:4.1.1");
```
kyguy: Does this need to be updated to 4.3.0 as well? (Hopefully, there is a 4.3.0 image available and works ok, otherwise will need to stick with this one)

kyguy (approved):
Looks good overall @egyedt, just one comment and a couple minor checkstyle failures to resolve and we should be clear with the CI!

cruise-control-metrics-reporter/src/test/java/com/linkedin/kafka/cruisecontrol/metricsreporter/utils/CCContainerizedKraftCluster.java
egyedt: good finding, I missed this part for this time. Updated the PR!

README.md
```diff
@@ -51,7 +51,8 @@ Cruise Control for Apache Kafka
   `2.6` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.11+`), `2.7` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.36+`),
   `2.8` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.66+`), `3.0` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.85+`),
   `3.1` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.85+`), `3.8` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.142+`),
-  `3.9` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.143+`), and `4.0` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.144+`)
+  `3.9` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.143+`), `4.0` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.144+`),
+  and `4.1` (i.e. [Releases](https://github.com/linkedin/cruise-control/releases) with `2.5.146+`)
```
kyguy: Does this need to be `4.3` as well?

Not huge deal, I imagine we will clean this README.md file up before the next release anyway

egyedt: good catch

egyedt:
@kyguy 
4.3.1 is also available, so I used that version in the final commit!

kyguy:
@egyedt one of the CI failures is due to a bug in the Vert.x webserver implementation. The [SAST security fix](https://github.com/cruise-control-for-kafka/cruise-control/pull/2395) added a global security block to base.yaml, which exposed that the Vert.x webserver implementation enforces OpenAPI security requirements regardless of the `webserver.security.enable config` is set. I have added a fix for it here https://github.com/cruise-control-for-kafka/cruise-control/pull/2404.

egyedt:
> @egyedt one of the CI failures is due to a bug in the Vert.x webserver implementation. The https://github.com/cruise-control-for-kafka/cruise-control/pull/2395 added a global security block to base.yaml, which exposed that the Vert.x webserver implementation enforces OpenAPI security requirements regardless of the webserver.security.enable config is set. I have added a fix for it here https://github.com/cruise-control-for-kafka/cruise-control/pull/2404.

Thanks a lot @kyguy!
I hope this time there will be no failure on this PR. Unfortunately, before my rebase of the SAST PR (#2395) everything was green, so I merged it right after the rebase, but the sec. config was not right, thanks for fixing it in #2404

egyedt:
@kyguy 
we can merge this if you agree!

kyguy:
Awesome work @egyedt Thanks for driving this forward!

PR #2349, opened 2025-12-17
Minor refactor in start script to remove Kafka-related leftovers

danielgospodinow:
## What

CC's start script defines variables with a `KAFKA_` prefix which is incorrect. This is likely a leftover from copy-pasting Kafka start scripts. 😁 

Looking at [kafka-run-class.sh](https://github.com/apache/kafka/blob/trunk/bin/kafka-run-class.sh) and [kafka-server-start.sh](https://github.com/apache/kafka/blob/trunk/bin/kafka-server-start.sh) further supports my claim. Lol.

## Why

Might cause a confusion when reading the code, especially to an engineer that's new to the codebase.

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [x] refactor
- [ ] security/CVE
- [ ] other

### DIFF
--- a/kafka-cruise-control-start.sh
+++ b/kafka-cruise-control-start.sh
@@ -62,13 +62,13 @@ fi
 shopt -u nullglob
 
 # JMX settings
-if [ -z "$KAFKA_JMX_OPTS" ]; then
-  KAFKA_JMX_OPTS="-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=false  -Dcom.sun.management.jmxremote.ssl=false "
+if [ -z "$JMX_OPTS" ]; then
+  JMX_OPTS="-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=false  -Dcom.sun.management.jmxremote.ssl=false "
 fi
 
 # JMX port to use
 if [  $JMX_PORT ]; then
-  KAFKA_JMX_OPTS="$KAFKA_JMX_OPTS -Dcom.sun.management.jmxremote.port=$JMX_PORT "
+  JMX_OPTS="$JMX_OPTS -Dcom.sun.management.jmxremote.port=$JMX_PORT "
 fi
 
 # Log directory to use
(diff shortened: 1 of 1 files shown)


kafka-cruise-control-start.sh
```diff
@@ -77,30 +77,26 @@ if [ "x$LOG_DIR" = "x" ]; then
...
 else
   # create logs directory
   if [ ! -d "$LOG_DIR" ]; then
     mkdir -p "$LOG_DIR"
   fi
 fi
 
-# If Cygwin is detected, LOG_DIR is converted to Windows format.
-(( CYGWIN )) && LOG_DIR=$(cygpath --path --mixed "${LOG_DIR}")
-KAFKA_LOG4J_OPTS="-Dkafka.logs.dir=$LOG_DIR $KAFKA_LOG4J_OPTS"
```
danielgospodinow, line 95: This is the only line I've removed given that `kafka.logs.dir` is not part of Cruise Control. Everything else is just renaming.

kyguy (approved):
+1

kyguy:
CircleCI failures are unrelated to the changes in this PR, they are due to some flaky tests which should be resolved by https://github.com/linkedin/cruise-control/pull/2338.

mimaison:
This is a major operational change for anyone using Cruise Control. I expect a (large?) portion of users set some of these environment variables to tune Cruise Control and this is going to break their environments. 

I can understand the desire to move away from Kafka names but it should be done in a compatible manner to give time for people to migrate.

I also wonder if names prefixed with something like `CC_` would be better than the names used i this PR.

danielgospodinow:
> I expect a (large?) portion of users set some of these environment variables to tune Cruise Control and this is going to break their environments. 

True. Yet I think it's the right thing to do since the current names are confusing if you operate both CC and Kafka (which obviously is quite common), and it's anyway a simple change to adopt if something breaks.

> I also wonder if names prefixed with something like CC_ would be better than the names used i this PR.

Sure, I'm fine with this suggestion too.

egyedt (approved):
LGTM!

egyedt:
@danielgospodinow 

Thank you for the PR.

Could you please rebase this PR and updated the commit messages based on the DCO rules!?
Thanks in advance!

mimaison:
Should we add a prefix (for example `CC_` or `CRUISECONTROL_`) to these new variables to avoid collisions?

PR #2348, opened 2025-12-17
Create Dockerfile

danielgospodinow:
## What

A simple two-staged Dockerfile that would provide a starting point for creating CC containers.

Resolves #2263. 

## Why

It eases experimentation and deployment, i.e. overall development, of CC.

## Suggestions

- We can perhaps follow up with creating a Helm chart to further ease CC's deployment on k8s. But I'm not sure how much the community wants this.
- We can also think about creating a CI job that builds CC Docker images and stores them in GitHub's registry for this repository.

I might open issues for both so that we can keep track of them and discuss them.

## Categorization
- [ ] documentation
- [ ] bugfix
- [x] new feature
- [ ] refactor
- [ ] security/CVE
- [ ] other

PR #2351, opened 2025-12-17
Add JMX Prometheus exporter agent

### LINKED ISSUE #2350: Cruise Control doesn't expose its JMX metrics
## Overview

As of today, there's no easy way to observe CC's metrics.

It would be great if it had a Prometheus metrics endpoint that can be scraped, for example.

An easy solution to this is to add a [JMX exporter](https://github.com/prometheus/jmx_exporter) agent to the CC process. This would provide a `/metrics` endpoint that can be used from Prometheus.

danielgospodinow:
## What

Provides a [JMX exporter](https://github.com/prometheus/jmx_exporter) agent (version [1.5.0](https://github.com/prometheus/jmx_exporter/releases/tag/1.5.0)) to CC. 

This allows CC to expose its JMX metrics in a Prometheus format via a `/metrics` endpoint on the `7071` port (hardcoded but can be made configurable if people think that'd be useful).

Resolves #2350.

## Why

Enables CC to be observed from a Prometheus/Grafana observability setup, which in turn enables creating dashboards for CC's health and performance, alerts for degradations, and so on.

## Categorization
- [ ] documentation
- [ ] bugfix
- [x] new feature
- [ ] refactor
- [ ] security/CVE
- [ ] other

### DIFF
--- a/kafka-cruise-control-start.sh
+++ b/kafka-cruise-control-start.sh
@@ -184,6 +184,9 @@ fi
 # If Cygwin is detected, classpath is converted to Windows format.
 (( CYGWIN )) && CLASSPATH=$(cygpath --path --mixed "${CLASSPATH}")
 
+# Configure the JMX exporter agent.
+JVM_OPTS="$JVM_OPTS -javaagent:observability/jmx_prometheus_javaagent-1.2.0.jar=7071:observability/jmx_exporter_config.yaml"
+
 # Launch mode
 if [ "x$DAEMON_MODE" = "xtrue" ]; then
   nohup $JAVA $KAFKA_HEAP_OPTS $KAFKA_JVM_PERFORMANCE_OPTS $KAFKA_GC_LOG_OPTS $KAFKA_JMX_OPTS $KAFKA_LOG4J_OPTS -cp $CLASSPATH $KAFKA_OPTS com.linkedin.kafka.cruisecontrol.KafkaCruiseControlMain "$@" > "$CONSOLE_OUTPUT_FILE" 2>&1 < /dev/null &
--- a/kafka-cruise-control-start.sh
+++ b/kafka-cruise-control-start.sh
@@ -185,7 +185,7 @@ fi
 (( CYGWIN )) && CLASSPATH=$(cygpath --path --mixed "${CLASSPATH}")
 
 # Configure the JMX exporter agent.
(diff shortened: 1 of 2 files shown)


danielgospodinow:
If https://github.com/linkedin/cruise-control/pull/2348 is merged when this PR is ready, we should also update here the `Dockerfile` to include the `observability` directory.

```Dockerfile
COPY --from=build /workspace/observability/ /cc/observability/
```

kafka-cruise-control-start.sh
```diff
@@ -184,6 +184,9 @@ fi
 # If Cygwin is detected, classpath is converted to Windows format.
 (( CYGWIN )) && CLASSPATH=$(cygpath --path --mixed "${CLASSPATH}")
 
+# Configure the JMX exporter agent.
+KAFKA_OPTS="$KAFKA_OPTS -javaagent:observability/jmx_prometheus_javaagent-1.5.0.jar=7071:observability/jmx_exporter_config.yaml"
```
danielgospodinow: If https://github.com/linkedin/cruise-control/pull/2349 gets merged first, we should rename this `KAFKA_OPTS` to `JVM_OPTS`.

danielgospodinow:
If https://github.com/linkedin/cruise-control/pull/2349 gets merged first, we should rename `KAFKA_OPTS` to `JVM_OPTS` in the start script.

kyguy (commented):
Isn't this possible via the JMX endpoint? From the Cruise Control Wiki: [1]

> You can easily expose Cruise Control metrics by [enabling](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_4/kafka-cruise-control-start.sh#L64) Java Management Extensions (JMX), and collect them by using your favourite JMX collection tool (e.g - [jmx_exporter](https://github.com/prometheus/jmx_exporter)).

Does embedding the Prometheus JMX Exporter provide sufficient advantages to justify including it by default, rather than leaving its configuration and deployment to the user?

[1] https://github.com/linkedin/cruise-control/wiki/Sensors#metrics-collection

danielgospodinow:
Oh, okay! That makes sense. I knew I'm likely missing something.

So enabling the JMX remote endpoint and deploying a standalone JMX exporter externally would probably work just as good.

I'll play around with this approach and compare the two options. If what currently exists is enough, then I'll close both the PR and the issue.

Thanks for the heads up, @kyguy!

danielgospodinow:
Closing because Kyle has a point. There's no need for this.

PR #2288, opened 2025-07-17
Implementation of intra-broker goal violation detection and self-healing

dgldn:
## Summary
This feature is completely based on work from @ecojan who worked on the [PR](https://github.com/linkedin/cruise-control/pull/1721). And, reviewed previoulsy by @alexander-falca @amuraru @zornhsu @CCisGG

1. Why:
This PR resolves [#1264](https://github.com/linkedin/cruise-control/issues/1264) by the goal violation detection and self-healing for intra-broker goals. This however does not account for race conditions between Cluster level Goals and will only trigger after said goals are finished.
The scope of this PR is to allow cruise control run IntraBroker goal violation detection and execute self-healing for now (as future work might make this obsolete as intra-broker goals and regular goals might be made to work together).

2. What:
Implementation of intra-broker goals violation detection and executing self-healing in case of JBOD(multiple logDirs) based kafka clusters.

### New Configurations:
```
anomaly.detection.intra.broker.goals= List of intra-broker goals for the Anomaly detector to look for
self.healing.intra.broker.goals= List of Intra broker Goals for self-healing
self.healing.intra.broker.goal.violation.enabled= If the self-healing should be enabled for intra-broker-goals. This configuration defaults to self.healing.enabled
```

The configs `anomaly.detection.intra.broker.goals` and `self.healing.intra.broker.goals` both default to `com.linkedin.kafka.cruisecontrol.analyzer.goals.IntraBrokerDiskCapacityGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.IntraBrokerDiskUsageDistributionGoal` since these are the only cruise control intra-broker goals available at the moment.

The `self.healing.intra.broker.goal.violation.enabled` can be disbaled explicilty overriding the config to `false`.

### New Anomaly Type: INTRA_BROKER_GOAL_VIOLATION
A new type of Anomaly that needs to be treated differently from the regular GOAL_VIOLATIONS.
When configuring for the rebalance parameters for this violation, we are skipping hard goals check and ignoring proposalsCache.

We are skipping hard goals check because intra-broker goals may not be hard goals (at the current given time).
We are ignoring proposalsCache because the proposals model takes into account all given default goals. If we add the intraBrokersGoals this will not work properly as currently, disk-granularity goals do not work with broker-granularity goals.

### New Anomaly detectors: IntraBrokerGoalViolationDetector
[...]

PR #2370, opened 2026-05-14
Add intra-broker replication throttle for log dir reassignment

dattagiri:
## Summary

This PR extends the work in https://github.com/linkedin/cruise-control/pull/2288 (intra-broker self-healing disk rebalance) by adding throttling support for intra-broker replica movements during log dir reassignment.

**Key changes:**
- New `IntraBrokerReplicationThrottleHelper` class that manages Kafka's `replica.alter.log.dirs.io.max.bytes.per.second` broker config (KIP-113)
- New Cruise Control config `default.intra.broker.replication.throttle` with fallback to `default.replication.throttle`
- New REST API parameter `intra_broker_replication_throttle` on the `/rebalance` endpoint
- Throttle lifecycle: set before disk moves, cleared incrementally per broker as tasks complete, final cleanup in `finally` block
- Pre-existing dynamic broker throttle values are preserved and restored after execution
- Stop execution waits for in-progress moves to complete before removing throttle (moves cannot be cancelled by Cruise Control)
- Externally-cancelled tasks are marked dead when stop is signalled to prevent infinite re-execute loops
- Explicit UTF-8 encoding for Java/Scala compilation to fix Docker build failures in POSIX/C locale environments

## Test plan

- [x] Unit tests for `IntraBrokerReplicationThrottleHelper` (10 tests covering set/clear/restore/timeout/static-config scenarios)
- [x] Existing `MaintenanceEventTest`, `AnomalyDetectorManagerTest`, `IdempotenceCacheTest` pass with updated mock expectations
- [x] `ExecutorTest` updated for new `executeProposals` parameter signature
- [ ] Integration testing with Docker-based `ExecutorTest` (requires Docker environment)
- [ ] Manual testing with a multi-disk Kafka cluster

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kyguy:
Hi @dattagiri, does this PR overlap with any of the work in PR #2145? I'm wondering if it might make more sense for us to combine our efforts on that PR and get it over the finish line, especially since it has already been reviewed. What do you think?

PR #2377, opened 2026-06-01
fix: add missing EasyMock.verify calls before reset in unit tests 

### LINKED ISSUE #1408: Missing EasyMock.verify in unit tests
In some unit tests, before resetting mocked objects or after mocked objects get used, `EasyMock.verify` does not get invoked on mocked objects. So, we should add it.

For example, in `UserTaskManagerTest.java` and `KafkaCruiseControlServletEndpointTest.java`. There could be more cases. So, I suggest thoroughly iterate all cases where `EasyMock` gets used.

sameer-sde:
## Summary
1. Why: EasyMock's API contract requires `verify()` to be called before `reset()` on a mock, to ensure all expected interactions occurred before the mock's state is cleared. Several unit tests were resetting mocks without verifying them first.
2. What: Added the missing `EasyMock.verify()` calls before `EasyMock.reset()` in `UserTaskManagerTest.java` and `KafkaCruiseControlServletEndpointTest.java`, and added `EasyMock.replay()` after reset where needed to keep the mocks in a valid state for subsequent test cases.

## Expected Behavior
All EasyMock mocks are verified before being reset, per EasyMock's contract, without breaking any existing test behavior.

## Actual Behavior
Some tests called `EasyMock.reset()` without a preceding `EasyMock.verify()`, skipping validation of prior mock interactions.

## Steps to Reproduce
1. Run `UserTaskManagerTest` and `KafkaCruiseControlServletEndpointTest` before this fix.
2. Observe that `reset()` calls occur without a prior `verify()`, meaning expected interactions on the mock were never validated.

## Known Workarounds
None — this is a test-hygiene fix; it doesn't affect production behavior.

## Additional evidence
1. All 6 `UserTaskManagerTest` tests pass.
2. `KafkaCruiseControlServletEndpointTest.testUserTaskParameters` is `@Ignored` upstream and remains skipped.

## Categorization
- [ ] documentation
- [x] bugfix
- [ ] new feature
- [ ] refactor
- [ ] security/CVE
- [ ] other

This PR resolves #1408.

### DIFF
--- a/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
+++ b/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
@@ -171,6 +171,7 @@ public class KafkaCruiseControlServletEndpointTest {
     List<UserTaskManager.UserTaskInfo> result1 = userTaskState.prepareResultList(parameters1);
     // Test Case 1 result
     Assert.assertEquals(3, result1.size());
+    EasyMock.verify(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
     EasyMock.reset(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
 
     // Test Case 2: Get all tasks from client 0.0.0.1
@@ -183,6 +184,7 @@ public class KafkaCruiseControlServletEndpointTest {
     List<UserTaskManager.UserTaskInfo> result2 = userTaskState.prepareResultList(parameters2);
     // Test Case 2 result
     Assert.assertEquals(3, result2.size());
+    EasyMock.verify(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
     EasyMock.reset(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
 
     // Test Case 3: Get all PROPOSALS and REMOVE_BROKERS from client 0.0.0.1
@@ -196,6 +198,7 @@ public class KafkaCruiseControlServletEndpointTest {
     List<UserTaskManager.UserTaskInfo> result3 = userTaskState.prepareResultList(parameters3);
     // Test Case 3 result
     Assert.assertEquals(2, result3.size());
+    EasyMock.verify(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
     EasyMock.reset(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
 
     // Test Case 4: Get all tasks limit to 4 entries
@@ -208,6 +211,7 @@ public class KafkaCruiseControlServletEndpointTest {
     List<UserTaskManager.UserTaskInfo> result4 = userTaskState.prepareResultList(parameters4);
     // Test Case 4 result
     Assert.assertEquals(4, result4.size());
+    EasyMock.verify(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
     EasyMock.reset(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
 
     // Transition UserTaskManager state: some tasks will move from ACTIVE to COMPLETED
@@ -235,6 +239,7 @@ public class KafkaCruiseControlServletEndpointTest {
     List<UserTaskManager.UserTaskInfo> result5 = userTaskState2.prepareResultList(parameters5);
     // Test Case 5 result
     Assert.assertEquals(1, result5.size());
+    EasyMock.verify(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
     EasyMock.reset(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
 
   }
--- a/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
+++ b/cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
@@ -75,6 +75,8 @@ public class UserTaskManagerTest {
 
     HttpServletRequest mockHttpServletRequest2 = prepareServletRequest(mockHttpSession, testUserTaskId.toString());
     EasyMock.reset(mockHttpServletResponse);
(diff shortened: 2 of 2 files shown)


cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
```diff
@@ -183,6 +184,7 @@ public void testUserTaskParameters() throws Exception {
     List<UserTaskManager.UserTaskInfo> result2 = userTaskState.prepareResultList(parameters2);
     // Test Case 2 result
     Assert.assertEquals(3, result2.size());
+    EasyMock.verify(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
```
kyguy, line 187: From what I understand, the `reset()` call above clears the expectations made in the static block and no other `expect()` calls are made before this and the other `verify()` calls below. If that is true, wouldn't these `verify()` calls be no-ops? Are they intended to check something specific?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
```diff
@@ -196,6 +198,7 @@ public void testUserTaskParameters() throws Exception {
     List<UserTaskManager.UserTaskInfo> result3 = userTaskState.prepareResultList(parameters3);
     // Test Case 3 result
     Assert.assertEquals(2, result3.size());
+    EasyMock.verify(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
```
kyguy, line 201: Same as above

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
```diff
@@ -208,6 +211,7 @@ public void testUserTaskParameters() throws Exception {
     List<UserTaskManager.UserTaskInfo> result4 = userTaskState.prepareResultList(parameters4);
     // Test Case 4 result
     Assert.assertEquals(4, result4.size());
+    EasyMock.verify(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
```
kyguy, line 214: Same as above

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
```diff
@@ -235,6 +239,7 @@ public void testUserTaskParameters() throws Exception {
     List<UserTaskManager.UserTaskInfo> result5 = userTaskState2.prepareResultList(parameters5);
     // Test Case 5 result
     Assert.assertEquals(1, result5.size());
+    EasyMock.verify(MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, MOCK_HTTP_SERVLET_RESPONSE);
```
kyguy, line 242: Same as above

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
```diff
@@ -75,6 +75,8 @@ public void testCreateUserTask() throws Exception {
 
     HttpServletRequest mockHttpServletRequest2 = prepareServletRequest(mockHttpSession, testUserTaskId.toString());
     EasyMock.reset(mockHttpServletResponse);
+    requestContext.setHeader(EasyMock.capture(userTaskHeader), EasyMock.capture(userTaskHeaderValue));
+    EasyMock.replay(mockHttpServletResponse);
```
kyguy, line 79: Isn't the header name the same across all the tests cases? If so, do these lines add any additional coverage beyond what we have in line 52?

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
```diff
@@ -85,6 +87,8 @@ public void testCreateUserTask() throws Exception {
     Assert.assertEquals(future, future3);
 
     EasyMock.reset(mockHttpServletResponse);
+    requestContext.setHeader(EasyMock.capture(userTaskHeader), EasyMock.capture(userTaskHeaderValue));
+    EasyMock.replay(mockHttpServletResponse);
```
kyguy, line 91: Same as above

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
```diff
@@ -100,6 +104,8 @@ public void testCreateUserTask() throws Exception {
     Assert.assertEquals(future4, future);
 
     EasyMock.reset(mockHttpServletResponse);
+    requestContext2.setHeader(EasyMock.capture(userTaskHeader), EasyMock.capture(userTaskHeaderValue));
+    EasyMock.replay(mockHttpServletResponse);
```
kyguy, line 108: Same as above

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/KafkaCruiseControlServletEndpointTest.java
sameer-sde: You're right, and thank you for pointing this out. After the first reset(), no new expect() calls are made on MOCK_UUID_GENERATOR, MOCK_HTTP_SESSION, or MOCK_HTTP_SERVLET_RESPONSE before the subsequent verify() calls for test cases 2–5. Since there are no expectations set, these verify() calls are effectively no-ops they trivially pass with an empty expectation set.
The only meaningful verify() is the one after result1, which validates the full setup from initializeServletRequests() and the static block (e.g., ensuring randomUUID() was called 6 times and setHeader was invoked as expected during populateUserTaskManager()).
I'll remove the redundant verify() calls for test cases 2–5 since they don't add any real verification value. Thanks for the catch!

cruise-control/src/test/java/com/linkedin/kafka/cruisecontrol/servlet/UserTaskManagerTest.java
sameer-sde: Good catch! The header name is always the same constant (USER_TASK_HEADER_NAME) across all sub-cases, so the capture doesn't assert anything new about the header name itself.
However, the reset() + setHeader capture + replay() blocks are necessary for correctness, not just coverage. After reset(mockHttpServletResponse), the mock reverts to record state. The subsequent getOrCreateUserTask() call internally invokes setHeader on the response mock without re-entering replay mode via replay(mockHttpServletResponse), this would result in an unexpected invocation error and break the test.
So these blocks are a required fix to keep the mock in a valid state for each sub-case. I agree the header name assertion is redundant I can simplify by using EasyMock.anyString() without the capture if you prefer, since we're not asserting anything new about it. Happy to update

mhratson:
Please follow PR template format.

sameer-sde:
Thanks @mhratson, I've reformatted the PR description to follow the template.

PR #2402, opened 2026-09-05
Document Linux Foundation generative AI policy for contributors.

viktorsomogyi:
The project needs a written rule for AI-generated submissions under the Linux Foundation guidance: tool terms must fit Apache 2.0, third-party material in the output needs a compatible license and attribution, and the human remains responsible under the DCO.


CONTRIBUTING.md
```diff
@@ -46,6 +46,51 @@ git commit -s -m "Your commit message"
...
+  project's licensing policies, and include notice, attribution, and the
+  applicable license terms.
+- Follow your employer's policies when they are stricter.
+
+If an AI tool prepared any part of a commit, add a `Co-Authored-By` (or
+`Generated-by` / `Assisted-by`) trailer that names the tool and its version:
+
+```
+Co-Authored-By: <AI tool name and version>
+```
```
ppatierno, line 72: within the Strimzi community we had a discussion about having these trailers within a commit when a developer uses AI tools and we decided to ban these trailers :-) You can read the motivations here https://github.com/strimzi/governance/pull/29#discussion_r3083082103
It was also mostly following the Kubernetes AI policy, given that Strimzi lives in the same land (which is different from Cruise Control of course).
In the above discussion, I was more for leaving `Assisted-by` only but the community agreed on forbidding them all.
We preferred a disclosure within the PR where the author has to say he used AI tool help to write the code. In the end the DCO will give anyway the author the responsibility of everything.

viktorsomogyi: It is an interesting discussion. My overall justification was just to make this transparent as I see it as the trend for AI contributions. Regarding the debate of its form, I don't really see how is the free-form PR description better than the trailer, to me there is no difference in meaning. I find the trailer better simply because of technical reasons as for instance we could easily pull these PRs if we wanted to show how much AI (and what models) contribute to our codebase, whereas it's much harder to do this with free-form descriptions. I agree too that in both cases the responsibility is the user's, given by the DCO.

What I don't necessarily find useful is to distinguish between the trailers as it can become quite subjective where does the "assist" end and where does "co-author" starts. To me, even a single line made by AI can be "co-authored" if it's a substantial part of the PR.

I can also see the other side too, who don't really want to display whether it was AI-assisted or not, afterall we can look at it as a very complex auto-complete, and I don't publish either which IDE and thus autocomplete I use 🙂 . From statistical standpoints it can still be interesting to see which IDE developers use though.

CONTRIBUTING.md
```diff
@@ -46,6 +46,51 @@ git commit -s -m "Your commit message"
...
+  conflicts with this project's Apache 2.0 license, its intellectual property
+  policies, or the [Open Source Definition](https://opensource.org/osd).
+- If the output includes pre-existing third-party copyrighted material, confirm
+  you have permission to contribute it under a license that complies with this
+  project's licensing policies, and include notice, attribution, and the
+  applicable license terms.
+- Follow your employer's policies when they are stricter.
+
+If an AI tool prepared any part of a commit, add a `Co-Authored-By` (or
+`Generated-by` / `Assisted-by`) trailer that names the tool and its version:
```
kyguy, line 68: Are the terms: `Co-authored-By`, `Generated-by`, `Assisted-by` interchangeable? Or do they imply different levels of AI assistance? If the latter, would it be worth defining the level of assistance for each term?

CONTRIBUTING.md
kyguy: I prefer the way that Strimzi project enforces AI usage disclossure in the PR templates. I feel it is very explicit about who the author is and where responsibility lies regardless of whether or not AI tools were used. With the author as the sole sign-off, it is explicitly clear they agree to the code CONTRIBUTING policy:

“As a contributor, you represent that the code you submit is your original work or that of your employer (in which case you represent you have the right to bind your employer). By submitting code, you (and, if applicable, your employer) are licensing the submitted code to the open source community subject to the Apache 2.0 license.”

To make it even more clear, we could even add a line there that this policy still stands even if the contributor used AI tools.

That being said, I don't mind using trailers as described.

viktorsomogyi: I do find trailers useful also for the said reason but also it's extra information for the reviewer: it often makes sense to review the PR with a different agent than used for implementation.

Looking at strimzi PRs, that kind of template checkmark might still make sense for the sake of clarity on the policy, so I would be fine if we did both: template check for disclosure and trailer for extra information to help the review and possibly statistics if we ever need that.

CONTRIBUTING.md
viktorsomogyi: I don't find value in differentiating them as it's quite subjective. For clarity, we could settle with "Assisted-by", which looks like the most generic one and allows both end of the spectrum (idea was given by AI or idea was given by human and generated by AI).

PR #2406, opened 2026-09-17
Bump Netty to 4.1.138.Final and Vertx to 4.5.34

### LINKED ISSUE #2405: Bump Netty to 4.1.138.Final and Vertx to 4.5.34
To reduce CVEs.

egyedt:
## Summary
1. Why: To reduce CVEs.
2. What: Bump Netty to 4.1.138.Final and Vertx to 4.5.34.

## Behavior
No behavioural change, but less CVEs.

## Categorization
- [ ] documentation
- [ ] bugfix
- [ ] new feature
- [ ] refactor
- [X] security/CVE
- [ ] other

This PR resolves #2405.


build.gradle
```diff
@@ -296,6 +296,7 @@ project(':cruise-control') {
     implementation "org.apache.logging.log4j:log4j-slf4j2-impl:2.25.5"
     implementation "io.netty:netty-handler:${nettyVersion}"
     implementation "io.netty:netty-transport-native-epoll:${nettyVersion}"
+    implementation "io.netty:netty-codec-http:${nettyVersion}"
```
utafrali: Adding `netty-codec-http` as an explicit dependency to pin the CVE-fixed version is the right call and is consistent with the existing `netty-handler` / `netty-transport-native-epoll` declarations above.

That said, Vertx also brings in `netty-codec-http2` for HTTP/2 support, which is in the same family of HTTP codec CVEs. If `netty-codec-http` needed explicit pinning to win version resolution, `netty-codec-http2` likely has the same risk. Worth adding:
```
implementation "io.netty:netty-codec-http2:${nettyVersion}"
```

utafrali: The project documents CVE-motivated pins in the `constraints` block with a `because()` clause (see lines 330-335 for the log4j example). For consistency and auditability, a short comment here — or a constraint entry — would make it clear why `netty-codec-http` is explicitly declared when the other two Netty modules above it are not:
```groovy
// Explicit pin to keep netty-codec-http at ${nettyVersion} regardless of what Vertx resolves transitively
implementation "io.netty:netty-codec-http:${nettyVersion}"
```
This is minor, but the `because()` pattern elsewhere in this file is worth following.

utafrali (approved):
Clean, minimal CVE-reduction PR: the version bumps in gradle.properties are straightforward, and the explicit netty-codec-http pin follows the existing pattern for netty-handler and netty-transport-native-epoll. The only substantive gap worth checking is whether netty-codec-http2 also needs pinning, since Vertx uses it for HTTP/2 and the same class of HTTP codec CVEs can affect both modules.

build.gradle
ppatierno: In general netty and Vert.x are pretty well aligned. That's my experience within the Strimzi project. Vert.x community always release a new version when there is a new Netty fixing CVEs. Are we sure we need this direct pinning? But I am fine if you want to take it.

ppatierno (approved):
LGTM. I left just one nit.

ppatierno:
@utafrali thanks for the fully AI-generated review. I think we don't need a review which explains what the PR does, which is a typical problem of verbose AI.

build.gradle
egyedt: In this case, I prefer to not pin it. Thanks for the details!

egyedt:
@ppatierno 
Thanks, I updated it, if everything will be green, then we can merge it!

kyguy (approved):

